PR #68, opened 2023-04-27
Fix typos

Sajjon:
Fixes ~40 typos of which two where incorrect spellings of `#if FOUNDATION_FRAMWORK`. 

I don't have Swift 5.9 toolchain setup, so I have not been able to run tests.

PR #72, opened 2023-04-27
fix for minor inconsistency

kaiserabliz:
Found a minor inconsistency.


jmschonfeld:
It looks like these two changes are also addressed in @sajjon's PR #68 - perhaps we should consolidate?

kaiserabliz:
> It looks like these two changes are also addressed in @Sajjon's PR #68 - perhaps we should consolidate?

Yep, it looks like they are included in that PR, I will close this one then.

kaiserabliz:
Changes exist in #68 . Closing this.

Issue #45, opened 2023-04-18
Implement `Locale.canonicalIdentifier` 


parkera:
`Locale`'s `canonicalIdentifier` depends on a table in CoreFoundation: https://github.com/apple/swift-corelibs-foundation/blob/main/CoreFoundation/Locale.subproj/CFLocaleIdentifier.c

Without support for constant collections in Swift, a naive port of this would result in a pretty large allocation and memory usage regression. We may be able to take advantage of some of ICU's support for canonicalization, but we will need to verify compatibility with the existing implementation. Some callers may be expecting some of the legacy conversions to continue to work.

PR #75, opened 2023-04-27
Implement Locale's identifier from components in Swift

parkera:
Implement `Locale`'s `identifier(from:)` in Swift.

Perf test:

```swift
        let c1 = ["kCFLocaleLanguageCodeKey" : "en"]
        let c2 = ["kCFLocaleLanguageCodeKey" : "zh",
                  "kCFLocaleScriptCodeKey" : "Hans",
                  "kCFLocaleCountryCodeKey" : "TW"]
        let c3 = ["kCFLocaleLanguageCodeKey" : "es",
                  "kCFLocaleScriptCodeKey" : "",
                  "kCFLocaleCountryCodeKey" : "409"]
        
        for _ in 0..<100_000 {
            let id1 = Locale.identifier(fromComponents: c1)
            let id2 = Locale.identifier(fromComponents: c2)
            let id3 = Locale.identifier(fromComponents: c3)
            assert(id1.isEmpty == false)
            assert(id2.isEmpty == false)
            assert(id3.isEmpty == false)
        }
```

And the results:

```
// Swift implementation
test_identifierFromComponents()	0.040965125 seconds
test_identifierFromComponents()	0.039983875 seconds
// Previous CF implementation, from Swift
test_identifierFromComponents()	0.35747675 seconds
test_identifierFromComponents()	0.357553 seconds
```

Part of issue #45
Resolves rdar://108618701

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale.swift
@@ -971,27 +971,26 @@ public struct Locale : Hashable, Equatable, Sendable {
             
             // Identifier keywords must be ASCII a-z, A-Z, 0-9
             // They are normalized to all-lowercase
-            var ok = true
-            var correctedKey = key.utf8.map { c in
-                if 0x41 <= c && c <= 0x5a {
+            var correctedKey : [CChar] = []
+            for char in key.utf8 {
+                if (0x41...0x5a).contains(char) {
                     // A-Z
                     // Convert to lowercase by adding 0x20
-                    return CChar(c + 0x20)
-                } else if 0x61 <= c && c <= 0x7a {
+                    correctedKey.append(CChar(char + 0x20))
+                } else if (0x61...0x7a).contains(char) {
                     // a-z
-                    return CChar(c)
-                } else if 0x30 <= c && c <= 0x39 {
+                    correctedKey.append(CChar(char))
+                } else if (0x30...0x39).contains(char) {
                     // 0-9
-                    return CChar(c)
+                    correctedKey.append(CChar(char))
                 } else {
-                    ok = false
-                    return CChar(0)
+                    // Skip this key
+                    return nil
                 }
             }
             // null-terminate
             correctedKey.append(CChar(0))
             
-            guard ok else { return nil }
             let correctedKeyString = String(cString: correctedKey)
             
             // Values must be non-empty
--- a/Sources/FoundationInternationalization/Locale/Locale.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale.swift
@@ -929,18 +929,27 @@ public struct Locale : Hashable, Equatable, Sendable {
         LocaleCache.cache.preferredLanguages(forCurrentUser: false)
     }
 
+    private static let languageCodeKey = "kCFLocaleLanguageCodeKey"
+    private static let scriptCodeKey = "kCFLocaleScriptCodeKey"
+    private static let countryCodeKey = "kCFLocaleCountryCodeKey"
+    private static let variantCodeKey = "kCFLocaleVariantCodeKey"
+    private static let calendarKey = "kCFLocaleCalendarKey"
 
     /// Constructs an identifier from a dictionary of components.
     public static func identifier(fromComponents components: [String : String]) -> String {
+        // Holds remaining keywords after we remove the CF-specific ones
+        var keywords = components
         var result = ""
-        if let language = components["kCFLocaleLanguageCodeKey"] {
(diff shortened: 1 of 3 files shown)


Sources/FoundationInternationalization/Locale/Locale.swift
```diff
@@ -932,21 +932,62 @@ public struct Locale : Hashable, Equatable, Sendable {
...
     public static func identifier(fromComponents components: [String : String]) -> String {
-#if FOUNDATION_FRAMEWORK
-        // TODO: Port to Swift
-        // n.b. the CFLocaleCreateLocaleIdentifierFromComponents API is normally [String: String], but for 'convenience' allows a `Calendar` value for "kCFLocaleCalendarKey"/"calendar".
-        return CFLocaleCreateLocaleIdentifierFromComponents(kCFAllocatorSystemDefault, components as CFDictionary).rawValue as String
-#else
-        return ""
-#endif // FOUNDATION_FRAMEWORK
+        var result = ""
+        if let language = components["kCFLocaleLanguageCodeKey"] {
```
iCharlesHu, line 936: Should we make theses constants?

iCharlesHu (approved):
I can't seem to find `test_identifierFromComponents`. Would you mind adding that test?

parkera:
Hm, do we have a good place to put perf tests yet?

Sources/FoundationInternationalization/Locale/Locale.swift
```diff
@@ -932,21 +932,62 @@ public struct Locale : Hashable, Equatable, Sendable {
...
+        
+        if let country {
+            result += country
+        }
+        
+        if let variant {
+            result += "_" + variant
+        }
+        
+        if let calendar = components["kCFLocaleCalendarKey"] as? Calendar {
```
itingliu, line 986: I believe we'll still have to iterate through all the rest of the keys in `components`  besides those that have been handled above, and append them as key-value pairs after `"@"`. For example `"[kCFLocaleCollationIdentifier: "search"]"` is a legitimate dictionary that CF implementation used to support.

itingliu (changes requested):
Marking this as "request changes" In case you missed my comment about supporting other keywords

parkera:
Yup, you're 100% correct. Will revise.

parkera:
@swift-ci test

parkera:
@swift-ci test

Sources/FoundationInternationalization/Locale/Locale.swift
```diff
@@ -929,27 +929,138 @@ public struct Locale : Hashable, Equatable, Sendable {
...
+                } else {
+                    ok = false
+                    return CChar(0)
+                }
+            }
+            // null-terminate
+            correctedKey.append(CChar(0))
+            
+            guard ok else { return nil }
+            let correctedKeyString = String(cString: correctedKey)
```
itingliu, line 995: I believe calling `init(cString:)` involves another loop to find the buffer length. Since we're already iterating the utf8 buffer we could use another initializer that takes the count like this one, `String(unsafeUninitializedCapacity: count...)`? cc @glessard does that sound right to you?

itingliu (approved):
Besides my minor performance comments, the overall strategy looks good to me.

Sources/FoundationInternationalization/Locale/Locale.swift
```diff
@@ -929,27 +929,138 @@ public struct Locale : Hashable, Equatable, Sendable {
...
+                    // Convert to lowercase by adding 0x20
+                    return CChar(c + 0x20)
+                } else if 0x61 <= c && c <= 0x7a {
+                    // a-z
+                    return CChar(c)
+                } else if 0x30 <= c && c <= 0x39 {
+                    // 0-9
+                    return CChar(c)
+                } else {
+                    ok = false
```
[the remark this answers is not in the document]
parkera: I'd have to make a custom error type and throw it to get outside the `map` function, or change the whole thing into a simple for loop. Maybe I'll do the latter. I don't expect invalid characters here to be very common, for what it's worth.

Sources/FoundationInternationalization/Locale/Locale.swift
parkera: For what it's worth, the first time I wrote this I forgot the `null` termination and this initializer gave me an assert, which I appreciated. So it's a bit safer.

itingliu:
@swift-ci please test

parkera:
@swift-ci test

parkera:
@swift-ci test

PR #76, opened 2023-04-27
Fix formatting and linting

Bellaposa:


parkera:
This was run through a script to lint, I assume. I think in the future we'll want to do something like this, but not until we have better built-in support to do the linting automatically for new changes. Then we can do it all in one batch.

PR #90, opened 2023-05-01
AttributeScope Cacheing and Performance Improvements

jmschonfeld:
Currently, when converting/serializing an `AttributedString` we walk the provided `AttributeScope` one time for every new attribute we encounter and we do not cache the information between conversions. However, walking an attribute scope is fairly slow (since it requires a conforms-to-protocol check on the type of each field in the scope) so this behavior has been demonstrated to be a performance bottleneck. This patch improves the behavior by:

- Eagerly walking a scope at the start of conversion rather than walking it after encountering each new attribute
- Cache the result of scope walking so that each scope is walked a maximum of one time
- Switch from `dlopen`/`dlsym` to `_typeByName` to avoid a dependency on framework paths and simplify scope loading code

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
@@ -1,0 +1,176 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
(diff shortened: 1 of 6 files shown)


jmschonfeld:
@swift-ci please test

Azoy (approved):
Looks good to me!

Issue #78, opened 2023-04-28
Calling Date.description might result in undefined behaviour


dhoepfl:
Date’s `description` function uses a fixed size buffer (26 bytes, [see here](
https://github.com/apple/swift-foundation/blob/fec24b0db1f427a5eaa63ef24480981ec162cc37/Sources/FoundationEssentials/Date.swift#L242)). This buffer is given to `strftime`.

Both, dates before “-999-01-01 00:00:00 +0000” and starting at “10000-01-01 00:00:00 +0000”, will require 27 bytes or more (including null byte at the end).

The return value of this function is checked by `guard strftime(…) >= 0` ([see](https://github.com/apple/swift-foundation/blob/fec24b0db1f427a5eaa63ef24480981ec162cc37/Sources/FoundationEssentials/Date.swift#L251)). While most library functions in C return a negative value on failure, `strftime` returns 0 if the buffer is not big enough to hold the formatted date. Thus this guard *never* executes.

The buffer in this case is not guaranteed to contain a ending null byte, as the [man page](https://manpages.org/strftime#return-value) states: “Otherwise, 0 shall be returned and the _contents of the array are unspecified_.” (Highlighting mine)

The undefined content (worst case: 26 bytes forming valid UTF-8, e.g. 26 spaces) is [then](https://github.com/apple/swift-foundation/blob/fec24b0db1f427a5eaa63ef24480981ec162cc37/Sources/FoundationEssentials/Date.swift#L255) given to `String(validatingUTF8:)` which probably will (in the worst case) run behind the end of the buffer, resulting in a crash.

I suggest to either replace the call to `strftime` by a better API or change the buffer size to be big enough to hold any description of any point in time that Date might represent. In addition to that, the guard should be changed to `guard strftime(…) >= 25` given that is the smallest valid size.

test cases:
```
Date(timeIntervalSinceReferenceDate: TimeInterval(-63145526400)).description == "0000-01-01 00:00:00 +0000"
Date(timeIntervalSinceReferenceDate: TimeInterval(-4200000000000)).description == "-131090-12-30 21:20:00 +0000"
Date(timeIntervalSinceReferenceDate: TimeInterval(4200000000000)).description == "15310-04-10 02:40:00 +0000"
```

valentinradu:
I'd rather start with 26 bytes like we do now and allow for one single reallocation if `strftime` returns 0 - which is guaranteed (in our specific case / given the format) to be due to undersizing the buffer.

[...]

PR #93, opened 2023-05-02
Fixes issue #78: Calling Date.description might result in undefined behaviour

dhoepfl:
Calling Date.description might result in undefined behaviour

The buffer used was too small in some cases.
Fixed by allocating (more than) enough space on the stack.

### DIFF
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -239,20 +239,31 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
     /// function `description(locale:)`.
     public var description: String {
         // NSDate uses the constant format `uuuu-MM-dd HH:mm:ss '+0000'`
-        let outputLength = 26 // Constant length of `uuuu-MM-dd HH:mm:ss '+0000'` + \0
         let format = "%Y-%m-%d %H:%M:%S +0000"
+        let unavailable = "<description unavailable>"
+
         var info = tm()
         var time = time_t(self.timeIntervalSince1970)
         gmtime_r(&time, &info)
-        let result = UnsafeMutablePointer<CChar>.allocate(capacity: outputLength)
-        defer {
-            result.deallocate()
-        }
-        guard strftime(result, outputLength, format, &info) >= 0 else {
-            return "<description unavailable>"
-        }
 
-        return String(validatingUTF8: result) ?? "<description unavailable>"
+        // This allocates stack space for range of 10^102 years
+        // That's more than Date currently supports.
+        let bufferSize = 128
+        return withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferSize) { buffer -> String in
+           guard let ptr = buffer.baseAddress else {
+              return unavailable
+           }
+
+           guard strftime(ptr, bufferSize, format, &info) != 0 else {
+              return unavailable
+           }
+
+           guard let result = String(validatingUTF8: ptr) else {
+              return unavailable
+           }
+
+           return result
+        }
     }
 #endif // !FOUNDATION_FRAMEWORK
 
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -249,7 +249,7 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
         // This allocates stack space for range of 10^102 years
         // That's more than Date currently supports.
         let bufferSize = 128
-        return withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferSize) { buffer -> String in
+        return withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferSize) { buffer in
             guard let ptr = buffer.baseAddress else {
                 return unavailable
             }
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -239,8 +239,21 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
     /// function `description(locale:)`.
     public var description: String {
         // NSDate uses the constant format `uuuu-MM-dd HH:mm:ss '+0000'`
+
+        // Glibc needs a non-standard format option to pad %Y to 4 digits
+#if canImport(Glibc)
+        let format = "%4Y-%m-%d %H:%M:%S +0000"
+#else
         let format = "%Y-%m-%d %H:%M:%S +0000"
+#endif
         let unavailable = "<description unavailable>"
+
+        guard self >= Date.distantPast else {
+            return unavailable
+        }
+        guard self <= Date.distantFuture else {
+            return unavailable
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Date.swift
```diff
@@ -239,20 +239,31 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
...
-        }
-        guard strftime(result, outputLength, format, &info) >= 0 else {
-            return "<description unavailable>"
-        }
 
-        return String(validatingUTF8: result) ?? "<description unavailable>"
+        // This allocates stack space for range of 10^102 years
+        // That's more than Date currently supports.
+        let bufferSize = 128
+        return withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferSize) { buffer -> String in
```
itingliu, line 252:
```suggestion
        return withUnsafeTemporaryAllocation(of: CChar.self, capacity: bufferSize) { buffer in
```

I think it's ok to omit the explicit return type here

Sources/FoundationEssentials/Date.swift
```diff
@@ -239,20 +239,31 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
...
+           guard let ptr = buffer.baseAddress else {
+              return unavailable
```
itingliu, line 254: Nit: The indentation looks off here and the easy returns below

itingliu (commented):
Can we add a unit test for this?

iCharlesHu:
Thanks for the fix! Ditto for adding a unit test.

Sources/FoundationEssentials/Date.swift
glessard: Another note: the temporary allocation function always provides a non-nil baseAddress (it crashes if it fails to allocate, like the other Swift dynamic allocators). Given that, it is okay to use `unsafelyUnwrapped` here, because this `else` branch cannot be tested.

Sources/FoundationEssentials/Date.swift
```diff
@@ -239,20 +239,31 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
...
+           }
+
+           guard strftime(ptr, bufferSize, format, &info) != 0 else {
+              return unavailable
+           }
+
+           guard let result = String(validatingUTF8: ptr) else {
```
glessard, line 261: Since we are touching this function, we could use this opportunity to switch to using `String._tryFromUTF8()`, which takes a buffer pointer as an argument. `strftime` returns the number of bytes it has output, then `String(validatingUTF8:)` has to call `strlen`. We can avoid that work.

Sources/FoundationEssentials/Date.swift
```diff
@@ -239,20 +239,31 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
...
-            result.deallocate()
-        }
-        guard strftime(result, outputLength, format, &info) >= 0 else {
-            return "<description unavailable>"
-        }
 
-        return String(validatingUTF8: result) ?? "<description unavailable>"
+        // This allocates stack space for range of 10^102 years
+        // That's more than Date currently supports.
+        let bufferSize = 128
```
itingliu, line 251: `struct tm` stores its year with an Int. That means that the farthest date we care about is the last moment of year `Int64.max`. Considering that a 128 buffer is probably overkill?

dhoepfl:
I just pushed commits that address most of the review comments:

* Indentation is fixed
* Explicit return type is removed
* Tests added

When writing the tests, I found some more things that needed to be addressed:

* glibc does not pad "%Y" to 4 digits. Worked around it by using non-standard "%4Y" for glibc.
* Huge negative offsets since reference time sometimes (not all!) resulted in positive years in the output. I guess that is related to `tm_year`’s type. As a work around, I limit description to `.distantPast….distantFuture`.
* That limitation leads to (currently) the result never using more than 26 bytes … which makes the change to temporary allocation no longer required and 128 bytes even more overkill. For now, I kept it.
* I tried to use `_tryFromUTF8` but I did not find a way without creating additional instances (to get a `UnsafeBufferPointer` from the `UnsafeMutableBufferPointer`). I’m not sure if that might trigger a copy in some cases, which would be worse than counting 26 chars. (I think using a non-validating init would be okay here.)

parkera:
@swift-ci test

glessard:
@dhoepfl Thanks! Creating an `UnsafeBufferPointer` from a start and a count never allocates. The suggestion to use one of the repairing initializers is interesting in this case: it should work in principle but we must be wary of changing code behaviour.

Sources/FoundationEssentials/Date.swift
itingliu: You also mentioned this in latest commit 

> * That limitation leads to (currently) the result never using more than 26 bytes … which makes the change to temporary allocation no longer required and 128 bytes even more overkill. For now, I kept it.

Why don't you just make this change while you're here?

itingliu (approved):
Left a comment about shrinking that buffer from 128. Other than that this works for me

Sources/FoundationEssentials/Date.swift
dhoepfl: I don’t think it really matters wether we allocate 26 bytes, 32, 64, or 128 bytes.

I don’t want to use 26 bytes because that will break when .distantFuture/Past is extended. If these limits get changed, 32 is also well within reach. So what would I change it to? A good value would be whatever TimeInterval allows. I’m not sure what this would be, probably somewhere around 330. Not super alignment friendly so I’d use 384 then.

64 would be enough, even for +/-Int128 but then, again, I don’t think it does matter.

What I _really_ would like to do is something like `"\(tm.tm_year, minWidth: 4, padding: "0")-\(tm.tm_month+1, width: 2, padding: "0")-\(tm. tm_mday, width: 2, padding: "0") \(tm.tm_hour, width: 2, padding: "0"):\(tm.tm_min, width: 2, padding: "0"):\(tm.tm_sec, width: 2, padding: "0")"` because that would remove the overhead of parsing a (constant) template, the overhead of validating UTF-8, and problems with non-standard `strftime` (I cannot test the WinSDK path).

dhoepfl: Just in case I failed to transport it in English: If you think this should be a different value, I’ll change it.

I just think that 26 is not future prove (and given the `%4Y` thing, I do no longer trust `strftime` implementations to not output a bit more than reasonable). Given that, I opted to be on the safe (safer) side by using 128 instead of 64. If stack space is a problem, I’d use 32 (alignment friendly version of 26 with a few bytes buffer).

compnerd: What do you mean by a non-standard `strftime`?  The windows version should be conformant as per the spec.

itingliu: I don't feel strongly towards any other number. It's just that I thought you agreed that 128 was overkill, so I was curious why you didn't change that number to something less likely being an overkill. I'm ok with this

PR #96, opened 2023-05-04
Replace `BufferViewIndex<Element>` with `Index` in BufferView.

nnabeyang:
This PR replaces `BufferViewIndex<Element>` with `Index` that is typealias of Collection in BufferView.

### DIFF
--- a/Sources/FoundationEssentials/JSON/BufferView.swift
+++ b/Sources/FoundationEssentials/JSON/BufferView.swift
@@ -178,12 +178,12 @@ extension BufferView:
     }
 
     @inline(__always)
-    func distance(from start: BufferViewIndex<Element>, to end: BufferViewIndex<Element>) -> Int {
+    func distance(from start: Index, to end: Index) -> Int {
         start.distance(to: end)
     }
 
     @inline(__always)
-    subscript(position: BufferViewIndex<Element>) -> Element {
+    subscript(position: Index) -> Element {
         get {
             _checkBounds(position)
             return self[unchecked: position]
@@ -191,7 +191,7 @@ extension BufferView:
     }
 
     @inline(__always)
-    subscript(unchecked position: BufferViewIndex<Element>) -> Element {
+    subscript(unchecked position: Index) -> Element {
         get {
             if _isPOD(Element.self) {
                 return position._rawValue.loadUnaligned(as: Element.self)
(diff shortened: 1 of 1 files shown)


Sources/FoundationEssentials/JSON/BufferView.swift
```diff
@@ -100,7 +102,8 @@ extension BufferView:
     RandomAccessCollection {
 
     typealias Element = Element
-    typealias Index = BufferViewIndex<Element>
+    // Index is already defined
+    // typealias Index = BufferViewIndex<Element>
```
iCharlesHu, line 106: (nit) we can probably remove these two comments.

iCharlesHu (approved):
Looks great to me (pending @glessard 's comments). Thanks for the fix!

Sources/FoundationEssentials/JSON/BufferView.swift
nnabeyang: @iCharlesHu Thank you. The comments is removed.
https://github.com/apple/swift-foundation/pull/96/commits/78161e185ca76e5c48953d5a27f36472a99eba05

Sources/FoundationEssentials/JSON/BufferView.swift
```diff
@@ -100,7 +102,6 @@ extension BufferView:
     RandomAccessCollection {
 
     typealias Element = Element
-    typealias Index = BufferViewIndex<Element>
```
glessard, line 103: I agree that the definitions below should be made in terms of `Index`, but I disagree with moving the definition of `Index` from here to the main struct. `Index` is part of the `Collection` conformance, while the struct simply holds two stored properties. While one of these _is_ `count`, if we were to change the internal representation of the struct to be two indices, I'd want to move the computed `count` to be among the `Collection` conformance.

nnabeyang: @glessard Thank you for your review. The following changes were added.

* revert Index
https://github.com/apple/swift-foundation/pull/96/commits/6900c61344048242878ccbb9b79b0a47b2b75e80

* move the computed count
https://github.com/apple/swift-foundation/pull/96/commits/6900c61344048242878ccbb9b79b0a47b2b75e80

glessard:
Thanks for noting this!

PR #97, opened 2023-05-05
LockedState: Fix lock not being released when body throws

nonsensery:
#### Summary

Currently, if the `body` closure passed to `withLockExtendingLifespanOfState(_:)` throws an error, the lock is not released. This PR fixes that.

#### Change

My fix was to switch the implementation of `withLockExtendingLifespanOfState(_:)` to delegate to `withLockUnchecked(_:)`, which does properly unlock if its body throws an error.

#### Verification

Added unit tests to verify this change, and backfilled tests that verify the lifespan-extending behavior of this method.

### DIFF
--- a/Sources/FoundationEssentials/LockedState.swift
+++ b/Sources/FoundationEssentials/LockedState.swift
@@ -94,12 +94,9 @@ internal struct LockedState<State> {
 
     // Ensures the managed state outlives the locked `body`.
     func withLockExtendingLifetimeOfState<T>(_ body: @Sendable (inout State) throws -> T) rethrows -> T {
-        try _buffer.withUnsafeMutablePointers { state, lock in
-            _Lock.lock(lock)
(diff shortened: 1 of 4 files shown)


glessard:
Thanks Alex! Unfortunately we have two copies of this file (until we have the `package` access keyword!). Could you copy this fix to `FoundationInternationalization/LockedState.swift`?

glessard (changes requested):
Thanks Alex! Unfortunately we have two copies of this file (until we have the package access keyword!). I applied the change to the other copy.

iCharlesHu:
@swift-ci please test

glessard:
@swift-ci please test

parkera:
We should get this fix in.

glessard:
@swift-ci please test

glessard (approved):
Thanks!

glessard:
Linked to rdar://109056170

PR #99, opened 2023-05-05
__CFUniChar*CharacterSetBitmaps should be `uint8_t const * const`

itingliu:
Also consolidate the position of `const`


kperryua (approved):
Looks good, thanks!

PR #102, opened 2023-05-09
Use `__owned` parameter modifier

glessard:
Using the `consuming` parameter modifier for copyable (aka normal) types is deferred from swift 5.9. This uses the older `__owned` modifier instead, and consumes the parameter into a `var` as the first statement. This should generate the same code. This fixes a compilation error in `swift-foundation`'s Linux CI.


glessard:
@swift-ci please test

jmschonfeld:
Based on CI, looks like `JSON5Scanner.swift` needs to be updated too

jmschonfeld (approved):
Looks like linux CI passes now, thanks!

iCharlesHu (approved):
It looks like Linux is passing now!

PR #103, opened 2023-05-09
Revert "Move count among the Collection conformance in BufferView"

glessard:
This reverts a change in the internal representation of `BufferView` which potentially has significant performance consequences. We might want to make this change, but we need to carefully evaluate its performance consequences first.

This reverts commit 6900c61344048242878ccbb9b79b0a47b2b75e80.

### DIFF
--- a/Sources/FoundationEssentials/JSON/BufferView.swift
+++ b/Sources/FoundationEssentials/JSON/BufferView.swift
@@ -15,7 +15,7 @@
 
 internal struct BufferView<Element> {
     let start: Index
-    let end: Index
+    let count: Int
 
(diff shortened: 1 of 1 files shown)


jmschonfeld (approved):

iCharlesHu (approved):

glessard:
@swift-ci please test

glessard:
@nnabeyang We might want to do this change, but we need to be convinced that it is worth it both in terms of code size and performance. We know that storing pointer+count is better for integer indexing, as in `UnsafeBufferPointer`. We should find out whether storing pointer+count or 2 pointers is better for the kind of index introduced in `BufferView`.

PR #105, opened 2023-05-10
Update README for toolchain requirement

itingliu:
The `consume` keyword requires `-enable-experimental-move-only` frontend flag, which was only enabled in recent snapshots. Specify a snapshot version that we know works in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Later this year, the porting effort will continue. It will bring high quality Sw
 
 ## Building and Testing
 
-Building the Foundation package requires the under-development [Swift 5.9 toolchain](https://www.swift.org/download/#swift-59-development) on macOS and Linux.
+Building the Foundation package requires the under-development [Swift 5.9 toolchain](https://www.swift.org/download/#swift-59-development) on or later than May 3rd 2023 (607f4eb), on macOS and Linux. 
 ### macOS
 
 macOS Ventura 13.3.1 is the minimum supported version.


glessard (approved):

README.md
```diff
@@ -48,7 +48,7 @@ Later this year, the porting effort will continue. It will bring high quality Sw
 
 ## Building and Testing
 
-Building the Foundation package requires the under-development [Swift 5.9 toolchain](https://www.swift.org/download/#swift-59-development) on macOS and Linux.
+Building the Foundation package requires the under-development [Swift 5.9 toolchain](https://www.swift.org/download/#swift-59-development) on or later than May 3rd 2023 (607f4eb), on macOS and Linux.
```
MaxDesiatov, line 51:
```suggestion
Building the Foundation package requires the under-development [Swift 5.9 toolchain](https://www.swift.org/download/#swift-59-development) or later than May 3rd 2023 (607f4eb), on macOS and Linux. 
```

glessard: @MaxDesiatov it's not a typo; the May 3rd snapshot is known to work. @compnerd This is an oversight, it should be the same on Windows.

itingliu: @compnerd I think you're right that it's the same for all platforms, but I'm not able to verify that. Actually we probably want to amend this README to include instructions for building on Windows. I'll leave it out for now

itingliu: @MaxDesiatov @glessard was correct that it's not a typo as that's an inclusive range

itingliu:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu:
Ahh it looks like snapshots younger than May 3rd should contain the availability macro feature.

iCharlesHu (approved):

itingliu:
@swift-ci please test

PR #111, opened 2023-05-11
Update `Date.description` tests for Foundation Framework

itingliu:
We use ICU for date calculation for Foundation Framework, and `gmtime` otherwise. It appears that the result starts to diverge at some time in Gregorian year 1582, which happens to be the year when Gregorian calendar went into effect. That in turn results in disparate representations of `Date` before year 1582.

It's not clear how `tm` and `strtime` handles before-Gregorian dates since it's entirely in the implementation details. This PR simply update test expectations accordingly.

### DIFF
--- a/Tests/FoundationEssentialsTests/DateTests.swift
+++ b/Tests/FoundationEssentialsTests/DateTests.swift
@@ -85,7 +85,11 @@ final class DateTests : XCTestCase {
     }
 
     func testDescriptionDistantPast() {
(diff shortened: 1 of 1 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #119, opened 2023-05-12
Remove String.reserveCapacity() calls with dubious value from JSONScanners

kperryua:
These calls to `String.reserveCapacity()` are wrong when the accumulated input is non-ASCII, resulting in requests for smaller capacities than is already being consumed. Changing them or removing them shows little to no affect on performance in benchmarks anyway. In the interest of simplicity, let's remove them.


glessard (approved):
Good idea.

glessard:
@swift-ci please test

iCharlesHu (approved):

PR #121, opened 2023-05-14
build: define `_CRT_SECURE_NO_WARNINIGS` on Windows

compnerd:
We need to define this macro to avoid warnings due to the use of non-bounded buffer operations.  Without this, we end up with an overwhelming number of warnings.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #122, opened 2023-05-15
Replace _CShims with RustShims

egdeeno:
Initial effort to replace C-files with rust-files.
it allows just type `swift build` to build the project.
Tested on Ubuntu 22.04.

TODO: 
macos support needs work.
Sources/RustShims/.cargo/config needs manual edit. Should be automated.
Sources/RustShims/build.sh needs manual edit. Should be automated.


egdeeno:
> Thanks for this PR @egdeeno. We have a high level goal in this project of replacing C code with Swift code. That unified implementation in one language will be easy for Swift developers to approach, and benefit from improvements to the Swift language and ecosystem.
> 
> Is it possible to rework this change to refactor some of the C code into Swift instead of into Rust? For files like `uuid.c`, that is the direction we intend to move.

egdeeno:
Thank you!
Agree, unified implementation is better.

PR #124, opened 2023-05-15
_CShims: add Windows variant of the string functions

compnerd:
Add a Windows path for the string functions that deal with locales. This is the first step towards supporting the Windows platform.


compnerd:
@swift-ci please test

parkera (approved):

PR #125, opened 2023-05-15
FoundationEssentials: correct `PlatformKey` for Windows

compnerd:
Windows prefers to use FLS (fibre local storage) for TLS (thread local storage).  The FLS key is of type `DWORD` rather than `UInt`.  Update the `typealias` for Windows.


jmschonfeld (approved):
Thanks for correcting this! Do any other pieces of the TLS implementation need to be updated here, or just the `PlatformKey` type?

jmschonfeld:
@swift-ci please test

PR #126, opened 2023-05-15
Remove @_implementationOnly import from Darwin

jmschonfeld:
This removes `@_implementationOnly` from one of the `Darwin` imports since the rest of the package imports `Darwin` as non-implementation-only


jmschonfeld:
@swift-ci please test

itingliu (approved):

parkera (approved):

compnerd:
Does it make sense to consider `@_implementationOnly` imports for the libc module?  I know that the existing Foundation implementations re-export the libc interface.  It is unclear to me what level of compatibility is intended here.

PR #127, opened 2023-05-15
Update the fallback locale from en to en-001

parkera:
We should use `en-001` (language identifier) / `en_001` (locale identifier) as a fallback instead of "just" English.

rdar://108958839


itingliu:
@swift-ci please test

itingliu (approved):

PR #129, opened 2023-05-16
Port _CShims UUID to swift

tonyarnold:
This is a spike for my own learning to see what's involved in removing the C-backed UUID implementation in favour of a Swift version. 

I'm happy to revise this, as I realise that the implementation of some of this isn't ideal.


Sources/FoundationEssentials/UUID.swift
```diff
@@ -23,26 +17,29 @@ public typealias uuid_string_t = (Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8
...
     /// Returns nil for invalid strings.
-    public init?(uuidString string: __shared String) {
-        let res = withUnsafeMutablePointer(to: &uuid) {
-            $0.withMemoryRebound(to: UInt8.self, capacity: 16) {
-                return uuid_parse(string, $0)
-            }
-        }
-        if res != 0 {
-            return nil
+    public init?(version: Version = .v4, uuidString string: __shared String) {
```
parkera, line 35: This would be an ABI breakage, since we're removing the previous initializer.

tonyarnold: That makes sense, thanks for letting me know Tony. 

I restored the original initializers in c3cc89cc5faf0bbd36c9123a73e996cab405569c, however I also understand that the changes in this PR are something that the team aren't ready to look at yet.

parkera:
One interesting security feature of our UUID implementation is that it is designed to do its work in constant time regardless of the content of the UUID. e.g., for equality checking. This is done in order to defeat timing attacks. We would need to verify we keep that behavior with a new implementation.

Sources/FoundationEssentials/UUIDv4.swift
```diff
@@ -0,0 +1,52 @@
...
+        randomBits[6] = (randomBits[6] & 0x0F) | 0x40
+        randomBits[8] = (randomBits[8] & 0x3F) | 0x80
+
+        return randomBits.withUnsafeBytes { buffer in
+            return buffer.bindMemory(to: uuid_t.self)[0]
+        }
+    }
+
+    static func v4_parse(uuidString string: __shared String) -> uuid_t? {
+        let components = string
```
stevapple, line 24: Based on https://github.com/apple/swift-foundation/pull/129#issuecomment-1577124404, one obvious optimization we can do here is to `guard` against `string.count`. However, both `String.count` and `String.UTF8View.count` is not O(1), so I guess we should parse the string manually and `throw` if it’s malformed, including too short or too long?

PR #131, opened 2023-05-16
FoundationEssentials: port `Date` to Windows

compnerd:
MSVCRT does not provide `gmtime_r`.  `gmtime` is re-entrant safe, but uses TLS storage.  Prefer to use `_gmtime64_s` which provides the re-entrancy with local buffers but swaps parameter orders.  Prefer to use the 64-bit time variant over the 32-bit `time_t`.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #133, opened 2023-05-16
Replace `__owned` with an `inout` parameter

glessard:
This is a second fix for the delay of the `consuming` parameter modifier for copyable types. The previous attempt, using `__owned`, resulted in a defensive retain of the parameter, which then caused an unnecessary copy. Using an `inout` parameter prevents this.

Addresses rdar://109425225


glessard:
@swift-ci please test

glessard:
@swift-ci please test

kperryua:
We probably don't want to be modifying this via `inout` *and* returning the string directly?

glessard:
It doesn't cause an unexpected retain to be added in the generated assembly, so I went with the slightly smaller change.

kperryua (approved):

PR #137, opened 2023-05-18
Foundationinternationalization: add `Codable` conformance

compnerd:
Add the missing conformance for `Codable` to Windows.  This fixes a large set of warnings when building the package.


compnerd:
@swift-ci please test

PR #138, opened 2023-05-18
FoundationInternationalization: capture with `self.` prefix

compnerd:
Use `self.` to mark the capture of `self` within the closure which is identified by the compiler.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #139, opened 2023-05-18
FoundationInternationalization: handle C bridging type differences

compnerd:
`enum`s are signed on Windows by default where as they are unsigned on Unices.  Adjust the bridged types to use the correct signedness.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1159,7 +1159,11 @@ internal final class _Calendar: Equatable, @unchecked Sendable {
 
             for i in 0..<7 {
                 var status = U_ZERO_ERROR
-                weekdayTypes[i] = ucal_getDayOfWeekType(ucalendar, UCalendarDaysOfWeek(rawValue: weekdaysIndex[i]), &status)
+#if os(Windows)
```
parkera, line 1162: Does it make sense instead here to provide an extension on `UCalendarDaysOfWeek` in our ICU shims file which takes either type?

parkera: (same for the other ifdefs)

compnerd: You cannot provide an overload in C IIRC.  We could change the signature to take a signed integer, and always cast to the signed type.  I'm not sure if that would be difficult to maintain or not.

parkera: In general I'm wondering if we can isolate the platform difference to something closer to the ICU "API" rather than having to put it in places that use it, which theoretically are abstracted from that. It may be that we need a slightly thicker layer of abstraction in the first place.

compnerd: How about something like an underscored variant of the API that is shared?  I'll try to sketch out something.

compnerd: The enums are more difficult than I anticipated ... maybe you have an idea?

parkera:
Hm, @iCharlesHu and @itingliu what do you think of this? Any other way we can avoid the cast, or the shim? I'm not sure if either are a great alternative...

compnerd:
@parkera I found that your idea for the enum extensions was rather insightful.  I've converted the shims into Swift instead of C.  I've preferred to keep the type as `CInt` rather than `CUnsignedInt` in case we want to use C to do the cast.

compnerd:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1158,8 +1158,7 @@ internal final class _Calendar: Equatable, @unchecked Sendable {
             var cease: UInt32?
 
             for i in 0..<7 {
-                var status = U_ZERO_ERROR
-                weekdayTypes[i] = ucal_getDayOfWeekType(ucalendar, UCalendarDaysOfWeek(rawValue: weekdaysIndex[i]), &status)
+                weekdaysIndex[i] = UInt32(_ucal_getDayOfWeekType(ucalendar, CInt(weekdaysIndex[i])))
```
parkera, line 1161: We lost the status here, but I guess we didn't check the result anyway?

parkera: I think I remember spelunking into ICU for this and verifying it was never set, but it's probably worth double checking my work there.

compnerd: Correct, and this was intentional.  If we care about the status, we should actually mark the function as `throws` and surface the status as an error.

parkera: I went back and forth on this for a bit, and in the end decided that I wouldn't try to Swiftify the ICU APIs, at least not here and now. Instead I added a `check` function to the error type and make it an explicit step after calling them.

compnerd: It _can_ return a `U_ILLEGAL_ARGUMENT_ERROR` as a status in the case that `dayOfWeek` is invalid.  But that we should be caught by the compiler, so 🤷‍♂️.

Sources/FoundationInternationalization/ICU/ICUShims.swift
```diff
@@ -0,0 +1,53 @@
...
+  internal init(_ rawValue: CInt) {
+    self.init(rawValue: EnumRawType(rawValue))
+  }
+}
+
+extension UNumberFormatFields {
+  internal init(_ rawValue: CInt) {
+    self.init(rawValue: EnumRawType(rawValue))
+  }
+}
```
parkera, line 53: If you'll permit me one nitpick - can we instead add these to the `ICU+Enums.swift` file, since I think that's actually the one meant to hold this kind of thing? I'm trying to keep these ICU extensions organized instead of ending up with a grab bag 'shims' file.

compnerd: Sure, I don't mind that at all.  I just didn't notice the file.

compnerd:
@swift-ci please test

parkera:
Curious test failure?

```
[251/547] Testing FoundationEssentialsTests.JSONEncoderTests/test_JSONPermitsTrailingCommas
error: Exited with signal code 11
```

kperryua:
> [251/547] Testing FoundationEssentialsTests.JSONEncoderTests/test_JSONPermitsTrailingCommas
> error: Exited with signal code 11

That's very curious as nothing in that test should be calling into FoundationInternationalization code…

compnerd:
Yeah, that is rather odd - I'm not sure what to make of it, and not much to go on either.  Being able to reproduce the failure locally so I can poke some at it under a debugger would be helpful.

parkera (approved):

parkera:
@swift-ci test

parkera:
Thanks, looks good.

compnerd:
@swift-ci please test

PR #140, opened 2023-05-18
Re-enable ICU tests

itingliu:
- unifdef FIXED_106570987
- Update test expectations due to different ICU version used by FOUNDATION_FRAMEWORK.
- Update test result on FOUNDATION_FRAMEWORK

### DIFF
--- a/Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
+++ b/Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
@@ -355,7 +355,7 @@ final class DateAttributedFormatStyleTests : XCTestCase {
                                                       ("3", .hour),
                                                       (":", nil),
                                                       ("04", .minute),
-                                                      (" ", nil),
+                                                      (expectedSeparator, nil),
                                                       ("PM", .amPM)],
         ]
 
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
```diff
@@ -38,6 +38,11 @@ extension AttributedString {
 @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
 final class DateFormatStyleTests : XCTestCase {
     let referenceDate = Date(timeIntervalSinceReferenceDate: 0)
+#if FOUNDATION_FRAMEWORK
+    let expectedSeparator = "\u{202f}"
+#else
+    let expectedSeparator = " "
+#endif
```
iCharlesHu, line 45: I very much love this solution!! Do you think we should just move it to TestSupport as a global variable?

iCharlesHu (approved):

Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
itingliu: I don't think so. It's very data dependent so I think it's better to keep it local. It's currently used for formatting Date also.

PR #144, opened 2023-05-23
Zero in "0 byte" displays like latn "O" in Chinese simplified.

itingliu:
Only a small number of languages prefer spelling out number in this context according to localization team. Chinese wasn't one of those, hence this bug. Fix it by hard-coding a list of language codes where we should format number 0 as a spellout number.

Also were using a rule-based number formatter to format number 0 into spellout "Zero", which defaults to an ordinal number. That default rule was incorrect. The number should be formatted as a cardinal number instead.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

parkera (approved):

PR #147, opened 2023-05-26
tests: disable currently failing test

compnerd:
This test is currently failing on CI.  It is unclear if there is a miscompile currently.  Temporarily disable the test to not lose signal on other tests.

### DIFF
--- a/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
@@ -1138,7 +1138,10 @@ final class JSONEncoderTests : XCTestCase {
         _test(JSONString: "[\"本日\"]", to: ["本日"])
     }
 
-    func test_JSONUnicodeEscapes() {
+    func test_JSONUnicodeEscapes() throws {
+#if os(Linux)
+        throw XCTSkip("current development swift builds cause a stack overflow")
+#endif
         let testCases = [
             // e-acute and greater-than-or-equal-to
             "\"\\u00e9\\u2265\"" : "é≥",


compnerd:
@swift-ci please test

iCharlesHu (approved):
Thanks Saleem! I was looking into this one as well. I can only reproduce it with nightly toolchain on Linux (not on macOS).

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
```diff
@@ -1138,7 +1138,9 @@ final class JSONEncoderTests : XCTestCase {
         _test(JSONString: "[\"本日\"]", to: ["本日"])
     }
 
-    func test_JSONUnicodeEscapes() {
+    func test_JSONUnicodeEscapes() throws {
+        try XCTSkipIf(true, "current swift builds cause a stack overflow")
```
iCharlesHu, line 1142: Actually can you only skip for Linux? It believe it still works fine on other platforms.

compnerd: I didn't manage to test it on macOS - let's disable it only on Linux if it is passing on macOS.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
> Thanks Saleem! I was looking into this one as well. I can only reproduce it with nightly toolchain on Linux (not on macOS).

Happy to help!  Sorry I couldn't seem to be able to make quick progress in tracking down what is going on.  It seems to be recusing, but I cannot tell why.

PR #149, opened 2023-05-26
FoundationEssentials: add `RawRepresentable` conformance to `_ErrorCo…

compnerd:
…deProtocol`

The `Self` constraint suggests that the protocol provide a `RawRepresentable` conformance.  Silence the compiler warning by adding the conformance.


parkera:
I think it boils down to a new requirement that would be imposed upon anyone who adopts this (outside the framework), which we don't have any mechanism to enforce. I could be wrong though.

iCharlesHu:
@swift-ci please test macOS

parkera:
@compnerd any idea what we can do about this one, given the ABI break?

compnerd:
I need to try to build this again to remember what is going on here; I wonder if the constraint might be something that we need on the typealias?  But either way, the new constraint is the issue.  I'm not sure if there is a way to adopt this type without that, but at the same time, we don't have a way to silence the warning.

PR #150, opened 2023-05-26
FoundationInternationalization: adjust the libc imports

compnerd:
Adjust the importing of the system libc library for Windows.  When possible, prefer to import the underlying C library interface through the `ucrt` (clang) module.  When needed, resort to use `CRT` which has an overlay which provides certain overloads for math functions.


iCharlesHu (approved):

compnerd:
@swift-ci please test

PR #151, opened 2023-05-26
FoundationEssentials: Use dlsym instead of _typeByName in Attribute scope lookup

Azoy:
`_typeByName` is actually quite slow compared to just `dlsym` these days. Use that instead.

Resolves: rdar://109785007

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
@@ -29,6 +29,7 @@ public enum AttributeScopes { }
 
 #if FOUNDATION_FRAMEWORK
 
+@_implementationOnly import Darwin
 @_implementationOnly import ReflectionInternal
 
 fileprivate struct ScopeDescription : Sendable {
@@ -54,7 +55,10 @@ fileprivate struct LoadedScopeCache : Sendable {
         lastImageCount = 0
     }
     
-    mutating func scopeType(for name: String) -> (any AttributeScope.Type)? {
+    mutating func scopeType(
+        for name: String,
+        in path: String
+    ) -> (any AttributeScope.Type)? {
         if let cached = scopeMangledNames[name] {
             if let foundScope = cached {
                 // We have a cached result, provide it to the caller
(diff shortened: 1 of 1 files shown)


Azoy:
@swift-ci please test

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/AttributedString/AttributeScope.swift
```diff
@@ -72,7 +76,13 @@ fileprivate struct LoadedScopeCache : Sendable {
             }
         }
         
-        let type = _typeByName(name) as? any AttributeScope.Type
+        guard let handle = dlopen(path, RTLD_NOLOAD),
```
compnerd, line 79: This doesn't work for Windows.  Could you please add a conditional path for this?  This is regressing the Windows build.

compnerd (changes requested):

Sources/FoundationEssentials/AttributedString/AttributeScope.swift
Azoy: This is inside `#if FOUNDATION_FRAMEWORK` which shouldn't be getting compiled in this package.

jmschonfeld: Yep, since these various system frameworks don't exist on Windows this code is currently (and will always be) `#if`'ed out for windows

compnerd: I missed the `#if FOUNDATION_FRAMEWORK`.  Sorry for the noise!

Sources/FoundationEssentials/AttributedString/AttributeScope.swift
```diff
@@ -29,6 +29,7 @@ public enum AttributeScopes { }
 
 #if FOUNDATION_FRAMEWORK
 
+@_implementationOnly import Darwin
```
iCharlesHu, line 32: We actually need just regular `import Darwin`. I believe this will cause the "inconsistently import" warning.

Azoy:
@swift-ci please test

iCharlesHu:
@swift-ci please test

PR #117, opened 2023-05-11
Move files shared by FoundationEssentials and FoundationInternationalization to _FoundationInternals

itingliu:
- Add a folder `Sources/_FoundationInternals` to host internal files needed by both modules. Move LockedState.swift there.
- Add `Sources/_FoundationInternals` as a symlink inside both modules, so the files under `Sources/_FoundationInternals` appear as their own sources respectively.

PR #153, opened 2023-05-27
Work towards: Stop calling into Foundation's implementation 

itingliu:
We continue disentangling the package from system Foundation API in this PR

- Use stdlib's `replacing(:with:)` instead of NSString's `replacingOccurrences(of:with:)`
- Move internal implementation of `_range(of:anchored:backwards:)` to _FoundationInternals. Update callsites to use this instead of NSString's `range(of:)` implementation.

### DIFF
--- a/Sources/_FoundationInternals/String/BidirectionalCollection.swift
+++ b/Sources/_FoundationInternals/String/BidirectionalCollection.swift
@@ -61,4 +61,72 @@ extension BidirectionalCollection {
         }
         return self[startOfNonTrimmedRange ... backIdx]
     }
+
+    // Equal to calling `index(&idx, offsetBy: -other.count)` with just one loop
+    func _index<S: BidirectionalCollection>(_ index: Index, backwardsOffsetByCountOf other: S) -> Index? {
+        var idx = index
+        var otherIdx = other.endIndex
+        while otherIdx > other.startIndex  {
+            guard idx > startIndex else {
+                // other.count > self.count: bail
+                return nil
+            }
+            other.formIndex(before: &otherIdx)
+            formIndex(before: &idx)
+        }
+        return idx
+    }
+
+    func _range<S: BidirectionalCollection>(of other: S, anchored: Bool = false, backwards: Bool = false) -> Range<Index>? where S.Element == Element, Element : Equatable {
+        var result: Range<Index>? = nil
+        var fromLoc: Index
+        var toLoc: Index
+        if backwards {
+            guard let idx = _index(endIndex, backwardsOffsetByCountOf: other) else {
+                // other.count > string.count: bail
+                return nil
+            }
+            fromLoc = idx
+
+            toLoc = anchored ? fromLoc : startIndex
+        } else {
+            fromLoc = startIndex
+            if anchored {
+                toLoc = fromLoc
+            } else {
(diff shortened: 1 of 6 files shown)


Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
```diff
@@ -15,7 +15,7 @@ import TestSupport
 #endif
 
 #if canImport(FoundationEssentials)
-@testable import FoundationEssentials
```
itingliu, line 18: We made `_FoundationInternals`'s internal functions available for both `FoundationEssentials` and `FoundationInternationalization` by adding its source files as symbolic links at both target's path (#117). A side effect of this approach is that we can no longer import both targets with `@testable import` -- the test sees the declaration of these internal functions from both targets, so callsites would get an "ambiguous use of \<function\>" error when it attempts to use them.  

So we have to stop importing other targets with `@testable import`. This is probably desired anyways

Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
```diff
@@ -29,12 +29,6 @@ import TestSupport
 import _RopeModule
 #endif
 
-extension AttributedString {
-    fileprivate var string: String {
-        String(self._guts.string)
```
itingliu, line 34: Since we can no longer `@testable import FoundationEssentials`, we can no longer access `_guts`. Let's just update the test to not require this.

itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
iCharlesHu: Dammit... what if we want to test `internal` method on `FoundationEssentials`?

Sources/_FoundationInternals/String/BidirectionalCollection.swift
```diff
@@ -61,4 +61,72 @@ extension BidirectionalCollection {
...
+                // other.count > self.count: bail
+                return nil
+            }
+            other.formIndex(before: &otherIdx)
+            formIndex(before: &idx)
+        }
+        return idx
+    }
+
+    func _range<S: BidirectionalCollection>(of other: S, anchored: Bool = false, backwards: Bool = false) -> Range<Index>? where S.Element == Element, Element : Equatable {
```
iCharlesHu, line 80: Can you add some tests for this version of `_range`?

itingliu: My bad for missing that. Will do

itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/DateFormatStyleTests.swift
itingliu: We move to-be-tested to `_FoundationInternal` so that test-importing either module grants access to those internal functions? I wouldn't worry too much about it now as long as the code and tests are structured well

iCharlesHu (approved):

PR #155, opened 2023-05-27
Implicit returns from single-expression

Roshankumar350:
Proposal: [SE-0255](https://github.com/apple/swift-evolution/blob/main/proposals/0255-omit-return.md) states that single expression can be Implicitly return.

Behaviour like this 
```
func foo() -> String {
    return "foo"
}
```
can be rewritten as:
```
func foo() -> String {
    "foo"
}
```
Other Impact: 
This also make code consistent.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
@@ -62,12 +62,12 @@ extension AttributeContainer {
     }
 
     public static subscript<K: AttributedStringKey>(dynamicMember keyPath: KeyPath<AttributeDynamicLookup, K>) -> Builder<K> {
-        return Builder(container: AttributeContainer())
+        Builder(container: AttributeContainer())
     }
 
     @_disfavoredOverload
     public subscript<K: AttributedStringKey>(dynamicMember keyPath: KeyPath<AttributeDynamicLookup, K>) -> Builder<K> {
-        return Builder(container: self)
+        Builder(container: self)
     }
 
     public struct Builder<T: AttributedStringKey> : Sendable {
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+Runs+AttributeSlices.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+Runs+AttributeSlices.swift
@@ -100,11 +100,11 @@ extension AttributedString.Runs {
     }
 
     public subscript<T : AttributedStringKey>(_ keyPath: KeyPath<AttributeDynamicLookup, T>) -> AttributesSlice1<T> {
-        return AttributesSlice1<T>(runs: self)
+        AttributesSlice1<T>(runs: self)
     }
 
     public subscript<T : AttributedStringKey>(_ t: T.Type) -> AttributesSlice1<T> {
-        return AttributesSlice1<T>(runs: self)
+        AttributesSlice1<T>(runs: self)
     }
 }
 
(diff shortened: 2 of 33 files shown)


Roshankumar350:
Hello @iCharlesHu 
Please review once :)

Roshankumar350:
@swift-ci  please test

iCharlesHu:
Hi @Roshankumar350! Thanks for the PR. Similar to https://github.com/apple/swift-foundation/pull/76 , I think in the future we will definitely want to do something like this (to keep our coding style consistent), but not until we can add automatic linting to this project. It's just too easy to regress the styling (and increases the chance of merge conflict) unless everyone runs through the same linter.

Roshankumar350:
Hello @iCharlesHu ,
One last query.
There is opportunity of work where statements can be combined of some behaviour.

Like here
https://github.com/apple/swift-foundation/blob/main/Sources/FoundationEssentials/Data/ContiguousBytes.swift#L93



Can I find and work on this?

iCharlesHu:
> https://github.com/apple/swift-foundation/blob/main/Sources/FoundationEssentials/Data/ContiguousBytes.swift#L93

Hi @Roshankumar350! Sorry, I don't quite follow what you meant by "statements can be combined". Could you elaborate?

In general, we are looking for help on:
- Documentation. Following Apple's [DocC initiative](https://developer.apple.com/documentation/docc), ideally, we would want header docs on all public symbols.
- Testing. Currently our test suite is quite limited. We are looking to expand our test coverage shortly.

Roshankumar350:
Hello @iCharlesHu :
Thank you for giving us right path to contribute :)

By this "statements can be combined" I mean to say this:
```
 public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
        let element = self.first!
        return try Swift.withUnsafeBytes(of: element) {
            return try body($0)
        }
    }
   ```
Here we have opportunity to reduce one line like this:

```
 public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
        return try Swift.withUnsafeBytes(of: self.first!) {
            return try body($0)
        }
    }
    ```

iCharlesHu:
> Here we have opportunity to reduce one line like this:
> 
> ```
>  public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
>         return try Swift.withUnsafeBytes(of: self.first!) {
>             return try body($0)
>         }
>     }
>     ```
> ```

Ahh I see. I would classify this as styling changes as well -- we would probably want to wait to have linter support before we make this kind of changes.

Roshankumar350:
> > Here we have opportunity to reduce one line like this:
> > ```
> >  public func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
> >         return try Swift.withUnsafeBytes(of: self.first!) {
> >             return try body($0)
> >         }
> >     }
> >     ```
> > ```
> 
> Ahh I see. I would classify this as styling changes as well -- we would probably want to wait to have linter support before we make this kind of changes.

With that being said..  I would like to present different perspective.

lets say API `func withUnsafeBytes` is being called recursively by any developer for `N` times based on his requirement.
Now for each stack frame it will allocate extra memory for `let element = self.first!` 
So effectively it will be allocating O(n) memory. Which can be reduces by  removing extra space allocation.


cc @iCharlesHu

PR #158, opened 2023-05-28
FoundationInternationalization: add explicit type signature

compnerd:
Add an explicit type signature for the closure as it was being inferred as `() -> WeekendRange` rather than `() -> WeekendRange?` which results in a failure to build as `nil` does not conform to `WeekendRange`.

### DIFF
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1147,7 +1147,7 @@ internal final class _Calendar: Equatable, @unchecked Sendable {
     }
 
     func weekendRange() -> WeekendRange? {
-        return lock.withLock {
+        return lock.withLock { () -> WeekendRange? in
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

jmschonfeld:
> it was being inferred as () -> WeekendRange rather than () -> WeekendRange?

Is this perhaps worthy of investigating as a potential bug in the compiler? Seems like there might be a type inference bug lurking in here, unless this is just an unsupported inference case

iCharlesHu (approved):

iCharlesHu:
> > it was being inferred as () -> WeekendRange rather than () -> WeekendRange?
> 
> Is this perhaps worthy of investigating as a potential bug in the compiler? Seems like there might be a type inference bug lurking in here, unless this is just an unsupported inference case

I wouldn't be surprised if that's the case. I've hit so many weird compiler errors like "can't resolve expression" only to have them resolved by explicitly annotate the types...

parkera (approved):

PR #159, opened 2023-05-30
FoundationInternationalization: improve Windows file path handling

compnerd:
This adjusts the Windows paths to account for the handling of file paths and link resolutions.  There is no `readlink` function, particularly in the C library.  Instead, we need to open the file, and then query the I/O Object Manager to determine the final path.  This emulates the behaviour of `readlink` for Windows.

### DIFF
--- a/Sources/FoundationInternationalization/TimeZone/TimeZone_Cache.swift
+++ b/Sources/FoundationInternationalization/TimeZone/TimeZone_Cache.swift
@@ -18,6 +18,10 @@ import Glibc
 import ucrt
 #endif
 
+#if os(Windows)
+import WinSDK
+#endif
+
 #if FOUNDATION_FRAMEWORK
 @_implementationOnly import _ForSwiftFoundation
 @_implementationOnly import CoreFoundation_Private.CFNotificationCenter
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/FoundationInternationalization/TimeZone/TimeZone_Cache.swift
```diff
@@ -106,6 +110,23 @@ struct TimeZoneCache : Sendable {
                 }
             }
 
+#if os(Windows)
+            let hFile = _TimeZone.TZDEFAULT.withCString(encodedAs: UTF16.self) {
+                CreateFileW($0, GENERIC_READ, DWORD(FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE), nil, DWORD(OPEN_EXISTING), 0, nil)
```
kperryua, line 115: Do we need all these permissions? Seems like `FILE_SHARE_READ` should suffice?

compnerd: They are not file permissions - the permission being requested is `GENERIC_READ`.  The bits that you are referencing (`FILE_SHARE_*`) are file open disposition flags.  They are all needed to permit other processes to simultaneously create handles to the file with read/write/delete access.

kperryua (approved):

PR #107, opened 2023-05-10
Correct implementation of `withLockExtendingLifetimeOfState`

glessard:
The method description was wrong: the `State` parameter must outlive the locked scope, not just the call to the closure. Thanks to @phausler for the pointer.

Addresses rdar://109056170

PR #160, opened 2023-05-30
Add test to verify lifetime extension

nonsensery:
#### Summary

This PR adds a test to verify the behavior of `LockedState.withLockExtendingLifetimeOfState(_:)`.

#### Discussion

The desired behavior is that the old state object not be deallocated until after the lock is released.

During the review of #107, we did not come up with a way to test the expected behavior of this method. An idea for testing it popped into my head, so I tried it out.

This test attempts to acquire the lock when the state object is deallocated. If the lock is still held at that point, the test fails. (Unfortunately, it fails by crashing, but I don't know a way around that.)

#### Verification

I ran this test with the implementations of `withLockExtendingLifetimeOfState` from both #97 (test failed) and #107 (test passed).

### DIFF
--- a/Tests/FoundationEssentialsTests/LockedStateTests.swift
+++ b/Tests/FoundationEssentialsTests/LockedStateTests.swift
@@ -19,7 +19,14 @@ import TestSupport
 #endif
 
 final class LockedStateTests : XCTestCase {
-    final class TestObject {}
+    final class TestObject {
+        var deinitBlock: () -> Void = {}
+
+        deinit {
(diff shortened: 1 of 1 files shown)


glessard:
@swift-ci please test

glessard:
@swift-ci please test

nonsensery:
Oops, I missed a line when committing my changes. It should compile now 😅

glessard (approved):
Great idea. Thanks!

PR #161, opened 2023-05-31
Provide `Predicate.true` and `Predicate.false`

jmschonfeld:
This provides static `true` and `false` properties on `Predicate` so that shorthand syntax of `.true` and `.false` can be used in APIs that accept a predicate

Resolves rdar://109162593


jmschonfeld:
@swift-ci please test

parkera (approved):

iCharlesHu (approved):

PR #162, opened 2023-06-01
Native implementation for initializing String with UTF16 code points

itingliu:
Continue disentangling the package from system Foundation API in this PR by adding a native implementation for `String(utf16CodeUnits:count:)`

Preliminary profiling shows approximately the same performance for multi-byte strings, and 3x improvement for single-byte ones.

### DIFF
--- a/Sources/_FoundationInternals/String/String+Internals.swift
+++ b/Sources/_FoundationInternals/String/String+Internals.swift
@@ -16,4 +16,31 @@ extension String {
             $0.properties.isWhitespace
         })
     }
+
+    static func _tryFromUTF16(_ input: UnsafeBufferPointer<UInt16>) -> String? {
+        withUnsafeTemporaryAllocation(of: UInt8.self, capacity: input.count * 3) { contents in
+            var ptr = contents.baseAddress!
+            var count = 0
+            let error = transcode(input.makeIterator(), from: UTF16.self, to: UTF8.self, stoppingOnError: true) { codeUnit in
+                ptr.pointee = codeUnit
+                ptr = ptr.advanced(by: 1)
+                count += 1
+            }
+
+            guard !error else {
+                return nil
+            }
+
+            return String._tryFromUTF8(UnsafeBufferPointer(rebasing: contents[..<count]))
+        }
+    }
+
+    static func _tryFromUTF16(_ input: UnsafeMutableBufferPointer<UInt16>, len: Int) -> String? {
+        _tryFromUTF16(UnsafeBufferPointer(rebasing: input[..<len]))
+    }
+
+    static func _tryFromUTF16(_ input: UnsafePointer<UInt16>, len: Int) -> String? {
+        _tryFromUTF16(UnsafeBufferPointer(start: input, count: len))
+    }
+
 }
--- a/Sources/FoundationInternationalization/ICU/ICU+Foundation.swift
+++ b/Sources/FoundationInternationalization/ICU/ICU+Foundation.swift
@@ -54,7 +54,7 @@ internal func _withResizingUCharBuffer(initialSize: Int32 = 32, _ body: (UnsafeM
                     var innerStatus = U_ZERO_ERROR
                     if let innerLen = body(innerBuffer.baseAddress!, len + 1, &innerStatus) {
                         if innerStatus.isSuccess && innerLen > 0 {
-                            return String(utf16CodeUnits: innerBuffer.baseAddress!, count: Int(innerLen))
+                            return String._tryFromUTF16(innerBuffer, len: Int(innerLen))
                         }
                     }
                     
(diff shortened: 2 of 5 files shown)


itingliu:
@swift-ci please test

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -16,4 +16,31 @@ extension String {
             $0.properties.isWhitespace
         })
     }
+
+    static func _tryFromUTF16(_ input: UnsafeBufferPointer<UInt16>) -> String? {
+        withUnsafeTemporaryAllocation(of: UInt8.self, capacity: input.count * 3) { contents in
+            var ptr = contents.baseAddress!
+            var count = 0
+            let error = transcode(input.makeIterator(), from: UTF16.self, to: UTF8.self, stoppingOnError: true) { codeUnit in
```
iCharlesHu, line 24: Is transcode provided by the standard library?

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -16,4 +16,31 @@ extension String {
...
+
+            guard !error else {
+                return nil
+            }
+
+            return String._tryFromUTF8(UnsafeBufferPointer(rebasing: contents[..<count]))
+        }
+    }
+
+    static func _tryFromUTF16(_ input: UnsafeMutableBufferPointer<UInt16>, len: Int) -> String? {
```
iCharlesHu, line 38: (Super super nit 😅) I believe Swift prefer `count` over `length`? I know this isn't a public API but I can already see us trying to remember whether it's count or length haha

iCharlesHu (approved):
THANK YOU TINAAAAA
I believe we can now upgrade ICU to remove the hidden Foundation import. I'll get started on that.

Sources/_FoundationInternals/String/String+Internals.swift
glessard: It is.

Sources/_FoundationInternals/String/String+Internals.swift
itingliu: I went back and forth with this actually, so `count` it is!

glessard (approved):

itingliu:
@swift-ci please test

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -16,4 +16,32 @@ extension String {
...
+    static func _tryFromUTF16(_ input: UnsafeBufferPointer<UInt16>) -> String? {
+        // Allocate input.count * 3 code points since one UTF16 code point may require up to three UTF8 code points when transcoded
+        withUnsafeTemporaryAllocation(of: UInt8.self, capacity: input.count * 3) { contents in
+            var ptr = contents.baseAddress!
+            var count = 0
+            let error = transcode(input.makeIterator(), from: UTF16.self, to: UTF8.self, stoppingOnError: true) { codeUnit in
+                ptr.pointee = codeUnit
+                ptr = ptr.advanced(by: 1)
+                count += 1
+            }
```
glessard, line 29: Since `transcode()` has no awareness of the size of the temporary allocation, I would suggest adding a debug-mode check that `count` is less than `contents.count`. I think it would be useful for our future selves.

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -16,4 +16,32 @@ extension String {
             $0.properties.isWhitespace
         })
     }
+
+    static func _tryFromUTF16(_ input: UnsafeBufferPointer<UInt16>) -> String? {
+        // Allocate input.count * 3 code points since one UTF16 code point may require up to three UTF8 code points when transcoded
+        withUnsafeTemporaryAllocation(of: UInt8.self, capacity: input.count * 3) { contents in
```
glessard, line 22:
```suggestion
        withUnsafeTemporaryAllocation(of: UTF8.CodeUnit.self, capacity: input.count * 3) { contents in
```
Extremely minor nit.

itingliu:
@swift-ci please test

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -16,4 +16,30 @@ extension String {
             $0.properties.isWhitespace
         })
     }
+
+    static func _tryFromUTF16(_ input: UnsafeBufferPointer<UInt16>) -> String? {
```
parkera, line 20: Nitpick about the name. `try` means something specific in Swift, so I feel we should just name this something like `init?(fromUTF16:)` if possible.

kperryua: Agreed, but sounds like we're going for parity with the already extant `_tryFromUTF8` from the stdlib?

glessard: I think that's it, yes. If we were to make it an `internal init` instead, it should have an underscored argument label.

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #165, opened 2023-06-06
(110353463) Support converting raw Data values in NSPredicate conversion

jmschonfeld:
When a `Predicate` captures a `Data` value, we should insert that into the converted `NSPredicate`


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #170, opened 2023-06-07
Date: protect against invalid use of `_gmtime_s`

compnerd:
In the case that `_gmtime_s` fails, the return value is initialised to an invalid value that will fail when invoking `_strftime`.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #171, opened 2023-06-07
Update API naming of PredicateExpressions.ForcedUnwrap

jmschonfeld:
This updates the API of `PredicateExpressions.ForcedUnwrap` to use `Inner` instead of `Wrapped` and `lhs:`. Since this is newer API that we don't expect anyone to be using yet, we can safely make this change without causing too much disruption

Resolves rdar://110119524


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #172, opened 2023-06-07
Guard an optional String before using it in String interpolation

itingliu:
`String(_utf16:count:)` returns an optional String. Guard it before using it.


itingliu:
@swift-ci please test

jmschonfeld (approved):
Thanks!

PR #174, opened 2023-06-09
Fix StringProtocol.lineRange(for:)/paragraphRange(for:)

lorentey:
These started by converting `self` into a `String`. This lead to crashes when `self` happened to be a Substring, as the input range isn’t necessarily a valid index range in the resulting string in that case.

While I’m here, review related helper functions to improve performance and to avoid more correctness issues.

Resolves rdar://109907151&110477287

### DIFF
--- a/Sources/FoundationEssentials/String/StringBlocks.swift
+++ b/Sources/FoundationEssentials/String/StringBlocks.swift
@@ -181,3 +181,26 @@ extension BidirectionalCollection where Element == UTF8.CodeUnit {
         return _StringBlock(start: start, end: end, contentsEnd: contentsEnd)
     }
 }
+
+extension Substring.UTF8View {
+    // Note: we could have these defined on BidirectionalCollection<UInt8>, like _getBlock.
+    // However, all current call sites are funneling through Substring.UTF8View, and it makes
+    // sense to avoid dealing with generics unless we're forced to. Feel free to convert these
+    // utilities to generic functions if needed.
+
+    internal func _lineBounds(
+        around range: Range<Index>
+    ) -> (start: Index, end: Index, contentsEnd: Index) {
+        let result = _getBlock(
+            for: [.findStart, .findEnd, .findContentsEnd, .stopAtLineSeparators], in: range)
+        return (result.start!, result.end!, result.contentsEnd!)
+    }
+
+    internal func _paragraphBounds(
+        around range: Range<Index>
+    ) -> (start: Index, end: Index, contentsEnd: Index) {
+        let result = _getBlock(
+            for: [.findStart, .findEnd, .findContentsEnd], in: range)
+        return (result.start!, result.end!, result.contentsEnd!)
+    }
+}
--- a/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
+++ b/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
@@ -149,54 +149,96 @@ extension StringProtocol {
     /// Returns the range of characters representing the line or lines
     /// containing a given range.
     @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
-    public func lineRange<R : RangeExpression>(for aRange: R) -> Range<Index> where R.Bound == Index {
-        return String(self).lineRange(for: aRange)
+    public func lineRange(for range: some RangeExpression<Index>) -> Range<Index> {
+        let r = _lineBounds(around: range)
+        return r.start ..< r.end
     }
 
     /// Returns the range of characters representing the
     /// paragraph or paragraphs containing a given range.
     @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
-    public func paragraphRange<R : RangeExpression>(for aRange: R) -> Range<Index> where R.Bound == Index {
-        return String(self).paragraphRange(for: aRange)
+    public func paragraphRange(for range: some RangeExpression<Index>) -> Range<Index> {
+        let r = _paragraphBounds(around: range)
+        return r.start ..< r.end
     }
 }
 
-extension String {
-    internal func lineRange<R : RangeExpression>(for aRange: R) -> Range<Index> where R.Bound == Index {
+extension StringProtocol {
+    @inline(never)
+    internal func _lineBounds(
+        around range: some RangeExpression<Index>
+    ) -> (start: Index, end: Index, contentsEnd: Index) {
+        // Avoid generic paths in the common case by manually specializing on `String` and
+        // `Substring`. Note that we're only ever calling `_lineBounds` on a `Substring`; this is
+        // to reduce the code size overhead of having to specialize it multiple times (at a slight
+        // cost to runtime performance).
+        if let s = _specializingCast(self, to: String.self) {
+            let range = s.unicodeScalars._boundaryAlignedRange(range)
+            return s[...].utf8._lineBounds(around: range)
+        } else if let s = _specializingCast(self, to: Substring.self) {
+            let range = s.unicodeScalars._boundaryAlignedRange(range)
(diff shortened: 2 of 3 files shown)


lorentey:
@swift-ci test

Azoy (approved):
These changes look good to me 👍

Sources/FoundationEssentials/String/StringBlocks.swift
```diff
@@ -181,3 +181,26 @@ extension BidirectionalCollection where Element == UTF8.CodeUnit {
...
+    // However, all current call sites are funneling through Substring.UTF8View, and it makes
+    // sense to avoid dealing with generics unless we're forced to. Feel free to convert these
+    // utilities to generic functions if needed.
+
+    internal func _lineBounds(
+        around range: Range<Index>
+    ) -> (start: Index, end: Index, contentsEnd: Index) {
+        let result = _getBlock(
+            for: [.findStart, .findEnd, .findContentsEnd, .stopAtLineSeparators], in: range)
+        return (result.start!, result.end!, result.contentsEnd!)
```
itingliu, line 196: oh yeah, you're right that we can force unwrap these indices here and below since it should match the options passed into `_getBlock`

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
```diff
@@ -149,54 +149,96 @@ extension StringProtocol {
...
+            // collections before & after the _lineBounds call.
+            let range = self.unicodeScalars._boundaryAlignedRange(range)
+
+            let startUTF8Offset = self.utf8.distance(from: self.startIndex, to: range.lowerBound)
+            let utf8Count = self.utf8.distance(from: range.lowerBound, to: range.upperBound)
+
+            let s = Substring(self)
+            let start = s.utf8.index(s.startIndex, offsetBy: startUTF8Offset)
+            let end = s.utf8.index(start, offsetBy: utf8Count)
+            let r = s.utf8._lineBounds(around: start ..< end)
```
itingliu, line 193: Can you explain more why we need to recalculate `start` and `end`? I thought `range` would be scalar aligned after `_boundaryAlignedRange` on line 185, so it'd also be utf8 aligned, so it'd be safe to use with `s.utf8`. What am I missing?

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
```diff
@@ -149,54 +149,96 @@ extension StringProtocol {
...
+
+            let resultUTF8Offsets = (
+                start: s.utf8.distance(from: s.startIndex, to: r.start),
+                end: s.utf8.distance(from: s.startIndex, to: r.end),
+                contentsEnd: s.utf8.distance(from: s.startIndex, to: r.contentsEnd))
+            return (
+                start: self.utf8.index(self.startIndex, offsetBy: resultUTF8Offsets.start),
+                end: self.utf8.index(self.startIndex, offsetBy: resultUTF8Offsets.end),
+                contentsEnd: self.utf8.index(self.startIndex, offsetBy: resultUTF8Offsets.contentsEnd))
```
itingliu, line 202: Following the above question -- Do we still need this if we already sanitize the indices before passing them to `_lineBounds(around:)`? Looking at the underlying `_getBlock`'s implementation, I think the returned indices should already be properly aligned if the indices passed to are already utf8-aligned. 

If it's still needed, do you think it would be reasonable to sink this index adjustment into `_lineBounds`'s implementation? The idea is that if the client uses `_lineBounds` on a utf8 view, they can be assured that the returned indices would be utf8-aligned. This way we can consolidate this and that in `_paragraphBounds(:)`

Sources/FoundationEssentials/String/StringBlocks.swift
lorentey: Yep! (I think at some point it would be worth splitting _getBlock up into separate functions (one each for the findStart/findEnd cases) to eliminate these optional results.)

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
lorentey: The issue here is that `s` is a different collection from `self`, so the indices we calculated in `s` aren't valid in `self` -- sadly we need to recalculate them from scratch using their UTF-8 offsets.

This whole third case is an exercise in frustration. 🙈 `StringProtocol.UTF8View` isn't guaranteed to be bidirectional (presumably to work around a temporary compiler bug 6-8 years ago, which we forgot to undo before setting things in stone), so we cannot call the generic `_getBounds` on it. And we do not expect or want to ever execute this case, so any and all code size we spend on it is pretty much pure waste. 😖

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
lorentey: Yes! These lines aren't for aligning the indices; rather, they are finding the positions in one collection that correspond to the positions we constructed in another.

(Indices are like pointers -- they are inherently tied to the particular collection instance that vended them, and aren't generally valid in any other. `String.Index` is mostly just an offset into UTF-8 (or UTF-16) storage; however, this offset value can be different between `self` and `s`, rendering index values incompatible between them. (For example, `self` could be using UTF-16 indices while `s` is UTF-8, or `self` can be a substring-like `StringProtocol` type where the `startIndex` isn't at offset zero. On this code path we already know the client is doing something weird, because we encountered a `StringProtocol` type that isn't `String` nor `Substring` -- we cannot safely assume anything about the internal structure of it.)

Sources/_FoundationInternals/String/BidirectionalCollection.swift
```diff
@@ -12,19 +12,16 @@
 
 extension BidirectionalCollection where Index == String.Index {
     internal func _alignIndex(roundingDown i: Index) -> Index {
-        return i < endIndex ? index(before: index(after: i)) : i
+        index(i, offsetBy: 0)
```
lorentey, line 15: This is a subtle change:

1.  We know we aren't running on a stdlib predating Swift 5.7, so `index(i, offsetBy: 0)` can safely be used as a substitute for `_index(roundingDown:)` -- it does the same thing, more efficiently than stepping forward-and-back.
2. If we're at the end, this returns the rounded off value, not the original input. (On the off chance someone tried to apply an unaligned index beyond the end index.

Sources/_FoundationInternals/String/BidirectionalCollection.swift
```diff
@@ -12,19 +12,16 @@
...
 
     internal func _alignIndex(roundingUp i: Index) -> Index {
         let truncated = _alignIndex(roundingDown: i)
-        if i > truncated && i < endIndex {
-            return index(after: i)
-        } else {
-            return i
-        }
+        guard i > truncated && truncated < endIndex else { return truncated }
+        return index(after: truncated)
```
lorentey, line 21: This change ensures we never return `i` as is without giving the underlying collection a chance to properly align it (and set performance flags etc.), like in point 2 above.

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
itingliu: Ah I see. This is to convert the indices of the `Substring` we constructed back to those of `self`. It has nothing to do with alignment.

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
itingliu: Ah yeah I somehow totally overlooked that we're executing this on this newly constructed `Substring`, not `self`. That makes sense. Thanks!

(This whole third case is sort of unavoidable when we added this function to  `StringProtocol` -- we basically already overpromised what we could do)

itingliu (approved):
Thanks for thorough explanation!

PR #179, opened 2023-06-14
Implement Date.description in pure Swift

FranzBusch:
# Motivation

The current implementation of `Date.description` is depending on the what platform we are running on. On Darwin it used the CFFoundation backed implementation and on Linux/Windows it used libc APIs to do the date calculations. This caused build issues on Windows and the libc APIs are not 32bit clean.

# Changes

This PR changes the implementation by using the methods implemented in swift-certificates that provide timestamp to UTC time calculations. These were added in swift-certificates for the exact same reason. Furthermore, it removes the conditional usage of CFFoundation since the output is the same.

# Result

Swift only implementation that works cross platform for `Date.description` which also should perform better than any libc API calls.


FranzBusch:
cc @compnerd This should fix your Windows issue hopefully

iCharlesHu:
@swift-ci please test

Sources/FoundationEssentials/TimeCalculations.swift
```diff
@@ -0,0 +1,292 @@
...
+
+        // Normalise out the values.
+        //
+        // Safe to do unchecked math on the months as we just checked its value above.
+        // Same for remdays, the loop only terminates if the number is smaller than at most 31.
+        //
+        // Note that, unlike struct tm, we return ordinal month numbers as well as days (i.e. 1 to 12).
+        // This fits us better when working with generalizedtime and friends.
+        return (
+            year: Int(years + 2000),
```
glessard, line 175: It seems like this should be years + 2031.

Sources/FoundationEssentials/Date.swift
```diff
@@ -260,36 +249,12 @@ extension Date : CustomDebugStringConvertible, CustomStringConvertible, CustomRe
...
-                return unavailable
-            }
-
-            guard let result = String(validatingUTF8: ptr) else {
-                return unavailable
-            }
-
-            return result
-        }
+        return "\(year)-\(month)-\(day) \(hours):\(minutes):\(seconds) +0000"
```
glessard, line 256: These string interpolations need to be padded with zeroes (0000-00-00 00:00:00 +0000)

compnerd:
Oddly enough, it does worse:

```
Build complete! (3.51s)
Test Suite 'Selected tests' started at 2023-06-14 20:16:52.729
Test Suite 'DateTests' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDateComparison' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDateComparison' passed (0.0 seconds)
Test Case 'DateTests.testDateMutation' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDateMutation' passed (0.0 seconds)
Test Case 'DateTests.testDescription1970' started at 2023-06-14 20:16:52.729
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:90: error: DateTests.testDescription1970 : XCTAssertEqual failed: ("1970-01-01 00:00:00 +0000") is not equal to ("1939-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescription1970' failed (0.0 seconds)
Test Case 'DateTests.testDescriptionBeyondDistantFuture' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDescriptionBeyondDistantFuture' passed (0.0 seconds)
Test Case 'DateTests.testDescriptionBeyondDistantPast' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDescriptionBeyondDistantPast' passed (0.0 seconds)
Test Case 'DateTests.testDescriptionDistantFuture' started at 2023-06-14 20:16:52.729
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:102: error: DateTests.testDescriptionDistantFuture : XCTAssertEqual failed: ("4001-01-01 00:00:00 +0000") is not equal to ("3970-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionDistantFuture' failed (0.0 seconds)
Test Case 'DateTests.testDescriptionDistantPast' started at 2023-06-14 20:16:52.729
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:97: error: DateTests.testDescriptionDistantPast : XCTAssertEqual failed: ("0000-12-30 00:00:00 +0000") is not equal to ("-31-12-30 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionDistantPast' failed (0.0 seconds)
Test Case 'DateTests.testDescriptionReferenceDate' started at 2023-06-14 20:16:52.729
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:84: error: DateTests.testDescriptionReferenceDate : XCTAssertEqual failed: ("2001-01-01 00:00:00 +0000") is not equal to ("1970-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionReferenceDate' failed (0.0 seconds)
Test Case 'DateTests.testDistantFuture' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDistantFuture' passed (0.0 seconds)
Test Case 'DateTests.testDistantPast' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.testDistantPast' passed (0.0 seconds)
Test Case 'DateTests.test_now' started at 2023-06-14 20:16:52.729
Test Case 'DateTests.test_now' passed (0.0 seconds)
Test Suite 'DateTests' failed at 2023-06-14 20:16:52.729
         Executed 11 tests, with 4 failures (0 unexpected) in 0.0 (0.0) seconds
Test Suite 'Selected tests' failed at 2023-06-14 20:16:52.729
         Executed 11 tests, with 4 failures (0 unexpected) in 0.0 (0.0) seconds
```

Sources/FoundationEssentials/TimeCalculations.swift
FranzBusch: I don't think that's it. I think the reference dates are just different here and I gotta fix that up; however, I am currently struggling to get the test to run locally.

FranzBusch:
@swift-ci please test

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -54,3 +54,12 @@ extension String {
     }
 
 }
+
+extension DefaultStringInterpolation {
```
FranzBusch, line 58: I am not really happy with this. I think we can do better here. Are we planning to add the `String(format)` init in this package?

FranzBusch:
With the latest changes I think we only have a problem on `distantPast` we are not getting `0000-01-01` but `0000-12-30`. I guess we are missing some leap days here. @Lukasa is that expected from the implementation?

Lukasa:
That would have to be a _lot_ of leap days to end up adding up to 12 months. What's the numerical value of `distantPast`?

Sources/_FoundationInternals/String/String+Internals.swift
glessard: Ultimately we will have access to all the interpolation-based formatters, so this seems like an appropriate approach.

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -54,3 +54,12 @@ extension String {
...
+    mutating func appendInterpolation(_pad value: Int, _toWidth width: Int, _using paddingString: String) {
+        for i in 0...(String(value).count - 1) {
+            appendLiteral(paddingString)
+        }
+        appendInterpolation(value)
+    }
```
glessard, line 64:
```suggestion
    mutating func appendInterpolation(_zeroPad value: Int, _toWidth width: Int) {
      precondition(width > 0)
      let representation = String(value)
      let count = width &- representation.utf8.count
      if count > 0 {
        appendLiteral(String(repeating: "0", count: count))
      }
      appendLiteral(representation)
    }
```
For a slightly simpler call site. If we really want to have the padding as a parameter, it should be a `Character`.

itingliu:
As far as I know `gmtime` doesn't seem to promise correctness before Gregorian calendar came into effect. That would also explain why we're only seeing test failures on `distantPast`, which *is* before Gregorian.

parkera:
It may be worth it to just verify that `distantPast` (and future) exactly matches an existing output just by special casing it.

NOTICE.txt
```diff
@@ -0,0 +1,43 @@
...
+* https://github.com/apple/swift-certificates
+
+---
+
+This product contains code to calculate and decompose UNIX timestamps derived from musl libc.
+
+* LICENSE (MIT):
+* https://git.musl-libc.org/cgit/musl/tree/COPYRIGHT
+* HOMEPAGE:
+* https://musl.libc.org
```
parkera, line 43: It doesn't seem standard to have a NOTICE.txt file in our other Swift repos. We have a license, which is reviewed for correctness.

Sources/FoundationEssentials/TimeCalculations.swift
```diff
@@ -0,0 +1,292 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Collections open source project
```
parkera, line 3: This doesn't seem correct

itingliu:
As I chat with @FranzBusch I realized that this would be a huge behavioral change for Darwin users who don't use Gregorian calendar -- previously we use the calendar of their current locale, but now we'd be using Gregorian unconditionally. I personally don't think it's a good idea, but I'm curious to hear what others think.

Also we took a look into CFDateFormatter's implementation, and it did look dates before Oct 15, 1582 are formatted with Julian calendar. I think it's fine to change the description for these dates to be in Gregorian if we can just clarify this in comments.

NOTICE.txt
FranzBusch: We do have a `NOTICE.txt` file in a bunch of repos for when we use other work or have to attribute it. Some examples:

- [swift-certificates](https://github.com/apple/swift-certificates/blob/main/NOTICE.txt)
- [swift-nio](https://github.com/apple/swift-nio/blob/main/NOTICE.txt)
- [swift-cassandra-client](https://github.com/apple/swift-cassandra-client/blob/main/NOTICE.txt)

Sources/FoundationEssentials/TimeCalculations.swift
FranzBusch: You are right. I copied the header from another file and it turns out that a bunch of files in here have the wrong copyright header.

parkera:
I think the main risk to changing the behavior would be if some app:

1. Puts `date.description` directly in a UI
2. The user of that app has a non-gregorian Calendar set
3. This change results in an apparent regression in behavior, although of course the app should have been using a formatter or format style in the first place

If someone was relying on programmatically parsing the value, then they would have likely failed anyway when the user had a non-gregorian calendar set.

It does also regress the behavior for a developer debugging their app using a non-gregorian calendar, in that the dates they see in their app will not match the ones they get when `po`ing in lldb.

itingliu:
> t does also regress the behavior for a developer debugging their app using a non-gregorian calendar, in that the dates they see in their app will not match the ones they get when `po`ing in lldb.

I don't worry too much about breaking app UI. But this second point is what I was concerned about. It would be a sad regression in developer experience. Would it be possible to ask lldb to format it differently themselves since they're already doing so for other types?

parkera:
That's possible, and this is actually one of the best types to do that for since it is a very simple memory layout. I'm not sure if lldb has the logic to format dates for other calendars, though, and I doubt they would also be able to read the AppleLocale default to know what the right setting is...

FranzBusch:
I leave this up to you two here, but if developer experience is really the only thing we are concerned about here than I am sure they can understand if they have to use a `DateFormatter` here. UTC formatted time is pretty common for developers around the world.

I am happy to change the PR to add back the conditional for Darwin here if that makes it easier. It will just mean we have differing behaviour between the platforms. @parkera @itingliu your call :)

itingliu (commented):
I know I'm back and forth on this -- but now I'm convinced that the performance gain from this change might outweigh the description being unlocalised. Let's give it a shot!

Sources/_FoundationInternals/String/String+Internals.swift
```diff
@@ -54,3 +54,15 @@ extension String {
     }
 
 }
+
+extension DefaultStringInterpolation {
+    mutating func appendInterpolation(_zeroPad value: Int, _toWidth width: Int) {
+      precondition(width > 0)
+      let representation = String(value)
+      let padding = width &- representation.utf8.count
```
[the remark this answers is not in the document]
FranzBusch: Moved it to the `Date` file itself and made it `fileprivate`

itingliu:
@swift-ci please test

itingliu (commented):
This test failure looks legitimate:

```
[587/689] Compiling FoundationEssentials TimeCalculations.swift
/build/swift-foundation/Sources/FoundationEssentials/TimeCalculations.swift:151:21: error: type 'Int64' has no member 'daysInMonth'
        while Int64.daysInMonth(months) <= remdays {
              ~~~~~ ^~~~~~~~~~~
```

FranzBusch:
@compnerd Could you test the latest commit of my branch on Windows?

FranzBusch:
@swift-ci please test

compnerd:
@FranzBusch sorry for the delay, I was travelling.  This doesn't seem to address all the problems:

I had to apply the following patch:

```
diff --git a/Sources/FoundationEssentials/TimeCalculations.swift b/Sources/FoundationEssentials/TimeCalculations.swift
index 481682a..62a4aed 100644
--- a/Sources/FoundationEssentials/TimeCalculations.swift
+++ b/Sources/FoundationEssentials/TimeCalculations.swift
@@ -148,8 +148,8 @@ extension Int64 {
         var years = remyears + (4 &* qCycles) + (100 &* cCycles) + (400 &* qcCycles)
 
         var months = 0
-        while daysInMonth(months) <= remdays {
-            remdays -= daysInMonth(months)
+        while Int64.daysInMonth(months) <= remdays {
+            remdays -= Int64.daysInMonth(months)
 
             // Unchecked because daysInMonth will crash if given a value greater than 12, so this
             // cannot exceed 12.

```

```
Test Suite 'Selected tests' started at 2023-07-05 17:37:54.430
Test Suite 'DateTests' started at 2023-07-05 17:37:54.435
Test Case 'DateTests.testDateComparison' started at 2023-07-05 17:37:54.435
Test Case 'DateTests.testDateComparison' passed (0.002 seconds)
Test Case 'DateTests.testDateMutation' started at 2023-07-05 17:37:54.437
Test Case 'DateTests.testDateMutation' passed (0.0 seconds)
Test Case 'DateTests.testDescription1970' started at 2023-07-05 17:37:54.437
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:90: error: DateTests.testDescription1970 : XCTAssertEqual failed: ("1970-01-01 00:00:00 +0000") is not equal to ("1939-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescription1970' failed (0.001 seconds)
Test Case 'DateTests.testDescriptionBeyondDistantFuture' started at 2023-07-05 17:37:54.438
Test Case 'DateTests.testDescriptionBeyondDistantFuture' passed (0.0 seconds)
Test Case 'DateTests.testDescriptionBeyondDistantPast' started at 2023-07-05 17:37:54.438
Test Case 'DateTests.testDescriptionBeyondDistantPast' passed (0.0 seconds)
Test Case 'DateTests.testDescriptionDistantFuture' started at 2023-07-05 17:37:54.439
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:102: error: DateTests.testDescriptionDistantFuture : XCTAssertEqual failed: ("4001-01-01 00:00:00 +0000") is not equal to ("3970-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionDistantFuture' failed (0.0 seconds)
Test Case 'DateTests.testDescriptionDistantPast' started at 2023-07-05 17:37:54.439
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:97: error: DateTests.testDescriptionDistantPast : XCTAssertEqual failed: ("0000-12-30 00:00:00 +0000") is not equal to ("-31-12-30 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionDistantPast' failed (0.0 seconds)
Test Case 'DateTests.testDescriptionReferenceDate' started at 2023-07-05 17:37:54.439
S:\SourceCache\swift-foundation\Tests\FoundationEssentialsTests\DateTests.swift:84: error: DateTests.testDescriptionReferenceDate : XCTAssertEqual failed: ("2001-01-01 00:00:00 +0000") is not equal to ("1970-1-1 0:0:0 +0000") -
Test Case 'DateTests.testDescriptionReferenceDate' failed (0.001 seconds)
Test Case 'DateTests.testDistantFuture' started at 2023-07-05 17:37:54.440
Test Case 'DateTests.testDistantFuture' passed (0.0 seconds)
Test Case 'DateTests.testDistantPast' started at 2023-07-05 17:37:54.440
Test Case 'DateTests.testDistantPast' passed (0.001 seconds)
Test Case 'DateTests.test_now' started at 2023-07-05 17:37:54.441
Test Case 'DateTests.test_now' passed (0.0 seconds)
Test Suite 'DateTests' failed at 2023-07-05 17:37:54.441
         Executed 11 tests, with 4 failures (0 unexpected) in 0.005 (0.005) seconds
Test Suite 'Selected tests' failed at 2023-07-05 17:37:54.441
         Executed 11 tests, with 4 failures (0 unexpected) in 0.005 (0.005) seconds
```

FranzBusch:
@compnerd Interesting. Not really sure where this difference is coming from since the Swift code is fully platform independent now. I sadly, don't have a Windows machine where I can reproduce this right now.

parkera:
See also #252

FranzBusch:
@swift-ci please test

parkera:
We're getting closer to being able to just use Foundation's own date/time calculations for this.

FranzBusch:
> We're getting closer to being able to just use Foundation's own date/time calculations for this.

Great to hear. Maybe we can re-use this code then for the ICU-free time calculations.

parkera:
@FranzBusch What do you think about refactoring this on top of the newly landed Gregorian calendar and ISO8601 formatting changes?

FranzBusch:
@parkera I updated the code to use the new `ISO801` formatter in `FoundationEssentials` and it works; however, the expected format is slightly different

> ("4001-01-01 00:00:00 +0000") is not equal to ("4001-01-01 00:00:00Z")

I don't see an option to configure the time zone formatting to align with the current formatting of `Date.description`. We have three options here:
- Break the format of `Date.description` slightly
- Copy the formatting code from the ISO8601 formatter and simplify/adapt it for `Date.description`
- Extend the `ISO801` format styles (this probably needs an evolution proposal)

What would you like to do?

itingliu:
According to [spec](https://en.wikipedia.org/wiki/ISO_8601#Time_zone_designators)
> If the time is in [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), add a Z directly after the time without a space...(omitted) A time zone where the civil time coincides with UTC is always designated as positive, though the offset is zero (see related specifications below). So the zone designation for London (on standard time) would be "+00:00", "+0000", or "+00".

But currently I don't think we distinguish between UTC and other GMT+0 timezones. These timezones would compare equal to each other
```swift 
 17> TimeZone(identifier: "GMT+0") == TimeZone(identifier: "UTC")
$R6: Bool = true
```
This means that we cannot tell from the time zone whether we should use "Z" or use "+0000".  And since we cannot tell those apart, I'm inclined to just expose the option directly publicly. But in the short run I think it's ok to expose an internal setting for `Date.description`

parkera:
I agree with @itingliu -- we can add an internal-only option to the format style to prefer `+0000` to `Z`. If it's interesting for other use cases beyond our one esoteric one we could pitch it as public API too.

PR #180, opened 2023-06-14
Introduce Locale.identifierCapturingPreferences to capture preferences in a single identifier

itingliu:
- Part 1: Clean up various usages of force-12-hour and force-24-hour
Consolidate `force24Hour` and `force12Hour` to `forceHourCycle`. Use this as the universal override option.
    
- Part 2: Introduce `identifierCapturingPreferences`
Add `Locale.identifierCapturingPreferences` to return a string identifier with preferences encoded in. This allows us to pass this identifier to ICU functions to request overrides rather than specifying options in individual API. This PR also adopts this identifier for `ICUNumberFormatter`.

Work towards resolving rdar://110211953

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
@@ -45,6 +45,7 @@ internal final class _Locale: Sendable, Hashable {
         var subdivision: Locale.Subdivision??
         var timeZone: TimeZone??
         var variant: Locale.Variant??
+        var identifierCapturingPreferences: String?
 
         // If the key is present, the value has been calculated (and the result may or may not be nil).
         var identifierDisplayNames: [String : String?] = [:]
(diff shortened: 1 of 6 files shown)


itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Locale/Locale_ICU.swift
```diff
@@ -399,6 +400,46 @@ internal final class _Locale: Sendable, Hashable {
         }
     }
 
+    internal var identifierCapturingPreferences: String {
```
parkera, line 403: This might be worth a comment explaining that it's not literally all preferences, but a subset of them which can be represented in the identifier. For example we don't put temperature unit in here.

parkera (approved):
Overall looks good to me.

itingliu:
@swift-ci please test

PR #181, opened 2023-06-15
Implement getting calendar symbols

itingliu:
We should include the calendar identifier via the "ca" keyword when getting calendar symbols from ICU, such as "en_US@ca=hebrew". It was mistakenly dropped when we rewrote Calendar, so now we always return the symbols of the default calendar of the locale.

We can fix this by updating the existing implementation with the correct identifier, but we can also just use `ICUDateFormatter`, bypassing `CFDateFormatter` altogether. We chose the latter approach here.

Fixes rdar://110585721

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
@@ -191,6 +191,24 @@ final class ICUDateFormatter {
         })
     }
 
+    // MARK: - Getting symbols
+
+    func symbols(for key: UDateFormatSymbolType) -> [String] {
+        let symbolCount = udat_countSymbols(udateFormat, key)
+        var result = [String]()
+        for i in 0 ..< symbolCount {
+            let s = _withResizingUCharBuffer { buffer, size, status in
+                udat_getSymbols(udateFormat, key, i, buffer, size, &status)
+            }
+
+            if let s {
+                result.append(s)
+            }
+        }
+
+        return result
+    }
+
     // -- Caching support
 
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
```diff
@@ -191,6 +191,24 @@ final class ICUDateFormatter {
         })
     }
 
+    // MARK: - Getting symbols
+
+    func symbols(for key: UDateFormatSymbolType) -> [String] {
+        let symbolCount = udat_countSymbols(udateFormat, key)
```
parkera, line 197: It looks like we cache the formatter, but do we also need to cache this result? Calling it in a loop and checking if it's cached inside ICU itself would be a good idea.

itingliu: I thought about it too but didn't pursue because it was already much better than what we had before, at which time we didn't even cache the formatter. But now that you mentioned it we might as well do it while we're here

itingliu: I'm actually not seeing any stack trace inside ICU. It looks like the counts and the symbols are loaded as constants upon initializing a `udat`, i.e. the known-to-be-expensive operation. The most expensive part of this function lies in `_withResizingUCharBuffer` due to utf-16 to utf-8 transcoding. I think we can forgo this cache here.

parkera (approved):

PR #182, opened 2023-06-16
Restore support of deprecated `DateComponent.week` when calculating date

itingliu:
We had been supporting `DateComponent.week` in Calendar's date calculation even though it's deprecated until recently. Let's keep doing that for the sake of compatibility.

Resolves rdar://110561999


itingliu:
@swift-ci please test

kperryua (approved):

PR #185, opened 2023-06-23
canonicalLocaleIdentifierFromString crashes with invalid input

itingliu:
`CFLocaleCreateCanonicalLocaleIdentifierFromString` may return nil, so it crashes when we force cast it to `String`. We're already handling this properly in the internal function, so just call into that one.

Resolves rdar://110547526


itingliu:
@swift-ci  please test

itingliu:
@swift-ci test

parkera (approved):

PR #191, opened 2023-07-05
Add support for non-Codable nil literals in Predicates

jmschonfeld:
Adds support for `nil` literals that are unconditionally `Sendable & Codable` in predicates even when the wrapped value type is not `Codable` or `Sendable`.

Resolves rdar://111180725


jmschonfeld:
@swift-ci please test

glessard (approved):

iCharlesHu (approved):

iCharlesHu:
It looks like they fixed the snapshot? 🎉

PR #192, opened 2023-07-06
Only use week if weekOfYear is missing

parkera:
Fixes a regression introduced when rewriting Calendar in Swift -- `CFCalendar` entry points expect this code to only use *one* of week of year or (deprecated) week.

Resolves rdar://111791782


parkera:
@swift-ci test

itingliu (approved):
Oops. Thanks!

PR #193, opened 2023-07-06
Allow nullable result for localizedString(forIdentifier:)

parkera:
Some CFLocale API require the result of `localizedString(forIdentifier:)` to remain `nullable`, even though the `NSLocale` entry point is non-nullable. Provide a mechanism to preserve compatibility.

Resolves rdar://110834805


parkera:
@swift-ci test

itingliu (approved):

PR #197, opened 2023-07-10
Predicate support for `RangeExpression.contains`

jmschonfeld:
This adds support for the `RangeExpression.contains` function. Previously, basic range expressions were supported via `Sequence.contains`, but this operator will support any range expression, including bounds whose strides are not `SignedInteger` such as `Date` or `Double`


jmschonfeld:
@swift-ci please test

glessard (approved):

Sources/FoundationEssentials/Predicate/Expressions/Range.swift
```diff
@@ -65,3 +65,106 @@ extension PredicateExpressions.Range : Codable where LHS : Codable, RHS : Codabl
 
 @available(macOS 9999, iOS 9999, tvOS 9999, watchOS 9999, *)
 extension PredicateExpressions.Range : Sendable where LHS : Sendable, RHS : Sendable {}
+
+@available(macOS 9999, iOS 9999, tvOS 9999, watchOS 9999, *)
+extension PredicateExpressions {
+    public struct ClosedRange<
```
jmschonfeld, line 71: I decided to delete the separate `ClosedRange.swift` file and just consolidate all range operators into this file

Sources/FoundationEssentials/Predicate/Expressions/Sequence.swift
```diff
@@ -123,6 +123,7 @@ extension PredicateExpressions {
         }
     }
     
+    @_disfavoredOverload
```
jmschonfeld, line 126: This is used to disambiguate between `Sequence.contains` and `RangeExpression.contains` - in this case, we prefer the more specific implementation provided by `RangeExpression`, just as the compiler would if you were calling `.contains` on an instance

Sources/FoundationEssentials/Predicate/KeyPath+Inspection.swift
```diff
@@ -53,8 +55,9 @@ extension AnyKeyPath {
         case 1: // struct/tuple/self stored property
             fallthrough
         case 3: // class stored property
-            // Stored property components are each one word
-            if header._keyPathHeader_bufferSize > Self.WORD_SIZE {
+            // Stored property components are either just the payload, or the payload plus 32 bits if the payload is the sentinel value
+            let size = (firstComponentHeader._keyPathComponentHeader_payload == .PROPERTY_OFFSET_TOO_LARGE) ? MemoryLayout<UInt64>.size : MemoryLayout<UInt32>.size
+            if header._keyPathHeader_bufferSize > size {
```
jmschonfeld, line 60: This is a bug fix that our tests started hitting with this change. For stored properties, there is always a component header (32 bits) and if the offset index doesn't fit inside of the header, the header is followed by a second 32 bits that contain the offset. Additionally, components are all word-aligned. Previously, we just checked if the buffer size was greater than one word (because on 64 bit platforms a single component is either 32 bits of header + 32 bits of padding or 32 bits of header and 32 bits of offset). But on watchOS which is 64_32, a word is only 32 bits so a single component is either just 32 bits of header (no padding) or 32 bits of header + 32 bits of offset. The updated arithmetic here accounts for this by comparing to 64/32 bits depending on whether the extra 32 bits of offset is used rather than comparing to word size.

PR #198, opened 2023-07-11
`Substring._rangeOf` fast path is incorrect

itingliu:
We had a fast path implemented incorrectly: We search the UTF8 code unit whenever possible, and convert the found UTF8 index back to String index. If the conversion fails, i.e. when the found UTF8 index does not fall on a character boundary, we simply bail and return nil. The correct implementation should be advancing the slice from the point of failure and keeping searching.

But we don't really need this fast path here as we already have another one that skips string transformation inside the helper function. Let's just remove this.


parkera (approved):

itingliu:
@swift-ci please test

PR #200, opened 2023-07-17
Add `Hashable` conformance to `StandardPredicateExpression`

stevapple:
This can enable collecting `StandardPredicateExpression`s into a fact list with `Set`.


jmschonfeld:
Hey @stevapple, thanks for posting this! I recently saw your comment on the original pitch thread and posted a reply there describing how I'm not quite sure this is the best approach because I'm worried that the behavior of this conformance wouldn't be the desired behavior: https://forums.swift.org/t/pitch-swift-predicates/62000/43. I'm curious what your thoughts are on this based on that context!

itingliu:
Closing this until we have a better idea of the appropriate approach. We should continue the discussion on whether we want this behavior on the forum.

PR #202, opened 2023-07-17
Locale.localeAsIfCurrentWithBundleLocalizations should respect user preferences

itingliu:
An "as-if-current" locale should reflect user's preferences. We do that correctly if `allowsMixedLocalizations` is true, but we missed that if otherwise.

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale_Cache.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale_Cache.swift
@@ -369,7 +369,8 @@ struct LocaleCache : Sendable {
             return nil
         }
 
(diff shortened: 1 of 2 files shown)


itingliu (commented):
I can't think of a straightforward way to test this change because it reads user preferences directly in the function. I manually tested this by setting an override on my Mac and make sure that `Locale.localeAsIfCurrentWithBundleLocalizations` returns that preference override.

itingliu:
@swift-ci please test

parkera (approved):

parkera:
The code in the `guard` and at the bottom are nearly the same now, but I think it's fine to leave it the way it is.

PR #203, opened 2023-07-17
Support Predicate construction with existential expression

jmschonfeld:
Developers may wish to construct predicates dynamically with arbitrary operators determined at runtime (for example, when making a predicate editor UI). The predicate builder closure should return an existential rather than a concrete type to allow for this dynamic construction. We annotate the old initializer as `@usableFromInline` to temporarily preserve the ABI of the old initializer until all call-sites have migrated to the source-compatible new initializer.

Resolves rdar://112423538


jmschonfeld:
@swift-ci please test

Azoy (approved):
Looks good to me!

glessard (approved):
I believe this is ABI-preserving.

jmschonfeld:
@swift-ci please test

PR #204, opened 2023-07-17
Predicate support for localized/case-insensitive string comparisons

jmschonfeld:
This adds support for the `localizedStandardContains`/`localizedCompare`/`caseInsensitiveCompare` functions in `Predicate`. These operators are all within `FOUNDATION_FRAMEWORK` blocks currently since we haven't ported the respective `String` APIs over to this repo yet. We also use `localizedCompare` instead of `localizedStandardCompare` here since the `Standard` comparison uses options such as forced ordering, width insensitive, etc. that destinations of these `Predicate`s do not yet support.

Resolves rdar://110083385


jmschonfeld:
@swift-ci please test

glessard (approved):

PR #168, opened 2023-06-07
Defaults to conversational hour options if the locale prefers it

itingliu:
Some locales prefer conversational day periods, ie. "12 at noon" to "12 pm". Currently clients need to use one of the `conversational` hour option to get this behavior, such as

```swift
let str = Date.now.formatted(.dateTime.hour(.conversationalDefaultDigits(amPM: .wide))
```

In some locales this is in fact the desired behavior, and should be the default. This PR adds a hook to swap "regular" hour with conversational one if we know the locale prefers it.

While we're here, clean up `ICUPatternGenerator`'s function interface and extra steps in `IntervalFormatStyle` that's now unnecessary.

rdar://110180766

PR #205, opened 2023-07-18
Consider both language and region when deciding whether to override day period with conversational style

itingliu:
We added an override to use the conversational day period option for Taiwan in #168. We should constrain it to when the language is Chinese.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #206, opened 2023-07-19
[WIP] Change DateComponent's storage layout

Azoy:
This is still marked as a WIP because I need to add tests to exercise all of the various branches this can take. This does pass the current preview tests.


naknut:
Hi @Azoy! I just posted in #302. I was thinking this could be a good first issue for me to contribute to. Do you still feel like this is a reasonable solution? If so I asked some questions in the issue on how to move forward.

PR #208, opened 2023-07-26
Date.FormatStyle does not autoupdate when 12-24 hour preferences changes when using autoupdating locale

itingliu:
We cache the generated date format per `Date.FormatStyle`. Normally `Date.FormatStyle`'s locale change would result in a cache miss because the hash has changed. However, using `.autoupdating` is an edge case where this does not apply -- `Locale.autoupdating` always compares equal to itself even when the locale identifier or preferences are different. Therefore we end up fetching the stale value when the locale is `.autoupdating`.
    
Fix this by using the locale identifier instead of the locale instance as the hash key.

Follow up: encapsulate hour preferences in the locale identifier instead of setting it as a special override. We should use this pattern wherever possible.


itingliu:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu (approved):

PR #211, opened 2023-07-27
Account for type parameter packs in Predicate coding

jmschonfeld:
This patch throws an error that ensures we won't crash if we encounter types with type parameter packs while encoding/decoding a predicate


jmschonfeld:
@swift-ci please test

itingliu (approved):
@swift-ci please test

Azoy (commented):
Looks good to me

PR #212, opened 2023-07-28
(112770879) “Washington, D.c.” capitalized incorrectly

itingliu:
Theoretically "." is a case-ignorable character, so "D.C." is conceptually the same as "DC", whose capitalized mapping would be "Dc". This behavior doesn't seem to align with real-world use cases well though. Workaround this by splitting the string with "." and titlecasing each substring individually.


itingliu:
@swift-ci test

itingliu:
@swift-ci please test

hristost (approved):

PR #215, opened 2023-08-07
(112778892) Unlocalized decimal separators

itingliu:
We regressed this behavior in 110211953; Now `IntegerFormatStyle` (and all other number format styles) and `Date.FormatStyle` ignores user's region settings if it uses the current locale.

The regression was caused by us changing to use `locale.identifierCapturingPreferences` for the identifier, which includes "Country" code from global preferences as the extra "rg" keyword. I misunderstood what that represents -- It has nothing to do with the Region setting in System Preferences. We should ignore it internally.

I'm very inclined to remove the reference to that key altogether, but let's keep it for compatibility reason for now.


itingliu:
@swift-ci please test

parkera (approved):

PR #216, opened 2023-08-10
Predicate incorrectly asserts on some keypaths

jmschonfeld:
In our key path inspection logic, I unfortunately got a few values wrong when inspecting the memory contents of keypaths. This leads to the inspection logic detecting stored property keypaths which use out-of-line storage (i.e. those with a very large offset in a large struct or those whose offset needs to be calculated at runtime due to properties of resilient types) as multiple components when they are only one component. This PR updates the mask values to ensure the appropriate logic, and the unit test (which accesses keypaths before, at, and after the resilient Date type property) ensures that predicate initialization with these keypaths no longer crashes.

### DIFF
--- a/Sources/FoundationEssentials/Predicate/KeyPath+Inspection.swift
+++ b/Sources/FoundationEssentials/Predicate/KeyPath+Inspection.swift
@@ -16,9 +16,6 @@ extension UInt32 {
     private static var COMPONENT_HEADER_KIND_MASK: UInt32 { 0x7F00_0000 }
     private static var COMPONENT_HEADER_PAYLOAD_MASK: UInt32 { 0x00FF_FFFF }
     
-    private static var STORED_COMPONENT_HEADER_PAYLOAD_MASK: UInt32 { 0x007F_FFFF }
-    fileprivate static var STORED_COMPONENT_PROPERTY_OFFSET_OUT_OF_LINE: UInt32 { 0x007F_FFFF }
-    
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

jckarter:
If you just need to know whether a key path is stored or not, then you can do that by using public API, since `MemoryLayout.offset(of: kp)` should only produce an offset for stored property key paths.

jmschonfeld:
> If you just need to know whether a key path is stored or not, then you can do that by using public API, since `MemoryLayout.offset(of: kp)` should only produce an offset for stored property key paths.

Unfortunately we need to know a little bit more than whether it is stored - `Predicate` requires that it only operates on single-component keypaths, so we need to check here to see whether or not this keypath only represents a single component (regardless of whether it's a stored or computed property)

Azoy:
We could rip out all of the keypath inspection for stored components however using `MemoryLayout.offset(of:)` and only do the unsafe stuff for the rest of the components. Using this API would have at least prevented this issue for stored component case.

jmschonfeld:
Ok yeah that's a fair point, I've updated this PR to do just that

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
macOS failure looks like it's due to a stale built package dependency after a snapshot update

PR #217, opened 2023-08-11
Fix JSONDecoder.unwrapString expectedType

swhitty:
`func unwrapString(from value: ...)` throws `DecodingError.valueNotFound` if the value is `null`, but it incorrectly reports that the expected type as `Array<String>` instead of `String`.
 
This PR fixes the bug, ensuring that the thrown error reports the expected type as `String` aligning the behaviour with Foundation on existing Darwin platforms.

```swift
DecodingError.valueNotFound(String.self, _)
```


iCharlesHu:
@swift-ci please test

kperryua (approved):
Good catch! Thanks!

Craz1k0ek (approved):

PR #218, opened 2023-08-15
Improve precision of internal rounding function for Duration values

theMomax:
Associated with rdar://84683018

The old rounding function defined on FixedWidthInteger was only used to round `Duration` values internally, basically requiring to round the `seconds` and `attoseconds` part separately. Furthermore, the old algorithm just converted to `Double`, rounded, and converted back, causing the results to not be precise in some scenarios. The new algorithm just implements rounding on `Duration` directly, making it more precise and simpler to use.


itingliu:
@swift-ci please test

itingliu (approved):

PR #240, opened 2023-08-17
Conditionalize Darwin imports with canImport instead of os

jmschonfeld:
We should ensure that our platform-specific imports are conditionalized with `canImport` rather than `os`. In this case, the file failed to build for watchOS because the `Darwin` import was excluded unintentionally.

### DIFF
--- a/Sources/FoundationEssentials/Data/DataProtocol.swift
+++ b/Sources/FoundationEssentials/Data/DataProtocol.swift
@@ -10,11 +10,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if os(macOS) || os(iOS)
+#if canImport(Darwin)
 import Darwin
-#elseif os(Linux)
+#elseif canImport(Glibc)
 import Glibc
-#elseif os(Windows)
+#elseif canImport(ucrt)
 import ucrt
 #endif
 


jmschonfeld:
@swift-ci please test

glessard (approved):

Sources/FoundationEssentials/Data/DataProtocol.swift
```diff
@@ -10,11 +10,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if os(macOS) || os(iOS)
+#if canImport(Darwin)
 import Darwin
-#elseif os(Linux)
+#elseif canImport(Glibc)
```
FranzBusch, line 15: @MaxDesiatov Do we need a `canImport` for musl here as well?

MaxDesiatov: Yes, of course, thanks for catching this!

Sources/FoundationEssentials/Data/DataProtocol.swift
```diff
@@ -10,11 +10,11 @@
...
 import Glibc
-#elseif os(Windows)
+#elseif canImport(ucrt)
 import ucrt
 #endif
```
MaxDesiatov, line 19: Would you mind adding support for Musl and WASI while we're at it?

```suggestion
#elseif canImport(Musl)
import Musl
#elseif canImport(ucrt)
import ucrt
#elseif canImport(WASILibc)
import WASILibc
#endif
```

glessard (approved):

glessard:
@swift-ci please test

iCharlesHu (approved):

PR #242, opened 2023-08-18
Empty string is returned instead of null for failure on CFLocaleCreateLocaleIdentifierFromWindowsLocaleCode

parkera:
In previous releases, `CFLocaleCreateLocaleIdentifierFromWindowsLocaleCode` would return `NULL` for a bad locale code. This restores that behavior instead of returning an empty string.

rdar://113874839


parkera:
@swift-ci test

itingliu (approved):

parkera:
@swift-ci test

PR #243, opened 2023-08-18
Getting localized calendar description via CFLocale is incorrectly returning NULL

parkera:
`CFCalendar` has an additional `calendar` key which `NSLocale` does not. The new Swift locale code needs to honor it as well.

rdar://113875255


parkera:
@swift-ci test

itingliu (approved):

PR #245, opened 2023-08-22
Move `#Predicate` macro to FoundationPreview

jmschonfeld:
This PR adds the `#Predicate` macro implementation and its unit tests to the open source swift-foundation repo. This will allow clients of swift-foundation to invoke the `#Predicate` macro to construct predicates using standard swift syntax.

A few notes:
- Adding a macro to a package manifest requires 5.9 tools. Since our current manifest only requires 5.8 tools, I created a copy of it (`Package@swift-5.8.swift`) for those using 5.8 tools that doesn't include the macro, and added the macro to the standard `Package.swift` after bumping its requirement to 5.9 tools
- Macros are currently only supported on Darwin platforms, so the package manifest does not add the macro on linux/windows. This isn't a robust solution (it doesn't work well with cross-compilation, etc.) but I expect it to be short lived as macro support on alternative platforms is currently in progress.
- The macro implementation does have some warnings that will be introduced into the project due to recent swift-syntax changes. After swift-syntax stabilizes the 5.9 release shortly, we'll quickly followup with another PR to address these warnings in the very near future.

### DIFF
--- a/Package@swift-5.8.swift
+++ b/Package@swift-5.8.swift
@@ -1,8 +1,7 @@
-// swift-tools-version: 5.9
+// swift-tools-version: 5.8
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
-import CompilerPluginSupport
 
 let package = Package(
     name: "FoundationPreview",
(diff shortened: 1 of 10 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

glessard (approved):

DougGregor (approved):

PR #246, opened 2023-08-23
Macro tests should run on linux

jmschonfeld:
While using macros within the compiler itself while building a project is not currently supported on linux/windows, building a macro target and running its unit tests _is_ possible on non-darwin platforms. This change updates the package manifest to build the macro target and its test on all platforms, but only adds the FoundationEssentials dependency on FoundationMacros (to allow the macro to be used during the build of downstream modules) when not on linux/windows.


jmschonfeld:
@swift-ci please test

glessard (approved):

PR #248, opened 2023-08-24
Move NSPredicate conversion in Predicate tests into a separate function

jmschonfeld:
To assist with maintenance of our `FOUNDATION_FRAMEWORK` `NSPredicate` conversion tests, we're moving the calls to the actual `NSPredicate` initializer into a separate function rather than calling it directly from each individual test.


jmschonfeld:
@swift-ci please test

jmschonfeld:
macOS failure looks to be unrelated due to a cache of the `_RopeModule` module from a prior snapshot

iCharlesHu (approved):

PR #249, opened 2023-08-24
Implement unlocalized range(of:) for AttributedString

itingliu:
`String` has a natively-implemented `_range(of:options:)`. This PR implements `AttributedString`'s `range(of:options:)` with that when localization support is not needed.

Ideally we should implement this on `BigString` to avoid iterating through `Characters`, which is tracked as a future TODO.
    
This function takes a `Locale` argument, which isn't available in FoundationEssentials. We could move this to FoundationInternalization where `Locale` is defined, but some clients use it with `locale: nil` and do not need localized results. To make the best of this let's add a non-localized version for FoundationEssentials.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedStringProtocol.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedStringProtocol.swift
@@ -257,13 +257,12 @@ extension AttributedStringProtocol {
 
         return self._utf8Index(at: startOffset) ..< self._utf8Index(at: endOffset) // O(log(n))
     }
-    
-#if FOUNDATION_FRAMEWORK
+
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

jmschonfeld (approved):
This LGTM, but I'd love to hear @lorentey's thoughts too in case I'm missing anything

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
Rebased on main and resolved @parkera's previous comment  -- Now that we have `Locale` in Essentials we can enable this API in Essentials where localized result isn't needed 🎉

parkera (approved):

iCharlesHu (approved):
🥳

PR #250, opened 2023-09-05
JSON5 parsing hex number with a trailing ‘E’ fails

iCharlesHu:
Number prevalidation is looking for a trailing "E" character in a number because Decimal string parsing accepts one, but it's not valid JSON (strtod rejects it). However, this should not apply to hex numbers.

Resolves rdar://114714976


iCharlesHu:
@swift-ci please test

parkera (approved):

PR #253, opened 2023-09-05
Enable the `#Predicate` macro on Linux

jmschonfeld:
With support for macros enabled on Linux, we can now enable the `#Predicate` macro on Linux. Using macros on Linux (and Darwin in Swift packages) requires a 5.9 toolchain so this patch replaces the `os(Linux)` checks with `compiler(>=5.9)` checks. This also bumps the minimum recommended 5.9 snapshot to the September 1st snapshot which added support for macros on Linux.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -109,8 +109,8 @@ package.targets.append(contentsOf: [
 ])
 #endif
 
-#if !os(Linux) && !os(Windows)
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

Craz1k0ek (approved):

iCharlesHu (approved):

PR #258, opened 2023-09-13
Update `Predicate` Unit Tests to use the Macro

jmschonfeld:
Previously due to a compiler limitation, the `Predicate` unit tests used manual constructions rather than the `#Predicate` macro. Now that the limitation has been lifted, we will update most of the unit tests to use the macro. However, we still leave the basic unit tests as-is until macros are supported on Windows.


jmschonfeld:
@swift-ci please test

hristost (approved):

PR #259, opened 2023-09-14
String paragraph/line searching can attempt to form invalid indices

jmschonfeld:
In our implementation of `_getBlock` which is used to find the ranges of paragraphs/lines within strings, we can potentially attempt to form an invalid index of a string resulting in a runtime crash. If the range provided begins at the start of the string and only contains one separator character (as is the case with a string that is only a separator character), then we attempt to form an index before the `startIndex` of the string while iterating backwards to find the contents end. This PR updates the `_matchesSeparators` function to use the `formIndex(_:offsetBy:limitedBy:)` function rather than calling `index(before:)` on what may potentially be the `startIndex` to ensure that we don't form invalid indices and instead break out of the loop as soon as we reach the end of the separator or the string.

### DIFF
--- a/Sources/FoundationEssentials/String/StringBlocks.swift
+++ b/Sources/FoundationEssentials/String/StringBlocks.swift
@@ -71,18 +71,18 @@ extension BidirectionalCollection where Element == UTF8.CodeUnit {
             }
         }
         for separator in separators {
-            var matches = true
             var strIdx = start
             var separatorIdx = reverse ? separator.count - 1 : 0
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

itingliu (approved):

parkera (approved):

jmschonfeld:
Addressed an off-by-one logic issue where separators would match even if their final UTF-8 scalar differed.

@swift-ci please test

PR #262, opened 2023-09-16
Fixed IntegerFormatStyle for values > Int64.max.

wadetregaskis:
If I understand the NumberFormatter implementation in the ICU library, this _should_ handle arbitrary-precision numbers, as long as they have a `description` which returns their POSIX-formatted representation, although that's not in the unit tests because it requires a working arbitrary-precision type, and I assume it's not acceptable to pull in a BigInt package for this purpose, and my efforts to make a quick-and-dirty one for testing purposes led only to tears and threats against Xcode's life.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
To be precise, ICU requires the string to be a "numeric string" as defined here: https://speleotrove.com/decimal/daconvs.html#refnumsyn

Ideally we would want to implement our own if we are to follow the spec, but realistically I doubt swift's implementation of `description` will ever change, so it'd probably work fine. Using `description` however doesn't feel right... I'm curious to hear what others think. @parkera @glessard ?

parkera:
We've certainly been known to change descriptions for various reasons over the years, so I would vote for using a dedicated internal function to be clear what the requirements are.

wadetregaskis:
> Ideally we would want to implement our own if we are to follow the spec, but realistically I doubt swift's implementation of `description` will ever change, so it'd probably work fine. Using `description` however doesn't feel right... I'm curious to hear what others think. @parkera @glessard ?

It's not just what the built-in integer types do for `description`, it's 3rd party types too.  My reason for doing this, for example, is because I want to use arbitrary-precision integers (via [BigInt](https://github.com/attaswift/BigInt)) while still having usable display in GUIs (because big numbers _especially_ need things like thousands separators).

There are [many 'BigInt' packages for Swift](https://swiftpackageindex.com/search?query=arbitrary+precision), and while I didn't exhaustively check them, I did look at most and they all do render the desired output from their `description` (because none of them remotely try to produce localised or "nicely" formatted results, they all do the same basic thing of just splatting out the latin decimal digits).

I did initially intend to write a BinaryInteger-to-Unicode-number-string function for this, but deferred it given:

* All my examples already produce the right output.
* I don't know what `BinaryInteger.description` intends to do as-is (e.g. maybe it's already defined to output in the desired format?  Sounds like no, though, from @parkera's comment).
* I wasn't sure how much effort it'd be to write an _efficient_ implementation.
* I couldn't conclusively locate the actual format specification, from ICU.  They refer vaguely to some other spec document, which doesn't seem to contain a clear number format specification.

If everyone's generally happy with this change otherwise, then I'm happy to dig into this last aspect and address that too.  I just didn't want to do that work pre-emptively without knowing if this fix would be welcome to begin with.

itingliu:
Thinking more about this I think it'd be more reasonable if we implement a string following the spec of on our own 

The problem with relying on `.description` is that there's no API contract of what that returns, so the fact that it works is pure coincidence. On the contrary we have all the internal values for `BinaryInteger` and `BinaryFloatingPoint` thanks to the protocol requirement, which would allow us to implement such function reliably. 

> I wasn't sure how much effort it'd be to write an efficient implementation.

I understand this concern, but I think it's totally fine to start with something that works intuitively, and defer non-obvious optimization later if needed. I'm all for avoiding unnecessary premature optimization

> I couldn't conclusively locate the actual format specification, from ICU. They refer vaguely to some other spec document, which doesn't seem to contain a clear number format specification.

https://speleotrove.com/decimal/daconvs.html#refnumsyn this is the spec. I agree that it's unclear, but in my opinion it's not much different from how opaque `.description` is

wadetregaskis:
Sorry for the delay - I installed Sonoma on my Macs last week and it took pretty much the entire week to repair the damage it did.  That's still a work in progress, but I can at least get Xcode to launch again, so I've picked this up.

I do have a working implementation of the additional changes, to not rely on `description`.

Unfortunately I rebased and I'm now blocked by the fact that top of tree doesn't build anymore (even before I reapply my patches).  1,513 errors, things like "'Predicate' is only available in macOS 9999 or newer" and "CircularReference".  I'll wait until those are all fixed before resuming my work.

parkera:
> > Unfortunately I rebased and I'm now blocked by the fact that top of tree doesn't build anymore (even before I reapply my patches). 1,513 errors, things like "'Predicate' is only available in macOS 9999 or newer" and "CircularReference". I'll wait until those are all fixed before resuming my work.
> 
> Thanks for bringing this up to our attention! We're going to resolve this soon

#271

wadetregaskis:
Does this project support 32-bit platforms (specifically, where UInt is 32-bit)?

wadetregaskis:
Getting there.  I believe I've address all the correctness issues for positive numbers, but there's at least one remaining bug for negative.

Also, this surely won't work if UInt is 32-bit, but only because I've hard-coded the assumption of 64-bitness into the unit tests (the non-test code should work fine in any bitness - I was careful not to hard-code any assumptions there).  I don't have any way (that I know of) to actually test the project in 32-bit mode.

I also still want to optimise the output generation - it's currently using a little over twice as much memory as it has to, due to the two-phase process of first assembling the 'word' blocks in reverse order, then glueing them together.  I think I can combine those two (possibly making the code a little simpler overall).

wadetregaskis:
@swift-ci please test

wadetregaskis:
Close to finished now.  I still need to do a final review and clean up any rough edges or outdated comments.

I added a dependency on attaswift/BigInt in order to unit test against a real arbitrary-precision integer.  I kept that to a single patch so it's easy to exclude or revert - though that dependency is only used for testing purposes, I don't know if it's considered kosher to include it.

The unit tests still only work in 64-bit, too.  Let me know if that needs to change.

wadetregaskis:
Okay, I think I'm done refining and reviewing it.

Let me know what you think.

parkera:
Sorry I didn't get your comment earlier -- but yes, we do support 32-bit platforms.

wadetregaskis:
In theory it now all works when UInt is 32-bit.  I have no way to test that, though, that I'm aware of.

itingliu:
> I added a dependency on attaswift/BigInt in order to unit test against a real arbitrary-precision integer. I kept that to a single patch so it's easy to exclude or revert - though that dependency is only used for testing purposes, I don't know if it's considered kosher to include it.

I'd prefer we don't pull in another package. It's still yet another dependency required for building the package even if it's used for testing. Also this makes the tests dependent on the package's implementation, which complicates the tests if the implementation changes. It might make sense to have that as integration test with that specific package, but not as unit tests in swift-foundation. The tests you added are already pretty comprehensive in my opinion.

Package.swift
```diff
@@ -22,7 +22,9 @@ let package = Package(
             revision: "0c1de7149a39a9ff82d4db66234dec587b30a3ad"),
         .package(
             url: "https://github.com/apple/swift-syntax.git",
-            from: "509.0.0")
+            from: "509.0.0"),
+        .package(url: "https://github.com/attaswift/BigInt.git",
+                 .upToNextMajor(from: "5.1.0")),
```
glessard, line 27: Could you get enough test coverage by using `Int128` from swift-numerics? This is something you could include. It would not be able to expose performance problems in extremely large numbers, but it should allow you to validate most of the algorithm.

wadetregaskis: `DoubleWidth` is (essentially) package-internal to swift-numerics; it's not exported.

wadetregaskis:
> I'd prefer we don't pull in another package. It's still yet another dependency required for building the package even if it's only used for testing. Also this makes the tests dependent on the package's implementation, which complicates the tests if the implementation changes. It might make sense to have that as integration test with that specific package, but not as unit tests in swift-foundation. The tests you added are already pretty comprehensive in my opinion.

Fair enough.  How about I put those bits behind `#if canImport(BigInt)` and comment out the import (in Package.swift), so that they're there for posterity if anyone wants to recheck with them, but they won't incur any overheads (dependent package fetching etc) for everyone, by default?

The BigInt-based tests caught most of the bugs, as I was implementing things.  There's not much useful surface area in the built-in integer types since nothing interesting happens until at least Int64.max + 1.

It's a shame that while the Swift Stdlib actually contains an Int128 type, it just doesn't export it.  Int128 wouldn't fully cover the testing desires, but it would be a lot better than just UInt64 and smaller.

Tests/FoundationInternationalizationTests/Formatting/Number/BinaryInteger+FormatStyleTests.swift
```diff
@@ -0,0 +1,146 @@
+// Copyright (c) 2023 Wade Tregaskis
```
glessard, line 1: The copyright should be assigned to the Swift project. While we're at it, updating the copyright years of the files being modified would be welcome.

wadetregaskis: Done.

wadetregaskis:
In addition to addressing feedback, I did some more proof-reading and improved some of the documentation.

I also realised it's possible to detect incorrect implementations of quotientAndRemainder(dividingBy:), and that by doing so it slightly simplifies the code anyway, so I added that in.  I know it's possible to get quotientAndRemainder(dividingBy:) wrong because when I implemented it my first time I did exactly what this checks for (using T-division instead of F-division, because the former is what I'd call the standard method, mathematically).

I was also reminded that `assert()` without a bespoke message gives pretty useless output, so I added detailed error messages to all the asserts.

Package.swift
glessard: We could copy the source for DoubleWidth into the test utilities, until such time as we have something either in the stdlib or in a package we are willing to import.

itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/Number/BinaryInteger+FormatStyleTests.swift
```diff
@@ -0,0 +1,146 @@
...
+        check(type: Int32.self, min: "-2147483648", max: "2147483647")
+        check(type: Int64.self, min: "-9223372036854775808", max: "9223372036854775807")
+
+        check(type: UInt8.self, min: "0", max: "255")
+        check(type: UInt16.self, min: "0", max: "65535")
+        check(type: UInt32.self, min: "0", max: "4294967295")
+        check(type: UInt64.self, min: "0", max: "18446744073709551615")
+    }
+
+    func testNumericStringRepresentation_buildinIntegersAroundDecimalMagnitude() throws {
```
glessard, line 44: Typo:
```suggestion
    func testNumericStringRepresentation_builtinIntegersAroundDecimalMagnitude() throws {
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -34,6 +40,199 @@ extension BinaryInteger {
...
+    ///
+    /// This is intended to be used only by the `numericStringRepresentation` property (both the specialised form below and the generic one for all BinaryIntegers, above).  Prefer the `numericStringRepresentation` property for all other use-cases.
+    ///
+    /// - Parameter intoEndOfBuffer: The buffer to write into, which _must_ contain enough space for the result.  The formatted output is placed into the _end_ of this buffer ("right-aligned", if you will), though the output is numerically still left-to-right.  The contents of this buffer do not have to be pre-initialised.
+    /// - Returns: How many entries (UInt8s) of the buffer were used.  Note that zero is a valid return value, as nothing is written to the buffer if `self` is zero (this may be odd but it's acceptable to `numericStringRepresentation` and it simplifies the overall implementation).
+    func numericStringRepresentation(intoEndOfBuffer buffer: inout Slice<UnsafeMutableBufferPointer<UInt8>>) -> Int {
+        guard .zero != self else { return 0 } // Easier to special-case this here than deal with it below (annoying off-by-one potential errors).
```
glessard, line 185: Nit: this case has already been handled in the caller, so the return statement is dead code in practice. Could be replaced by an assertion.

glessard (approved):
This is a great contribution, thank you! Since this particular kind of formatting is independent of locale, it would be interesting to discuss whether a `String` initializer that provides the numeric string representation of a `BinaryInteger` belongs in the standard library.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -34,6 +40,199 @@ extension BinaryInteger {
...
+extension UInt {
+    /// Formats `self` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html) which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
+    ///
+    /// This is intended to be used only by the `numericStringRepresentation` property (both the specialised form below and the generic one for all BinaryIntegers, above).  Prefer the `numericStringRepresentation` property for all other use-cases.
+    ///
+    /// - Parameter intoEndOfBuffer: The buffer to write into, which _must_ contain enough space for the result.  The formatted output is placed into the _end_ of this buffer ("right-aligned", if you will), though the output is numerically still left-to-right.  The contents of this buffer do not have to be pre-initialised.
+    /// - Returns: How many entries (UInt8s) of the buffer were used.  Note that zero is a valid return value, as nothing is written to the buffer if `self` is zero (this may be odd but it's acceptable to `numericStringRepresentation` and it simplifies the overall implementation).
+    func numericStringRepresentation(intoEndOfBuffer buffer: inout Slice<UnsafeMutableBufferPointer<UInt8>>) -> Int {
```
iCharlesHu, line 184: Since it's a implementation detail this method should probably be `private`.

Tests/FoundationInternationalizationTests/Formatting/Number/BinaryInteger+FormatStyleTests.swift
```diff
@@ -0,0 +1,146 @@
...
+// REQUIRES: executable_test
+
+import XCTest
+
+import FoundationEssentials
+@testable import FoundationInternationalization
+
+#if canImport(BigInt) // Not included by default as it's a 3rd party library; requires https://github.com/attaswift/BigInt.git be added the package dependencies.  Proved useful in the past for finding bugs that only show up with large numbers.
+import BigInt
+#endif
```
iCharlesHu, line 19: Instead of using `BigInt`, we should wait until `Decimal` is implemented and use it instead: https://github.com/apple/swift-foundation/issues/236

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -34,6 +40,199 @@ extension BinaryInteger {
...
+    ///
+    /// This produces output that (at time of writing) looks identical to the `description` for many `BinaryInteger` types, such as the built-in integer types.  However, the format of `description` is not specifically defined by `BinaryInteger` (or anywhere else, really), and as such cannot be relied upon.  Thus this purpose-built method, instead.
+    ///
+    /// It might be worth moving this into the Swift standard library one day, so that it can be used as the basis for the default `description` instead of duplicating that conversion process.  At least while `description`'s output happens to match this one's.
+    ///
+    /// This property's type is an ArraySlice rather than a straight ContiguousArray because it's computationally more expensive to size the array exactly right (whether by pre-calculating the required size more precisely, or by dynamically resizing it during execution of the iterative algorithm).  Since this is intended to only be a transient value, for the purposes of translating from BinaryInteger to the ICU libraries, this is the right compromise of runtime speed (CPU time) and memory efficiency (it's usually only off by a byte or two, if that).
```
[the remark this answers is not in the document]
iCharlesHu: Also, could you break down the comments into multiple lines? I know we don't have an official style guide yet, but these comments will definite exceed that limit.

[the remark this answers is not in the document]
wadetregaskis: I can.  What is the line length limit?

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -34,6 +40,199 @@ extension BinaryInteger {
...
+    internal var numericStringRepresentation: ArraySlice<UInt8> {
+        // Fast-path for values that fit into a UInt, as the conversion to a UInt should be virtually free if it's possible (it's essentially just self.words[0]) and there's a specialisation of this function for UInt that's faster.
+        if let fastForm = UInt(exactly: self) {
+            return fastForm.numericStringRepresentation
+        }
+
+        assert(.zero != self, "Value of zero (for self) should have been handled by fast path, but wasn't.") // Zero isn't handled correctly in the algorithm below (no numbers will actually be emitted) because it's more work to do so, which is unnecessary as the fast path above should handle that case.
```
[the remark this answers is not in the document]
wadetregaskis: Perhaps; I'll make it so.

It should never trip as long as `UInt(exactly:)` and the implementation of the concrete type's `words` property aren't buggy.  If either is, then it's likely that the result of this method will be wrong anyway.  So the benefit of the check is probably just a faster and clearer failure.

But the downside of the precondition check seems quite small - slight increase in code size and a [hopefully very small] runtime cost.

wadetregaskis:
> This is a great contribution, thank you! Since this particular kind of formatting is independent of locale, it would be interesting to discuss whether a `String` initializer that provides the numeric string representation of a `BinaryInteger` belongs in the standard library.

I think that's a good direction.  As I noted in the code, another benefit of moving this into the standard library is that various `description`s can use it, eliminating duplicate code (I haven't looked at how `description` is implemented for the standard numeric types, but it's likely _very_ similar to what I've implemented here).

Tests/FoundationInternationalizationTests/Formatting/Number/BinaryInteger+FormatStyleTests.swift
wadetregaskis: To be clear, are you saying to remove this code entirely?  It's up to you, but FWIW I feel it's useful to keep accessible - e.g. while addressing feedback I locally re-enabled the BigInt package dependency in order to re-test things.

Package.swift
wadetregaskis: To clarify, I don't have strong feelings on this, I'd just like to be sure it's a fully informed decision.  Copy-pasting `DoubleWidth` is mechanically easy but it's against general best practice.

[I did suggest putting `DoubleWidth` into its own package](https://forums.swift.org/t/status-check-int128-uint128/67694/17), so that it can be more cleanly reused, but @stephentyrone discouraged it.

Having use of 128-bit integers would improve the unit test's effective coverage significantly.  It's not enough to provide full confidence, since some bugs by nature are only hit with _much_ larger numbers, but it's fairly close.

tbkka:
> ... benefit of moving this into the standard library is that various descriptions can use it, eliminating duplicate code ...

I think it would be useful to have a standard fallback implementation that can handle user-defined types.

I'd want to do some performance profiling before adopting it for the standard numeric types, as the `description` methods are used in some performance-critical code (such as JSON formatting).

wadetregaskis:
> I'd want to do some performance profiling before adopting it for the standard numeric types, as the `description` methods are used in some performance-critical code (such as JSON formatting).

Yes, that would be wise.

I didn't do any comparative performance analysis with this implementation.  I did take some pains to _try_ to make it quite fast - the code could be a lot simpler otherwise - but that's all.  A significant factor in its performance is how well the compiler optimises it, of course, but an even bigger one is how the actual type in question performs as a `BinaryInteger` (particularly re. `quotientAndRemainder(dividingBy:)`).  And that raises the question of which BigInt implementation you use for reference, etc.  So, potentially quite an undertaking to really evaluate robustly.

For built-in types, this will likely be _slower_ than dedicated implementations (as currently written) because of things like `decimalDigitsAndMagnitudePerWord` which are fairly fast but can be just hard-coded constants in non-generic implementations.

Tests/FoundationInternationalizationTests/Formatting/Number/BinaryInteger+FormatStyleTests.swift
wadetregaskis: Well spotted!  Fixed.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Done.

Package.swift
```diff
@@ -30,7 +30,9 @@ let package = Package(
             revision: "0c1de7149a39a9ff82d4db66234dec587b30a3ad"),
         .package(
             url: "https://github.com/apple/swift-syntax.git",
-            from: "509.0.0")
+            from: "509.0.0"),
+//        .package(url: "https://github.com/attaswift/BigInt.git",
+//                 .upToNextMajor(from: "5.1.0")),
```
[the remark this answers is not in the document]
wadetregaskis: Done.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
wadetregaskis: Done.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Done (well, `fileprivate` by necessity, but yes).

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
parkera: We don't have a line length limit in this project, preferring to let users wrap it to their own preference in their editors.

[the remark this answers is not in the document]
parkera: That said - I agree that we should keep header doc and implementation comments split up.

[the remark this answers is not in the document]
wadetregaskis: So I think this is resolved, then?  I didn't hard-wrap anything, but I moved the couple of paragraphs in question into the function body as regular comments.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -34,6 +40,199 @@ extension BinaryInteger {
...
+
+                // By definition the remainder has to be a single word (since the divisor, `wordMagnitude`, fits in a single word), so we can avoid working on a BinaryInteger generically and just use the first word directly, which is concretely UInt.
+                assert(remainder.bitWidth - (negative ? 1 : 0) <= Words.Element.bitWidth,
```
oscbyspro, line 82: The bit-width of fixed-width integers is not normalized, so `remainder.bitWidth` may be fixed to a value larger than `UInt.bitWidth` (see `Int256`).

oscbyspro (changes requested):
Hi, I saw this pull request mentioned on the Swift forums and wanted to try it out. While doing so, I noticed that a debug assertion fails for fixed-width integers wider than `UInt` (see comment). Without this assertion, the rest works with my double-width integer in [Numberick](https://github.com/oscbyspro/Numberick).

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Thanks for catching that.

This would presumably also make Int64 & UInt64 fail in 32-bit mode.

It's very unfortunate that it's documented to behave that way; it seems to render that property useless.  The full width of fixed-width integers is already provided by the type-level property `bitWidth`.

The claimed bit width is relied upon for more than just that assert; I'll have to think a bit about how to work around this problem…

wadetregaskis: …which, it turns out, is not easy.  I'm still trying to find a way; no luck today, I'll have to pick this up again some other day.

Mathematically it's trivial, but because of the structure of the `BinaryInteger` and associated types (in particular that `Magnitude` is merely `Numeric`, which is virtually useless) it is hard to actually implement the necessary functionality.

Though in looking at this I realised the size estimation, based on bit width, was technically wrong anyway.  Even when `bitWidth` returns what I was expecting.  I don't _think_ it could ever actually cause an off-by-one bug because I think it would require a power of ten and a power of two to be equal (which is impossible), but still, it'd be nice to eliminate that possibility as a matter of principle.  I'm still looking into that, too.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Actually, no, reverted.  We'd both overlooked the fact that it _can_ be called on zero because it's used for the individual words of the BinaryInteger (from the generic `numericStringRepresentation` directly, not the UInt specialisation!), any of which can actually be zero.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Okay, life found a way.  The end result isn't that complicated, actually, and might even be more efficient than before since we can take advantage of the fact that we only care about the _magnitude_, really, as that's what we're essentially converting to decimal digits.  So a lot of the complexities of two's complement representation can be side-stepped.

A positive coincidental side-effect of this is that the algorithm no longer relies on the `bitWidth` instance property at all (though still the type-level version for UInt, which is fine).  That lack of reliance is is good because it turns out some 3rd party implementations of `BinaryInteger` don't implement it correctly anyway (e.g. https://github.com/attaswift/BigInt/issues/115).

wadetregaskis: Oh, and also, [attaswift/BigInt](https://github.com/attaswift/BigInt) conforms both its types (`BigInt` and `BigUInt`) to `FixedWidthInteger` even though they're not fixed width.  Perhaps just to make it more convenient to use (I actually [ab]use this in the unit tests here, to use the `init` method that takes a string with configurable radix)?  So my initial attempts to special-case `bitWidth` interpretation for non-`FixedWidthInteger` `BinaryInteger`s was doomed to fail anyway (or, at least, to be irrelevant).

wadetregaskis:
I added unit tests using Numberick (https://github.com/oscbyspro/Numberick) in the same vein as the ones using BigInt, which found one additional bug.  Mostly the code doesn't care about fixed-width vs flexible-width but there is an underlying difference as the former may zero-pad `words` to the expected length, whereas the latter doesn't have any reason to (but also may have redundant, zero-padding words, I suppose - but attaswift/BigInt doesn't).

wadetregaskis:
@iCharlesHu & @glessard, I think this is just waiting on your clarifications regarding your earlier comments.  Can you take a quick look, please?

wadetregaskis:
So… should I drop this PR?  It seems like it's not going anywhere. 😞

I'm not sure how best to put this; I don't want to offend… this was my first attempt to contribute to an Apple open source project, and I had quite some enthusiasm towards some further contributions, but now I think that would be a bad idea.

This is fundamentally a trivially bug fix, for something that was _really_ obviously broken for a long time.  It ended up being many days of work for me - which would have been fine, it was educational at least, _iff_ it actually went through.  Doing all that work to no actual avail is disheartening.

itingliu (approved):
Sorry @wadetregaskis I was absorbed in my own work and this just slipped my mind. This looks good to merge to me. I'm going to kick off another CI run and then I'll merge it

If @iCharlesHu or @glessard has further comments we can always address them in a follow up PR

itingliu:
@swift-ci please test

parkera:
Thank you so much @wadetregaskis! I really appreciate this contribution and the detailed work you put into it.

stephentyrone:
Yeah, I don't see anyway for this change to manifest that error.

PR #263, opened 2023-09-18
Update swift-syntax tag to 509.0.0

jmschonfeld:
Now that Swift 5.9 is released and swift-syntax has tagged their 509.0.0 release, we should update our dependency on swift-syntax to specify this tag rather than a snapshot.


iCharlesHu (approved):

iCharlesHu:
@swift-ci please test

ahoppen (approved):

PR #273, opened 2023-09-27
Fix mismatched Locale hash/equality implementation

jmschonfeld:
Recently, we accidentally updated `Locale`'s hashing and equatable implementations to differ - the `hash(into:)` function sometimes took more information (such as the identifier and preferences) into account than the equatable checking leading to multiple `Locale`s being equal but having different hashes. This change updates the two implementations to be compatible and simplifies the logic.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #275, opened 2023-09-28
FoundationInternationalizationTests: explicitly cast to `CLong`

compnerd:
This adds support for Windows which is a LLP64 environment.  That maps `Long` to 32-bit, and thus we need to explicitly cast the value to `CLong` which matches the `labs` signature.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #276, opened 2023-09-28
_CShims: excise tgmath on Windows

compnerd:
While Windows does have tgmath support, it is difficult to use due to the conflicting definitions originating from clang and MSVC.  Excise this for the time being which allows us to build the full project on Windows.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #280, opened 2023-10-04
Remove `FOUNDATION_FRAMEWORK` restriction on macOS Predicate tests

jmschonfeld:
Now that `Predicate` and its associated types have appropriate availability, we can remove the `FOUNDATION_FRAMEWORK` condition around the `Predicate` tests. Previously, this was needed because `Predicate` had `macOS 9999` availability, but due to variadic generics it didn't actually backdeploy to macOS <14. Now that the availability is correct, the tests will not run unless the host is macOS 14. This will allow these tests to run on macOS 14+ hosts (at desk currently, and in CI in the future when the CI devices are updated)

### DIFF
--- a/Tests/FoundationEssentialsTests/PredicateTests.swift
+++ b/Tests/FoundationEssentialsTests/PredicateTests.swift
@@ -14,9 +14,6 @@
 import TestSupport
 #endif
 
-// Predicate does not back-deploy to older Darwin versions
-#if FOUNDATION_FRAMEWORK || os(Linux) || os(Windows)
-
 @available(macOS 14, iOS 17, tvOS 17, watchOS 10, *)
 final class PredicateTests: XCTestCase {
     
@@ -754,5 +751,3 @@ final class PredicateTests: XCTestCase {
         }
     }
 }
-
-#endif
--- a/Tests/FoundationEssentialsTests/PredicateTests.swift
+++ b/Tests/FoundationEssentialsTests/PredicateTests.swift
@@ -14,8 +14,13 @@
 import TestSupport
 #endif
 
-@available(macOS 14, iOS 17, tvOS 17, watchOS 10, *)
 final class PredicateTests: XCTestCase {
+    
+    override func setUp() async throws {
+        guard #available(macOS 14, iOS 17, tvOS 17, watchOS 10, *) else {
+            throw XCTSkip("This test is not available on this OS version")
+        }
+    }
     
     struct Object {
         var a: Int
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

hassila:
Same for the Codable support that also had such a check IIRC?

jmschonfeld:
> Same for the Codable support that also had such a check IIRC? (Sources/FoundationEssentials/Predicate/Archiving/Predicate+Codable.swift)

The `Predicate` `Codable` code is indeed behind `#if FOUNDATION_FRAMEWORK`, but for a different reason than the availability of `Predicate` itself - unfortunately archiving of predicates relies on some reflection capabilities that we haven't yet ported over to FoundationPreview, so that will need to stay as `#if FOUNDATION_FRAMEWORK` for now

hassila:
Ok, thanks for clarification, we are looking forward then!

Tests/FoundationEssentialsTests/PredicateTests.swift
```diff
@@ -14,9 +14,6 @@
 import TestSupport
 #endif
 
-// Predicate does not back-deploy to older Darwin versions
-#if FOUNDATION_FRAMEWORK || os(Linux) || os(Windows)
-
 @available(macOS 14, iOS 17, tvOS 17, watchOS 10, *)
```
grynspan, line 17: FYI `@available` does not prevent tests from running when using XCTest. You'll need a guard in each test method, e.g.:

```swift
func testFoo() throws {
  guard #available(macOS 14, iOS 17, tvOS 17, watchOS 10, *) else {
    throw XCTSkip("Unavailable")
  }

  ...
}
```

jmschonfeld: Thanks for the tip, I was just diving into the CI logs to find out why these tests were still running. Instead of guarding within every test, is there a way I can guard at the test case level? Is there perhaps something in `XCTestCase` that is safe to override where I can write this guard and have it apply to every test in the test case?

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/PredicateTests.swift
jmschonfeld: I added this check to `setUp() async throws` to only check once, and moved the availability annotation from the test class to each individual function to avoid a warning for usage of `#if available(macOS 14, ...)` within an `@available(macOS 14, ...)` class.

grynspan: That'll work.

iCharlesHu (approved):

PR #282, opened 2023-10-04
Move String additions from _FoundationInternals to FoundationEssentials + package

parkera:
This moves a bunch of `String` internal implementations to FoundationEssentials instead of the "internal" group (which ends up symlinked into each package). Use `package` to make them available to FoundationInternationalization.


parkera:
@swift-ci test

itingliu (approved):
I'm so glad we can do this

iCharlesHu (approved):
Yesss!!!

PR #283, opened 2023-10-05
Add DocC documentation catalogs to Essentials, Internationalization, and Preview modules.

invalidname:
This PR is the first step of providing DocC-based documentation for Swift Foundation.

This PR adds `.docc` directories to FoundationEssentials, FoundationInternationalization, and Foundation (the top-level superset module). Along with providing an overview of each module, this PR performs an initial curation of top-level symbols into topic groups.

It doesn't curate all symbols in all types; that can come later when each type is documented. However, some types (such as `AttributedString` and `Predicate`) do get an initial curation for the sake of giving their helper types someplace to go.

Generate documentation with Xcode's "Product -> Build Documentation" menu item, or via the `xcodebuild` command-line:

````
xcodebuild docbuild -destination 'name=Any Mac' -scheme FoundationPreview-Package -derivedDataPath YOUR_OUTPUT_PATH
````

On other platforms, you can use the [Swift-DocC Plugin](https://apple.github.io/swift-docc-plugin/documentation/swiftdoccplugin/) to generate documentation.

### DIFF
--- a/Sources/Foundation/Documentation.docc/Curation/FoundationEssentials.md
+++ b/Sources/Foundation/Documentation.docc/Curation/FoundationEssentials.md
@@ -6,13 +6,13 @@ Access core types and essential functionality used by most applications.
 
 The Foundation Essentials package provides the most important types for everyday development:
 
-* Core types beyond those provided by the Swift standard library, such as ``Data``, ``Decimal``, and ``UUID``.
+* Core types beyond those provided by the Swift standard library, such as ``Data`` and ``UUID``.
 * Dates and time intervals.
 * An ``AttributedString`` type for strings with runs of metadata, such as styling information.
 * JSON encoding and decoding.
 * A ``URL`` type for network and local resources.
 
-Foundation Essentials provides these features without system dependencies, so developers can count on consistent behavior, regardless of platform.
+Foundation Essentials provides a canonical implementation of these types, so developers can count on consistent behavior, regardless of platform.
 
 ## Topics
 

### CODE REVIEW

File: Sources/Foundation/Documentation.docc/Curation/FoundationEssentials.md [hunk 1]
```diff
@@ -0,0 +1,37 @@
+# Foundation Essentials
+
+Access core types and essential functionality used by most applications.
+
+## Overview
+
+The Foundation Essentials package provides the most important types for everyday development:
+
+* Core types beyond those provided by the Swift standard library, such as ``Data``, ``Decimal``, and ``UUID``.
```
itingliu, line 9: Alas, we don't have `Decimal` or a concrete plan of what to do about it. Its fate is TBD. Do we still want it here?

Sources/Foundation/Documentation.docc/Curation/FoundationEssentials.md
```diff
@@ -0,0 +1,37 @@
...
+* Dates and time intervals.
+* An ``AttributedString`` type for strings with runs of metadata, such as styling information.
+* JSON encoding and decoding.
+* A ``URL`` type for network and local resources.
+
+Foundation Essentials provides these features without system dependencies, so developers can count on consistent behavior, regardless of platform.
```
itingliu, line 15: I'm not sure if this "without system dependencies" is entirely correct. We still use other libraries. It's just that FoundationEssentials abstracts away the underlying dependencies and chooses the one depending on the platform for developers, so they don't need to worry about platform check.

Sources/Foundation/Documentation.docc/Curation/FoundationEssentials/Fundamentals/NumbersAndBasicValues/Decimal.md
```diff
@@ -0,0 +1,45 @@
+# ``FoundationPreview/Decimal``
```
itingliu, line 1: There is indeed a `Decimal`  in the repo, but that is just a stub and does nothing. As I said in another comment I think we might want to consider holding this off until we decide its fate

Sources/Foundation/Documentation.docc/Curation/FoundationEssentials.md
invalidname: I was trying to include this concept from the [original announcement](https://www.swift.org/blog/future-of-foundation/):

> a new FoundationEssentials package will provide the most important types in Foundation with no system dependencies to help accomplish this.

But yeah, without the context of the rest of that announcement, there are lots of ways to read "without system dependencies".

PR #287, opened 2023-10-05
(116548255) test decoding dynamic subclasses

glessard:
Adapt a test from swift-corelibs-foundation, where JSON decoding could produce the wrong dynamic type. Add this test to prevent regressions. Original issue: https://github.com/apple/swift-corelibs-foundation/issues/4806

Addresses rdar://116548255

### DIFF
--- a/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
@@ -724,6 +724,21 @@ final class JSONEncoderTests : XCTestCase {
         _ = try? JSONEncoder().encode(ReferencingEncoderWrapper([Float.infinity]))
     }
 
+    func testDecodeDynamicSubclass() throws {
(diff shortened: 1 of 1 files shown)


glessard:
@swift-ci please test

kperryua (commented):
I think this extant test verifies the same behavior. Can you verify? 

```
    func testDecodingConcreteTypeParameter() {
        let encoder = JSONEncoder()
        guard let json = try? encoder.encode(Employee.testValue) else {
            XCTFail("Unable to encode Employee.")
            return
        }

        let decoder = JSONDecoder()
        guard let decoded = try? decoder.decode(Employee.self as Person.Type, from: json) else {
            XCTFail("Failed to decode Employee as Person from JSON.")
            return
        }

        expectEqual(type(of: decoded), Employee.self, "Expected decoded value to be of type Employee; got \(type(of: decoded)) instead.")
    }
```

glessard:
@kperryua you're right, that does the same thing. <sigh> Thanks!

PR #289, opened 2023-10-09
Date comparison handles nanoseconds incorrectly

parkera:
The comparison result for nanoseconds was backwards. This also takes the chance to switch to a more Swift-friendly version of `modf`.

rdar://116433981

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -721,8 +721,8 @@ public struct Calendar : Hashable, Equatable, Sendable {
             }
         case .minute:
             // assumes that time zone or other adjustments are always whole minutes
-            var (int1, _) = modf(date1.timeIntervalSinceReferenceDate)
-            var (int2, _) = modf(date2.timeIntervalSinceReferenceDate)
+            var int1 = date1.timeIntervalSinceReferenceDate.rounded(.down)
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci test

parkera:
We have a tranche of calendar tests that cover this but are written in ObjC which we hope to port soon.

Catfish-Man (approved):
lgtm

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -733,8 +733,8 @@ public struct Calendar : Hashable, Equatable, Sendable {
                 return .orderedAscending
             }
         case .second:
-            let (int1, _) = modf(date1.timeIntervalSinceReferenceDate)
-            let (int2, _) = modf(date2.timeIntervalSinceReferenceDate)
+            let int1 = date1.timeIntervalSinceReferenceDate.rounded(.down)
+            let int2 = date2.timeIntervalSinceReferenceDate.rounded(.down)
```
Catfish-Man, line 737: Oh I just realized: do we want .towardZero here too?

Catfish-Man: (This would be a behavior change from the old code, I'm just wondering if it's more correct in edge cases)

parkera: Hm, I'll need to think about it separately.

PR #293, opened 2023-10-12
Add String tests from stdlib

itingliu:
Batch add `String` and `Substring` tests from stdlib. These were removed from stdlib in https://github.com/apple/swift/pull/67252/files and https://github.com/apple/swift/pull/67450/. These tests were added to test Foundation's `StringProtocol` extension that called into `NSString` API.
    
Now that some of the tests were implemented with Swift natively, they should be made available for FoundationPreview, but we'll track that in a separate PR.

### DIFF
--- a/Tests/FoundationEssentialsTests/StringTests.swift
+++ b/Tests/FoundationEssentialsTests/StringTests.swift
@@ -348,7 +348,8 @@ extension String {
 
 final class StringTestsStdlib: XCTestCase {
 
-   public func withOverriddenLocaleCurrentLocale<Result>(
+    // This only works for String functions that calls into `NSString`, which calls the Obj-C `-[NSLocale current]`, which gets swizzled here. This does not work for String functions implemented natively in Swift.
+    public func withOverriddenLocaleCurrentLocale<Result>(
         _ temporaryLocale: NSLocale,
         _ body: () -> Result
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

Tests/FoundationEssentialsTests/StringTests.swift
```diff
@@ -327,3 +327,1925 @@ final class StringTests : XCTestCase {
...
+        }
+        do {
+            let s = "абвг"
+            let r = s.index(after: s.startIndex)..<s.endIndex
+            expectEqual(ComparisonResult.orderedSame,
+                        s.compare("бвг", range: r))
+        }
+
+        expectEqual(ComparisonResult.orderedSame,
+                    "abc".compare("abc", locale: Locale.current))
```
jmschonfeld, line 703: Should we make this an explicit locale so that the results of this test don't depend on the test host device's locale (in actually, I guess the current locale would probably never actually affect this comparison since they're equal, but might be a good change for correctness' sake?)

itingliu: Good catch. Definitely

itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #295, opened 2023-10-18
Remove a few extensions to `BinaryFloatingPoint` that are unused.

stephentyrone:
These are internal extensions, and are not used anywhere except in tests. They also make questionable use of the `pow` function in ways that would eventually lead to interesting errors on Linux and Windows. Best to simply excise them (and the tests for them as well).


stephentyrone:
@swift-ci test

PR #296, opened 2023-10-19
Fix issue with escaping of literals in Date.VerbatimFormatStyle

theMomax:
The previous logic for escaping the CLDR escape character for literal strings in the date pattern string (') when building the `Date.FormatString` was incorrect, producing invalid date pattern strings, and thus causing the formatting to fail under certain circumstances.

This PR fixes the logic along with the test expectations and adds tests that verify the functionality.


itingliu:
@swift-ci please test

itingliu (approved):
This is clever. Thanks!

PR #297, opened 2023-10-20
Work around SwiftSyntax crash due to keypath usage

jmschonfeld:
In some situations, using the `\.argumentList` key path here causes a crash. While that crash is being investigated, switch to assigning to the property directly to avoid the crash.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #299, opened 2023-10-25
Update remaining Predicate tests to use the #Predicate macro

jmschonfeld:
This updates the last of the predicate tests to use `#Predicate` instead of manual construction and removes the windows restrictions for usage of the macro

### DIFF
--- a/Tests/FoundationEssentialsTests/PredicateTests.swift
+++ b/Tests/FoundationEssentialsTests/PredicateTests.swift
@@ -14,6 +14,13 @@
 import TestSupport
 #endif
 
+#if !FOUNDATION_FRAMEWORK
+// Resolve ambiguity between Foundation.#Predicate and FoundationEssentials.#Predicate
+@freestanding(expression)
+@available(macOS 14, iOS 17, tvOS 17, watchOS 10, *)
+macro Predicate<each Input>(_ body: (repeat each Input) -> Bool) -> Predicate<repeat each Input> = #externalMacro(module: "FoundationMacros", type: "PredicateMacro")
+#endif
+
 final class PredicateTests: XCTestCase {
     
     override func setUp() async throws {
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/PredicateTests.swift
```diff
@@ -14,6 +14,13 @@
 import TestSupport
 #endif
 
+#if !FOUNDATION_FRAMEWORK
+// Resolve ambiguity between Foundation.#Predicate and FoundationEssentials.#Predicate
+@freestanding(expression)
+@available(macOS 14, iOS 17, tvOS 17, watchOS 10, *)
+macro Predicate<each Input>(_ body: (repeat each Input) -> Bool) -> Predicate<repeat each Input> = #externalMacro(module: "FoundationMacros", type: "PredicateMacro")
+#endif
+
```
iCharlesHu, line 23: Sigh... we can't switch to swift-testing fast enough.. do we have to do this for every test file?

iCharlesHu (approved):

Tests/FoundationEssentialsTests/PredicateTests.swift
jmschonfeld: Luckily this is the only file that uses this macro so at least for now it's a non-issue, and I _think_ it still might work in other files too (since the `macro`'s visibility is `internal`) but I'm not certain

jmschonfeld:
Looks like the usage of the macro triggered a compiler crash. That crash _should_ be resolved by https://github.com/apple/swift-syntax/pull/2312 (since similar crashes were successfully resolved by a similar change for SwiftPM plugins) so I'll hang onto this until a snapshot with the swift-syntax fix lands

jmschonfeld:
We've determined that the Linux CI failure is indeed due to this swift-syntax issue, and is affecting the `main` branch already (i.e. it is not directly caused by this PR). Seeing that macOS passed, I'm merging this to update our tests with the known linux failure already happening on `main`.

PR #303, opened 2023-10-27
WIP Gregorian Calendar in FoundationEssentials

itingliu:
### Part 1: `dateComponents(from date:)`
Implement `dateComponents(from date:)`. The implementation largely follows that of ICU in calendar.cpp and gregocal.cpp. Julian day calculation algorithm is referenced [The Explanatory Supplement to the Astronomical Almanac](https://aa.usno.navy.mil/publications/exp_supp).

First, calculate the Julian day number from a given `Date`. Then, convert the Julian day number to Gregorian calendar date fields[^1]. The remaining part is the time within the day.

There are a few things to look out for
- Julian day starts at noon, while `Date` uses midnight as the reference time, so we need to adjust it acccordingly at every conversion. Note that ICU uses Modified Julian Day number throughout their codebase, which starts at midnight, so
- Dates close to Gregorian calendar adoption, which is also referred as "cutover" or "transition" date interchangeably, needs to be handled with care. Date before the adoption date were represented in Julian Calendar and Gregorian Calendar after the adoption date. The common Gregorian Calendar adoption date is Friday, 15 October 1582. It's also customizable in the API.
- The week number of a month and a year depends on `firstWeekday` and `minimumDaysInFirstWeek` values. The week starts at `firstWeekday`. Week one must contain `minimumDaysInFirstWeek` number of days.

[^1]: A nice readable version of the Julain - Gregorian converting algorithm: https://en.wikipedia.org/wiki/Julian_day#Julian_day_number_calculation.

Co-authored by @parkera 

### Part 2: `date(from components:)`

The implementation also largely follows that of ICU. The logic follows this pattern generally
- Calculate the Julian day number from a given year, month, day of month number
- Add time field values if they are set
- Adjust for timezone offset

The complexity lies in the following situations
(1) The passed-in date components lacks essential fields to determine the Julian day number
(2) The date components has week-related fields set, but no year, month, day of month fields set
(3) The date components has both week-related fields and month, day of month fields set
(4) The day is near Gregorian calendar transition date

For (1), simply assume the start of the year/month/day if the field is missing. The value would be either 0 or 1, depending on if it's 0 or 1 indexed.
For (2), calculate the julian day of the start of the month. Advance by multiples of the given week number.
For (3), we need to determine which fields take over precedence of others. ICU tracks the order of when the fields are modified; newly modified ones always take precedence over past ones. We have not been using this mechanism at all in existing Calendar_ICU's implementation. Instead we've been setting the fields arbitrarily. Therefore, here we simplify ICU's heuristics by  assuming the timestamps of modified fields are all equal.
For (4), recalculate the date using Julian calendar if the found date is before Gregorian transition date.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -10,10 +10,66 @@
 //
 //===----------------------------------------------------------------------===//
 
+/// Julian date helper
+/// Julian dates are noon-based. Gregorian dates are midnight-based.
+/// JulianDates        `2451544.5 ..< 2451545.5` (Jan 01 2000, 00:00 - Jan 02 2000, 00:00)
+/// maps to JulianDay  `2451545`                 (Jan 01 2000, 12:00)
+extension Date {
+    static let julianDayAtDateReference: Double = 2_451_910.5 // 2001 Jan 1, midnight, UTC
+
+    var julianDate: Double {
+        timeIntervalSinceReferenceDate / 86400 + Self.julianDayAtDateReference
+    }
+
+    var julianDay: Int {
+        Int(julianDate.rounded())
+    }
+
+    init(julianDay: Int) {
+        self.init(julianDate: Double(julianDay))
+    }
+
+    init(julianDate: Double) {
+        let secondsSinceJulianReference = (julianDate - Self.julianDayAtDateReference) * 86400
+        self.init(timeIntervalSinceReferenceDate: secondsSinceJulianReference)
+    }
+}
 /// This class is a placeholder and work-in-progress to provide an implementation of the Gregorian calendar.
 internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
+    
+    let julianCutoverDay: Int// Julian day (noon-based) of cutover
+    let gregorianStartYear: Int
+    let gregorianStartDate: Date
+
+    // Only respects Gregorian identifier
     init(identifier: Calendar.Identifier, timeZone: TimeZone?, locale: Locale?, firstWeekday: Int?, minimumDaysInFirstWeek: Int?, gregorianStartDate: Date?) {
-        fatalError()
+
+        self.timeZone = timeZone ?? .gmt
+        if let gregorianStartDate {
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
I believe this should allow us to implement ISO 8601 formatting and build Date.Description (#179)  on top of this one  (cc @FranzBusch

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -10,26 +10,108 @@
...
+    init(julianDay: Int) {
+        self.init(julianDate: Double(julianDay))
+    }
+
+    init(julianDate: Double) {
+        let secondsSinceJulianReference = (julianDate - Self.julianDayAtDateReference) * 86400
+        self.init(timeIntervalSinceReferenceDate: secondsSinceJulianReference)
+    }
+}
 /// This class is a placeholder and work-in-progress to provide an implementation of the Gregorian calendar.
```
parkera, line 37: Maybe we can remove this?

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -10,26 +10,108 @@
...
-    var firstWeekday: Int
-    
-    var minimumDaysInFirstWeek: Int
-    
+    var _firstWeekday: Int?
+    var firstWeekday: Int {
+        set {
+            if newValue >= 1 && newValue <= 7 {
+                _firstWeekday = newValue
+            }
```
parkera, line 90: Is there any value in a `fatalError` here instead of silently dropping the value?

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
parkera: I guess we still have more work to do, actually.

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

parkera (approved):

PR #306, opened 2023-11-01
A word-based binary integer encoding approach.

oscbyspro:
Salutations! I come in peace ✌️

> [!NOTE]
> This pull request builds on [an earlier contribution](https://github.com/apple/swift-foundation/pull/262).

### Description

These changes aim to achieve two things. The first is making the binary integer text encoding algorithm non-generic so that models such as DoubleWidth perform well. The second is performing mutating division, making arbitrary precision faster. Copying the binary integer's words to a temporary allocation achieves both. Since the new algorithm takes a collection of words, it can also test big integers without big integer models.

### Measurements

I have published a small [package][PRP] with some [benchmarks][PRT]. 

- `v0`: the `old` version
- `v1`: the `new` version
- `v2`: the `new` version + 2x `@inlinable` + 1x `@usableFromInline`

#### The results on my machine in seconds:

|         | v0     | v1    | v2    | [Numberick][NBK] |
|---------|-------:|------:|------:|-----------------:|
| Int     |  1.755 | 0.314 | 0.188 | 0.172            |
| UInt    |  0.261 | 0.302 | 0.181 | 0.166            |
| Int256  | 42.807 | 0.875 | 0.443 | 0.430            |
| UInt256 | 34.844 | 0.815 | 0.428 | 0.426            |

[PRP]: https://github.com/oscbyspro/my-swift-foundation-pull-request
[PRT]: https://github.com/oscbyspro/my-swift-foundation-pull-request/blob/main/Tests/PullRequestTests/Tests.swift
[NBK]: https://github.com/oscbyspro/Numberick

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
+++ b/Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
@@ -43,250 +43,183 @@ extension BinaryInteger {
 // MARK: - BinaryInteger + Numeric string representation
 
 extension BinaryInteger {
-    /// Formats `self` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html) which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
-    ///
-    /// This produces output that (at time of writing) looks identical to the `description` for many `BinaryInteger` types, such as the built-in integer types.  However, the format of `description` is not specifically defined by `BinaryInteger` (or anywhere else, really), and as such cannot be relied upon.  Thus this purpose-built method, instead.
-    internal var numericStringRepresentation: ArraySlice<UInt8> {
-        // It might be worth moving this method into the Swift standard library one day, so that it can be used as the basis for the default `description` instead of duplicating that conversion process.  At least while `description`'s output happens to match this one's.
-        //
-        // This property's type is an ArraySlice rather than a straight ContiguousArray because it's computationally more expensive to size the array exactly right (whether by pre-calculating the required size more precisely, or by dynamically resizing it during execution of the iterative algorithm).  Since this is intended to only be a transient value, for the purposes of translating from BinaryInteger to the ICU libraries, this is the right compromise of runtime speed (CPU time) and memory efficiency (it's usually only off by a byte or two, if that).
-
-        // Fast-path for values that fit into a UInt, as the conversion to a UInt should be virtually free if it's possible (it's essentially just self.words[0]) and there's a specialisation of this function for UInt that's faster.
-        if let fastForm = UInt(exactly: self) {
-            return fastForm.numericStringRepresentation
-        }
-
-        precondition(.zero != self, "Value of zero (for self) should have been handled by fast path, but wasn't.") // Zero isn't handled correctly in the algorithm below (no numbers will actually be emitted) because it's more work to do so, which is unnecessary as the fast path above should handle that case.
-
-        // The algorithm here is conceptually fairly simple.  In a nutshell, the value of self is broken down into Word-sized chunks, each of which is divided by ten repeatedly until the value dimishes to zero.  The remainder of each division is the next digit of the result (starting with the least significant).
-        //
-        // A conceptually simpler approach is to skip the first step of breaking things into Word-sized chunks, and just do the division by 10 on the whole value of `self`.  The difference is performance - native integer division (for machine-word-sized integers) is essentially O(1), whereas division of arbitrary-precision integers is essentially O(log2(bitWidth)) since it's composed of _multiple_ machine-word-sized divides proportionate to its binary magnitude.
-        //
-        // So we replace some of those expensive O(log2(bitWidth)) divides with simpler O(1) divides, by first dividing by the largest multiple of ten such that the remainder fits in a single machine word (UInt), and then using regular integer division CPU instructions to further divide that simple machine-word-sized remainder down into individual digits.
-
-        let (decimalDigitsPerWord, wordMagnitude) = Self.decimalDigitsAndMagnitudePerWord()
-        let negative = 0 > self
-        let maximumDigits = (Self.maximumDecimalDigitsForUnsigned(bitWidth: self.magnitudeBitWidth)
-                             + (negative ? 1 : 0)) // Include room for "-" prefix if necessary.
-        var actualDigits: Int = Int.min // Actually initialised inside the closure below, but the compiler mistakenly demands a default value anyway.
-
-        return ContiguousArray<UInt8>(unsafeUninitializedCapacity: maximumDigits) { buffer, initialisedCount in
-            var tmp = self
-            var wordInsertionPoint = buffer.endIndex - 1
-
-            while .zero != tmp {
-                let (quotient, remainder) = tmp.quotientAndRemainder(dividingBy: wordMagnitude)
-                precondition(.zero == remainder || (negative == (0 > remainder)), "Starting value \(tmp) is \(negative ? "negative" : "positive (or zero)") yet the remainder of division by \(wordMagnitude) is not: \(remainder).  quotientAndRemainder(dividingBy:) is not implemented correctly for \(type(of: self)) (it might be using F-division instead of T-division).") // It's an entirely understandable error for an implementor to use F-division for their integer quotient and remainder, but they're supposed to use T-division.  i.e. the quotient is supposed to be rounded towards zero rather than down (and that effects the modulus correspondingly, since either way the results must satisfy r = d ⨉ (r idiv i) + (r mod i)).  T-division is convenient because its remainder is neatly the value of interest to this algorithm, rather than being offset by the divisor if r is negative.  While it would be technically possible to assume F-division if the remainder's sign doesn't match, the incorrect implementation of quotientAndRemainder(dividingBy:) will probably still break other algorithms and so we shouldn't encourage it.
-
-                // By definition the remainder has to be a single word (since the divisor, `wordMagnitude`, fits in a single word), so we can avoid working on a BinaryInteger generically and just use the first word directly, which is concretely UInt.
-                assert(remainder.magnitudeBitWidth <= Words.Element.bitWidth,
-                       "The remainder of dividing \(tmp) by \(wordMagnitude), \(remainder), should fit into a single word, yet it does not (its magnitude bit width is \(remainder.magnitudeBitWidth) which is greater than the \(Words.Element.bitWidth) bits of Words.Element (\(Words.Element.self))).")
-                var word = remainder.words.first ?? 0
-
-                if negative {
-                    // Luckily for us `words` is defined to be in two's complement form, so we can manually flip the sign.  This doesn't normally work because two's complement cannot represent the positive version of its most negative value, but we know we won't have that here because it's the remainder from division by `wordMagnitude`, which is always going to be less than UInt.max because `wordMagnitude` itself has to fit into UInt (and the remainder of division is always at least one smaller than the divisor).
-                    // Note that for a word of zero (no remainder) this does technically overflow but it's intentional - zero is special since there's no distinct representation for -0 vs +0, but ~UInt(0) &+ 1 is 0, conveniently!  So we can trade a conditional (for avoiding this block if `word` is .zero) for two trivial arithmetic instructions.
-                    word = ~word &+ 1
-                }
-
-                let digitsAdded = word.numericStringRepresentation(intoEndOfBuffer: &buffer[...wordInsertionPoint])
-
-                if .zero != quotient { // Not on the last word, so need to fill in leading zeroes etc.
-                    wordInsertionPoint -= decimalDigitsPerWord
-
-                    let leadingZeroes = decimalDigitsPerWord - digitsAdded
-                    assert(0 <= leadingZeroes, "Negative leading zeroes \(leadingZeroes)!  Expected \(decimalDigitsPerWord) digits per word and added \(digitsAdded).")
-
-                    if 0 < leadingZeroes {
-                        buffer[(wordInsertionPoint + 1)...(wordInsertionPoint + leadingZeroes)].initialize(repeating: UInt8(ascii: "0"))
-                    }
-                } else {
-                    wordInsertionPoint -= digitsAdded
-                }
-
-                tmp = quotient
-            }
-
-            if negative {
-                buffer[wordInsertionPoint] = UInt8(ascii: "-")
-                wordInsertionPoint -= 1
-            }
-
-            actualDigits = wordInsertionPoint.distance(to: buffer.endIndex) - 1
-
-            let unusedDigits = maximumDigits - actualDigits
-            assert(0 <= unusedDigits, "Negative unused digits \(unusedDigits)!  Expected at most \(maximumDigits) digit(s) but emitted \(actualDigits).")
-
-            if 0 < unusedDigits {
-                buffer[0..<unusedDigits].initialize(repeating: 0) // The buffer is permitted to be partially uninitialised, but only at the end.  So we have to initialise the unused portion at the start, in principle.  Technically this probably doesn't matter given we never subsequently read this part of the buffer, but there's no way to express that such that the compiler can ensure it for us.
-            }
-
-            initialisedCount = maximumDigits
-        }[(maximumDigits - actualDigits)...]
-    }
     
-    /// - Parameter bitWidth: The bit width of interest.  Must be zero or positive.
-    /// - Returns: The maximum number of decimal digits that an unsigned value of the given bit width may contain.
-    internal static func maximumDecimalDigitsForUnsigned(bitWidth: Int) -> Int {
-        guard 0 < bitWidth else { return 0 }
-        guard 1 != bitWidth else { return 1 } // Algorithm below only works for bit widths of _two_ and above.
-
-        return Int((Double(bitWidth) * log10(2)).rounded(.up)) + 1 // https://www.exploringbinary.com/number-of-decimal-digits-in-a-binary-integer
-    }
-    
-    /// The bit width of the magnitude of `self`.
+    /// Formats `self` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html)
+    /// which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
     ///
-    /// This is useful for determining how many bits are needed to store the magnitude of `self` (an unsigned integer) _without_ actually determining the magnitude (via the `magnitude` property) since that is relatively expensive (in memory if not also runtime, depending on the size and implementation of the underlying type).
+    /// This produces output that (at time of writing) looks identical to the `description` for
+    /// many `BinaryInteger` types, such as the built-in integer types.  However, the format of
+    /// `description` is not specifically defined by `BinaryInteger` (or anywhere else, really),
+    /// and as such cannot be relied upon.  Thus this purpose-built method, instead.
     ///
-    /// It is never less than one.
-    internal var magnitudeBitWidth: Int {
-        // `BinaryInteger` does provide a `bitWidth` property which could be used to help with this, but for three things:
-        //
-        //   1. For `FixedWidthInteger`s it returns the fixed (maximum) size of the type, not the (minimum) size required to represent `self`.
-        //
-        //   2. It returns the size of the value in two's complement representation, which includes the sign bit that we don't care about.  But we can't just subtract one from that value [for signed types], because for negative powers of two the signed form is one bit shorter than the [unsigned] magnitude's (signed types can represent -(2^N) through +(2^N)-1 - note the asymmetry).  That special case can technically be handled, but it requires determining if `self` is a negative power of two which is relatively expensive for large `BinaryInteger`s.
-        //
-        //   3. Some `BinaryInteger` implementations implement `bitWidth` wrong, due to its terse and ambiguous documentation.  e.g. some are also `FixedWidthInteger`s yet _don't_ return the maximum bit width of the type (or vice versa), some have off-by-one errors for negative powers of two, etc.  Although it's not this code's responsibility to allow for implementation errors, it's nice to.
-        //
-        // So, it's both necessary and preferable (respectively) to just examine `words` directly.
-
-        if .zero == self {
-            return 1
-        } else if .zero <= self {
-            // Find the highest-order word with any bits set, determine the overall index of the highest set bit, and return that plus one (index to count conversion).
-
-            for (i, word) in words.reversed().enumerated() {
-                if .zero != word {
-                    let fullWidth = type(of: word).bitWidth
-                    return ((words.count - i - 1) * fullWidth) + (fullWidth - word.leadingZeroBitCount)
-                }
-            }
-
-            preconditionFailure("\(type(of: self)) \(self) compared as not zero yet all its words are zero.")
-        } else { // `self` is negative.
-            // Perform two's complement one word at a time, keeping track of the index of the highest set bit seen so far.  After enumerating all the words, return that index plus one (index to count conversion).
-
-            var carryingOne = true // Covers both the initial +1 (as part of two's complement) and overflow between words.
-            var indexOfHighestSetBitSeenSoFar = 0
-
-            for (i, word) in words.enumerated() {
-                var positiveWord = ~word
-
-                if carryingOne {
-                    (positiveWord, carryingOne) = positiveWord.addingReportingOverflow(1)
-                }
-
-                let fullWidth = type(of: word).bitWidth
-
-                if carryingOne {
-                    indexOfHighestSetBitSeenSoFar += fullWidth
-                } else if .zero != positiveWord {
-                    indexOfHighestSetBitSeenSoFar = (fullWidth * i) + (fullWidth - positiveWord.leadingZeroBitCount) - 1
-                }
-            }
-
-            return indexOfHighestSetBitSeenSoFar + 1
-        }
-    }
-
-    /// Determines the magnitude (the largest decimal magnitude that fits in Word, e.g. 100 for UInt8) and "maximum" digits per word (e.g. two for UInt8).
-    ///
-    /// Note that 'maximum' in this case is context-specific to the `numericStringRepresentation` algorithm.  It is not necessarily the maximum digits required for _any_ Word, but rather any value of Word type which is less than the maximum decimal magnitude.  Typically this is one digit less.
-    internal // For unit test accessibility, otherwise would be fileprivate.
-    static func decimalDigitsAndMagnitudePerWord() -> (digits: Int, magnitude: Self) {
-        // This method cannot be defined statically because it depends on the types of both Self and Word.  The compiler can in principle fold this down to the resulting values at compile time - since it knows the concrete types for any given call site - and then just inline those into the caller.
-
-        // First, a fast-path that works for any type (for `Self`) which can (essentially) represent a UInt (or larger).
-        let guessDigits = Int(Double(Words.Element.bitWidth) * log10(2))
-        let guessMagnitude = pow(10, Double(guessDigits))
-
-        if let magnitudeAsSelf = Self(exactly: guessMagnitude) {
-            return (guessDigits, magnitudeAsSelf)
-        }
-
-        // Alas `Self` is smaller than UInt, so fall back to a truly generic - but slower - algorithm to find the results.  This is because BinaryIntegers - unlike e.g. FixedWidthIntegers - don't provide APIs for questions like "what is the maximum bit width of Self?".
-
-        var count = 1
-        var value: Words.Element = 1
-
-        while true {
-            var (nextValue, overflowed) = value.multipliedReportingOverflow(by: 10)
-
-            // Words.Element might be wider than the actual type, e.g. if Self is UInt8 (and UInt is not).  The magnitude is limited by the smallest of the two.
-            if !overflowed && nil == Self(exactly: nextValue) {
-                overflowed = true
-            }
-
-            if !overflowed || .zero == nextValue {
-                count += 1
-            }
-
-            if overflowed {
-                return (count - 1, Self(value))
-            }
-
-            value = nextValue
-        }
+    public var numericStringRepresentation: ArraySlice<UInt8> {
+        numericStringRepresentationForBinaryInteger(words: self.words, isSigned: Self.isSigned)
     }
 }
 
-extension UInt {
-    /// Formats `self` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html) which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
-    ///
-    /// This is intended to be used only by the `numericStringRepresentation` property (both the specialised form below and the generic one for all BinaryIntegers, above).  Prefer the `numericStringRepresentation` property for all other use-cases.
-    ///
-    /// - Parameter intoEndOfBuffer: The buffer to write into, which _must_ contain enough space for the result.  The formatted output is placed into the _end_ of this buffer ("right-aligned", if you will), though the output is numerically still left-to-right.  The contents of this buffer do not have to be pre-initialised.
-    /// - Returns: How many entries (UInt8s) of the buffer were used.  Note that zero is a valid return value, as nothing is written to the buffer if `self` is zero (this may be odd but it's acceptable to `numericStringRepresentation` and it simplifies the overall implementation).
-    fileprivate func numericStringRepresentation(intoEndOfBuffer buffer: inout Slice<UnsafeMutableBufferPointer<UInt8>>) -> Int {
-        guard .zero != self else { return 0 } // Easier to special-case this here than deal with it below (annoying off-by-one potential errors).
-
-        var insertionPoint = buffer.endIndex - 1
-        var tmp = self
-
-        // Keep dividing by ten until the value disappears.  Each time we divide, we get one more digit for the output as the remainder of the division.  Since with this approach digits "pop off" from least significant to most, the output buffer is filled in reverse.
-        while .zero != tmp {
-            let (quotient, remainder) = tmp.quotientAndRemainder(dividingBy: 10)
-
-            buffer[insertionPoint] = UInt8(ascii: "0") + UInt8(remainder)
-
-            if .zero != quotient {
-                insertionPoint -= 1
-                assert(insertionPoint >= buffer.startIndex, "Buffer is too small (\(buffer.count) UInt8s) to contain the result.")
-            }
-
-            tmp = quotient
-        }
-
-        return insertionPoint.distance(to: buffer.endIndex)
+/// Formats `words` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html)
+/// which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
+///
+/// - Parameters:
+///   - words: The binary integer's words.
+///   - isSigned: The binary integer's signedness.
+///
+internal // In order to test big integers without big integer models.
+func numericStringRepresentationForBinaryInteger(
+words: some Collection<UInt>, isSigned: Bool) -> ArraySlice<UInt8> {
+    // Copies the words and then passes them to a non-generic, mutating, word-based algorithm.
+    withUnsafeTemporaryAllocation(of: UInt.self, capacity: words.count) { copy in
+        let count =  copy.initialize(fromContentsOf: words)
+        defer{ copy.baseAddress!.deinitialize(count: count) }
+        return numericStringRepresentationForMutableBinaryInteger(words: copy, isSigned: isSigned)
     }
+}
 
-    /// Formats `self` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html) which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
-    ///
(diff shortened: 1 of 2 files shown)


itingliu:
@wadetregaskis Do you want to take a look?

wadetregaskis:
> @wadetregaskis Do you want to take a look?

Yeah, I'm happy to, but it probably won't be until later this week.  I already looked over it a bit when it was sent to me originally, and I think overall it looks good.  I might have some questions, but don't wait on me, if everyone else is happy with this patch already.

iCharlesHu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+    var index = buffer.endIndex
+    
+    repeat {
+        precondition(index > buffer.startIndex)
+        buffer.base.formIndex(before: &index)
+        
+        let remainder: UInt
+        (magnitude, remainder) = magnitude.quotientAndRemainder(dividingBy: 10)
+        buffer.base[index] = UInt8(ascii: "0") &+ UInt8(truncatingIfNeeded: remainder)
+    }   while magnitude != 0 as UInt
```
wadetregaskis, line 150: Is the `as UInt` necessary?  `magnitude` is defined as `Uint` (since it's initialised from `word` which is `UInt`), so there should be no ambiguity nor any significant type inference cost.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+///   - word: An unsigned binary integer's only word.
+///   - buffer: A buffer with enough memory to accommodate the format.
+///
+private func formTrailingNumericStringRepresentationForUnsignedInteger(
+word: UInt, into buffer: Slice<UnsafeMutableBufferPointer<UInt8>>) -> Int {
+    var magnitude = word
+    var index = buffer.endIndex
+    
+    repeat {
+        precondition(index > buffer.startIndex)
```
wadetregaskis, line 144: `precondition` doesn't really log anything useful when it fails - it'd be wise to include a human-readable failure message with relevant context included.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+/// The `exponent` is also the maximum number of decimal digits needed to represent a binary integer
+/// in the range of `0 ..< power`. Another method is used to estimate the total number of digits, however.
+/// This is so that binary integers can be rabased and encoded in the same loop.
+///
+/// ```
+/// 32-bit: (exponent:  9, power:           1000000000)
+/// 64-bit: (exponent: 19, power: 10000000000000000000)
+/// ```
+///
+private func maxDecimalExponentAndPowerForUnsignedIntegerWord() -> (exponent: Int, power: UInt) {
```
wadetregaskis, line 196: As a sidenote, this gets evaluated at compile time and reduced down to the actual constants, which is pretty awesome.  Its use later becomes two inline instructions, to load the immediates 0x8ac7230489e80000 and 0x13 (on x86-64).

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-
-                if carryingOne {
-                    indexOfHighestSetBitSeenSoFar += fullWidth
-                } else if .zero != positiveWord {
-                    indexOfHighestSetBitSeenSoFar = (fullWidth * i) + (fullWidth - positiveWord.leadingZeroBitCount) - 1
-                }
-            }
-
-            return indexOfHighestSetBitSeenSoFar + 1
+    let radix: (exponent: Int, power: UInt) = maxDecimalExponentAndPowerForUnsignedIntegerWord()
```
wadetregaskis, line 96: It might be better to fetch that in the closure below, as it appears the compiler doesn't automatically move it there, so this is incurring unnecessary closure argument passing overhead.

(which is minor - some wasted stack space and register juggling overhead, mainly - but might be precluding further optimisations through inlining)

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+///
+/// [algorithm]: https://www.exploringbinary.com/number-of-decimal-digits-in-a-binary-integer
+///
+/// - Parameter bitWidth: The unsigned binary integer's bit width.
+///
+private func maxDecimalDigitCountForUnsignedInteger(bitWidth: Int) -> Int {
+    // - Int.init(some BinaryFloatingPoint) rounds to zero.
+    // - Double.init(exactly:) and UInt.init(_:) for correctness.
+    // - log10(2.0) is: 1.0011010001000001001101⌈01...⌉ * 2^(-2).
+    return Int(Double(exactly: UInt(bitWidth))! * log10(2.0)) + 1
```
wadetregaskis, line 222: This doesn't work correctly for `bitWidth`s of 1 or less.  Those aren't values that will appear at runtime, but it might be worth putting an assert here to help ensure that just in case some future refactor exposes this to wider use.

And/or document this limitation in the function documentation.

Also, changing the parameter type to `UInt` will at least preclude negative values.

wadetregaskis (commented):
Looks pretty good overall, to me.

Since we're now focusing on performance, I spent a fair bit of time analysing the actual machine code (for x86-64) that this results in (in 'release' build configuration).  It's impressive what the compiler is able to do to reduce all this down to some pretty simple and short loops.

Although if I see it use `lea` to decrement a value by one again, I shall be very annoyed. 😜

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+///   - divisor:  An unsigned binary integer's only word.
+///
+/// - Returns: The `remainder`, which is a value in the range of `0 ..< divisor`.
+///
+private func formQuotientWithRemainderForUnsignedInteger(
+words dividend: Slice<UnsafeMutableBufferPointer<UInt>>, dividingBy divisor: UInt) -> UInt {
+    var remainder = UInt.zero
+    
+    for index in (dividend.startIndex ..< dividend.endIndex).reversed() {
+        (dividend.base[index], remainder) = divisor.dividingFullWidth((high: remainder, low: dividend.base[index]))
```
wadetregaskis, line 168: It’s a shame that this results in the use of `udivti3` from compiler-rt instead of a simple `divq` instruction (on x86-64; presumably there’s an equivalent instruction on AArch64).  Evidently the compiler doesn’t believe that the result will never overflow (`udivti3` does the extra work of checking for and trapping on overflow, which is a waste here).

Perhaps there's some way to convince the compiler that the checks are unnecessary?  Overflow should be impossible in practice, by the inherent constraints of its use here.

oscbyspro:
Oh, neat. I appreciate your feedback. I'll address each point in time.

>[!NOTE]
> The algorithm is fundamentally sign-magnitude, and it would be simpler to call `magnitude.words` and skip the form-two's-complement method. I have tried it, and it performs similarly when inlinable, but DoubleWidth\<T\> performs 3x worse when warring against module boundaries compared calling `self.words`.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-
-            tmp = quotient
-        }
-
-        return insertionPoint.distance(to: buffer.endIndex)
+/// Forms the `two's complement` of a binary integer.
+///
+/// - Parameter words: A binary integer's mutable words.
+///
+private func formTwosComplementForBinaryInteger(words: UnsafeMutableBufferPointer<UInt>) {
```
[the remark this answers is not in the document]
oscbyspro: I don't have a strong opinion about it, but I found your comment a bit perplexing because I don't ascribe signedness to it. The 2's complement format is named after the 2's complement operation `~x &+ 1` (unsigned, fixed-width, `2 ^ x.bitWidth - x`). It took me a while to pinpoint, but I believe this snippets highlight our different viewpoints:


```swift
for x in Int8.min ... Int8.max {
    var magnitude = UInt8(bitPattern: x) // some bit pattern
    
    if type(of: x).isSigned && magnitude >> (x.bitWidth - 1) == 1 {
        magnitude = ~magnitude &+ 1 // form 2's complement when needed
    }
    
    precondition(x.magnitude == magnitude) // bit pattern is magnitude
}

// MARK: Dunno, but how I think you see it:

for x in Int8.min ... Int8.max {
    let magnitude: UInt8
    
    magnitude = UInt8(bitPattern: x < 0 ? ~x &+ 1 : x) // some representation conversion
    
    precondition(x.magnitude == magnitude)
}
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
oscbyspro: Dunno. Swift's BinaryInteger protocol comes with has generic comparison operators, Swift's preferred integer literal is Int, and Xcode does not show what integer literals resolve to.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+            // Because the array is pre-filled with ASCII zeros, we can skip ahead; only the final loop's zeros are trimmed.
+            asciiIndex -= radix.exponent
+            // The entire magnitude has been encoded when the formed quotient is empty.
+        }   while wordsIndex > words.startIndex
+        
+        //  Jump to the most significant digit.
+        asciiIndex = writeIndex
+        
+        //  Add a minus sign to negative values.
+        if  isLessThanZero {
```
itingliu, line 121:
```suggestion
        if isLessThanZero {
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+            ascii.formIndex(before:  &asciiIndex)
```
itingliu, line 122:
```suggestion
            ascii.formIndex(before: &asciiIndex)
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-            initialisedCount = maximumDigits
-        }[(maximumDigits - actualDigits)...]
+/// Formats `words` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html)
+/// which is the required input form for certain ICU functions (e.g. `unum_formatDecimal`).
+///
+/// - Parameters:
+///   - words: The binary integer's words.
+///   - isSigned: The binary integer's signedness.
+///
+internal // In order to test big integers without big integer models.
```
itingliu, line 67:
```suggestion
internal
```

Nit: I'd prefer leave out this kind of comment, especially when it's already an established pattern throughout the project

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+func numericStringRepresentationForBinaryInteger(
```
itingliu, line 68: Would it make sense to declare this inside an extension on `BinaryInteger` so that we can have `words: Self.Words` instead of it being an existential `some Collection<UInt>`? Also a bonus point for less cluttering of global namespace?

oscbyspro: You would lose the ability to unit test it without adding or importing a big integer model. Big integers, as defined by Swift's BinaryInteger protocol, are basically fancy `some Collection<UInt>`. I added this method to allow unit testing using `Array<UInt>` since the need for external big integers was an earlier complaint. The other helper methods are single-use,  however. So you can merge them into one intimidating blob if clutter is an issue.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+words: some Collection<UInt>, isSigned: Bool) -> ArraySlice<UInt8> {
+    // Copies the words and then passes them to a non-generic, mutating, word-based algorithm.
+    withUnsafeTemporaryAllocation(of: UInt.self, capacity: words.count) { copy in
+        let count =  copy.initialize(fromContentsOf: words)
+        defer{ copy.baseAddress!.deinitialize(count: count) }
```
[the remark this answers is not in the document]
oscbyspro: I changed it to the following since [withUnsafeTemporaryAllocation(of:capacity:)](https://github.com/apple/swift/blob/b72f0c95e4f711158b9b1c015cc22fa6653b2c86/stdlib/public/core/TemporaryAllocation.swift#L335) does not guarantee the size of the buffer (although the current implementation covers exactly the requested number of elements).

```swift
internal func numericStringRepresentationForBinaryInteger(
words: some Collection<UInt>, isSigned: Bool) -> ArraySlice<UInt8> {
    // Copies the words and then passes them to a non-generic, mutating, word-based algorithm.
    withUnsafeTemporaryAllocation(of: UInt.self, capacity: words.count) {
        let initializedEndIndex = $0.initialize(fromContentsOf: words)
        let initialized = UnsafeMutableBufferPointer(rebasing: $0[..<initializedEndIndex])
                
        defer {
            initialized.deinitialize()
        }
        
        return numericStringRepresentationForMutableBinaryInteger(words: initialized, isSigned: isSigned)
    }
}
```

oscbyspro:
Hm. The documentation of [Swift.withUnsafeTemporaryAllocation(of:capacity:)](https://developer.apple.com/documentation/swift/withunsafetemporaryallocation(of:capacity:_:)) states:

> The implementation may allocate a larger buffer pointer than is strictly necessary to contain capacity values of type type. The behavior of a program that attempts to access any such additional storage is undefined.

Does that speak of the buffer in the closure, or is it just implementation details leaking through documentation?

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
itingliu: Ah, that makes sense. So, that was what your previous comment was about, but you removed because of my ask, sorry 😬

itingliu (commented):
@glessard Do you want to take a look, specifically about the usage around the `unsafemutableBufferPointer`? And also if we should consider sinking these into stdlib?

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
oscbyspro: That's an astute observation! I'd love to know if there's a better way to do it. I have an open issue in Numberick ([#59](https://github.com/oscbyspro/Numberick/issues/59)) about division being slower than expected in some cases. I wrote a note about udivmodti4 taking up most of the time in Instruments, but I did not know how to proceed beyond that. It's a bit funky that it performs oveflow checking when it truncates the result anyways:

```swift
print(UInt(123).dividingFullWidth((high: 122, low: UInt.max))) // (quotient: UInt.max, remainder: 122)
print(UInt(123).dividingFullWidth((high: 123, low: 00000000))) // (quotient: 00000000, remainder: 000)
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
wadetregaskis: I think the key point is that you can't fit `x.magnitude` into `Self` for any signed `Self`, specifically because of `type(of: x).min`.  e.g. for `Int8` (just to keep the numbers small) `Int8.min` is -128.  The inverse of that is 128, which is greater than `Int8.max` (of 127).  You have to upgrade to `Int16` to represent it _in two's complement form_.

Aside: this is presumably why the type of the `magnitude` property is _not_ `Self`, but rather an associated value defined by each specific type.  All the built-in signed integer types use their unsigned partners for it.

So, the function in question _cannot_ perform two's complement correctly for all values; it's not a generic two's complement implementation.  What it _can_ do is:

1. Invert any _positive_ two's complement-encoded value.
2. Invert any negative, two's complement-encoded value into its positive form _as an unsigned-type of the same bit width_.

(1) isn't relevant to the function's actual use in this file, only (2).  Thus why I suggested a name that is more specific to its intended use (even if it technically ignores another use it can work for, (1)). 

This is all just to try to clarify.  I don't feel that this is a particular important point in the grand scheme of things, either.  It just seems like a sensible little refinement, to me.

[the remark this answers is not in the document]
wadetregaskis: Or at least, not for _all_ signed `BinaryInteger`s.  You can certainly have some where that limitation doesn't apply - e.g. because they're flexible-width - or I suppose hypothetically even a `FixedWidthInteger` that constraints it minimum value to something less (in magnitude) than 2^(bitWidth-1).  But the code in question has to work for _all_ `BinaryInteger`s.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: Yeah, I assume it's fine to rely on this compiler optimisation.  Plus I spent quite a while trying to come up with _any_ alternative that's remotely scalable and elegant, but found nothing.

That's also why in my earlier implementation I specifically commented about that, in the code.  I just hadn't discovered, at that point, that the compiler _does_ in fact optimise it, so I believe my comment was along the lines of "I _hope_ the compiler can optimise this…". 😁

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: I apologise, you're quite right.  I misunderstood the algorithm here (I think re. the rounding behaviour).  I've now tested it empirically and indeed it works as intended.

The value of 1 for bit width 0 is okay - that case shouldn't ever arise in practice anyway - although a `precondition(0 < bitWidth, …)` might be wise (or an assert).

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
wadetregaskis: I was mistaken about `__udivti3` re. the overflow case.  It's doesn't actually trap, it just truncates, as you observed.  That's what one gets for trusting StackOverflow. 😝

For completeness… it's essentially using [`__udivmodti4`](https://github.com/apple/llvm-project/blob/76a94e518e019fa3766b8a95d61f46cc474695d1/compiler-rt/lib/builtins/udivmodti4.c#L100C24-L100C36) because `__udivti3` is basically an alias; it's two instructions: one to zero out r8 to effect a `NULL` for the out parameter that reports the remainder, and then a jump to `__udivmodti4` (which incidentally is a waste of space since `__udivmodti4` is the _very_ next procedure, so it'd flow into it automatically without the `jmp`).

`__udivmodti4` is a little complicated, both at the C level and in the resulting assembly (67 instructions, on x86-64, and branchy).  It does use `divq`, ultimately (it's actually hard-coded to on x86-64, via inline assembly).  But you still have the overhead in the function call and unnecessary checks to use other paths - it really would benefit from being inlined.  I don't know how to make the compiler inline it, though.

Which raises the question:  why on earth isn't it just using `__udivmodti4` directly, since we _do_ care about the remainder here.  It appears that it's using the result of the division to calculate the remainder manually:

```assembly
cmp  r12, rdi ; Check that the index is > 0.  On first entry to this loop it's endIndex.
jle  loc_19eb ; Basically, crash if index is <= 0.
lea  rbx, qword [r12-1] ; Decrement index of current word in `words`.
mov  r13, qword [r15+r12*8-8] ; Load next word from `words` argument.
mov  rdi, r13 ; low half of 128-bit argument "a" (dividend) for __udivti3.  rsi is high half, and is initially set to zero but is the remainder in subsequent iterations.
mov  rdx, r14 ; low half of 128-bit argument "b" (divisor) for __udivti3.
xor  ecx, ecx ; high half of argument "b".  Top 32 bits are known zero from earlier in the function.
call __udivti3 ; rax = a / b
mov  rdi, qword [rbp+var_38] ; words.count.
mov  rcx, rax
mul  r14 ; Quotient * divisor (the function argument, not the argument to __udivti3.
mov  rsi, r13 ; The aforementioned next word from `words`.
sub  rsi, rax ; Remainder = current word - (Quotient * divisor)
mov  qword [r15+r12*8-8], rcx ; Save the quotient over the current word in `words`.
mov  r12, rbx ; Just shifting the current index around registers.
cmp  rdi, rbx ; Check that index is != words.count.
jne  loc_19a0 ; Repeat this whole block while it isn't.
```

I don't see how that could be faster than using `__udivmodti4`, since the latter calculates the remainder no matter what - it merely drops it silently if its 'rem' out argument is `NULL`, as it is in the use via `__udivti3`.  Weird.  Seems like an optimiser bug.

wadetregaskis:
> Hm. The documentation of [Swift.withUnsafeTemporaryAllocation(of:capacity:)](https://developer.apple.com/documentation/swift/withunsafetemporaryallocation(of:capacity:_:)) states:
> 
> > The implementation may allocate a larger buffer pointer than is strictly necessary to contain capacity values of type. The behavior of a program that attempts to access any such additional storage is undefined.
> 
> Does that speak of the buffer in the closure, or is it just implementation details leaking through documentation?
> 
> It sounds like it refers to the buffer in the closure because that's how Swift uses the term **capacity** elsewhere:
> 
> ```swift
> for capacity in 0 ... 32 {
>     String(unsafeUninitializedCapacity: capacity) {
>         print("capacity: \(capacity), count: \($0.count)")
>         precondition($0.count == Swift.max(16, capacity))
>         return 0
>     }
> }
> ```

I believe it means that `$0.count` is guaranteed to be >= `capacity`, not merely ==.

It makes sense as such.  Hypothetically for a capacity <= 15 it can provide a buffer pointing to the actual stack space allocated for the `String` (using the `String`'s "inline" form to avoid a heap allocation), so it might as well just pass 15 in that case.  And of course for its non-inline forms its heap allocations have to be multiples of 16 at least (malloc on Apple platforms has never supported allocation granularities smaller than 16 bytes).

It turns out [the current implementation](https://github.com/apple/swift/blob/cabb5e109fa373ff82b3d335fd6055d93ada98d8/stdlib/public/core/String.swift#L560) _does_ do this optimisation - of directly initialising the stack memory for capacities <= 15 - but oddly it doesn't pass the _actual_ capacity in that case (which would be 15), instead pretending it's always exactly the requested capacity.  An oversight, perhaps?

Talking in 64-bit here.  32-bit the same principles apply but the numbers might be different.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
oscbyspro: It performs the *two's complement operation* `~x &+ 1`, which is defined for all binary integers. It's signedness agnostic, and it does not reinterpret anything. Whether the result of `~x &+ 1` is useful to us is determined at the call site.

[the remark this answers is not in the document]
oscbyspro: This is only to clarify that you cannot find an input where `~x &+ 1` does not fit. In the one place it's used, the `words` has—to your point—been reinterpreted as an unsigned `magnitude`. The type system isn't there to help, so there's no bit cast reinterpreting it. I get the feeling that you'd be happier with renaming the `magnitude` as `reinterpretedUnsignedMagnitude` or somesuch, instead of renaming this method :man_shrugging:

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+///
+/// - Parameters:
+///   - words: The binary integer's words.
+///   - isSigned: The binary integer's signedness.
+///
+internal // In order to test big integers without big integer models.
+func numericStringRepresentationForBinaryInteger(
+words: some Collection<UInt>, isSigned: Bool) -> ArraySlice<UInt8> {
+    // Copies the words and then passes them to a non-generic, mutating, word-based algorithm.
+    withUnsafeTemporaryAllocation(of: UInt.self, capacity: words.count) { copy in
```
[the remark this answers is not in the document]
oscbyspro: Hm. That is fair. I should measure some small values and perhaps retrofit the fast path you wrote. As a side note, it is possible to return a String instead of an ArraySlice through an additional temporary allocation. Given that String and withUnsafeTemporaryAllocation(\_:) can stack allocate, it might be beneficial. I haven't measured the difference, however.

oscbyspro:
I have added a signed and unsigned tiny integer benchmark to see whether fast pathing is worth it. I've also measured an `@inlinable` version of the original implementation to see what is and isn't a module boundary issue. The original fast path does, indeed, improve the non-`@inlinable` performance of tiny integers, but Swift's standard library beats it. It seems, to me, that returning a String and stack allocating small integers is more impactful.

> [!NOTE]
> I've changed to Xcode 15.0.1, so everything is faster.

glessard:
> > The implementation may allocate a larger buffer pointer than is strictly necessary to contain capacity values of type. The behavior of a program that attempts to access any such additional storage is undefined.
> 
> Does that speak of the buffer in the closure, or is it just implementation details leaking through documentation?

I think this is a leak of implementation details, coupled with an API documentation deserving some improvements.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,187 @@ extension BinaryInteger {
...
+            asciiIndex -= radix.exponent
+            // The entire magnitude has been encoded when the formed quotient is empty.
+        }   while wordsIndex > words.startIndex
+        
+        // Jump to the most significant digit.
+        asciiIndex = writeIndex
+        
+        // Add a minus sign to negative values.
+        if isLessThanZero {
+            ascii.formIndex(before: &asciiIndex)
```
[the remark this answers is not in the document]
oscbyspro: Sorry, I didn't see your question earlier. The `capacity` is given by the following expression so please have a look at [number-of-decimal-digits-in-a-binary-integer](https://www.exploringbinary.com/number-of-decimal-digits-in-a-binary-integer). 

```swift
let capacity = maxDecimalDigitCountForUnsignedInteger(bitWidth: words.count * UInt.bitWidth) + (isLessThanZero ? 1 : 0)
```

Since `capacity` is only an upper bound, I've used the fullest possible width as an uncomplicated approximaton (it can be trimmed a bit). In the loop, we then divide by `10` repeatedly and write each digit back-to-front. The loop is on `repeat`, so we write at least one. The minimum capacity is `1` when `words.count == 0`, and `1` is added when `isLessThanZero == true`.

```
let isLessThanZero = isSigned && Int(bitPattern: words.last ?? 0) < 0 // tests the last bit when isSigned == true
```

#### Important edge case thing-y

The loop branchlessly decrements `asciiIndex` by `radix.exponent` (the digits written plus zeros for padding) each iteration, so it might be negative at the end of the loop. This is one reason why we set it to `writeIndex` afterwards, the other is trimmings the last iteration's padding. This assumes `Index == Int` (alternatives exist).

oscbyspro:
Quick update: I'm working on a new version that should be speedy for small integer too. I'll update this pull request within the next few days or so. It looks promising. It involves returning a String instead of an ArraySlice, so I hope that's OK. The ASCII sequence can be accessed using `withUTF8(_:)`.

oscbyspro:
As a side note, I saw that BinaryInteger's `formatted()` method casts to Int, which traps big integers:

```swift
extension BinaryInteger {
    /// Format `self` using `IntegerFormatStyle()`
    public func formatted() -> String {
        IntegerFormatStyle().format(Int(self))
    }
}
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
oscbyspro: I see that  puts a lot of emphesis on avoiding type assumptions. My concern was more about the expression resolving to an Int-to-UInt conversion. In any case, I've changed it to `.zero` since Xcode shows it resolving to the correct type.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
oscbyspro: The zero bit width case occurs when an integer's `words` collection is empty. I don't have an integer like that, but you could perhaps justify an unsigned big integer normalizing zero to `[]`. In any case, I have unit tests for empty collections, which result in `"0"`.

oscbyspro:
I have added the 1,000,000th Fibonacci sequence element (210k decimal digits) to the table :smiley:

glessard:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+///
+/// This method consumes the `words` such that the buffer is filled with zeros when it returns.
+///
+private func numericStringRepresentationForMutableBinaryInteger(
+words: UnsafeMutableBufferPointer<UInt>, isSigned: Bool) -> ArraySlice<UInt8> {
+    //  Note that negative values are in two's complement form.
+    let isLessThanZero = isSigned && Int(bitPattern: words.last ?? 0) < 0
+    //  The magnitude is formed when the words represent a negative value.
+    if  isLessThanZero {
+        formTwosComplementForBinaryInteger(words: words)
```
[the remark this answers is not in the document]
oscbyspro: While it's possible to move this transformation to the `BinaryInteger` extension, it's only viable if `@inlinable` because it won't be specialized otherwise. The unspecialized performance penalty of evaluating something like `Self.isSigned && self < Self.zero` is quite apparent for `Int256`, but it can also seen in `Int`. If you are all OK with `@inlinable`, removing this method and calling `magnitude.words` is the way to go about it.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-            }
+/// Writes `word` in "Numeric string" format (https://speleotrove.com/decimal/daconvs.html)  to
+/// the trailing edge of `buffer`, then returns the least significant `buffer` index written to.
+///
+/// - Parameters:
+///   - word: An unsigned binary integer's only word.
+///   - buffer: A buffer with enough memory to accommodate the format.
+///
+private func formTrailingNumericStringRepresentationForUnsignedInteger(
+word: UInt, into buffer: Slice<UnsafeMutableBufferPointer<UInt8>>) -> Int {
```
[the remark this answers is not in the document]
oscbyspro: I dropped this function and inlined its body in `v2` because it was a bit arcane.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-                    indexOfHighestSetBitSeenSoFar += fullWidth
-                } else if .zero != positiveWord {
-                    indexOfHighestSetBitSeenSoFar = (fullWidth * i) + (fullWidth - positiveWord.leadingZeroBitCount) - 1
-                }
-            }
-
-            return indexOfHighestSetBitSeenSoFar + 1
+    let radix: (exponent: Int, power: UInt) = maxDecimalExponentAndPowerForUnsignedIntegerWord()
+    let capacity = maxDecimalDigitCountForUnsignedInteger(bitWidth: words.count * UInt.bitWidth) + (isLessThanZero ? 1 : 0)
+    var ascii = ContiguousArray(repeating: UInt8(ascii: "0"), count: capacity) // Set initial ASCII zeros (see later steps).
```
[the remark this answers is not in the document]
oscbyspro: It's simpler, and it removes a branch from the loop. I tried it both ways and did not see a benefit to in-loop fiddling.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
+/// 32-bit: (exponent:  9, power:           1000000000)
+/// 64-bit: (exponent: 19, power: 10000000000000000000)
+/// ```
+///
+private func maxDecimalExponentAndPowerForUnsignedIntegerWord() -> (exponent: Int, power: UInt) {
+    var exponent = 1 as  Int
+    var power = 0010 as UInt
+    
+    while true {
+        let next = power.multipliedReportingOverflow(by: 10)
```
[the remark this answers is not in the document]
oscbyspro: ...and doing nothing saves 10 characters.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
oscbyspro: I changed the call site to the following for clarity:

```swift
//  We reinterpret the words as an unsigned binary integer.
var magnitude = /* consume */ words
//  Note that negative values are in two's complement form.
let isLessThanZero = isSigned && Int(bitPattern: magnitude.last ?? .zero) < .zero
//  The **unsigned** magnitude is formed when the words represent a negative value.
if  isLessThanZero {
    formTwosComplementForBinaryInteger002(words: magnitude)
}
```

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
```diff
@@ -43,250 +43,183 @@ extension BinaryInteger {
...
-                    indexOfHighestSetBitSeenSoFar = (fullWidth * i) + (fullWidth - positiveWord.leadingZeroBitCount) - 1
-                }
-            }
-
-            return indexOfHighestSetBitSeenSoFar + 1
+    let radix: (exponent: Int, power: UInt) = maxDecimalExponentAndPowerForUnsignedIntegerWord()
+    let capacity = maxDecimalDigitCountForUnsignedInteger(bitWidth: words.count * UInt.bitWidth) + (isLessThanZero ? 1 : 0)
+    var ascii = ContiguousArray(repeating: UInt8(ascii: "0"), count: capacity) // Set initial ASCII zeros (see later steps).
+    
+    var wordsIndex = words.endIndex
```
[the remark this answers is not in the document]
oscbyspro: Slicing sounds reasonable, even if it does not change the algorithm. I've made it so the magnitude is sliced (rebased) instead of indexed. It's a bit slower, but not by much, and I'm not sure why. It was an opportunity to reinterpret the words as an unsigned binary integer, at least by variable name, so it's probably worth it for clarity. The break condition is now `magnitude.isEmpty`, for example.

Sources/FoundationInternationalization/Formatting/Number/BinaryInteger+FormatStyle.swift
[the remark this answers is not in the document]
oscbyspro: For what it's worth, no index goes out of bounds or assumes typing in v2. I've also replaced all assertions with preconditions since the difference in performance was insignificant.

oscbyspro:
I sense no enthusiasm—and nobody has made it a priority—should I close it? I know  has a habit of keeping things in limbo, but I don't. I'd much rather cross this PR off my list and go somewhere my contributions are appreciated :man_shrugging:

itingliu (approved):
Apologies for the delayed reply. I personally do want to take this one ✌️ 

@swift-ci please test

itingliu:
@swift-ci please test

PR #310, opened 2023-11-09
Swift testing proof of concept

aleksproger:
- Added dependency on `swift-testing`
- Migrated `BufferViewTests` to `swift-testing` as a showcase
- Note that `swift-testing` requires 5.10 toolchain to work, so the `swift-foundation` toolchain then should be bumped accordingly

### DIFF
--- a/Tests/FoundationEssentialsTests/BufferViewTests.swift
+++ b/Tests/FoundationEssentialsTests/BufferViewTests.swift
@@ -19,17 +19,16 @@ import TestSupport
 #endif
 
 final class BufferViewTests: XCTestCase {
+    func testAll() async {
+        await XCTestScaffold.runAllTests(hostedBy: self)
+    }
 
     func testOptionalStorage() {
-        XCTAssertEqual(
-            MemoryLayout<BufferView<UInt8>>.size, MemoryLayout<BufferView<UInt8>?>.size
-        )
-        XCTAssertEqual(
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

Tests/FoundationEssentialsTests/BufferViewTests.swift
```diff
@@ -19,17 +19,16 @@ import TestSupport
 #endif
 
 final class BufferViewTests: XCTestCase {
```
iCharlesHu, line 21: Since this is a proof of concept, could you fully adopt `swift-testing` for this file? For example, we should switch to using `@Suite` and `@Testing` etc with a struct and avoid using `XCTestCase` directly?

Tests/FoundationEssentialsTests/BufferViewTests.swift
```diff
@@ -19,17 +19,16 @@ import TestSupport
...
+    func testAll() async {
+        await XCTestScaffold.runAllTests(hostedBy: self)
+    }
```
iCharlesHu, line 24: This should be in it's own file so other `swift-testing` tests can use it as well.

parkera:
Thanks @aleksproger for this - we are pushing forward on this in future PRs and also working through the cyclical dependency in swift-foundation <-> swift-testing.

parkera:
( #545 )

PR #311, opened 2023-11-09
[GregorianCalendar] Implement TimeZone support for `date(from components: DateComponents)`

itingliu:
To support DST-observing time zone, add a helper function for TimeZone to return the raw offset and DST offset individually so we can fine-tune the behavior for the time during the skipped time frame and the repeated time frame.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -90,7 +90,6 @@ enum ResolvedDateComponents {
 
     init(dateComponents components: DateComponents) {
         var (year, month) = Self.yearMonth(forDateComponent: components)
-        let minMonth = 1
         let minWeekdayOrdinal = 1
         if let d = components.day {
             if components.yearForWeekOfYear != nil, let weekOfYear = components.weekOfYear {
@@ -125,7 +124,6 @@ enum ResolvedDateComponents {
     }
 
     init(preferComponent c: Calendar.Component, dateComponents components: DateComponents) {
-        let minMonth = 1
         let minWeekdayOrdinal = 1
         switch c {
         case .day:
@@ -464,7 +462,7 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
     }
 
 
-    func date(from components: DateComponents, inTimeZone timeZone: TimeZone?, resolvedComponents: ResolvedDateComponents? = nil) -> Date? {
+    func date(from components: DateComponents, inTimeZone timeZone: TimeZone, resolvedComponents: ResolvedDateComponents? = nil) -> Date? {
 
         let resolvedComponents = resolvedComponents ?? ResolvedDateComponents(dateComponents: components)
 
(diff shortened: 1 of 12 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -464,7 +462,7 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
     }
 
 
-    func date(from components: DateComponents, inTimeZone timeZone: TimeZone?, resolvedComponents: ResolvedDateComponents? = nil) -> Date? {
```
itingliu, line 467: Drive-by fix: Remove the optional as we already always pass a non-nil timeZone

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -90,7 +90,6 @@ enum ResolvedDateComponents {
 
     init(dateComponents components: DateComponents) {
         var (year, month) = Self.yearMonth(forDateComponent: components)
-        let minMonth = 1
```
itingliu, line 93: Drive-by fix: Remove unused variable

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -498,20 +496,14 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
             secondsInDay += Double(nanosecond) / Double(nano_coef)
         }
 
-        let timeZoneOffset: Int
```
itingliu, line 501: This part is the real relevant fix

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1729,7 +1729,7 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
         let start = date - 48.0 * 60.0 * 60.0
 
 
-        guard let nextDSTTransition = _locked_nextDaylightSavingTimeTransition(startingAt: start, limit: start + 4 * 8600 * 1000.0) else {
```
itingliu, line 1732: Drive-by fix: It was 86400 before The Rewrite™️

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationEssentials/TimeZone/TimeZone_Protocol.swift
```diff
@@ -17,6 +17,12 @@ package protocol _TimeZoneProtocol : AnyObject, Sendable, CustomDebugStringConve
     
     var identifier: String { get }
     func secondsFromGMT(for date: Date) -> Int
+
+    /// The returned `daylightSavingOffset` is different from that returned by `daylightSavingTimeOffset(:)` in that this interprets `Date` in GMT.
+    /// Essentially this is equivalent to adjusting `date` to this time zone using `rawOffset`, then passing the adjusted date to `daylightSavingTimeOffset(for: <adjusted date>)`.
+    /// This also handles the skipped time frame on DST start day differently from `daylightSavingTimeOffset(:)`, where dates in the skipped time frame are considered *not* in DST here, hence the DST offset would be 0.
+    func rawAndDaylightSavingTimeOffset(forGMTDate date: Date) -> (rawOffset: Int, daylightSavingOffset: Int)
```
parkera, line 24: I have a naming nitpick, which you can take or leave. Calling this a "GMT" date may be confusing, since (as you well know) `Date` doesn't really have a time zone associated with it. I would just call it:

```
func rawAndDaylightSavingTimeOffset(for date: Date) -> (rawOffset: Int, daylightSavingOffset: Int)
```

parkera: Interestingly the below `daylightSavingTimeOffset` returns a `TimeInterval`. I imagine the offset is always an integer, but it may be useful to keep that for consistency.

itingliu: I know!! I actually started with what you suggested, but I found it confusing when comparing this to the result of `daylightSavingTimeOffset(:)`. However I also agree that this current name doesn't help much disambiguating anyways, so I'm going to go back to the previous name and just leave the comment there.

itingliu: And yes `daylightSavingTimeOffset(:)` is always an integer if you look at the implementation. But that's an API so we can't change that anyways

itingliu:
@swift-ci

Sources/FoundationEssentials/TimeZone/TimeZone_Protocol.swift
itingliu: Sure I don't mind having it return a `TimeInterval` for the sake of consistency. We're converting it to Double anyways to add it to a `Date`

itingliu:
@swift-ci please test

parkera (approved):

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #317, opened 2023-11-15
[WIP Gregorian Calendar] Implement `date(byAdding components: DateComponents, to date: Date, wrappingComponents: Bool)`

itingliu:
The implementation largely follows ICU's calendar::add and calendar::roll.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -179,6 +179,10 @@ enum ResolvedDateComponents {
 /// This class is a placeholder and work-in-progress to provide an implementation of the Gregorian calendar.
 internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
     
+    let kSecondsInWeek = 604_800
+    let kSecondsInDay = 86400
+    let kSecondsInHour = 3600
+
     let julianCutoverDay: Int// Julian day (noon-based) of cutover
     let gregorianStartYear: Int
     let gregorianStartDate: Date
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

parkera (approved):

parkera:
Awesome!

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -727,9 +731,521 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
...
+
+        case .month:
+            var dc = dateComponents(allComponents, from: date, in: timeZone)
+            dc.month = (dc.month ?? 0) + amount
+            capDay(in: &dc) // adding 1 month to Jan 31 should return Feb 29, not Feb 31
+            let resolvedComponents = ResolvedDateComponents(preferComponent: .month, dateComponents: dc)
+            return self.date(from: dc, inTimeZone: timeZone, resolvedComponents: resolvedComponents)!
+
+        case .quarter:
+            // TODO: This isn't supported in Calendar_ICU either. We should do it here though.
```
parkera, line 827: I wonder if part of the reason this is unsupported (for adding, at least) is that it may be unclear what adding "1 quarter" to a date might mean. Is it the same as 3 months? I think it's far more likely to use in matching where you want to know when the start of some specific quarter is, for example.

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -18,6 +18,44 @@ import TestSupport
 @testable import FoundationEssentials
 #endif
 
+extension DateComponents {
+    init(field: Calendar.Component, value: Int) {
```
parkera, line 22: Hah - I added this in my other PR here:

https://github.com/apple/swift-foundation/pull/322/files#diff-10a4fcffe1d9245a1a5371a1ce086a0a67726b8dfde2454781613fecea34469bR82

```swift
    package init?(component: Calendar.Component, value: Int) {
        switch component {
        case .calendar, .timeZone, .isLeapMonth:
            return nil
        default:
            setValue(value, for: component)
        }
    }
```

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -686,6 +686,46 @@ final class CalendarBridgingTests : XCTestCase {
 }
 #endif
 
+
+extension DateComponents {
+    init(field: Calendar.Component, value: Int) {
```
parkera, line 691: We actually don't need this in the tests - they import as testable so they can just use the internal one in the package.

parkera: Perhaps we should consider adding this as API, though - it's clearly useful.

itingliu: But yours from https://github.com/apple/swift-foundation/pull/322/files#diff-10a4fcffe1d9245a1a5371a1ce086a0a67726b8dfde2454781613fecea34469bR82 isn't merged yet so I can't use it yet!

itingliu: I'll replace this later when yours is ready.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
itingliu: I don't think it's an unreasonable API though. It's as useful as figuring out the date when someone says, "Let's revisit the plan **next quarter**". I think it's reasonable to treat adding one quarter as adding three months, but whether we want to preserve the day of the month or the weekday is open to discussion.

Tests/FoundationInternationalizationTests/CalendarTests.swift
parkera: Maybe just copy/paste the implementation so the merge is obvious?

itingliu: Done!

PR #319, opened 2023-11-17
Workaround compatibility issue with nanoseconds fix

parkera:
We recently (#289) updated this code to compare nanoseconds correctly, but we found some existing apps depending on the wrong order. This puts in a compatibility check (framework only) to restore the broken behavior until those apps have a chance to update.


parkera:
@swift-ci test

itingliu (approved):

PR #320, opened 2023-11-17
Workaround type checker timeout

parkera:
A recent change in the type checker seems to have caused this code to time out on build. For now specify the explicit type.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #321, opened 2023-11-17
Propose API improvements to Calendar

parkera:
In macOS 14 / iOS 17, `Calendar` was rewritten entirely in Swift. One of the many benefits of this change is that we can now more easily create Swift-specific `Calendar` API that feels more natural than the existing `enumerate` methods. In addition, we are taking the opportunity to add a new field to the `DateComponents` type to handle one case that was only exposed via the somewhat esoteric CoreFoundation API `CFCalendarDecomposeAbsoluteTime`.

PR #322, opened 2023-11-17
Calendar Sequence API

parkera:
See #321 for API proposal.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -84,11 +84,9 @@ extension DateComponents {
             guard calendar.value(v, isValidFor: .year) else { return false }
         }
         
-        if #available(FoundationPreview 0.4, *) {
-            if let v = self.dayOfYear {
-                hasAtLeastOneFieldSet = true
-                guard calendar.value(v, isValidFor: .dayOfYear) else { return false }
-            }
+        if let v = self.dayOfYear {
+            hasAtLeastOneFieldSet = true
+            guard calendar.value(v, isValidFor: .dayOfYear) else { return false }
         }
 
         if let v = self.quarter {
@@ -162,9 +160,7 @@ extension DateComponents {
         // A note on performance: this approach is much faster than using key paths, which require a lot more allocations.
         if self.era != nil { return .era }
         if self.year != nil { return .year }
-        if #available(FoundationPreview 0.4, *) {
-            if self.dayOfYear != nil { return .dayOfYear }
-        }
+        if self.dayOfYear != nil { return .dayOfYear }
         if self.quarter != nil { return .quarter }
         if self.month != nil { return .month }
         if self.day != nil { return .day }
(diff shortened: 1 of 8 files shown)


parkera:
@swift-ci test

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -261,86 +279,182 @@ private func _handleCalendarResultNotFound(date: Date, calendar: Calendar, comps
...
+                    // We have nothing to add
+                    return nil
+                }
+                
+                guard let startAt = previousResult else {
+                    // We have finished
+                    return nil
+                }
+                                
+                let next = calendar.date(byAdding: components, to: startAt, wrappingComponents: wrappingComponents)
```
davedelong, line 436: This approach is susceptible to the "1+1≠2" problem in calendar math. By always adding from the previous result, we're allowing for drift to occur.

As a simple example, if the start date is "January 31" and the stride is "1 month", then the first addition will result in "February 28". The addition after that will result in "March 28", which has drifted from the likely expected value of "March 31".

The typical way to solve this is to always add from the start date, and to scale the value being added. So you start by adding one month, then two months, then three months etc; instead of adding one month every time.

This, however, gives rise to other questions about which units can be used as a stride.

`DateComponents` allows you to specify values that aren't relevant to addition, such as the `.calendar`, `.timezone`, `.weekday`, etc. It seems to me that the sequence and iterator initializers should include checks to `fatalError()` if the components include values that are illogical to use as a stride.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -261,86 +279,182 @@ private func _handleCalendarResultNotFound(date: Date, calendar: Calendar, comps
...
+            }
+        }
+        
+        public func makeIterator() -> Iterator {
+            return Iterator(calendar, start: start, range: range, matching: matchingComponents, matchingPolicy: matchingPolicy, repeatedTimePolicy: repeatedTimePolicy, direction: direction)
+        }
+    }
+    
+    /// A `Sequence` of `Date`s, calculated by iterative addition of `DateComponents`.
+    struct DatesByAdding : Sendable, Sequence {
```
davedelong, line 391: All of these types appear to be missing visibility annotations.

davedelong: Ah of course; I glossed over the type-erased return types. Thanks :)

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -407,47 +407,43 @@ extension Calendar {
...
-                                
-                let next = calendar.date(byAdding: components, to: startAt, wrappingComponents: wrappingComponents)
+                let scaled = components.scaled(by: iteration)
+                let next = calendar.date(byAdding: scaled, to: start, wrappingComponents: wrappingComponents)
                 guard let next else {
                     // We have finished
-                    previousResult = nil
+                    finished = false
                     return nil
                 }
```
davedelong, line 438: Shouldn't this be `finished = true`, since "We have finished"?

parkera: Yes!

parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

itingliu (approved):
Just wanted to double check that one spot

parkera:
@swift-ci test

parkera:
@swift-ci test

jmschonfeld:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

Issue #326, opened 2023-11-22
`DateInterval` convenience for "sidedness"?


kielgillard:
I was reviewing some code today that checked two date intervals, say `Monday` and `Tuesday`, were arranged like this on a timeline:
```
----|---Monday---|---Tuesday---|----
```

These intervals intersect, but it would be nice to easily check _how_ they intersect; if they "touch" at their boundaries; if the intervals are "beside" each other. This resulted in the following implementation:
```
func isBeside(_ other: DateInterval) -> Bool {
   self.start == other.end || self.end == other.start
}
```

Naming aside*, how viable would this be as an incredibly tiny but nice convenience method on `DateInterval` which I could contribute?

It seems it would have to go through evolution which, as the tiniest of conveniences, would be practically embarrassing. While I think it would be delightful to have this quality assured convenience available when you realise you need it, I do not have any data that demands its presence in Foundation, either. I did no more than a search of GitHub for Swift or Objective-C code for date intervals, dates and times. There's no compelling third party redundancy to solve. Is this a dead end?

\* pun not intended.

kielgillard:
I was compelled by the edit my comment above to updated the referenced PR to replace the notion of "sidedness" with a notion of "continuity".

kielgillard:
I thought more about a motivation for it and I am considering something more useful.

A user might have a logbook, employment or time planner app that requires them to enter the start and end time of particular activities in a particular sequence with some acceptable gap between (e.g. work 30 minutes here, do 30 minutes there and take a 10 minute break, at most five minutes between segments). The app has to check both the intersections of intervals and the measure of time between them. 

The current intersection API can only get you so far with this (it can compute intervals of intersections) but it would be good to know the time interval between two date intervals to know if the intersection is at the boundary (they share opposing start/end times) or if there was no intersection at all.

After putting this into words, I see a more useful API here: a difference API for computing a `TimeInterval` (or `DateInterval`!) between two `DateInterval` values:
```
a.   |-----|
           |<--compute this interval-->|
b.                                     |-----|

let timeBetweenBA = b.timeIntervalSince(a) // positive time interval
let timeBetweenAB = a.timeIntervalSince(b) // negative time interval

let datesBetweenBA = b.dateIntervalSince(a) // (a.end, b.start)
let datesBetweenAB = a.dateIntervalSince(b) // the same: (a.end, b.start)
```
And in this case:
```
c.   |--------|
d.         |-----|

[...]

PR #327, opened 2023-11-22
Compute differences between `DateInterval`s with `timeInterval(to:)` and `dateInterval(to:)` methods.

kielgillard:
Implements #326.

Please read #326 where I explain how I expect this needs a pitch and review and asks what I should do next.

### What?
Returns the seconds between `self` and `dateInterval` or `nil` if there is no difference in time between them.
For example, given these two intervals on a timeline:
```
|-----|  <-- time interval --> |-----|
```
Returns a negative time interval when `self` ends before `dateInterval` starts. A postive time interval indicates `self` starts after `dateInterval` ends.
A return value of `0` indicates `self` starts or ends where `dateInterval` ends or starts (in other words, they intersect at their opposing start/end moments):
```
|-----|-----|
```
A return value of `nil` indicates `self` and `dateInterval` do not have any time between them:
```
|--|-----|--|
```

### Testing
I have included unit tests.

I was not able to build and test on Xcode 15b3:
```
No such module 'FoundationMacros'
```
As such, I am relying on any CI to build and test these changes.

### DIFF
--- a/Tests/FoundationEssentialsTests/DateIntervalTests.swift
+++ b/Tests/FoundationEssentialsTests/DateIntervalTests.swift
@@ -50,6 +50,140 @@ final class DateIntervalTests : XCTestCase {
         let testInterval3 = DateInterval(start: start, duration: 100.0)
         XCTAssertNotEqual(testInterval1, testInterval3)
     }
+    
+    func test_intervalsBetweenDateIntervalAndDate() {
+        let earlier = Date(timeIntervalSince1970: 0)
+        let middle = Date(timeIntervalSince1970: 5)
+        let later = Date(timeIntervalSince1970: 10)
+
+        let start = Date(timeIntervalSince1970: 1)
+        let duration: TimeInterval = 8
+        let end = start.addingTimeInterval(duration) // 9
+        let testInterval1 = DateInterval(start: start, duration: duration)
+
+        // * --- |testInterval1|
+        let t1 = testInterval1.timeInterval(to: earlier)
+        let d1 = testInterval1.dateInterval(to: earlier)
+        XCTAssertEqual(t1, 1)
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@swift-ci test

Tests/FoundationEssentialsTests/DateIntervalTests.swift
```diff
@@ -50,6 +50,105 @@ final class DateIntervalTests : XCTestCase {
...
+        XCTAssertEqual(d10, nil)
+        
+        // |testInterval11| on itself
+        let testInterval11 = DateInterval(start: Date(timeIntervalSince1970: 0), end: Date(timeIntervalSinceReferenceDate: 0))
+        let t11 = testInterval11.timeInterval(to: testInterval11)
+        XCTAssertEqual(t11, nil)
+        
+        let d11 = testInterval11.dateInterval(to: testInterval11)
+        XCTAssertEqual(d11, nil)
+    }
```
kielgillard, line 151: I would have broken these down into separate test functions (one expecting non-nil and another nil intervals) but I implemented it like this to follow the lead of pre-existing test functions.

iCharlesHu:
Hi @kielgillard! It looks like you are proposing a change to `TimeInterval`'s public interface? Any changes to Foundation's API surface requires `API Proposal`: https://github.com/apple/swift-foundation/blob/main/Proposals/0000-template.md

Could you go ahead and draft up a proposal for this change so we can review them both here and on the forum? Thanks!

kielgillard:
@iCharlesHu Hello, thank you for the comment. Would still love any feedback you might have!

I'm now working on a proposal in #331 and the implementation can now be found in #330.

PR #329, opened 2023-11-27
Add import for _dyld_image_count

parkera:
Adds an import for use of `_dyld_image_count` in `FOUNDATION_FRAMEWORK` block.


parkera:
@swift-ci test

PR #334, opened 2023-12-01
Do not attempt to cache based on a signature that uses a `Locale`

parkera:
`autoupdatingCurrentLocale` only considers its autoupdating-ness as far as hashing and equality goes. That means if we include it in the `Signature` of the caches used for number formatters (which is how we unique them), an underlying change to the current locale does not result in flushing the cache correctly.

Resolve this for number formatting by storing the identifier and preferences directly. Although we do not use the preferences yet, we will need to in the near future to resolve an issue where we do not honor users customized grouping separators.


parkera:
@swift-ci test

parkera:
@swift-ci test

PR #337, opened 2023-12-05
Implement base64 for Data in Swift

glessard:
This brings the swift-corelibs-foundation implementation of base64 to swift-foundation and Darwin.

Furthermore we use `BufferView` alongside a new abstraction for memory initialization, `OutputBuffer`, to perform the encoding with better memory safety: no unsafe types used in the business logic, and bounds-checking always enabled.

Addresses rdar://116882832 and https://github.com/apple/swift-foundation/issues/222


FranzBusch (commented):
Amazing! I would be interested to see how this compares in performance against https://github.com/swift-extras/swift-extras-base64 which is used a lot in the server ecosystem. Maybe we can add a benchmark for the new base64 implementation here?

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+                )
+            }
+        }
+        catch {
+            precondition(error is Base64Error)
+            return nil
+        }
+    }
+
+
```
Harry-KNIGHT, line 47: Double empty space
```suggestion

```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+        let capacity = Self.estimateBase64Size(length: self.count)
+        if #available(macOS 11, *) {
+            return self.withBufferView { inputBuffer in
+                String(utf8Capacity: capacity) {
+                    Self.base64EncodeBytes(inputBuffer, &$0, options: options)
+                }
+            }
+        }
+        else {
```
Harry-KNIGHT, line 107:
```suggestion
        } else {
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+                }
+            }
+
+            if invalid {
+                if ignoreUnknown {
+                    continue
+                } else {
+                    throw Base64Error.cannotDecode
+                }
+            }
```
Harry-KNIGHT, line 214: Why do not make something like that ?
```suggestion
            if invalid, ignoreUnknown  {
                    continue
                } else {
                    throw Base64Error.cannotDecode
                }
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+        let base64CharsPtr = base64Chars.utf8Start
+
+        let lineLength: Int
+        var currentLineCount = 0
+        let separatorByte1: UInt8
+        var separatorByte2: UInt8?
+
+        if options.isEmpty {
+            lineLength = 0
+            separatorByte1 = 0
```
Harry-KNIGHT, line 278: Why don't you make early return for do not have nested `if else statement` ?

Sources/FoundationEssentials/JSON/OutputBuffer.swift
```diff
@@ -0,0 +1,250 @@
...
+    }
+
+    mutating func moveAppend(
+        fromContentsOf source: Slice<UnsafeMutableBufferPointer<T>>
+    ) {
+        moveAppend(fromContentsOf: UnsafeMutableBufferPointer(rebasing: source))
+    }
+}
+
+extension OutputBuffer<UInt8> /* where T: BitwiseCopyable */ {
```
Harry-KNIGHT, line 131: Delete commentary ?
```suggestion
extension OutputBuffer<UInt8> {
```

Sources/FoundationEssentials/JSON/OutputBuffer.swift
```diff
@@ -0,0 +1,250 @@
...
+        let p = UnsafeMutableRawPointer(start.advanced(by: initialized))
+        p.storeBytes(of: value, as: Value.self)
+        initialized &+= q
+    }
+}
+
+extension OutputBuffer {
+    var initializedPrefix: /*borrowed*/ BufferView<T> {
+        /* _read */ get /* borrowing(self) */ {
+            /* yield */ return BufferView(
```
Harry-KNIGHT, line 155: Delete commentaries ?
```suggestion
    var initializedPrefix: /*borrowed*/ BufferView<T> {
        /* _read */ get /* borrowing(self) */ {
            /* yield */ return BufferView(
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+     If the input contains any bytes that are not valid Base64 characters,
+     this will throw a `Base64Error`.
+
+     - parameter bytes:      The Base64 bytes
+     - parameter output:     An OutputBuffer to be filled with decoded bytes
+     - parameter options:    Options for handling invalid input
+     - returns:              Whether decoding was successful
+     */
+    static func base64DecodeBytes(
+        _ bytes: borrowing BufferView<UInt8>,
```
karwa, line 160: `bytes` is only iterated over; there doesn't seem to be any benefit to requiring a (contiguous) `BufferView`.

This could be generic. In fact, it could even be a single-pass `Sequence`.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+    }
+
+    /**
+     This method encodes data in Base64.
+
+     - parameter dataBuffer: A BufferView of the bytes to encode
+     - parameter options:    Options for formatting the result
+     - parameter buffer:     The buffer to write the bytes into
+
+       NOTE: dataBuffer would be better expressed as a <T: Collection> where T.Element == UInt8, T.Index == Int but this currently gives much poorer performance.
```
karwa, line 257: Interesting - how much worse is the performance? Any ideas about what's going on?

The way this function reads data from the buffer is probably not optimal and could allow dropping the `T.Index == Int` requirement.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+
+    /// Initialize a `Data` from a Base-64 encoded String using the given options.
+    ///
+    /// Returns nil when the input is not recognized as valid Base-64.
+    /// - parameter base64String: The string to parse.
+    /// - parameter options: Encoding options. Default value is `[]`.
+    public init?(
+        base64Encoded base64String: consuming String,
+        options: Base64DecodingOptions = []
+    ) {
```
parkera, line 31: This is a style nitpick, but I think it's important to maintain it across as much of our code as we can until we figure out an automatic formatting story. We don't split up the initializers across multiple lines, or limit to 80 chars, unless we've got an extreme number of arguments.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+            self = try base64String.withUTF8 {
+                // String won't pass an empty buffer with a `nil` `baseAddress`.
+                try Data(
+                    decodingBase64: BufferView(unsafeBufferPointer: $0)!,
+                    options: options
+                )
+            }
+        }
+        catch {
+            precondition(error is Base64Error)
```
parkera, line 42: Why bother throwing an error at all if we simply ignore it here?

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+
+
+    /// Initialize a `Data` from a Base-64, UTF-8 encoded `Data`.
+    ///
+    /// Returns nil when the input is not recognized as valid Base-64.
+    ///
+    /// - parameter base64Data: Base-64, UTF-8 encoded input data.
+    /// - parameter options: Decoding options. Default value is `[]`.
+    public init?(
+        base64Encoded base64Data: borrowing Data,
```
parkera, line 55: Is this ABI compatible with the existing public API?

```
    @inlinable // This is @inlinable as a convenience initializer.
    public init?(base64Encoded base64Data: __shared Data, options: Data.Base64DecodingOptions = [])
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,374 @@
...
+        )
+    }
+
+    // MARK: - Create base64
+
+    /// Returns a Base-64 encoded string.
+    ///
+    /// - parameter options: The options to use for the encoding. Default value is `[]`.
+    /// - returns: The Base-64 encoded string.
+//    @available(macOS 11, *)
```
parkera, line 95: What's the plan with this availability annotation?

The existing one is again `@inlinable`

Sources/FoundationEssentials/JSON/OutputBuffer.swift
```diff
@@ -0,0 +1,250 @@
...
+                try initializer(&output)
+                let initialized = output.relinquishBorrowedMemory()
+                assert(initialized.baseAddress == buffer.baseAddress)
+                count = initialized.count
+            }
+        )
+    }
+}
+
+extension String {
```
parkera, line 201: An organizational thing, but I think I would prefer the extensions for String and Data go with files that are closer those types than this one. Otherwise this file tends to become a 'dumping ground' for extensions.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -0,0 +1,370 @@
...
+    /// Returns a Base-64 encoded string.
+    ///
+    /// - parameter options: The options to use for the encoding. Default value is `[]`.
+    /// - returns: The Base-64 encoded string.
+    @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
+    public func base64EncodedString(options: Base64EncodingOptions = []) -> String {
+        if self.isEmpty { return "" }
+
+        let capacity = Self.estimateBase64Size(length: self.count)
+        if #available(macOS 11, *) {
```
parkera, line 102: I don't think we need this at all, actually. Our min deployment target is macOS 13.3: 

https://github.com/apple/swift-foundation/blob/main/Package.swift#L17

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: `__shared` and `borrowing` are ABI-compatible. `__owned` and `consuming` are not.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: This one turned out to be unnecessary, and I forgot to delete it. The underlying `String` initializer requires macOS 11.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: The fact that the deployment target is stringly-defined in the package leads to warnings in some situations.

Sources/FoundationEssentials/JSON/OutputBuffer.swift
glessard: I agree with that, though littering `String` and `Data`'s files with experimental extensions seems undesirable as well.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: I had typed throws on the mind, but it doesn't help anything here.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: That would be a different control flow. We want to either `continue` or `throw` if `invalid` is true, but  the rest of the loop must execute when `invalid` is false.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: We do want to leave room for a byte-parallel implementation. We have not done that yet, but it is eminently doable, and more so with contiguous storage. Currently we are replicating the existing API (and ABI) from the existing version of Foundation. We may add more API at a later time.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: I reverted to `__shared` after checking on the actual mangling.

parkera (approved):

PR #342, opened 2023-12-06
Fix remaining FormatStyle caches of autoupdating locales

parkera:
Fix up remaining issues where we cache a `Locale` for format styles, and an autoupdating locale will not cause a refreshed format.

Follows up on 18ac5ac (#334 )

Fixes rdar://119010483

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Date/ICURelativeDateFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Date/ICURelativeDateFormatter.swift
@@ -22,7 +22,13 @@ package import FoundationICU
 
 @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
 internal final class ICURelativeDateFormatter {
-
+    struct Signature : Hashable {
+        let localeIdentifier: String
+        let numberFormatStyle: UNumberFormatStyle.RawValue?
+        let relativeDateStyle: UDateRelativeDateTimeFormatterStyle.RawValue
+        let context: UDisplayContext.RawValue
+    }
+    
     static let sortedAllowedComponents : [Calendar.Component] = [ .year, .month, .weekOfMonth, .day, .hour, .minute, .second ]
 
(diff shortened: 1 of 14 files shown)


parkera:
@swift-ci test

Sources/FoundationInternationalization/Formatting/Date/ICURelativeDateFormatter.swift
```diff
@@ -73,9 +79,10 @@ internal final class ICURelativeDateFormatter {
         }
     }
 
-    internal static func formatterCreateIfNeeded(format: Date.RelativeFormatStyle) -> ICURelativeDateFormatter {
-        let formatter = Self.cache.formatter(for: format) {
-            ICURelativeDateFormatter(uNumberFormatStyle: format.unitsStyle.icuNumberFormatStyle, uRelDateStyle: format.unitsStyle.icuRelativeDateStyle, locale: format.locale, context: format.capitalizationContext.icuContext)
+    internal static func formatter(for style: Date.RelativeFormatStyle) -> ICURelativeDateFormatter {
+        let signature = Signature(localeIdentifier: style.locale.identifier, numberFormatStyle: style.unitsStyle.icuNumberFormatStyle?.rawValue, relativeDateStyle: style.unitsStyle.icuRelativeDateStyle.rawValue, context: style.capitalizationContext.icuContext.rawValue)
```
itingliu, line 83: Should this be the preference-capturing locale identifier to account for, say, custom first weekday? I imagine that information would be useful for calculating relative date in terms of weeks

itingliu (approved):
Only one comment regarding the relative formatted. It doesn't look like we are registering preferences with ICU formatter currently anyways, but wanted to check

Sources/FoundationInternationalization/Formatting/Date/ICURelativeDateFormatter.swift
parkera: Hm, that may be another bug. rdar://119343432 (Check if ICURelativeDateFormatter should be using preference-capturing locale)

PR #343, opened 2023-12-07
Support combining Predicates

jmschonfeld:
This PR adds support for combining predicates together to form compound predicates. Predicates can be combined like the following:

```swift
let predicateA: Predicate<String>
let predicateB: Predicate<Int>

let predicateC = #Predicate<String, Int> {
    predicateA.evaluate($0) && predicateB.evaluate($1)
}
```

This creates a compound `predicateC` which only evaluates to true when `predicateA` evaluates to true given the first input, and `predicateB` evaluates to true given the second input.

### DIFF
--- a/Sources/FoundationMacros/PredicateMacro.swift
+++ b/Sources/FoundationMacros/PredicateMacro.swift
@@ -24,7 +24,7 @@ package import SwiftSyntaxBuilder
 // A list of all functions supported by Predicate itself, any other functions called will be diagnosed as an error
 // This allows for checking the function name, the number of arguments, and the argument labels, but the types of the arguments will need to be validated by the post-expansion type checking pass
 // The trailingClosure parameter indicates whether the final argument is a closure and therefore supports dropping the final argument label in favor of a trailing closure
-private let knownSupportedFunctions: Set<FunctionStructure> = [
+private var _knownSupportedFunctions: Set<FunctionStructure> = [
     FunctionStructure("contains", arguments: [.unlabeled]),
     FunctionStructure("contains", arguments: [.closure(labeled: "where")]),
     FunctionStructure("allSatisfy", arguments: [.closure(labeled: nil)]),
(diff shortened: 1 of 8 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu:
Sigh... looks like compiler regression again?

jmschonfeld:
@iCharlesHu it looks like the compiler only hits this issue when asserts are enabled, but without asserts it compiles successfully. I've put this code behind `#if FOUNDATION_FRAMEWORK` for now until we can resolve this assertion (it looks like it may very well be a false positive given it compiles successfully without the assertion)

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Sources/FoundationMacros/PredicateMacro.swift
```diff
@@ -40,13 +40,64 @@ private let knownSupportedFunctions: Set<FunctionStructure> = [
     FunctionStructure("caseInsensitiveCompare", arguments: [.unlabeled])
 ]
 
+private var knownSupportedFunctions: Set<FunctionStructure> {
```
iCharlesHu, line 43: Sorry that you have to do this 💀

iCharlesHu (approved):

Magnuti:
Great work on this guys! Really helpful stuff.

It may be early due to to 17.4 version requirement, but could you put some information about the fact that this is possible on the Swift Data docs? Right under here https://developer.apple.com/documentation/foundation/predicate#overview

I had to search for ages just to find out that this is possible.

bdkjones:
The Swift forum admins hate posts that are about Foundation rather than "Swift the actual language", so there's not really a better place to put this comment:

The combination approach shown here is, in practice, very limited. If you try to use it for more than just a couple predicates, the compiler's type-checker falls over and code won't compile. This makes anything other than very trivial filtering quite difficult.

Imagine an app where we want to dynamically filter some model objects by any/all of: 

- name (String)
- creationDate (Date)
- starRating (Int)
- isFavorite
- category (Int)

We might ask, "Fetch all objects with `foo` in their name, created after 1 January 2025, with 3+ stars, in the `dogs` category." The key point is: we don't know the conditions until runtime and the user can change any/all of them with a search interface in our app.

`#Predicate` won't do this. The compiler just can't handle the complexity involved in the macro. Dynamic filtering of model objects is a pretty basic task of almost every app, so it ought to be very intuitive, performant, and easy. Unfortunately, the current macro-reliant implementation of `#Predicate` isn't.

I dunno what the right fix is, but I'd just like to suggest that Apple not consider the current implementation ideal.

PR #345, opened 2023-12-08
Add RangeExpressionContains to the standard PredicateCodableConfiguration

### LINKED ISSUE #344: Predicate with range expression can't be encoded
If I create a predicate like this:

```
#Predicate<Dummy> {
    (0...10).contains($0.value)
}
```

And try to encode it with a `JSONEncoder`, I get this error:

`The 'Foundation.PredicateExpressions.RangeExpressionContains<Foundation.PredicateExpressions.ClosedRange<Foundation.PredicateExpressions.Value<Swift.Int>, Foundation.PredicateExpressions.Value<Swift.Int>>, Foundation.PredicateExpressions.KeyPath<Foundation.PredicateExpressions.Variable<PredicateRangeCoding.Dummy>, Swift.Int>>' type is not in the provided allowlist (required by /)`

Even though I've provided a codable configuration:

```
struct Dummy: PredicateCodableKeyPathProviding {
    static var predicateCodableKeyPaths: [String: PartialKeyPath<Dummy>] {
        ["value": \.value]
    }

    let value: Int
}

var configuration = PredicateCodableConfiguration.standardConfiguration
configuration.allowKeyPathsForPropertiesProvided(by: Dummy.self)
```

Encoding succeeds if I explicitly allow this type, like so:

```
configuration.allowType(PredicateExpressions.RangeExpressionContains<PredicateExpressions.ClosedRange<
                        PredicateExpressions.Value<Int>, PredicateExpressions.Value<Int>>,
                        PredicateExpressions.KeyPath<PredicateExpressions.Variable<Dummy>, Int>>.self)
```

More strangely, encoding also succeeds if I first encode a different predicate that does not contain a range, like this one:

```
#Predicate<Dummy> {
    $0.value > 0
}
```

Other types of predicates can be encoded without issue, so it seems specific to the range expression.

I've created a sample project here: https://github.com/ordo-one/external-reproducers/tree/main/swift/predicate-range-coding. I had a issue when building this on the command line, but it worked when building in Xcode, not sure why:

`main.swift:30:24: error: external macro implementation type 'FoundationMacros.PredicateMacro' could not be found for macro 'Predicate'`

Using Xcode 15.0 (15A240d) with Swift version 5.9 (swiftlang-5.9.0.128.108 clang-1500.0.40.1).

jmschonfeld:
It looks like we missed an expression type in the standard configuration. This adds it to allow for encoding/decoding with predicates that contain these expressions without requiring it to be manually allowed.

Resolves #344

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
+++ b/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
@@ -352,6 +352,7 @@ extension PredicateCodableConfiguration {
         // Foundation-defined PredicateExpression types
         configuration.allowPartialType(PredicateExpressions.Arithmetic<PredicateExpressions.Value<Int>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.Arithmetic")
         configuration.allowPartialType(PredicateExpressions.ClosedRange<PredicateExpressions.Value<Int>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.ClosedRange")
+        configuration.allowPartialType(PredicateExpressions.RangeExpressionContains<PredicateExpressions.Value<Range<Int>>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.RangeExpressionContains")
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

Issue #186, opened 2023-06-25
Inconsistent formatting behavior when BinaryInteger cannot be converted into Int64


joey-gm:
ICUNumberFormatter doesn't support `UInt64`  number formatting. Integers not representable by `Int64` are clamped as per documentation.

Under current implementation, however, if a `BinaryInteger` cannot be converted into `Int64`, IntegerFormatStyle will first attempt to convert the integer into Decimal (represented by `Double`) before falling back to clamping. This may result in inconsistent formatting behavior.

One edge case is 9223372036854775808 (Int64.max + 1). This integer can be converted to a floating-point value (9.223372036854776E+18), which will then be formatted into "9" instead of the the clamped number of 9,223,372,036,854,775,807 (Int64.max).

Should we drop the Decimal conversion in IntegerFormatStyle?
- IntegerFormatStyle: Lines 221-222
- IntegerFormatStyle.Percent: Lines 253-254
- IntegerFormatStyle.Currency: Lines 285-286
- AttributedString: Lines 512-513

Personally, I would prefer to drop the `Int64` clamping as well and fall back to the Swift Standard Library's LosslessStringConvertible: `String(value)`. This will allow us to preserve the correct numeric value albeit incorrect format. Ideally, we should have a proper implementation for `UInt64`  number formatting.

https://github.com/apple/swift-foundation/blob/fa61cd8aeb316727f9c7d740caa675e8c6f75abf/Sources/FoundationInternationalization/Formatting/Number/IntegerFormatStyle.swift#L211-L225

https://github.com/apple/swift-foundation/blob/fa61cd8aeb316727f9c7d740caa675e8c6f75abf/Sources/FoundationInternationalization/Formatting/Number/IntegerFormatStyle.swift#L253-L254

https://github.com/apple/swift-foundation/blob/fa61cd8aeb316727f9c7d740caa675e8c6f75abf/Sources/FoundationInternationalization/Formatting/Number/IntegerFormatStyle.swift#L285-L286

https://github.com/apple/swift-foundation/blob/fa61cd8aeb316727f9c7d740caa675e8c6f75abf/Sources/FoundationInternationalization/Formatting/Number/IntegerFormatStyle.swift#L512-L513

jmschonfeld:
[...]

PR #346, opened 2023-12-11
Improved big-number formatting

oscbyspro:
Howdy! I've made some changes to improve your big-number formatting.

### Changes

- Removed trapping conversion to Int in BinaryInteger/formatted().
  - IntegerFormatStyle can format big integers since ([#262](https://github.com/apple/swift-foundation/pull/262)).
- Removed rounding conversion to Double in BinaryFloatingPoint/formatted()
  - This method now does whatever FloatingPointFormatStyle/format(\_:) does.
- Reenabled "Decimal Tests" in NumberFormatStyleTests.swift.
  - The normal styles worked as-is, but the attributed one needed some love.
- Added a "numeric string representation" case to ICUNumberFormatterBase.Value.
  - IntegerFormatStyle.Attributed now uses this value rather than Int64(clamping:).
- Removed conversions to Decimal in each integer format style ([#186](https://github.com/apple/swift-foundation/issues/186)).
  - IntegerFormatStyle, [...].Attributed, [...].Currency, [...].Percent
- Removed isZero and doubleValue from ICUNumberFormatterBase.Value.
  - This makes it easier to accommodate non-numeric payloads, e.g. String.
  - Both were used in ByteCountFormatStyle, which now queries FormatInput.
- Added internal \_format(\_:doubleValue:) method to ByteCountFormatStyle.
  - Parameterized function body for use with values convertible to Double.
  - Prior method was removed because it relied on [...].Value/doubleValue.

### Comments

 [RESOLVED] I was not sure what to do with ICUNumberFormatterBase.Value's isZero and doubleValue – please share your thoughts. I solved it :tm: by force unwrapping Double.init(\_:). The force unwrap should never fail, assuming the numeric string representation is well-formed, but large values may result in ± Double.infinity. It is not a problem for isZero, but I do not know the semantics of doubleValue.

### Alternatives

 [OUTDATED] You can add some thing like fallbackInteger(any BinaryInteger) instead of numericStringRepresentation(String). Perhaps it makes sense if you don't want to check isZero through Double.

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/ByteCountFormatStyle.swift
+++ b/Sources/FoundationInternationalization/Formatting/ByteCountFormatStyle.swift
@@ -152,9 +152,9 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
             return false
         }
 
-        func _format(_ value: ICUNumberFormatter.Value) -> AttributedString {
+        public func format(_ value: Int64) -> AttributedString {
             let unit: Unit = allowedUnits.contains(.kb) ? .kilobyte : .byte
-            if spellsOutZero && value.isZero {
+            if spellsOutZero && value == .zero {
                 let numberFormatter = ICUByteCountNumberFormatter.create(for: "measure-unit/digital-\(unit.name)\(unit == .byte ? " unit-width-full-name" : "")", locale: locale)
                 guard var attributedFormat = numberFormatter?.attributedFormat(.integer(.zero), unit: unit) else {
                     // fallback to English if ICU formatting fails
@@ -192,7 +192,7 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
                 maxSizes = Self.maxBinarySizes
             }
 
-            let absValue = abs(value.doubleValue)
+            let absValue = abs(Double(value))
             let bestUnit: Unit = {
                 var bestUnit = allowedUnits.smallestUnit
                 for (idx, size) in maxSizes.enumerated() {
@@ -209,7 +209,7 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
             }()
 
             let denominator = decimal ? bestUnit.decimalSize : bestUnit.binarySize
-            let unitValue = value.doubleValue/Double(denominator)
+            let unitValue = Double(value)/Double(denominator)
 
             let precisionSkeleton: String
             switch bestUnit {
@@ -231,7 +231,7 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
                 let localizedParens = localizedParens(locale: locale)
                 attributedString.append(AttributedString(localizedParens.0))
 
-                var attributedBytes = byteFormatter!.attributedFormat(value, unit: .byte)
+                var attributedBytes = byteFormatter!.attributedFormat(.integer(value), unit: .byte)
                 for (value, range) in attributedBytes.runs[\.byteCount] where value == .value {
                     attributedBytes[range].byteCount = .actualByteCount
                 }
@@ -242,13 +242,7 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
 
             return attributedString
         }
-
-
-        public func format(_ value: Int64) -> AttributedString {
-            _format(.integer(value))
-        }
     }
-
 }
 
 @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
--- a/Sources/FoundationInternationalization/Formatting/ByteCountFormatStyle.swift
+++ b/Sources/FoundationInternationalization/Formatting/ByteCountFormatStyle.swift
@@ -151,10 +151,10 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
 
             return false
         }
-
-        public func format(_ value: Int64) -> AttributedString {
+        
+        func _format(_ formatterValue: ICUNumberFormatter.Value, doubleValue: Double) -> AttributedString {
             let unit: Unit = allowedUnits.contains(.kb) ? .kilobyte : .byte
-            if spellsOutZero && value == .zero {
+            if spellsOutZero && doubleValue.isZero {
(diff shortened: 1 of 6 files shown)


oscbyspro:
Just to be clear: this pull request is a smörgåsbord. I wanted to fix the trapping conversion (and I suggest you keep that part). But, after that, I reenabled some of your tests and looked for ways to fix `IntegerFormatStyle.Attributed`'s clamping behavior. Adding a new case to `ICUNumberFormatterBase.Value` is one way, but I don't know if that's what you want. The numeric string representation is perhaps not something you want to query. But, since I have no way of knowing your plans for IntegerFormatStyle.Attributed, I thought I would leave it up for review.

oscbyspro:
I don't want to bombard this thread. But after looking more closely at it, I see that `isZero` and `doubleValue` are only used in `ByteCountFormatStyle`. So you could perhaps remove `isZero` and `doubleValue` from `ICUNumberFormatterBase.Value` and get those from `FormatInput` instead. In that case, `[...].Value` could contain anything convertible to `ICUNumberFormatterBase.FormatResult`. It's one option that circumvents the current awkwardness of using a `String` payload in `[...].Value`.

Sources/FoundationInternationalization/Formatting/Number/ICUNumberFormatter.swift
```diff
@@ -77,6 +78,8 @@ internal class ICUNumberFormatterBase {
                 return num == 0
             case .decimal(let num):
                 return num == 0
+            case .numericStringRepresentation(let num):
+                return Double(num)!.isZero
```
itingliu, line 82: I'd probably make it conditional instead of force unwrap since there's no guarantee that the string would always be in correct format when passed into `ICUNumberFormatterBase.Value`

itingliu:
> I don't want to bombard this thread. But after looking more closely at it, I see that `isZero` and `doubleValue` are only used in `ByteCountFormatStyle`. So you could perhaps remove `isZero` and `doubleValue` from `ICUNumberFormatterBase.Value` and get those from `FormatInput` instead. In that case, `[...].Value` could contain anything convertible to `ICUNumberFormatterBase.FormatResult`. It's one option that circumvents the current awkwardness of using a `String` payload in `[...].Value`.
> 
> Edit: here's a separate [commit](https://github.com/oscbyspro/swift-foundation/commit/f065a2d8beb964e19a9c9300cebda9bbe1736904) with this approach (`+5`, `-37`).

Thank you! I think this is indeed the most reasonable approach.

Sources/FoundationInternationalization/Formatting/Number/ICUNumberFormatter.swift
itingliu: But if we do end up removing `isZero` and `doubleValue` then we won't need this at all

oscbyspro:
> Thank you! I think this is indeed the most reasonable approach. I'd be ok either if you include that as part of this change or a separate PR. No bombarding at all either

Cool. I've added f065a2d to this pull request.

Sources/FoundationInternationalization/Formatting/Number/ICUNumberFormatter.swift
oscbyspro: I removed `isZero` and `doubleValue` in f065a2d and took these values from `FormatInput` instead.

itingliu (commented):
@swift-ci please test

itingliu (approved):
LGTM! Thanks!

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/ByteCountFormatStyle.swift
```diff
@@ -152,9 +152,9 @@ public struct ByteCountFormatStyle: FormatStyle, Sendable {
             return false
         }
 
-        func _format(_ value: ICUNumberFormatter.Value) -> AttributedString {
```
itingliu, line 155: Sorry I spoke too soon. This is actually used internally by Measurement formatting (source code not available yet but will be soon) where we're passing in `value: .floatingPoint()`. Do you think we can find a way to keep this around?

The ideas I have in mind are (1) keep the initial approach you took in this PR, where we'd just make an educated guess with `isZero`, or(2) add an explicit `func _format(value: Double)` and flesh out sharable guts from the format function to share between these two types. 

What do you think? Any other ideas?

itingliu: I feel bad asking for (2) since that's really not part of this PR. I can make a follow up change on that if you want to revert to (1)

oscbyspro: I see, no worries. I've tried a few solutions, and this one was the least complicated: ba8bf28aba71e70b476ab303203487346f7e1d4c (`+10`, `-6`). It's basically (2), but parameterized as `_format(_:doubleValue:)`. It's a small change.

```swift
_format(.floatingPoint(value), doubleValue: value)
```

oscbyspro:
Here's a solution that lets you use Double in ByteCountFormatStyle:

- [changes](https://github.com/apple/swift-foundation/commit/ba8bf28aba71e70b476ab303203487346f7e1d4c) (`+10`, `-6`)


```swift
_format(.integer(int64), doubleValue: Double(int64))
_format(.floatingPoint(double), doubleValue: double)
```

If you think it's reasonable approach, then I'll add it to this pull request.

oscbyspro:
Hm. I suppose I'll add the proposed solution and amend it if needed. I think that might be easier.

parkera (approved):
I'll let @itingliu have the final say here but this looks good to me.

itingliu (approved):
LGTM!

itingliu:
@swift-ci please test

itingliu:
CI is down due to a regression in recent toolchain. That should be fixed in https://github.com/apple/swift/pull/70559. I'll rerun the test after CI updates to a newer toolchain containing the fix

itingliu:
@swift-ci please test

PR #350, opened 2023-12-12
AttributeScope cache does not correctly store negative results

jmschonfeld:
The `LoadedScopeCache` had a bug in its handling of `Optional` dictionary values in which it would fail to store negative lookup results rather than storing a `nil` value in the dictionary. This switches the representation to use a custom `enum` rather than an `Optional` to ensure clarity and avoid this bug causing many unnecessary `dlopen` calls for lookup misses.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #351, opened 2023-12-13
Predicate to NSPredicate conversion should support captured URLs

jmschonfeld:
`NSPredicate` supports embedded URLs when passing the `NSPredicate` to CoreData, so we should ensure that when converting a `Predicate` to an `NSPredicate` it successfully embeds captured URLs


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #352, opened 2023-12-18
[Gregorian Calendar] Handle adding date components across DST transition like Calendar_ICU

itingliu:
Currently we implement adding units larger than `.day` as follows:
- Convert the date to date components
- Update the value of the added component
- Convert the date components back to date

The time of the day information is reduced to a `DateComponents` with, e.g. `hour == 1`, `minute == 30`, at step 1. Then when we convert the `DateComponents` back to `Date` at step 3, we always use the *first* occurrence of the time if it falls into the DST repeated time frame.

This is different from `Calendar_ICU`'s implementation, which uses the *latter* and rewind back the date by the DST transition interval (e.g. 1 hour for Pacific Time).

These yeild the same result except for when the input time and the output time are *both* during DST transition time frame. Update the implementation to match Calendar_ICU's behavior.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -147,6 +147,7 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
     let kSecondsInWeek = 604_800
     let kSecondsInDay = 86400
     let kSecondsInHour = 3600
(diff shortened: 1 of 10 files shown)


itingliu:
@swift-ci please test

parkera (approved):
That's a lot of new tests!

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #357, opened 2023-12-20
Fix off-by-one in Duration.UnitsFormatStyle list pattern getter

gwynne:
Fixes an off-by-one error in `Duration.UnitsFormatStyle._getFullListPattern(length:)` for lengths greater than 2 which prevents applying the "end piece" pattern.

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
+++ b/Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
@@ -381,7 +381,7 @@ extension Duration {
                 let middle = _getListPattern(UATIMEUNITLISTPAT_MIDDLE_PIECE)
                 let end = _getListPattern(UATIMEUNITLISTPAT_END_PIECE)
 
-                for i in 0 ..< length - 1 {
+                for i in 0 ..< length {
                     if i == 0 {
                         pattern = start
                     } else if i != length - 1 {
--- a/Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
+++ b/Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
@@ -371,25 +371,25 @@ extension Duration {
             let lastPlaceholder = "{1}"
 
             var pattern: String!
-            if length == 1 {
-                pattern = placeholder
-            } else if length == 2 {
-                pattern = _getListPattern(UATIMEUNITLISTPAT_TWO_ONLY)
-                pattern.replace(lastPlaceholder, with: placeholder)
-            } else {
-                let start = _getListPattern(UATIMEUNITLISTPAT_START_PIECE)
-                let middle = _getListPattern(UATIMEUNITLISTPAT_MIDDLE_PIECE)
-                let end = _getListPattern(UATIMEUNITLISTPAT_END_PIECE)
 
-                for i in 0 ..< length {
-                    if i == 0 {
-                        pattern = start
-                    } else if i != length - 1 {
-                        pattern.replace(lastPlaceholder, with: middle)
-                    } else {
-                        pattern.replace(lastPlaceholder, with: end)
-                    }
+            switch length {
+            case 1:
+                pattern = placeholder
+            case 2:
+                pattern = self._getListPattern(UATIMEUNITLISTPAT_TWO_ONLY)
+                pattern.replace(lastPlaceholder, with: placeholder)
+            case let length:
+                let middle = self._getListPattern(UATIMEUNITLISTPAT_MIDDLE_PIECE)
+
+                pattern = self._getListPattern(UATIMEUNITLISTPAT_START_PIECE)
+                // Each of the three pieces provides _two_ placeholders each,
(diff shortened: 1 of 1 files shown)


Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
```diff
@@ -381,7 +381,7 @@ extension Duration {
                 let middle = _getListPattern(UATIMEUNITLISTPAT_MIDDLE_PIECE)
                 let end = _getListPattern(UATIMEUNITLISTPAT_END_PIECE)
 
-                for i in 0 ..< length - 1 {
+                for i in 0 ..< length {
                     if i == 0 {
                         pattern = start
                     } else if i != length - 1 {
```
itingliu, line 387: I believe the right fix is to change this to `length - 2` instead of changing the number of loops . We do only want to run this loop only `(length - 1)` times. The effect is clear when running `DurationUnitsFormatStyleTests` with this change. (Unfortunately CI is down at the moment, but we shall see when it comes back up soon)

gwynne: The way I read this code, wouldn't that only make the problem worse? The condition `if i != length - 1` can already never be true (`..<` is an open range exclusive of its upper bound, meaning `I` will never be `length - 1`; the final iteration of the loop is at `length - 2`).

I originally wrote the fix in a different form that was a bit more explicit, but I had opted for the current version under the principle of making the most minimal necessary change. The "better" version (which if I'm correct, addresses your concern) looks like this:

```swift
    func _getFullListPattern(length: Int) -> String {
        let placeholder = "{0}"
        let lastPlaceholder = "{1}"

        var pattern: String!

        switch length {
        case 1:
            pattern = placeholder
        case 2:
            pattern = self._getListPattern(UATIMEUNITLISTPAT_TWO_ONLY)
            pattern.replace(lastPlaceholder, with: placeholder)
        case let length:
            let middle = self._getListPattern(UATIMEUNITLISTPAT_MIDDLE_PIECE)

            pattern = self._getListPattern(UATIMEUNITLISTPAT_START_PIECE)
            for _ in 1 ... (length - 2) {
                pattern.replace(lastPlaceholder, with: middle)
            }

            pattern.replace(lastPlaceholder, with: self._getListPattern(UATIMEUNITLISTPAT_END_PIECE))
            pattern.replace(lastPlaceholder, with: placeholder)
        }
        return pattern
    }
```

gwynne: Update: I stand corrected; this version doesn't work properly either. It seems that despite the logic seeming semantically correct, the end result is the pattern is substituted one time too many 😕

gwynne: Update 2: I have pushed a version that passes the tests correctly. Sorry for the miscue!

itingliu: > The condition `if i != length - 1` can already never be true

I think what you meant is that `if i != length - 1` is **always true**? As you said, "`i` will never be `length - 1`" because that's the upper bound that we're never going to step onto. It's probably easier to read if we swapped the `if i != length - 1` block with the `else` block below though....

I like your "better version" given its readability, but this test would still fail  https://github.com/apple/swift-foundation/blob/aad084892a3208621049730c36f33f82ca8a291e/Tests/FoundationInternationalizationTests/Formatting/DurationUnitsFormatStyle.swift#L29


The gotcha is that for any given length, we only want to manipulate the pattern `length - 1` times, including both the start and the end piece. 

For example for `length: 4`
1. To start, assign pattern to the start piece, e.g. `"{0} <start separator> {1}"`
2. Then: replace `"{1}"` with middle piece, e.g. `"{0} <start separator> {0} <middle separator> {1}"`
3. Finally: replace `"{1}"` with the end piece, e.g. `"{0} <start separator> {0} <middle separator> {0} <end separator> {1}"`

So we actually only modify the pattern 3 times. The middle piece is only handled **once**. Your approach however runs the middle piece loop twice, resulting in one extra placeholder in the end

itingliu (approved):
It looks like we had a race condition when writing our comments 🙂 
This updated version LGTM

Sources/FoundationInternationalization/Formatting/Duration+UnitsFormatStyle.swift
gwynne: > I think what you meant is that if i != length - 1 is always true? 

Yep, that's what I meant! 😅 My distracted brain keeps swapping the condition bodies because if I were writing that logic as an if clause I would have written `if I == length - 1 { pattern.replace(end) }` rather than the inverse 🤪

gwynne:
> It looks like we had a race condition when writing our comments 🙂 This updated version LGTM

As long as it all works now! 😬 You'll have to merge, though; I haven't had write access to the Apple org since 2021 or so 😅

itingliu:
@swift-ci please test

PR #361, opened 2023-12-29
Add support for Dictionary Key decoding

miku1958:
When the Key of a Dictionary is not String, it will cause Decode to fail and try to decode it as an array before failing, e.g.: https://github.com/apple/swift/issues/63778


miku1958:
@swift-ci please test

YOCKOW:
I guess, if we want some compatibility with *current Foundation*, we should use [SE-0320](https://github.com/apple/swift-evolution/blob/main/proposals/0320-codingkeyrepresentable.md) rather than something like this.

miku1958:
@YOCKOW I think [SE-0320](https://github.com/apple/swift-evolution/blob/main/proposals/0320-codingkeyrepresentable.md) is for the old JSONDecoder design; In the new implementation, JSONDecoder only allows strings to be used as keys.

miku1958:
Wait, SE-320 has been implemented. Ok let me close this

PR #364, opened 2024-01-02
AttributedString --> NSAS conversion is expensive due to dictionary merging and setting empty attribute dictionaries

jmschonfeld:
This is a small patch that improves the performance of `AttributedString` to `NSAttributedString` conversion. First, dynamically merging the descriptions of each attribute scope can be expensive, so we now only do that once and cache the result rather than combining scopes dynamically for each request. Additionally, setting an empty attribute dictionary over a run can also be expensive, so we now ensure that we only set attributes for a range of the created NSAS if we have attributes to set (since we know that the string starts with no attributes)


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #356, opened 2023-12-20
[Gregorian Calendar] Implement `dateInterval(of:for:)`

itingliu:
This mostly follows the implementation of Calendar_ICU.

PR #365, opened 2024-01-02
[Gregorian Calendar] Implement `range(of:in:for:)`.

itingliu:
The logic is identical to that in Calendar_ICU.

This builds on #356. I'll rebase after that is merged.


Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -258,30 +262,259 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
...
+        guard let ord1 = ordinality(of: smaller, in: larger, for: interval.start + 0.1) else {
+            return nil
+        }
+
+        guard let ord2 = ordinality(of: smaller, in: larger, for: interval.start + interval.duration - 0.1) else {
+            return nil
+        }
+
+        guard ord2 >= ord1 else {
+            // Protect against an unexpected value from ICU for ord2
```
parkera, line 309: Do we still need this protection here? Maybe if we refactor them together...

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -258,30 +262,259 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
...
+                return _algorithmD(at: capped)
+            default:
+                break
+            }
+        default:
+            break
+        }
+
+        return nil
     }
```
parkera, line 518: I do want to factor this out, but I'm totally ok with landing this as-is until we figure out how to do that.

parkera (approved):

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
itingliu: I removed the comment since that's no longer correct. Although, what refactor are you referring to?

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
parkera: Just the eventual possibility that we put this algorithmA,B,C,D stuff into one place. Not a blocker here.

itingliu: Agreed. That's definitely on my plate

PR #367, opened 2024-01-04
Fix an Off-By-One Error in String's File System Representation

jmschonfeld:
String's file system representation uses a buffer of size `PATH_MAX`, so strings up to size `PATH_MAX - 1` should fit (to leave room for the null byte). However, due to an off by one error, we accidentally failed for strings of exactly length `PATH_MAX - 1`.


parkera (approved):

jmschonfeld:
@swift-ci please test

PR #368, opened 2024-01-04
[Gregorian Calendar] Implement weekend-related functions

itingliu:
Calendar's weekend support requires the following functions:
1. `isDateInWeekend`: answers whether a given date is in weekend
2. `weekendRange`: function that returns the days and times of the weekend
    
This implements the first. The second one, weekend range, is in fact related to calendar's locale rather than calendar itself. Move it over to `_LocaleProtocol` and have locale handle it.


itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Locale/Locale_ICU.swift
```diff
@@ -1229,6 +1230,115 @@ internal final class _LocaleICU: _LocaleProtocol, Sendable {
...
+                weekdaysIndex[i] = (weekdaysIndex[i - 1] % 7) + 1
+            }
+
+            var weekdayTypes : [UCalendarWeekdayType] = [UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY]
+
+            var onset: UInt32?
+            var cease: UInt32?
+
+            var status = U_ZERO_ERROR
+            let cal = ucal_open(nil, 0, identifier, UCAL_DEFAULT, &status)
```
parkera, line 1255: Is it going to be more expensive (CPU time) to open and close the calendar here, vs on `Calendar_ICU` where we reuse the `ucalendar` we already had?

I don't think it is a showstopper, because this makes a lot more sense, but we should be aware.

itingliu: I don't see an obvious performance regression. This value is cached inside this `_LocaleICU` instance. So we'd only call `ucal_open` once per Calendar-Locale combination. Since we're also caching locale by its identifier, we'd be fetching from the cache most of the time.

parkera (approved):

PR #369, opened 2024-01-04
Provide manual name override for local swift-foundation in the benchmark package

jmschonfeld:
The benchmark package relies on the package at `../` for the `FoundationEssentials` module. By default the package name is inferred to be the directory name, so the dependency on `"swift-foundation"` works if your directory is named `swift-foundation` but fails to resolve the package if your directory is named differently. This adds a manual override for the package name to ensure the name of the directory doesn't influence whether you can build benchmarks

While I'm at it, I added `Benchmarks/.swiftpm` to the `.gitignore` in addition to `/.swiftpm`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,7 +44,7 @@ Package.resolved
 #
 # Xcode automatically generates this directory with a .xcworkspacedata file and xcuserdata
 # hence it is not needed unless you have added a package configuration file to your project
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #291, opened 2023-10-11
Few benchmarks for predicate evaluation

ser-0xff:
# Motivation

Predicate matching performance is 10x slower [over custom implementation](https://forums.swift.org/t/pitch-swift-predicates/62000/54), this PR adds benchmark plugin to get a baseline performance and to help future Predicates development.

Benchmark is added in the same fashion as for [swift-nio](https://github.com/apple/swift-nio/pull/2536) and [swift-certifcates](https://github.com/apple/swift-certificates/pull/125).

There are 5 benchmarks.
1) Benchmark with a very simple predicate giving a 'true' for everything shows we can evaluate about 4 million objects per second on M1 Pro.
2) predicate with single KeyPath condition with variable evaluates ~ 2.5 million/sec
3) predicate with single KeyPath condition with computed property evaluates ~ 1.5 million/sec
4) predicate with single KeyPath condition with nested computed property evaluates ~ 1.3 million/sec
5) predicate with three KeyPath conditions with nested computed properties evaluates ~ 600 K/sec

The most common case for our product is more than one condition, and we widely use computed properties, so if we will use more than 3 conditions then we will have throughout less than 600K/sec, which not too much for us. 

All benchmarks can be executed with `swift package benchmark --scale` in the project 'Benchmarks' directory.

We observe `swift_getAtKeyPath` execution takes significant time relative to whole execution time.
image
The issue can be reproduced with a nested computed properties in benchmark (4) and (5). A single benchmark can be executed with command `swift package benchmark --scale --filter "Predicate #5 - 3 KeyPath nested computed property conditions"`
image

Also we observe that predicate evaluation entails few malloc calls, what is also possibly the cause of poor performance.

Each benchmark report a set of information, the most interesting for now is a 'Throughput' line, which shows how many evaluations per second can be executed.

To run with profile in Xcode:

[Running benchmarks in Xcode and using Instruments for profiling benchmarks](https://swiftpackageindex.com/ordo-one/package-benchmark/1.11.2/documentation/benchmark/runningbenchmarks#Running-benchmarks-in-Xcode-and-using-Instruments-for-profiling-benchmarks)
Profiling benchmarks or building the benchmarks in release mode in Xcode with jemalloc is currently not supported (as Xcode currently doesn’t support interposition of the malloc library) and requires disabling jemalloc.
Make sure Xcode is closed and then open it from the CLI with the BENCHMARK_DISABLE_JEMALLOC environment variable set e.g.:

```
[...]

PR #370, opened 2024-01-05
Add some benchmarks for Calendar and Locale

parkera:
Using our nice new benchmarking suite (#291), add some basic benchmarks for Calendar and Locale. We can expand this with more tests over time of course.


parkera:
@swift-ci test

itingliu (approved):

PR #371, opened 2024-01-05
Improve performance of UUID equality

parkera:
Improves the performance of UUID equality. Also, restores desired behavior (for security reasons) of doing the comparison in constant time.


kperryua (approved):

parkera:
@swift-ci test

PR #374, opened 2024-01-11
Fix out of order result check in ICUPatternGenerator

gwynne:
Simple, straightforward fix - check for success _after_ attempting the operation rather than before it. I can't imagine `udatpg_clone()` fails even remotely often enough for this to ever get noticed under normal conditions (nor can I think of a meaningful unit test, given that a failure trips a fatal error).


itingliu (approved):
Thanks!

itingliu:
@swift-ci please test

parkera (approved):

PR #376, opened 2024-01-12
[Proposal] RangeSet/IndexSet Conversion

jmschonfeld:
This proposal adds new API to `RangeSet<Int>` and `IndexSet` to allow conversion between the two types

### DIFF
--- a/Proposals/0005-rangeset-indexset-conversion.md
+++ b/Proposals/0005-rangeset-indexset-conversion.md
@@ -1,8 +1,9 @@
 # `RangeSet`/`IndexSet` Conversion
 
-* Proposal: SF-NNNN
+* Proposal: SF-0005
 * Author(s): [Jeremy Schonfeld](https://github.com/jmschonfeld)
-* Status: **Pitch**
+* Review Manager: [Charles Hu](https://github.com/iCharlesHu)
+* Status: **Active Review: Jan 26, 2024 ... Feb 2, 2024**
 * Implementation: _coming soon_
 
 ## Introduction/Motivation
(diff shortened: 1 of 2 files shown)


stephentyrone:
Would it be useful to have an `IndexSet` initializer that dropped any negative indices, and only converted the non-negative?

jmschonfeld:
> Would it be useful to have an `IndexSet` initializer that dropped any negative indices, and only converted the non-negative?

I'd be potentially open to that, but I'm not sure exactly how useful it would be. I can't think of a good scenario in which you'd want to drop the negative indices so I'm worried it could become an accidental trap if it were easier to find/use that API than the failable one for the many cases where you'd want to specially handle negative indices. The equivalent could be written as:

```swift
myRangeSet.remove(contentsOf: Int.min ..< 0)
let myIndexSet = IndexSet(integersIn: myRangeSet)

// - OR -

let myIndexSet = IndexSet(integersIn: myRangeSet.intersection(RangeSet(Int.min ..< 0)))
```

Those aren't exactly the most trivial things to spell out, but it's doable if we decide that a separate API isn't worth it.

jmschonfeld:
This proposal has been pitched at https://forums.swift.org/t/pitch-rangeset-indexset-conversion/69405 - there have been a few comments about the label of one of the initializers but after discussion I feel that it'd be best to keep the label consistent with existing Foundation APIs and keep the proposal as-is. This should be ready for a formal review

iCharlesHu (approved):
Active review: Jan. 26, 2024 ~ Feb. 2, 2024

PR #377, opened 2024-01-12
Implement Data reading and writing

parkera:
Implements file reading and writing for `Data`. For now use `String` for path instead of `URL`.

### DIFF
--- a/Benchmarks/Benchmarks/DataIO/DataIO.swift
+++ b/Benchmarks/Benchmarks/DataIO/DataIO.swift
@@ -1,0 +1,87 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import func Benchmark.blackHole
+@testable import FoundationEssentials
+
+#if canImport(Glibc)
+import Glibc
+#endif
+#if canImport(Darwin)
+import Darwin
+#endif
+
+func testPath() -> String {
+    // Generate a random file name
+    String.temporaryDirectoryPath.appendingPathComponent("testfile-\(UUID().uuidString)")
+}
+
+func generateTestData() -> Data {
+    // 16 MB file, big enough to trigger things like chunking
+    let count = 1 << 24
+    
+    let memory = malloc(count)!
+    let ptr = memory.bindMemory(to: UInt8.self, capacity: count)
+    
+    // Set a few bytes so we're sure to not be all zeros
+    let buf = UnsafeMutableBufferPointer(start: ptr, count: count)
+    for i in 0..<128 {
+        buf[i] = UInt8.random(in: 1..<42)
+    }
+    
+    return Data(bytesNoCopy: ptr, count: count, deallocator: .free)
+}
+
+func cleanup(at path: String) {
+    _ = unlink(path)
+    // Ignore any errors
+}
+
+let data = generateTestData()
+let readMe = testPath()
+
+let benchmarks = {
+    Benchmark.defaultConfiguration.maxIterations = 1_000_000_000
+    Benchmark.defaultConfiguration.maxDuration = .seconds(3)
+    Benchmark.defaultConfiguration.scalingFactor = .kilo
+//    Benchmark.defaultConfiguration.metrics = .arc + [.cpuTotal, .wallClock, .mallocCountTotal, .throughput] // use ARC to see traffic
+//  Benchmark.defaultConfiguration.metrics = [.cpuTotal, .wallClock, .mallocCountTotal, .throughput] // skip ARC as it has some overhead
+  Benchmark.defaultConfiguration.metrics = .all // Use all metrics to easily see which ones are of interest for this benchmark suite
+
+    Benchmark("read-write-emptyFile") { benchmark in
+        let path = testPath()
+        let data = Data()
+        try data.write(to: path)
+        let read = try Data(contentsOf: path, options: [])
+        cleanup(at: path)
+    }
+
+    Benchmark("write-regularFile") { benchmark in
+        let path = testPath()
+        try data.write(to: path)
+        cleanup(at: path)
+    }
+    
+    Benchmark("read-regularFile", 
+              configuration: .init(
+                setup: {
+                    try! data.write(to: readMe)
+                },
+                teardown: {
+                    cleanup(at: readMe)
+                }
+              )
+    ) { benchmark in
+        blackHole(try Data(contentsOf: readMe))
+    }
+}
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -47,7 +47,7 @@ let package = Package(
         .executableTarget(
             name: "DataIOBenchmarks",
             dependencies: [
-                .product(name: "FoundationEssentials", package: "swift-foundation"),
+                .product(name: "FoundationEssentials", package: "swift-foundation-local"),
                 .product(name: "Benchmark", package: "package-benchmark"),
             ],
             path: "Benchmarks/DataIO",
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -44,5 +44,17 @@ let package = Package(
                 .plugin(name: "BenchmarkPlugin", package: "package-benchmark")
             ]
         ),
+        .executableTarget(
+            name: "DataIOBenchmarks",
+            dependencies: [
+                .product(name: "FoundationEssentials", package: "swift-foundation"),
+                .product(name: "Benchmark", package: "package-benchmark"),
+            ],
+            path: "Benchmarks/DataIO",
+            plugins: [
+                .plugin(name: "BenchmarkPlugin", package: "package-benchmark")
+            ]
+        ),
+
     ]
 )
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -1,0 +1,345 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if FOUNDATION_FRAMEWORK
+@_implementationOnly import _ForSwiftFoundation
+@_implementationOnly import _CShims
+#else
+package import _CShims
+#endif
+
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Glibc)
+import Glibc
+#endif
+
+private func readExtendedAttributesFromFileDescriptor(_ fd: Int32, attrsToRead: [String]) -> [String : Data] {
+#if canImport(Darwin) && !NO_FILESYSTEM
+    var output: [String : Data] = [:]
+    for key in attrsToRead {
+        key.withCString { keyStr in
+            let maxXAttrLength = 1000
+            withUnsafeTemporaryAllocation(of: CUnsignedChar.self, capacity: maxXAttrLength) { buf in
+                let result = fgetxattr(fd, keyStr, buf.baseAddress, maxXAttrLength, 0, 0)
+                if result != -1 {
+                    // -1 means no such attribute
+                    // Force unwrap buffer - if we do not have a base address, assert is appropriate.
+                    output[key] = Data(bytes: buf.baseAddress!, count: result)
+                } else if errno == ERANGE {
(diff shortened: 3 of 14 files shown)


parkera:
@swift-ci test

Benchmarks/Package.swift
```diff
@@ -44,5 +44,17 @@ let package = Package(
                 .plugin(name: "BenchmarkPlugin", package: "package-benchmark")
             ]
         ),
+        .executableTarget(
+            name: "DataIOBenchmarks",
+            dependencies: [
+                .product(name: "FoundationEssentials", package: "swift-foundation"),
```
jmschonfeld, line 50: I think you likely want `swift-foundation-local` here instead of `swift-foundation`

Benchmarks/Benchmarks/DataIO/DataIO.swift
```diff
@@ -0,0 +1,87 @@
...
+    
+    Benchmark("read-regularFile", 
+              configuration: .init(
+                setup: {
+                    try! data.write(to: readMe)
+                },
+                teardown: {
+                    cleanup(at: readMe)
+                }
+              )
```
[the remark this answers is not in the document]
hassila: See https://swiftpackageindex.com/ordo-one/package-benchmark/1.22.0/documentation/benchmark/writingbenchmarks#Customize-startupteardown-code-for-benchmarks

Benchmarks/Package.swift
parkera: Indeed, thanks,.

Benchmarks/Benchmarks/DataIO/DataIO.swift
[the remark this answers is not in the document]
parkera: By the way, do you have any advice on one-target-many-benchmarks vs many-targets-one-benchmark?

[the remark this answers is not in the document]
hassila: That's a good question and the answer is "it depends", but a few reflections:

I think it in general makes sense to group benchmarks together in one target which you want to check together (ie. they depend on the same class, set of classes, of module / component which is under test) just as a convenience for the engineer who wants to validate performance iteratively while making changes to a specific part of the software. For CI the structure just doesn't matter as one would typically run all of the benchmarks anyway, so the structure is just optimising for a) ease-of-running during interactive development and b) ease-of-updating/changing benchmarks as they are related.

It is similar to unit tests in nature that you usually want to have tests that are related to each other in a "suite" (one target for benchmarks).

The Benchmark package supports filtering/skipping of both targets and benchmarks with regexes:
```
--filter <filter>
Benchmarks matching the regexp filter that should be run
--skip <skip>
Benchmarks matching the regexp filter that should be skipped
--target <target>
Benchmark targets matching the regexp filter that should be run
--skip-target <skip-target>
Benchmark targets matching the regexp filter that should be skipped
```

Which makes it easy to run partial sets of targets/benchmarks if using a thoughtful naming convention too.

One difference is that you might have benchmarks that are both similar to unit tests, but also much more complex benchmarks that are more similar to integration tests, so the answer depends a bit also on the flavour of benchmarks you do.

For longer more complex benchmarks "integration test"-style, it's a good idea to do as swift-nio did and just break out the actual code under test into separate files/targets and just have the benchmark file itself be benchmark configuration and basically a driver - see e.g. https://github.com/apple/swift-nio/blob/main/Benchmarks/Benchmarks/NIOPosixBenchmarks/Benchmarks.swift
 
While simpler benchmarks can be done inline...

So more concretely for Foundation, I'd probably consider a structure with targets with a naming convention, something along the lines:

`FoundationEssentials`
`FoundationEssentials-Date`
`FoundationEssentials-Serialization`
`FoundationEssentials-Predicate`
`FoundationInternationalization`
`FoundationNetworking`
`FoundationNetworking-URLSession`

Etc, basically a common targets (e.g. `FoundationEssentials`) where most benchmarks for that module reside, but then having additional targets for whatever types that might need more complex sets of benchmarks (I just picked some examples above how it could look)) - this allows for regex filtering if wanting to run benchmarks for a given part of Foundation easily with e.g. `swift package benchmark --target "FoundationEssentials.*"` but doesn't make the benchmarks to balloon out of control in terms of size in a single file. But thats me, YMMV :-)

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -0,0 +1,345 @@
...
+    var deallocator: Data.Deallocator?
+}
+
+internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, maxLength: Int?, options: Data.ReadingOptions, attributesToRead: [String], attributes: inout [String: Data]) throws -> ReadBytesResult {
+    if inPath.isEmpty {
+        // For compatibility, throw a different error than the perhaps-expected 'file not found' here (41646641)
+        throw CocoaError(.fileReadInvalidFileName)
+    }
+    
+    let fd = try inPath.withFileSystemRepresentation { inPathFileSystemRep in
```
karwa, line 175: I wonder if `withFileSystemRepresentation` is necessary here and elsewhere, even if it only has an effect on Darwin platforms.

I actually think that method should be deprecated. Fundamentally, Foundation should not be trying to second-guess how the filesystem is going to transform a file name. Even the operating system doesn't really know what kind of transformations are going to happen, because that's handled by the (possibly third-party) filesystem driver. You just give it a file name, and if it needs to be transformed, the driver will do it.

AFAICT, other libraries which wrap the POSIX `open` function do not perform this transformation on Darwin platforms (e.g. NodeJS doesn't), so why is it necessary for Foundation to do it?

At the very least, I think this needs documenting, but as explained I think we should investigate avoiding this work altogether.

karwa:
It is unfortunate that this does not use `swift-system` -- that library is supposed to expose OS APIs such as `open` and `read` in their full fidelity, which platform-abstraction libraries such as Foundation can use as a safe, idiomatic interface to the OS (hence the name "system"). Its maintainers at Apple have invested a significant amount of time on it.

It would be much nicer if this patch could introduce a set of functions to integrate `Data` with `FilePath` (even as an internal implementation detail), building on the functionality already present in `swift-system`, and then layer its public API on top by converting strings and URLs to `FilePath`s.

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -0,0 +1,345 @@
...
+        localProgress?.completedUnitCount = 1
+        
+        result = ReadBytesResult(bytes: bytes, length: Int(fileSize), deallocator: .unmap)
+#else
+        // This was disabled above
+        fatalError("mapping should not be enabled")
+#endif
+    } else {
+        // We've verified above that fileSize will fit in `Int`
+        guard let bytes = malloc(Int(fileSize)) else {
```
karwa, line 264: It's unfortunate that this always allocates. `Data` can store its contents [in an inline buffer](https://github.com/apple/swift-foundation/blob/87fdbaef20095a5c04dc952b291cc5f20d01bca6/Sources/FoundationEssentials/Data/Data.swift#L1265). 

It's not huge - on 64-bit systems, the inline buffer is 14 bytes, but for very small files it could be useful.

Benchmarks/Benchmarks/DataIO/DataIO.swift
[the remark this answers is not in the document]
parkera: Thanks for that detailed response. I think the approach we'll take in the short term is to just focus on getting a base level of tests in. Once we see how/if/when we use them, we can always refactor it to match requirements for CI or easier at-desk usage.

parkera:
> It is unfortunate that this does not use `swift-system` -- that library is supposed to expose OS APIs such as `open` and `read` in their full fidelity, which platform-abstraction libraries such as Foundation can use as a safe, idiomatic interface to the OS (hence the name "system"). Its maintainers at Apple have invested a significant amount of time on it.
> 
> It would be much nicer if this patch could introduce a set of functions to integrate `Data` with `FilePath` (even as an internal implementation detail), building on the functionality already present in `swift-system`, and then layer its public API on top by converting strings and URLs to `FilePath`s.

This is a long-term goal. Our shorter term one is to provide a drop in replacement implementation for this that works on all platforms (including Darwin), and the best way to do that is to make sure we replicate the existing functionality exactly.

Once we address some limitations in swift system (including some that are internal to Darwin builds but affect us here) we intend to figure out how to expose `FilePath` based API via Foundation.

Sources/FoundationEssentials/Data/Data+Reading.swift
parkera: I suspect the overhead of doing the I/O is much higher than a <=14 byte malloc, but your point is still valid. I think plumbing through the possibility of using inline storage would require some internal `Data` initializers, which we can follow up with.

parkera:
@swift-ci test

Sources/FoundationEssentials/Data/Data+Reading.swift
jmschonfeld: I added `withFileSystemRepresentation` in a prior PR in order to mirror the functionality of the existing APIs that Foundation provides on `String`/`NSString` (but with a swifty interface). I think that we should use `withFileSystemRepresentation` here and other places when calling down to C APIs that accept paths for a few reasons. First, it maintains compatibility with existing Foundation implementations. The existing Foundation framework performs this transformation and changing the implementation as we port it to swift could lead to compatibility-breaking behavior, so a safer bet is to maintain the compatibility here.

But perhaps more importantly, I think it's still the correct thing to do. You are right that the transformations can be handled by the file system driver and that different file systems can behave differently within the same OS, so this implementation isn't _completely_ accurate. That being said, I think this is still the best choice for Darwin. While APFS handles unicode normalization, HFS+ (which is also commonly used on Darwin platforms) does not and failing to normalize strings here (using this special character table) can result in incorrect paths being used. If we don't normalize HFS+ paths, we might not find the file we're looking for due to a mismatch which could be surprising to some developers. In the worst case, we perform a normalization that something like APFS would have done for us, but in the best case we ensure that we normalize for file systems where that is a requirement.

Another benefit is that `withFileSystemRepresentation` is portable - for example, my understanding is that Windows should use a UTF-16 representation and this allows us to have one entry point for conversion to file system representations that can vary based on the platform.

I'm happy to add some documentation to that function as a followup to explain this for maintainers since I can see how that history/context might not be clear from the API itself.

jmschonfeld: Despite the surprising behavior that was my understanding of how it worked. I had some older documentation I was looking to explain this but sadly I can't seem to find that document, and writing a sample app operating on an HFS volume actually seems to prove (from what I can tell) that the HFS+ driver does automatically decompose when creating/reading from the file. So after taking a further look, it does seem that perhaps this normalization might not have been originally necessary (or perhaps this is an artifact of a time when HFS didn't perform that normalization for us). I'll see if I can gather some more info there with examples to thoroughly document it.

That being said - although it looks like I was incorrect about the necessity for interaction with HFS, the compatibility justification still stands for file systems that are normalization-preserving like APFS. If an app exists today that stores a file name and reads/writes to that file using Foundation APIs, the filename will always be the fully decomposed format. Changing this implementation to avoid the decomposition would be a behavioral difference on an APFS volume where normalization is preserved, and I think we'd likely see fallout from changing Foundations' behavior here despite it being different behavior than other libraries that might call `open without this normalization.

I'll do some more investigation on the necessity of this (in case other clients calling `open` might be incorrect in not localizing) but I think that can be handled separately since I think we need to land this `Data` implementation as-is.

parkera:
@swift-ci test

Sources/FoundationEssentials/Data/Data+Reading.swift
karwa: Thanks for investigating. I hope that this Foundation reimplementation effort can provide an opportunity to reexamine some of these behaviours. I appreciate that this may take time, but I still like to raise the issue so people know about it and look in to it.

I'll just give a quick response to this point:

> That being said - although it looks like I was incorrect about the necessity for interaction with HFS, the compatibility justification still stands for file systems that are normalization-preserving like APFS. If an app exists today that stores a file name and reads/writes to that file using Foundation APIs, the filename will always be the fully decomposed format. Changing this implementation to avoid the decomposition would be a behavioral difference on an APFS volume where normalization is preserved, and I think we'd likely see fallout from changing Foundations' behavior here despite it being different behavior than other libraries that might call `open without this normalization.

While APFS is normalisation-preserving, I believe it is still normalisation-aware. So if you save a file `caf\u{00E9}` (NFC), you can open it using `cafe\u{0301}` (NFD), and it won't allow both versions to exist in the same folder.

The only visible difference between HFS+ and APFS in this matter is if you create a file, then iterate the contents of its containing directory and look for the file using a Unicode-unaware search such as `memcmp`. HFS+ will have normalised the filename, so you'll need to look for the "file system representation" in order to get a match, while APFS will have stored the filename as you gave it.

The problem is that this whole process is ill-advised. Even if you can open a file using the name "X", that doesn't necessarily mean you will find a file literally named "X" if you iterate the directory's contents. The filesystem might transform the filename in any number of ways we can't predict (e.g. a case-insensitive filesystem might normalise everything to uppercase, and some filesystems need to truncate filenames). Users shouldn't be looking up files in that way.

That's why I argue that `withFileSystemRepresentation` should be deprecated. We should put an end to this idea that a directory will contain a particular byte-string, because it isn't true. Once applications no longer operate with that assumption, we wouldn't need to do this normalisation here, even on APFS -- we could respect the APFS designers intent to have their filesystem be normalisation-preserving.

Anyway, that's enough for now. I'd be very interested to know what your investigations uncover; please do @ me here or on the Swift forums if possible.

parkera:
@swift-ci test

jmschonfeld (approved):

PR #384, opened 2024-01-17
Use deployment target instead of 9999 for min OS version

parkera:
Instead of using the placeholder `9999` availability, use the min deployment target. This simplifies implementation inside the package -- we no longer need to check for availability of something which is entirely contained within the package itself due to an availability marker on public API.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -6,10 +6,13 @@ import CompilerPluginSupport
 
 // Availability Macros
 let availabilityMacros: [SwiftSetting] = [
-    "FoundationPreview 0.1:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
-    "FoundationPreview 0.2:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
-    "FoundationPreview 0.3:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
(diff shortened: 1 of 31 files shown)


parkera:
@swift-ci test

Package.swift
```diff
@@ -6,10 +6,13 @@ import CompilerPluginSupport
 
 // Availability Macros
 let availabilityMacros: [SwiftSetting] = [
-    "FoundationPreview 0.1:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
-    "FoundationPreview 0.2:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
-    "FoundationPreview 0.3:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
-    "FoundationPreview 0.4:macOS 9999, iOS 9999, tvOS 9999, watchOS 9999",
+    "FoundationPreview 0.1:macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4",
+    "FoundationPredicate 0.1:macOS 14, iOS 17, tvOS 17, watchOS 10",
```
parkera, line 10: Predicate has an actual OS dependency of macOS 14 (et. al.) and higher. We use a different macro for it.

jmschonfeld (approved):

PR #385, opened 2024-01-18
[Gregorian Calendar] Support day of year calculation

itingliu:


itingliu:
@swift-ci please test

parkera (approved):

PR #388, opened 2024-01-26
Use `_GregorianCalendar` as the backing calendar for `struct Calendar`

itingliu:
This patch does the following
- Use `_GregorianCalendar` in place of `_CalendarICU` as the backing calendar when `foundation_essentials_feature` is enable
- Add a second backing calendar variable to compare the results of the two implementation and log when the result differs.
- Add `Calendar_Gregorian_Autoupdating` just like `Calendar_Autoupdating`, which forwards every call to the internally cached calendar


itingliu:
@swift-ci please test

PR #391, opened 2024-01-29
Feat: implemented divide extension for string

siamakrostami:
Extension on String to calculate the remainder value of two strings.
This extension allows you to calculate the remainder when dividing one string
(representing a decimal number) by another string (representing the divisor).
The strings are internally converted to NSDecimalNumber for precision.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Divide.swift
+++ b/Sources/FoundationEssentials/String/String+Divide.swift
@@ -1,0 +1,45 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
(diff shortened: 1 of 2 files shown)


karwa:
Have you read https://github.com/apple/swift-foundation/blob/main/Evolution.md?

In particular:

> How to propose a change
>
> - Consider the goal: Before proposing a change, please consider how your idea fits into the goals and themes for the upcoming release.
> - Socialize the idea: Propose a rough sketch of the idea in the "pitches" section of the Swift forums, the problems it solves, what the solution looks like, etc., to gauge interest from the community.

PR #393, opened 2024-01-30
[json] implementation note regarding Double

glessard:
Large integers are valid JSON, and we accept loss of precision when converting them to Double.

Since this is an implementation-specific decision, we should document it somewhere.
(The RFC does not specify behaviour for this case, one way or the other.)


glessard:
@swift-ci please test

kperryua (approved):

iCharlesHu (approved):

PR #395, opened 2024-02-01
Calendar.date(_: , matchesComponents: ) does not check for leap months correctly

hristost:
When checking if a date matches components, we were checking whether isLeapMonth has been set, but were not reading its actual value. This meant that this method would sometimes return wrong results in calendars that have leap months.


hristost:
@swift-ci please test

parkera (approved):

PR #396, opened 2024-02-02
Rename the UUID functions so they do not conflict with other definitions

parkera:
Since we export the `uuid` functions from the library, we should prefix them with something so they don't conflict with the (commonly used) names from another UUID library.


parkera:
@swift-ci test

iCharlesHu (approved):
Thanks Tony!! I also noticed that all the module names are global as well. At some point we should probably change the _CShims name to be more specific as well.

PR #397, opened 2024-02-02
Introduce Swift Subprocess Proposal

iCharlesHu:
This PR introduces a proposal for `Subprocess`.

### DIFF
--- a/Proposals/NNNN-swift-subprocess.md
+++ b/Proposals/NNNN-swift-subprocess.md
@@ -10,6 +10,12 @@
 ## Revision History
 
 * **v1**: Initial draft
+* **v2**: Minor Updates:
+    - Switched `AsyncBytes` to be backed by `DispatchIO`.
+    - Introduced `resolveExecutablePath(withEnvironment:)` to enable explicit lookup of the executable path.
+    - Added a new option, `closeWhenDone`, to automatically close the file descriptors passed in via `.readFrom` and friends.
+    - Introduced a new parameter, `shouldSendToProcessGroup`, in the `sendSignal` function to control whether the signal should be sent to the process or the process group.
+    - Introduced a section on "Future Directions."
 
 ## Introduction
 
@@ -280,8 +286,9 @@ public struct Subprocess: Sendable {
     // The standard error of the child process, expressed as AsyncSequence<UInt8>
     // This property is `nil` if the standard error is discarded or written to disk
     public var standardError: AsyncBytes? { get }
-
-    public func sendSignal(_ signal: Signal) throws
+    // If `shouldSendToProcessGroup` is `true`, the signal will be send to the entire process
+    // group instead of the current process.
+    public func sendSignal(_ signal: Signal, toProcessGroup shouldSendToProcessGroup: Bool) throws
 }
 
 extension Subprocess {
@@ -491,7 +498,7 @@ _(We welcome community input on which Linux and Windows "escape hatches" we shou
 
 In addition to supporting the direct passing of `Sequence<UInt8>` and `AsyncSequence<UInt8>` as the standard input to the child process, `Subprocess` also provides a `Subprocess.InputMethod` type that includes two additional input options:
 - `.noInput`: Specifies that the subprocess does not require any standard input. This is the default value.
-- `.readingFrom`: Specifies that the subprocess should read its standard input from a file descriptor provided by the developer.
+- `.readFrom`: Specifies that the subprocess should read its standard input from a file descriptor provided by the developer. Subprocess will automatically close the file descriptor after the process exits if `closeWhenDone` is set to `true`.
 
 ```swift
 extension Subprocess {
@@ -501,7 +508,7 @@ extension Subprocess {
     @available(watchOS, unavailable)
     public struct InputMethod: Sendable, Hashable {
         public static var noInput: Self
-        public static func readFrom(_ fd: FileDescriptor) -> Self
+        public static func readFrom(_ fd: FileDescriptor, closeWhenDone: Bool) -> Self
     }
 }
 ```
@@ -514,7 +521,7 @@ let ls = try await Subprocess.run(executing: .named("ls"))
 
 // Alteratively, developers could pass in a file descriptor
 let fd: FileDescriptor = ...
-let cat = try await Subprocess.run(executing: .named("cat"), input: .readingFrom(fd))
+let cat = try await Subprocess.run(executing: .named("cat"), input: .readFrom(fd, closeWhenDone: true))
 
 // Pass in a async sequence directly
 let sequence: AsyncSequence = ...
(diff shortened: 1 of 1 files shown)


Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,853 @@
...
+        /// such as `/bin/ls`
+        public static func at(_ filePath: FilePath) -> Self
+    }
+}
+```
+
+
+### `Subprocess.Environment`
+
+`struct Environment` is used to configure how should the process being lunched receive its environment values:
```
davedelong, line 672: typo: `launched` 🥪

iCharlesHu:
Updated based on community feedbacks. Thanks all!

Proposals/NNNN-swift-subprocess.md
iCharlesHu: Opps

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,882 @@
...
+
+We propose a new `struct Subprocess`. Developers primarily interact with `Subprocess` via the static `run` methods which asynchronously executes a subprocess.
+
+```swift
+@available(FoundationPreview 0.4, *)
+@available(iOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+extension Subprocess {
+    public static func run(
```
weissi, line 135: @iCharlesHu This needs to specify the cancellation behaviour. For Swift Concurrency to work it's crucial that under task cancellation (`taskGroup.cancelAll()`) the `async` methods return quickly. For sub processes that will mean that it needs to get rid of the process which requires sending a signal -- likely `SIGKILL`. But this needs to be clearly specified.

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,882 @@
...
+
+```swift
+extension Subprocess {
+    @available(FoundationPreview 0.4, *)
+    @available(iOS, unavailable)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    public struct InputMethod: Sendable, Hashable {
+        public static var noInput: Self
+        public static func readFrom(_ fd: FileDescriptor, closeWhenDone: Bool) -> Self
```
weissi, line 511: It's crucial here to specify who closes this file descriptor in the parent and precisely when and how. Please note that these file descriptors must usually be closed in the parent _right after_ fork/posix_spawn. It is _not_ acceptable to wait with the closing until after the process exits because that may lead to a deadlock. For example, the parent commonly inherits a pipe into the child. But the child can only see EOF on the pipe if there are no more writers left. Also: It's crucial to keep the number of open file descriptors low (macOS default limit is 256).

I would assume that `Subprocess` either always closes or has the ability to close file descriptors that it inherits into a child at the earliest possible point. But there doesn't seem to be API or docs around this (unless I missed it).

weissi: (same for `writeTo`).

weissi (changes requested):
This looks like a great basis for an API. Requesting changes because two crucial lifecycle things are not specified (from what I can tell):

1. What happens under Swift Concurrency cancellation?
2. Who manages the lifetime (i.e. who closes) file descriptors inherited into sub processes

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,882 @@
...
+```
+
+_(We welcome community input on which Linux and Windows "escape hatches" we should add)_
+
+
+### `Subprocess.InputMethod`
+
+In addition to supporting the direct passing of `Sequence<UInt8>` and `AsyncSequence<UInt8>` as the standard input to the child process, `Subprocess` also provides a `Subprocess.InputMethod` type that includes two additional input options:
+- `.noInput`: Specifies that the subprocess does not require any standard input. This is the default value.
+- `.readFrom`: Specifies that the subprocess should read its standard input from a file descriptor provided by the developer. Subprocess will automatically close the file descriptor after the process exits if `closeWhenDone` is set to `true`.
```
weissi, line 501: After the process is done is too late, this should happen as soon as possible.

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,882 @@
...
+
+For reference, Python uses [`shlex.split`](https://docs.python.org/3/library/shlex.html), which could serve as a valuable starting point for implementation.
+
+## Combined `stdout` and `stderr`
+
+In Python's `Subprocess`, developers can merge standard output and standard error into a single stream. This is particularly useful when an executable improperly utilizes standard error as standard output (or vice versa). We should explore the most effective way to achieve this enhancement without introducing confusion to existing parameters—perhaps by introducing a new property.
+
+## Automatic `Subprocess` Cancellation
+
+Currently, `Subprocess` does not terminate the spawned process, even if the `Subprocess` instance goes out of scope. It would be beneficial to investigate whether it is more sensible to attempt terminating the spawned process when the `Subprocess` itself goes out of scope or when the parent task is canceled.
```
weissi, line 855: It really ought to terminate the process. Or else you'll deadlock super easily. Imagine a task group which reads chunks from the sub process's output. Under cancellation that `AsyncSequence` will stop being iterated. That however means that the sub process won't be able to write a lot because the pipe will run full (as the parent is no longer reading). And that means the sub process will never exit --> deadlock.

Proposals/NNNN-swift-subprocess.md
FranzBusch: I mentioned this in the [pitch](https://forums.swift.org/t/pitch-swift-subprocess/69805/38) as well and Charles said that he is going to change this to send a SIGKILL

iCharlesHu:
Updated for v3

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,878 @@
...
+}
+```
+
+Here are some examples:
+
+```swift
+// By default `InputMethod` is set to `.noInput`
+let ls = try await Subprocess.run(executing: .named("ls"))
+
+// Alteratively, developers could pass in a file descriptor
```
itingliu, line 513:
```suggestion
// Alternatively, developers could pass in a file descriptor
```

itingliu (commented):
General comment, but non blocking for review: We should document the expected behaviors and their arguments next to each function since these behaviors might not be super straightforward

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,878 @@
+# Introducing Swift Subprocess
+
+* Proposal: [SF-NNNN](NNNN-swift-subprocess.md)
+* Authors: [Charles Hu](https://github.com/iCharlesHu)
+* Review Manager: TBD
```
itingliu, line 5:
```suggestion
* Review Manager: [Tina Liu](https://github.com/itingliu)
```

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,878 @@
+# Introducing Swift Subprocess
+
+* Proposal: [SF-NNNN](NNNN-swift-subprocess.md)
```
itingliu, line 3:
```suggestion
* Proposal: [SF-0007](0007-swift-subprocess.md)
```

Proposals/NNNN-swift-subprocess.md
```diff
@@ -0,0 +1,878 @@
...
+* Authors: [Charles Hu](https://github.com/iCharlesHu)
+* Review Manager: TBD
+* Status: **Draft**
```
itingliu, line 6:
```suggestion
* Status: **Active Review: Feb 27, 2024...Mar 05, 2024**
```

itingliu (approved):
It looks to me that pending questions were answered in V3. I'd like to move this to formal review. @iCharlesHu will keep taking questions during review period.

weissi (approved):
Approving to allow merging which will unlock the actual discussion

PR #399, opened 2024-02-02
Fix public visibility of Data writing and reading options

parkera:
These options need to be public in order to use them outside the module!


parkera:
@swift-ci test

kperryua (approved):

PR #400, opened 2024-02-03
Use Glibc or Darwin for POSIX error codes

parkera:
We don't need to redeclare POSIXError here, we can just use the one from Glibc or Darwin.

Also use the platform ifdefs from swift-corelibs-foundation

### DIFF
--- a/Sources/FoundationEssentials/ErrorCodes+POSIX.swift
+++ b/Sources/FoundationEssentials/ErrorCodes+POSIX.swift
@@ -9,343 +9,11 @@
  //
  //===----------------------------------------------------------------------===//
 
-#if !FOUNDATION_FRAMEWORK || canImport(C)
-// Define the POSIXErrorCode for all platforms here.
-public enum POSIXErrorCode : Int32 {
-    /// Operation not permitted.
-    case EPERM = 1
-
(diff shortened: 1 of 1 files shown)


parkera:
@swift-ci test

Sources/FoundationEssentials/ErrorCodes+POSIX.swift
```diff
@@ -394,469 +62,603 @@ extension POSIXErrorCode : _ErrorCodeProtocol {
...
-        return .EPERM
+    /// Operation not permitted.
+    public static var EPERM: POSIXErrorCode { 
+        return .EPERM 
     }
 
     /// No such file or directory.
-    public static var ENOENT: POSIXErrorCode {
-        return .ENOENT
+    public static var ENOENT: POSIXErrorCode {
```
jmschonfeld, line 71: Looks like this might have added some trailing spaces at the end of these lines, was that intentional?

parkera: Nope, will fix.

parkera:
@swift-ci test

jmschonfeld (approved):

PR #406, opened 2024-02-08
Export system Foundation for Darwin platforms

iCharlesHu:
Export system Foundation on Darwin platforms. This allows us to lower the supported platforms and allow this package to be used as a dependency on Darwin.

---
While I was here: deleted `Package-5.8.swift` since it's no longer used and really diverged from the main file.


Package.swift
```diff
@@ -4,120 +4,170 @@
 import PackageDescription
 import CompilerPluginSupport
 
-// Availability Macros
-let availabilityMacros: [SwiftSetting] = [
```
iCharlesHu, line 8: @parkera  This might have to default to be `true` otherwise tests won't run on CI... or maybe we should set some special environment variables for testing?

parkera: Hm, how does the swift-crypto project do this? We should certainly resolve before merging because we don't want to effectively disable building for macOS in CI.

Package.swift
```diff
@@ -4,120 +4,170 @@
...
-        // Foundation (umbrella)
-        .target(
-            name: "FoundationPreview",
-            dependencies: [
+#if canImport(Darwin)
+let useSystemFoundation = !developmentOnDarwin
+#else
+// On non Darwin platforms always use package
+let useSystemFoundation = false
+#endif
```
FranzBusch, line 15: I don't recommend doing it this way since this will break cross compilation from Darwin to another platform. `swift-crypto` just defines a variable `isDevelopment = false` here and package maintainers manually set that `true`. Setting it to `true` allows to run the code in the package on Darwin platforms. In general using `#if canImport` in `Package.swift` is not a good idea cc @MaxDesiatov

MaxDesiatov: +1 to what Franz said, this check will just break all packages depending on `swift-foundation` when cross-compiling.

MaxDesiatov: Keep in mind that for some platforms one can only cross-compile, e.g. Wasm/WASI, embedded etc. Presence of `#if canImport` or `#if os` checks in `Package.swift` make this package and everything that depends on it unusable for such platforms.

parkera:
Closing this for now, since we're going a different direction in the short term.

PR #408, opened 2024-02-09
Use correct start index when accessing Data slices

parkera:
`Data` slices don't start with zero indices.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -75,7 +75,7 @@ private func writeToFileDescriptorWithProgress(_ fd: Int32, data: Data, reportPr
         }
     }
 
-    var nextRange = 0..<length
+    var nextRange = data.startIndex..<data.startIndex.advanced(by: length)
     var numBytesRemaining = length
     while numBytesRemaining > 0 {
         if let localProgress, localProgress.isCancelled {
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci test

jmschonfeld (approved):

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -75,7 +75,7 @@ private func writeToFileDescriptorWithProgress(_ fd: Int32, data: Data, reportPr
         }
     }
 
-    var nextRange = 0..<length
+    var nextRange = data.startIndex..<data.startIndex.advanced(by: length)
```
itingliu, line 78: Nit: I can't comment there, but on line 90 I think it's safer to write it as `index.advanced(:)` instead of adding the raw indices? I guess it doesn't make actual difference, so it's just a style nitpick which I'm ok with leaving it asis right now

itingliu (approved):

Sources/FoundationEssentials/Data/Data+Writing.swift
parkera: I think you're right, but since Data uses int indexes it's functionally identical. If Data ever switched its index type (which it can't do anyway) then it would just end up being a compile error. So it's a fairly benign nitpick.

PR #409, opened 2024-02-09
File system representations should not be limited to PATH_MAX

jmschonfeld:
There are certain strings that historically use the file system representation that aren't strictly paths (such as environment values and arguments to a launched process). Since these can exceed the size of `PATH_MAX`, we should use a variable size for the buffer rather than a fixed `PATH_MAX` size


jmschonfeld:
@swift-ci please test

itingliu (approved):

iCharlesHu (approved):

PR #410, opened 2024-02-09
Port FileManager to swift-foundation

jmschonfeld:
This PR adds a swift-native implementation of `FileManager` that also serves as the new underlying implementation of `NSFileManager`/`FileManager` in Foundation itself on Darwin. With a swift-first approach, there are definitely some areas around this API that we could improve to make more natural in swift (such as the delegate pattern, `async`/`await` support, etc.) but we'd like to land the API as-is currently to provide compatibility with the existing `FileManager` APIs, and leave API improvements as a future direction.

There are some areas of `FileManager` that aren't fully implemented here - notably those APIs that require `URL` implementations. For now the `FileManager` operates on `String`-based path APIs and I'll circle back to add the `URL` functionality later once we can port more of that to Swift.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -64,6 +64,9 @@ let package = Package(
             "FoundationMacros",
             .product(name: "_RopeModule", package: "swift-collections"),
           ],
+          cSettings: [
+            .define("_GNU_SOURCE", .when(platforms: [.linux]))
+          ],
           swiftSettings: [
             .enableExperimentalFeature("VariadicGenerics"),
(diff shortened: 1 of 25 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

iCharlesHu (approved):
THANK YOU JEREMYYYYYY

jmschonfeld:
@swift-ci please test

PR #412, opened 2024-02-12
Update test expectation for upcoming ICU-74 change

itingliu:
ICU 74 made these changes:
1. "en_GB@hours=h12" used to return "am", but now it returns "AM".
2. The separator used to be a simple dash, but now it's dash with NNSP. This would also be an intentional change.

Update the test expectations, but disable the tests for now until ICU74 landed in all platforms.
Resolves 122765941


itingliu:
@swift-ci please test

parkera (approved):

PR #415, opened 2024-02-12
Added Uniqueness test for UUID

Roshankumar350:
We tried to add Uniqueness test for UUID.

### DIFF
--- a/Tests/FoundationEssentialsTests/UUIDTests.swift
+++ b/Tests/FoundationEssentialsTests/UUIDTests.swift
@@ -116,4 +116,24 @@ final class UUIDTests : XCTestCase {
         XCTAssertFalse(uuid2 > uuid1)
         XCTAssertTrue(uuid2 == uuid1)
     }
+    
+    func test_UUIDUniqueness() {
+        let uuidsAtTimeRangeT0 = (0...9999).reduce(into: [UUID: UUID]()) { result, _ in
+            let uuid = UUID()
+            return result[uuid] = uuid
+        }
+        
+        let uuidsAtTimeRangeT1 = (0...9999).reduce(into: [UUID: UUID]()) { result, _ in
+            let uuid = UUID()
+            return result[uuid] = uuid
+        }
+        
+        uuidsAtTimeRangeT1.values.forEach { uuid in
+            if uuid == uuidsAtTimeRangeT0[uuid] {
+                XCTFail("Uniqueness failed")
+            } else {
+                XCTAssertTrue(true, "Uniqueness passed")
+            }
+        }
+    }
 }


Roshankumar350:
@iCharlesHu: Please review.

Roshankumar350:
@swift-ci test

parkera:
Can I get some background on why this test was added?

parkera:
@swift-ci test

Roshankumar350:
Hello @parkera 

Waiting for your review.


Thanks.

parkera:
It is of course impossible to test that it is always "universally unique" - so is, then, the purpose of this test to validate that we don't have a very basic error of always generating the same UUID 10,000 times in a row? 

With the current construction of the test case, isn't it true that if we generate the same uuid 9,999 times and then 1 different it will pass? Is that a useful test case?

Roshankumar350:
Hello @parkera 
Thanks for reviewing. Please checks my understanding and review. Thanks

- It is of course impossible to test that it is always "universally unique" - so is, then, the purpose of this test to validate that we don't have a very basic error of always generating the same UUID 10,000 times in a row?

Lets say pipeline run in a day is 24 times (Assuming 1 hr per  pipeline execution) I.e Constant Time `C`

And Lifetime of this framework is `N` days (Assuming N is number of days it will be maintained and consumed by client)

so each test cases in pipeline will be executed till the life time will be  `N*C = NC` which is equivalent to `N`

Taking one random test, say `test_UUIDuuidString` 
https://github.com/apple/swift-foundation/blob/main/Tests/FoundationEssentialsTests/UUIDTests.swift#L36 

`This test cases will be executed N times and will produces same result since its value is pre-defined (Known value)`

But in UUID we don’t have pre-defined values of test. Since its universally unique and will produce unique value every time.

To validate this we have agreed by definition that its universally uniques. So we can have checks till the life time of this framework that it will pass this and if it fails once, definition is no longer valid.


- With the current construction of the test case, isn't it true that if we generate the same uuid 9,999 times and then 1 different it will pass? Is that a useful test case?

In UUID, Even if its fails once, its a failure. But I can have `break` in this loop to have overall failure and can have log too.


Thanks :)

parkera:
I'm not sold on the idea of non-deterministic testing to catch failures. We will often dismiss one-off failures as some kind of external factor rather than a failure of the code under test.

Roshankumar350:
> I'm not sold on the idea of non-deterministic testing to catch failures. We will often dismiss one-off failures as some kind of external factor rather than a failure of the code under test.

This reasoning hold true to me for general testing. UUID(In contrary) cannot be tested with the same reasoning.  
Since purpose of testing is to have expected behaviour for expected input. We are doing the same.

`non-deterministic testing to catch failures` :  I have given reason with explanation. How it become non-deterministic, Not sure.. Can you please provide any reference where i resonate your statement. 

Thanks you in advance. :)

PR #420, opened 2024-02-14
Add lock when retrieving ProcessInfo.environment

iCharlesHu:
Lock the environment when resolving.

While I was here: also added nullability annotations.


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #423, opened 2024-02-16
CalendarGregorian search functions may loop indefinitely when reaching the capped date

itingliu:
We are capping the input `Date` to the internal upper bound in various internal functions. This caused the incremental search functions to loop indefinitely because the search results are always the same since the effective input never changes.
    
Remove the capping from internal functions entirely. The capping was added to improving performance when calling into ICU. We should instead move the capping up to Calendar.swift, the call site to reach consistent behavior, but not inside the implementation.
    
Also fix miscellaneous other problems found in testing
    
Resolves 122946218

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -204,8 +204,6 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
 
         if let firstWeekday, (firstWeekday >= 1 && firstWeekday <= 7) {
             _firstWeekday = firstWeekday
-        } else if let localePrefs, let forcedWeekdayNumber = localePrefs.firstWeekday?[identifier], Locale.Weekday(forcedWeekdayNumber) != nil {
-            _firstWeekday = forcedWeekdayNumber
         }
 
         if var minimumDaysInFirstWeek {
@@ -215,8 +213,6 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
                 minimumDaysInFirstWeek = 7
             }
             _minimumDaysInFirstWeek = minimumDaysInFirstWeek
-        } else if let localePrefs, let forcedMinimumDaysInFirstWeek = localePrefs.minDaysInFirstWeek?[identifier] {
-            _minimumDaysInFirstWeek = forcedMinimumDaysInFirstWeek
         }
     }
 
@@ -248,7 +244,13 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
         }
 
         get {
-            _firstWeekday ?? 1
+            if let _firstWeekday {
+                return _firstWeekday
+            } else if let localePrefs, let forcedWeekdayNumber = localePrefs.firstWeekday?[identifier], Locale.Weekday(forcedWeekdayNumber) != nil {
+                return forcedWeekdayNumber
+            } else {
+                return 1
+            }
         }
     }
 
@@ -265,7 +267,13 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
         }
 
         get {
-            _minimumDaysInFirstWeek ?? 1
+            if let _minimumDaysInFirstWeek {
+                return _minimumDaysInFirstWeek
+            } else if let localePrefs, let forcedMinimumDaysInFirstWeek = localePrefs.minDaysInFirstWeek?[identifier] {
+                return forcedMinimumDaysInFirstWeek
+            } else {
+                return 1
+            }
         }
     }
 
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -37,7 +37,7 @@ extension Date {
     }
 
     var julianDay: Int {
-        Int(julianDate.rounded())
+        Int((julianDate + 0.5).rounded(.down))
     }
 
     init(julianDay: Int) {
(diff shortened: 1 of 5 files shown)


itingliu:
@swift-ci please test

itingliu:
Tests are failing because this change requires https://github.com/apple/swift-foundation/pull/418

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -37,7 +37,7 @@ extension Date {
     }
 
     var julianDay: Int {
-        Int(julianDate.rounded())
+        Int((julianDate + 0.5).rounded(.down))
```
jmschonfeld, line 40: Is there a behavioral difference between these two? `.rounded()` at face value seems simpler but perhaps I'm not thinking of the right number that would differ in behavior here

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -192,10 +193,20 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
...
+        // We do not store the Locale here, as Locale stores a Calendar. We only keep the values we need that affect Calendar's operation.
+        if let locale {
+            localeIdentifier = locale.identifier
+            localePrefs = locale.prefs
+        } else {
+            localeIdentifier = ""
+            localePrefs = nil
+        }
+
         self.localeIdentifier = locale?.identifier ?? ""
```
jmschonfeld, line 205: Is this line effectively duplicate of the lines above that set `localeIdentifier`?

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -1069,20 +1088,24 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
 
             case .hour:
                 guard let day = ordinality(of: .day, in: .yearForWeekOfYear, for: date) else { return nil }
-                let hour = dateComponent(.hour, from: date)
+
+                let startOfLarger = start(of: .yearForWeekOfYear, at: date)!
```
jmschonfeld, line 1092: nit: should this be a part of the `guard` statement above just to defensively protect against `nil` unwrapping here?

jmschonfeld:
It looks like the tests are failing with this change too - perhaps there might be a latent bug or two here or maybe tests need to be updated if this is new correct behavior?

itingliu:
> It looks like the tests are failing with this change too - perhaps there might be a latent bug or two here or maybe tests need to be updated if this is new correct behavior?

Tests are failing because it depends on https://github.com/apple/swift-foundation/pull/418. I've rebased. Let's see if they pass now

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
itingliu: The default rounding behavior for `rounded()` is even or away from zero. They behave differently if `julianDate` is negative.

For example, if `julianDate == -0.5` 
```
Int(julianDate.rounded()) == -1 // wrong
Int((julianDate + 0.5).rounded(.down)) == 0 // correct
```

What this is doing to match a floating point `julianDate: Double` (a point in time) to a `julianDay: Int` to represent a calendar day (a span of time). 

I find it easier to start with a positive day number, so here's an example

Given `julianDate: 2451545.0 == 2000-01-01 12:00:00` and its `julianDay == 2451545`, find some function `fn` so that every point in calendar day 2000-01-01 has `julianDay == 2451545`

So:
```
fn(2451544.5..<2451545.5) -> 2451545

// Literally
fun(2000-01-01 00:00:00 ..< 2000-01-02 00:00:00) -> 2000-01-01
```

Similarly for negative Julian days:
```
fn(-0.5..<0.5) -> 0

// Literally
fun(-4712-01-01 00:00:00 ..< -4712-01-02 00:00:00) -> -4712-01-01 
```

That is why `-0.5` needs to be rounded to `0`, not `-1`

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
itingliu: oh yes. This should be removed

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -192,10 +193,20 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
             self.gregorianStartDate = Date(timeIntervalSince1970: -12219292800) // 1582-10-15T00:00:00Z
         }
 
-        self.locale = locale
+        // We do not store the Locale here, as Locale stores a Calendar. We only keep the values we need that affect Calendar's operation.
```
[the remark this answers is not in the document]
itingliu: You're right that the behavior is now different. But this is also what we're doing currently -- i.e. even if the calendar's locale is autoupdating, the value  read from locale doesn't get autoupdated. It only autoupdates if the calendar itself is a `Calendar.autoupdatingCurrent` calendar.

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
jmschonfeld: Ah I see, makes sense - thanks for the explanation!

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -884,7 +922,6 @@ final class CalendarBridgingTests : XCTestCase {
 #endif
 
 
-#if ENABLE_CALENDAR_COMPATIBILITY_TEST // These tests take a long time to run, so disable them for now
```
jmschonfeld, line 887: Just want to make sure - enabling these tests was intentional right?

jmschonfeld (approved):
LGTM! Left one small question

Tests/FoundationInternationalizationTests/CalendarTests.swift
itingliu: Yes! Enabling them for CI to test the changes. I'm disabling them in a follow up PR

PR #424, opened 2024-02-16
[Proposal] Expression Macro and Types

jmschonfeld:
This PR introduces the Expression Macro and Types proposal


Proposals/NNNN-expression-macro.md
```diff
@@ -0,0 +1,201 @@
+# `#Expression` Macro and Type
+
+* Proposal: [SF-NNNN](NNNN-expression-macro.md)
```
itingliu, line 3:
```suggestion
* Proposal: [SF-0006](0006-expression-macro.md)
```

Proposals/NNNN-expression-macro.md
```diff
@@ -0,0 +1,201 @@
...
+* Author(s): [Jeremy Schonfeld](https://github.com/jmschonfeld)
+* Review Manager: TBD
```
itingliu, line 5:
```suggestion
* Review Manager: [Tina Liu](https://github.com/itingliu)
```

Proposals/NNNN-expression-macro.md
```diff
@@ -0,0 +1,201 @@
...
+* Bug: rdar://122026982
+* Status: **Pitch**
```
itingliu, line 7:
```suggestion
* Status: **Active review: Feb 22, 2024...Feb 29, 2024**
```

Proposals/NNNN-expression-macro.md
```diff
@@ -0,0 +1,201 @@
...
+
+- Aggregate functions such as `min()` and `max()` which are function calls and cannot be represented as a `KeyPath`
+- Conditional logic such as using if/else expressions or ternaries to vend different data based on some property
+
+### Defining `Predicate` in terms of `Expression`
+
+For most purposes, `Predicate` is effectively the same as `Expression where Output == Bool`. In an ideal world, perhaps there would be an avenue for definining `Predicate` in terms of `Expression` (i.e. via a `typealias` of sorts) such that conversion between the two types would be unnecessary and we would not need to duplicate some of the similar API surfaces between `Expression` and `Predicate`. However, since `Predicate` is already shipping as ABI in the SDK we cannot change its definition trivially and the reasons for this approach mentioned above are not significant enough to warrant jumping through hoops to perform this re-coring of `Predicate` on `Expression`. Instead, I don't expect the presence of both of these types to have significant impact on the developer experience or maintenance burden so I propose adding the two types side-by-side as discussed above.
```
itingliu, line 201:
```suggestion
For most purposes, `Predicate` is effectively the same as `Expression where Output == Bool`. In an ideal world, perhaps there would be an avenue for defining `Predicate` in terms of `Expression` (i.e. via a `typealias` of sorts) such that conversion between the two types would be unnecessary and we would not need to duplicate some of the similar API surfaces between `Expression` and `Predicate`. However, since `Predicate` is already shipping as ABI in the SDK we cannot change its definition trivially and the reasons for this approach mentioned above are not significant enough to warrant jumping through hoops to perform this re-coring of `Predicate` on `Expression`. Instead, I don't expect the presence of both of these types to have significant impact on the developer experience or maintenance burden so I propose adding the two types side-by-side as discussed above.
```

itingliu (approved):

itingliu (approved):

PR #425, opened 2024-02-17
Make timeZoneDataVersion available in FoundationInternationalization

parkera:
Now that `TimeZone` has sunk to Essentials, we can make this API available as an extension in FoundationInternationalization.


parkera:
@swift-ci test

Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
```diff
@@ -478,7 +478,6 @@ internal final class _TimeZoneICU: _TimeZoneProtocol, Sendable {
         return result
     }
 }
-#endif //canImport(FoundationICU)
```
parkera, line 481: I moved the endif down to cover the rest of the file

itingliu (approved):

PR #426, opened 2024-02-19
Exclude some paths from automount standardization

jmschonfeld:
Standardizing a path has historically unconditionally changed some paths in certain directories to support auto mount paths. However, there are certain directories on Darwin that exist when the system is installed that we know will never be an auto mount path, so we now instead exclude them from this behavior to avoid standardizing a path to something that has different semantic meaning.


jmschonfeld:
@swift-ci please test

kperryua (approved):

jmschonfeld:
@swift-ci please test

PR #427, opened 2024-02-19
FileManager's setAttributes:forItemAtPath: is unnecessarily slow

jmschonfeld:
This PR fixes two small issues:

1. Resolving symlinks in a path should use `getattrlist` for absolute paths, not relative paths (this resulted in slower-than-expected performance for absolute paths)
2. Setting cat info attributes for a file should be a no-op when no relevant attributes are specified

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -97,7 +97,7 @@ extension _FileManagerImpl {
         #endif
     }
     
-    private static let _catInfoKeys = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
+    private static let _catInfoKeys: [FileAttributeKey] = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
     static func _setCatInfoAttributes(_ attributes: [FileAttributeKey : Any], path: String) throws {
         let hasRelevantKeys = attributes.keys.contains(where: { _catInfoKeys.contains($0) })
         guard hasRelevantKeys else { return }
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -97,9 +97,9 @@ extension _FileManagerImpl {
         #endif
     }
     
+    private static let _catInfoKeys = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
     static func _setCatInfoAttributes(_ attributes: [FileAttributeKey : Any], path: String) throws {
-        let usedKeys: Set<FileAttributeKey> = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
-        let hasRelevantKeys = attributes.keys.contains(where: { usedKeys.contains($0) })
+        let hasRelevantKeys = attributes.keys.contains(where: { _catInfoKeys.contains($0) })
         guard hasRelevantKeys else { return }
         
         #if !FOUNDATION_FRAMEWORK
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -98,12 +98,13 @@ extension _FileManagerImpl {
     }
     
     static func _setCatInfoAttributes(_ attributes: [FileAttributeKey : Any], path: String) throws {
+        let usedKeys: Set<FileAttributeKey> = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
```
karwa, line 101: This could be hoisted out of the function, to make sure we don’t actually allocate and populate a new `Set` each time. Also the name `usedKeys` isn’t very clear IMO.

jmschonfeld: I chose to leave the list of attributes that we use here within this function in order to minimize the risk that this list becomes out-of-sync with the implementation of this function. This function is currently only called in one place, but if it were called in multiple I'd be worried that we'd end up skipping it incorrectly if we changed the implementation and having the list + implementation side-by-side helps.

But I can change the name of this variable to make it clearer, perhaps `relevantKeys` to match the `hasRelevantKeys` variable?

jmschonfeld: Oh unless you meant hoist it out to a static global rather than a local variable instead of moving it to the caller - that should be possible and you're right it would reduce cost of creating the array/set here. I can do that

karwa: Yes, I meant hoisting it to a type/global constant.

The compiler can't create constant sets, because the hash function starts with a random seed at runtime. If you check out the [Godbolt](https://swift.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:swift,selection:(endColumn:68,endLineNumber:2,positionColumn:68,positionLineNumber:2,selectionStartColumn:68,selectionStartLineNumber:2,startColumn:68,startLineNumber:2),source:'func+hasRelevantNumber(_+input:+%5BInt%5D)+-%3E+Bool+%7B%0A++++let+relevantNumbers+%3D+%5B1,+2,+99,+101,+102,+199,+42,+123,+12309%5D%0A++++return+input.contains+%7B+relevantNumbers.contains($0)+%7D%0A%7D%0A%0Afunc+hasRelevantNumber_set(_+input:+%5BInt%5D)+-%3E+Bool+%7B%0A++++let+relevantNumbers:+Set+%3D+%5B1,+2,+99%5D%0A++++return+input.contains+%7B+relevantNumbers.contains($0)+%7D%0A%7D'),l:'5',n:'0',o:'Swift+source+%231',t:'0')),k:44.51538015089959,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:swiftnightly,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:swift,libs:!(),options:'-O',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+swiftc+nightly+(Editor+%231)',t:'0')),k:55.484619849100405,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4), it can constant-fold the array, but it then calls `Set.init(_nonEmptyArrayLiteral: [A])` each time the enclosing function is called.

Alternatively, since this is a small number of elements, using an Array instead of a Set would also work well. In that case, the compiler doesn't even bother to create a physical array, and just inlines the values directly in to the `.contains` check (also shown in the Godbolt).

jmschonfeld: Makes sense, thanks for catching that! I've moved the array out to a static property, dropped the set initialization, and renamed the variable to be clearer about its purpose

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

parkera (approved):

PR #429, opened 2024-02-20
Enable implicit casting of file attribute values without NSNumber

jmschonfeld:
In Foundation.framework, file attributes are stored as an `NSNumber` to ensure that swift callers dynamically casting to any applicable numeric type can receive a valid value. However, we previously did not do this in swift-foundation due to the lack of `NSNumber`. `AnyHashable` has the same implicit casting behavior between numeric types, for example:

```
print(AnyHashable(1) as? Bool) // "true"
print(AnyHashable(Int(3)) as? UInt // "3"
```

In order to enable the same behavior in all uses of FileManager, this PR changes the implementation to box file attribute values in an `AnyHashable` to allow for dynamic, implicit conversions between numeric types when requested

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -58,11 +58,10 @@ extension mode_t {
 
 func _readFileAttributePrimitive<T: BinaryInteger>(_ value: Any?, as type: T.Type) -> T? {
     guard let value else { return nil }
-    #if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
It turns out (as shown by the linux failure) that the implicit casting behavior demonstrated by `AnyHashable` on Darwin is actually backed by the ObjC runtime and `NSNumber` and does not work on linux. Closing this PR since using `AnyHashable` doesn't actually work here like I'd hoped.

PR #430, opened 2024-02-20
Use `_CalendarGregorian` as the backing calendar for `struct Calendar`

itingliu:
- Fix precision issues around nanoseconds calculation.

Update test expectations.

- Lessen the chance of overflowing `func add(:)` by using `Double`

We eventually convert to Double before return anyways.

- Disable TestDateComponentsDiscreteConformance.testRandomSamples temporarily for 32-bit platforms

This test triggers code path that has wrong assumption on Calendar API's return value, and overflows when the input dates are distant from each other. Disable it for now. Will track the fix in 123262305

Resolves 123318367

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -24,13 +24,18 @@ import Glibc
 import CRT
 #endif
 
+#if FOUNDATION_FRAMEWORK
+// For feature flag
+@_implementationOnly import _ForSwiftFoundation
+#endif
+
 /**
  `Calendar` encapsulates information about systems of reckoning time in which the beginning, length, and divisions of a year are defined. It provides information about the calendar and support for calendrical computations such as determining the range of a given calendrical unit and adding units to a given absolute time.
 */
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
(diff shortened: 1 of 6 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -483,7 +488,7 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - returns: A new `DateInterval` if the starting time and duration of a component could be calculated, otherwise `nil`.
     @available(macOS 10.12, iOS 10.0, tvOS 10.0, watchOS 3.0, *)
     public func dateInterval(of component: Component, for date: Date) -> DateInterval? {
-        _calendar.dateInterval(of: component, for: date)
+        _calendar.dateInterval(of: component, for: date.capped)
```
parkera, line 491: Is this the right place for this or in the specific calendar implementation?

itingliu: I do think so. I think the limits that `_calendar` can support is its implementation detail. The inner `_calendar` may be able to support beyond the capped date, but we still want to enforce the cap for `struct Calendar`'s clients for compatibility reasons here, regardless of what `_calendar` is

PR #431, opened 2024-02-21
FileManager directory creation doesn't call to subclasses

jmschonfeld:
The file manager implementation needs to call the `fileManager` variable when calling other FileManager functions instead of using `self` in order to call down to an overriden subclass implementation if present. I accidentally used `self` instead of `fileManager` in one instance which resulted in subclasses experiencing different behavior with this swift implementation of `FileManager`. I'll followup with a future change to make this mistake harder to create by potentially renaming the internal functions to be different than the public API functions.


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #432, opened 2024-02-21
Introduce `#Expression` macro and type

jmschonfeld:
This implements the API proposed in https://github.com/apple/swift-foundation/pull/424


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Tests/FoundationMacrosTests/MacroTestUtilities.swift
```diff
@@ -138,5 +147,32 @@ func AssertMacroExpansion(macros: [String : Macro.Type], testModuleName: String
...
+        file: file,
+        line: line
+    )
+}
+
+extension String {
+    func _replacing(_ text: String, with other: String) -> Self {
+        if #available(macOS 13.0, *) {
+            // Use the stdlib API if available
+            self.replacing(text, with: other)
```
itingliu, line 172: Nit: Is this the function from [String Processing](https://github.com/apple/swift-experimental-string-processing/blob/cd0ce57a69310b419218f26f1656f206d5995c3c/Sources/_StringProcessing/Algorithms/Algorithms/Replace.swift#L89)? If so, would it make sense to match their availability with `@available(SwiftStdlib 5.7, *)` instead of `macOS 13.0`?

itingliu (approved):

Tests/FoundationMacrosTests/MacroTestUtilities.swift
jmschonfeld: Yes, but `SwiftStdlib` is a macro available only within the Swift stdlib projects (just like `FoundationPreview` is only available within our project) so unfortunately we can't use that from here

itingliu: Gotcha, thanks!

PR #434, opened 2024-02-22
[Predicate] Use a global atomic counter for VariableID

Azoy:
Instead of a global lock, we can use a global atomic counter instead for incrementing PredicateExpression.VariableID. Let's use it!


Azoy:
@swift-ci please test

Azoy:
@swift-ci please test

jmschonfeld (approved):

PR #436, opened 2024-02-23
Add framework-style builds to Benchmarks

parkera:
Adds use of the `FOUNDATION_FRAMEWORK` define to the benchmarks so we can use them in more build configurations.

I also took the opportunity to clean up some of the settings (don't use ARC counting, for example), and further segment the predicate tests until we get some attention on that compile time issue with predicates with 3 items.

### DIFF
--- a/Benchmarks/Benchmarks/Predicates/Predicates.swift
+++ b/Benchmarks/Benchmarks/Predicates/Predicates.swift
@@ -12,18 +12,62 @@
 
 import Benchmark
 import func Benchmark.blackHole
+
+#if FOUNDATION_FRAMEWORK
+import Foundation
+#else
 import FoundationEssentials
+#endif
 
-let benchmarks = {
-    Benchmark.defaultConfiguration.maxIterations = 1_000_000_000
-    Benchmark.defaultConfiguration.maxDuration = .seconds(3)
-    Benchmark.defaultConfiguration.scalingFactor = .kilo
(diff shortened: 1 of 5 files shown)


Benchmarks/Benchmarks/Predicates/Predicates.swift
```diff
@@ -12,18 +12,62 @@
...
+#else
 import FoundationEssentials
+#endif
+
+let monster = Monster(name: "Orc", level: 80, hp: 100, mana: 0, weapon: .sword(Sword(p1: 1, p2: 2, p3: 3, p4: 4, p5: 5)))
+let monster2 = Monster(name: "Orc", level: 80, hp: 100, mana: 0, weapon: .sword(Sword(p1: 1, p2: 2, p3: 3, p4: 4, p5: 5)))
+
+// These tests are disabled, as enabling them will make compilation fail due to https://github.com/apple/swift/issues/69277
+#if false
+func registerPredicateTests_disabled() {
```
itingliu, line 27: Does this `_disabled()` suffix indicate other issues besides the issue 69277? Asking because  `#if false` should suffice in disabling the test, unless there's something else also going on

itingliu (approved):

Benchmarks/Benchmarks/Predicates/Predicates.swift
parkera: No, same issue.  I was just making it a little more clear why it's inside a `if false`.

PR #437, opened 2024-02-26
Resolving symlinks can remove prefix from path

jmschonfeld:
When resolving symlinks via the quick-path using `getattrlist`, we accidentally dropped ~8 characters from the start of the path due to incorrectly offsetting by `attr_dataoffset`. Since this offset is the offset from the start of the returned buffer struct and not from the path buffer itself, we don't need to use this offset here and instead can just use the pointer directly without advancing it. Previously, the unit test would drop the leading 8 characters from the path (and we call the internal function from the unit test to avoid standardizing `/private/var` paths to `/var` which coincidentally also looks like dropping the leading 8 characters but is intentional.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -361,11 +361,10 @@ extension String {
             return nil
         }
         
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #439, opened 2024-02-26
Introduce Swift Subprocess

iCharlesHu:
This is the implementation accompanying https://github.com/apple/swift-foundation/pull/397


iCharlesHu:
@swift-ci please test

iCharlesHu:
@swift-ci please test

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -0,0 +1,232 @@
...
+        if spawnAttributeError == 0 && self.platformOptions.qualityOfService == .utility{
+            spawnAttributeError = posix_spawnattr_set_qos_class_np(&spawnAttributes, QOS_CLASS_UTILITY)
+        } else if spawnAttributeError == 0 && self.platformOptions.qualityOfService == .background {
+            spawnAttributeError = posix_spawnattr_set_qos_class_np(&spawnAttributes, QOS_CLASS_BACKGROUND)
+        }
+
+        // Setup cwd
+        var chdirError: Int32 = 0
+        if intendedWorkingDir != .currentWorkingDirectory {
+            chdirError = intendedWorkingDir.withPlatformString { workDir in
```
jmschonfeld, line 122: Currently SwiftSystem's `withPlatformString` behaves differently than Foundation's `withFileRepresentation` (namely that it doesn't perform normalization/decomposition on Darwin). Do we want to use System's implementation here, or do we want to be consistent with code like `Task`/`FileManager` and use Foundation's here? (this goes for all uses of `withPlatformString`)

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -0,0 +1,232 @@
...
+
+        // Error handling
+        if chdirError != 0 || spawnAttributeError != 0 {
+            try self.cleanupAll(input: input, output: output, error: error)
+            if spawnAttributeError != 0 {
+                throw POSIXError(.init(rawValue: result) ?? .ENODEV)
+            }
+
+            if chdirError != 0 {
+                throw CocoaError(.fileNoSuchFile, userInfo: [
```
jmschonfeld, line 135: I'm not sure `fileNoSuchFile` is appropriate here since the directory may exist but, for example, might not be permitted for the current user. Should we instead use the `CocoaError` static functions that create a cocoa error based on `chdirError`?

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -0,0 +1,232 @@
...
+        // Run additional config
+        if let spawnConfig = self.platformOptions.additionalSpawnAttributeConfigurator {
+            try spawnConfig(&spawnAttributes)
+        }
+        if let fileAttributeConfig = self.platformOptions.additionalFileAttributeConfigurator {
+            try fileAttributeConfig(&fileActions)
+        }
+        // Spawn
+        var pid: pid_t = 0
+        let spawnError: CInt = executablePath.withCString { exePath in
```
jmschonfeld, line 149: I suspect that this `withCString` should really be `withFileSystemRepresentation` because it is a path

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Darwin.swift
```diff
@@ -0,0 +1,232 @@
...
+            executionInput: input,
+            executionOutput: output,
+            executionError: error
+        )
+    }
+}
+
+// Special keys used in Error's user dictionary
+extension String {
+    static let debugDescriptionErrorKey = "NSDebugDescription"
```
jmschonfeld, line 177: We have a pattern of defining keys like this at the bottom of `CocoaError.swift` instead of defining them as extensions on `String`, should we follow that same pattern for this one?

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -0,0 +1,135 @@
...
+
+        var workingDirectory: String?
+        if intendedWorkingDir != FilePath.currentWorkingDirectory {
+            // Only pass in working directory if it's different
+            workingDirectory = intendedWorkingDir.string
+        }
+        // Spawn
+        var pid: pid_t = 0
+        let spawnError: CInt = executablePath.withCString { exePath in
+            return workingDirectory.withOptionalCString { workingDir in
```
jmschonfeld, line 55: I think `workingDirectory` and `executablePath` should be using `withFileSystemRepresentation` here instead of `withCString`

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Linux.swift
```diff
@@ -0,0 +1,135 @@
...
+                createSession: false,
+                createProcessGroup: false
+            )
+        }
+    }
+}
+
+// Special keys used in Error's user dictionary
+extension String {
+    static let debugDescriptionErrorKey = "DebugDescription"
```
jmschonfeld, line 131: I think this can be moved out of platform-specific logic since it's the same cross-platform, and similar comment about the one I left on the Darwin version

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+            let rawByteValue: UnsafeMutablePointer<CChar> = valueContainer.createRawBytes()
+            defer {
+                rawByteKey.deallocate()
+                rawByteValue.deallocate()
+            }
+            /// length = `key` + `=` + `value` + `\null`
+            let totalLength = keyContainer.count + 1 + valueContainer.count + 1
+            let fullString: UnsafeMutablePointer<CChar> = .allocate(capacity: totalLength)
+            #if canImport(Darwin)
+            _ = snprintf(ptr: fullString, totalLength, "%s=%s", rawByteKey, rawByteValue)
```
jmschonfeld, line 98: I wonder if there's any way to do this in a safer manner that doesn't rely on string formatting (which has a lot of pitfalls / sneaky issues in Swift)

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+        case .inherit(let updates):
+            var current = ProcessInfo.processInfo.environment
+            for (keyContainer, valueContainer) in updates {
+                if let stringKey = keyContainer.stringValue {
+                    // Remove the value from current to override it
+                    current.removeValue(forKey: stringKey)
+                }
+                // Fast path
+                if case .string(let stringKey) = keyContainer,
+                   case .string(let stringValue) = valueContainer {
```
jmschonfeld, line 116: Previously, `Task` used the `fileSystemRepresentation` of all environment keys/values when spawning a process. Do we want to maintain that behavior with `Subprocess` or do we intentionally want to avoid doing that here?

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// MARK: Args Creation
+extension Subprocess.Arguments {
+    // This method follows the standard "create" rule: `args` needs to be
+    // manually deallocated
+    internal func createArgs(withExecutablePath executablePath: String) -> [UnsafeMutablePointer<CChar>?] {
+        var argv: [UnsafeMutablePointer<CChar>?] = self.storage.map { $0.createRawBytes() }
+        // argv[0] = executable path
+        if let override = self.executablePathOverride {
```
jmschonfeld, line 153: Same goes here - `Task` previously used the `fileSystemRepresentation` for all string arguments rather than just using the as-is cString value. Do we want to do the same here or diverge for `Subprocess`?

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+            executablePath: executablePath,
+            env: env, argv: argv,
+            intendedWorkingDir: intendedWorkingDir,
+            uidPtr: uidPtr, gidPtr: gidPtr,
+            supplementaryGroups: supplementaryGroups
+        )
+    }
+
+    internal static func pathAccessible(_ path: String, mode: Int32) -> Bool {
+        return path.withCString {
```
jmschonfeld, line 263: Should we use `FileManager` to perform this here which will correctly use the `fileSystemRepresentation` instead of `withCString`?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+        >(minimumCapacity: self.count)
+        for (key, value) in self {
+            result[.rawBytes(key.toArray())] = .rawBytes(value.toArray())
+        }
+        return result
+    }
+}
+
+fileprivate extension Data {
+    func toArray<T>() -> [T] {
```
jmschonfeld, line 467: Could you elaborate a bit more on why this is necessary? I would have naively assumed that anything that we need to do with the `Data` shouldn't require copying the data out to an `Array`

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+}
+
+// MARK: - Internal
+extension Subprocess {
+    internal enum StringOrRawBytes: Sendable, Hashable {
+        case string(String)
+        case rawBytes([CChar])
+
+        // Return value needs to be deallocated manually by callee
+        func createRawBytes() -> UnsafeMutablePointer<CChar> {
```
jmschonfeld, line 528: There are a decent number of functions here that return pointers allocated that the caller needs to manually deallocate. In Swift I think we normally prefer the closure-taking-function approach so that we can allocate, call the closure with the pointer, then deallocate and not rely on the caller to deallocate. This feels like a great use for that with something like a `withUnsafeBytes` function - do you think that'd work well here?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                return string
+            case .rawBytes(let rawBytes):
+                return String(validatingUTF8: rawBytes)
+            }
+        }
+
+        var count: Int {
+            switch self {
+            case .string(let string):
+                return string.count
```
jmschonfeld, line 549: Are you sure you want the count of the `Character`s here instead of the UTF-8 count or otherwise?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+            case .rawBytes(let rawBytes):
+                return strnlen(rawBytes, Int.max)
+            }
+        }
+
+        func hash(into hasher: inout Hasher) {
```
jmschonfeld, line 555: The hash/equality for this type will normalize all of the strings so that strings that contain different underlying bytes may compare equal. Is that intentional here, or should we be comparing/hashing the UTF-8 contents directly to perform a literal comparison like we'd do for the bytes. I'm not sure I agree with the "If Raw bytes is valid UTF8, hash it as so" comment below

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                } else {
+                    hasher.combine(bytes)
+                }
+            }
+        }
+    }
+}
+
+extension FilePath {
+    static var currentWorkingDirectory: Self {
```
jmschonfeld, line 572: Should we use `FileManager` for this instead of calling `getcwd` directly?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+        let path = getcwd(nil, 0)!
+        defer { free(path) }
+        return .init(String(cString: path))
+    }
+}
+
+extension Optional where Wrapped : Collection {
```
jmschonfeld, line 579: Since this extension only works for an `Array`, can we enforce that statically in the type signature?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                return try body(ptr)
+            }
+        case .none:
+            return try body(nil)
+        }
+    }
+}
+
+extension Optional where Wrapped == String {
+    func withOptionalCString<R>(_ body: ((UnsafePointer<Int8>)?) throws -> R) rethrows -> R {
```
jmschonfeld, line 596: Is there a meaningful difference between `optString.withOptionalCString` and `optString?.withCString` at the call sites?

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+        case .some(let wrapped):
+            return try wrapped.withCString {
+                return try body($0)
+            }
+        }
+    }
+}
+
+// MARK: - Stubs for the one from Foundation
+public enum QualityOfService: Int, Sendable {
```
jmschonfeld, line 609: Should this be inside of a `#if !FOUNDATION_FRAMEWORK` block?

Sources/FoundationEssentials/Subprocess/Subprocess+IO.swift
```diff
@@ -0,0 +1,267 @@
...
+                try writeFd.close()
+            }
+        }
+    }
+}
+
+// MARK: - Private Helpers
+extension FileDescriptor {
+    internal func read(upToLength maxLength: Int) throws -> [UInt8] {
+        let buffer: UnsafeMutableBufferPointer<UInt8> = .allocate(capacity: maxLength)
```
jmschonfeld, line 262: This buffer is never deallocated - is there a way to do this a bit safer without allocating the buffer ourself? Also, should this return a `Data` instead of a `[UInt8]`

Sources/FoundationEssentials/Subprocess/Subprocess.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+}
+
+
+extension Subprocess.Result: Equatable where T : Equatable {}
+
+extension Subprocess.Result: Hashable where T : Hashable {}
+
+extension POSIXError : Swift.Error {}
```
jmschonfeld, line 163: Is this not already defined elsewhere in the package?

Tests/FoundationEssentialsTests/SubprocessTests.swift
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    func testComplex() async throws {
+        struct Address: Codable {
+            let ip: String
+        }
+
+        let result = try await Subprocess.run(
+            executing: .named("curl"),
+            arguments: ["http://ip.jsontest.com/"]
```
jmschonfeld, line 37: Is it ok for our test to rely on this external address, or should we use something like `apple.com`?

Sources/FoundationEssentials/Subprocess/Subprocess+IO.swift
```diff
@@ -0,0 +1,267 @@
...
+        let readCount = try self.read(into: .init(buffer))
```
FranzBusch, line 263: This is potentially blocking and the method should be async or at least all callers must be async and handle it correctly.

Sources/FoundationEssentials/Subprocess/Subprocess+AsyncBytes.swift
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        private let fileDescriptor: FileDescriptor
+
+        init(fileDescriptor: FileDescriptor) {
+            self.fileDescriptor = fileDescriptor
+        }
+
+        public func makeAsyncIterator() -> Iterator {
+            return Iterator(fileDescriptor: self.fileDescriptor)
```
FranzBusch, line 94: We must enforce here that only a single iterator is created ever and `fatalError` on subsequent iterator creation otherwise there might be concurrent reads on the FD.

Sources/FoundationEssentials/Subprocess/Subprocess+AsyncBytes.swift
```diff
@@ -0,0 +1,110 @@
...
+                if currentPosition < buffer.count {
+                    let value = buffer[currentPosition]
+                    self.currentPosition += 1
+                    return value
+                }
+                return try await self.reloadBufferAndNext()
+            }
+
+            private func read(from fileDescriptor: FileDescriptor, maxLength: Int) async throws -> [UInt8] {
+                return try await withCheckedThrowingContinuation { continuation in
```
FranzBusch, line 68: This has to handle task cancellation otherwise a process might get wedged while waiting for a read which will never happen.

Sources/FoundationEssentials/Subprocess/Subprocess+AsyncBytes.swift
```diff
@@ -0,0 +1,110 @@
...
+                    DispatchIO.read(
```
FranzBusch, line 69: I personally don't think we should use the `Dispatch` APIs here for our async reading/writing of pipes. We can model this with task executors and can therefore keep all the code in concurrency land and properly handle cancellation.

Sources/FoundationEssentials/Subprocess/Subprocess+API.swift
```diff
@@ -0,0 +1,225 @@
...
+
+import SystemPackage
+
+extension Subprocess {
+    public static func run(
+        executing executable: Executable,
+        arguments: Arguments = [],
+        environment: Environment = .inherit,
+        workingDirectory: FilePath? = nil,
+        platformOptions: PlatformOptions = .default,
```
FranzBusch, line 20: Do we really need all of those methods that don't take a configuration? To me it seems simpler and cleaner if we just had methods that take a `configuration` and then overload based on input/output/error

Sources/FoundationEssentials/Subprocess/Subprocess+API.swift
```diff
@@ -0,0 +1,225 @@
...
+            platformOptions: platformOptions
+        )
+        .run(output: output, error: error, body)
+    }
+}
+
+// MARK: - Configuration Based
+extension Subprocess {
+    public static func run<R>(
+        withConfiguration configuration: Configuration,
```
FranzBusch, line 217: Naming NIT: 
```suggestion
        configuration: Configuration,
```

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+@_implementationOnly import _CShims
+#else
+package import _CShims
+#endif
+
+
+// MARK: - Signals
+extension Subprocess {
+    public struct Signal : Hashable, Sendable {
+        public let rawValue: Int32
```
FranzBusch, line 33: Do we really want to enforce `Int32` here or just make that an implementation detail? For most of our public APIs we prefer `Int` over specific `IntX` types.

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+            }
+            if let errorError = errorError {
+                throw errorError
+            }
+        }
+
+        internal func run<R>(
+            output: RedirectedOutputMethod,
+            error: RedirectedOutputMethod,
+            _ body: @Sendable @escaping (Subprocess, StandardInputWriter) async throws -> R
```
FranzBusch, line 184: The `body` really shouldn't require to be `@Sendable` nor `@escaping`. This will limit the way how this is used and make it very hard to use from within an actor.

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                output: executionOutput,
+                error: executionError)
+            // After spawn, cleanup child side fds
+            try self.cleanup(
+                process: process,
+                childSide: true,
+                parentSide: false,
+                attemptToTerminateSubProcess: false
+            )
+            return try await withTaskCancellationHandler {
```
FranzBusch, line 201: IMO this is not the right way to do task cancellation here. I personally think we should check for cancellation only when we do a `read` or `write` syscall. This makes sure that user code in the `body` still continues to execute and only when they try to make a sys call we check for it.

Sources/FoundationEssentials/AsyncLineSequence.swift
```diff
@@ -0,0 +1,168 @@
...
+        @_specialize(where Base == Subprocess.AsyncBytes)
+        public mutating func next() async rethrows -> String? {
+            /*
+             0D 0A: CR-LF
+             0A | 0B | 0C | 0D: LF, VT, FF, CR
+             E2 80 A8:  U+2028 (LINE SEPARATOR)
+             E2 80 A9:  U+2029 (PARAGRAPH SEPARATOR)
+             */
+            let _CR: UInt8 = 0x0D
+            let _LF: UInt8 = 0x0A
```
weissi, line 43: I'd write `let _LF = UInt8(ascii: "\n")` or even remove these constants all together, there's no perf benefit of doing this.

Sources/FoundationEssentials/Subprocess/Platforms/Subprocess+Unix.swift
```diff
@@ -0,0 +1,286 @@
...
+            "/sbin",
+            "/usr/local/bin"
+        ])
+    }
+
+    internal func resolveExecutablePath(withPathValue pathValue: String?) -> String? {
+        switch self.storage {
+        case .executable(let executableName):
+            // If the executableName in is already a full path, return it directly
+            if Subprocess.Configuration.pathAccessible(executableName, mode: X_OK) {
```
weissi, line 179: This is racy and can block

Sources/_CShims/process_shims.c
```diff
@@ -0,0 +1,298 @@
...
+
+        rc = posix_spawnattr_setflags(
+            (posix_spawnattr_t *)spawn_attrs, flags | POSIX_SPAWN_SETEXEC);
+        if (rc != 0) {
+            return rc;
+        }
+    }
+
+    // Spawn
+    return posix_spawn(pid, exec_path, file_actions, spawn_attrs, args, env);
```
weissi, line 118: if this fails with `POSIX_SPAWN_SETEXEC`, how's the error communicated to the parent?

Sources/_CShims/process_shims.c
```diff
@@ -0,0 +1,298 @@
...
+        if (rc != 0) { return rc; }
+    }
+    if (file_descriptors[4] != 0) {
+        rc = close(file_descriptors[5]);
+        if (rc != 0) { return rc; }
+    }
+
+    // Finally, exec
+    execve(exec_path, args, env);
+    // If we got here, something went wrong
```
weissi, line 295: Indeed, how's the error code communicated back to the parent? I'd expect a pipe between parent and child here that has O_CLOEXEC. So if the parent reads something there, then that's the child's exit code. If the parent reads EOF, then everything's okay.

Here's code I've written in a previous life taking care of this: https://github.com/BromiumInc/BromiumCoreUtils/blob/2ec1c52a09831893618b9620383015e4362f7aa5/BromiumCoreUtils/BRUTask.m#L338-L341

Sources/_CShims/process_shims.c
```diff
@@ -0,0 +1,298 @@
...
+        if (rc != 0) { return rc; }
+    }
+    if (file_descriptors[3] != 0) {
+        rc = close(file_descriptors[3]);
+        if (rc != 0) { return rc; }
+    }
+    if (file_descriptors[4] != 0) {
+        rc = close(file_descriptors[5]);
+        if (rc != 0) { return rc; }
+    }
```
weissi, line 291: this fails to
- close the other file descriptors
- reset the signal masks

example: https://github.com/BromiumInc/BromiumCoreUtils/blob/2ec1c52a09831893618b9620383015e4362f7aa5/BromiumCoreUtils/BRUTask.m#L320-L334

Tests/FoundationEssentialsTests/SubprocessTests.swift
```diff
@@ -0,0 +1,69 @@
...
+#endif // FOUNDATION_FRAMEWORK
+
+#if canImport(TestSupport)
+import TestSupport
+#endif
+
+final class SubprocessTests: XCTestCase {
+    func testSimple() async throws {
+        let ls = try await Subprocess.run(executing: .named("ls"), output: .collect, error: .discard)
+        let result = String(data: ls.standardOutput!, encoding: .utf8)!
```
weissi, line 25:
```suggestion
        let result = String(decoding: ls.standardOutput!, as: UTF8.self)
```

Sources/_CShims/process_shims.c
MaxDesiatov, line 118: While `posix_spawn` is fine in some circumstances, it is not always ideal. In some cases one needs to replace the current process with `execv` instead of launching a new one, for example the `swift run` command in SwiftPM works exactly this way. Is there anything in the current `Subprocess` API that would allow that? If not, is there anything in the API that would prevent us from adding it in the future?

Sources/FoundationEssentials/Subprocess/Subprocess.swift
```diff
@@ -0,0 +1,163 @@
...
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+import SystemPackage
+
+public struct Subprocess: Sendable {
```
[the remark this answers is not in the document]
MaxDesiatov: From this perspective, adding `Sendable` in the future seems possible, while taking it away would not be an option at all? In the case of uncertainty like this maybe it's better to be conservative and not add it at least until we're certain that the compiler features are sufficient to make this work well in the future?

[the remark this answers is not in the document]
jmschonfeld: I don't believe `~Copyable` and `~Escapable` are mutually exclusive with `Sendable`, they represent fundamentally different concepts so even if `Subprocess` were designed to be `~Copyable`/`~Escapable` then it still can be `Sendable` in my understanding. Since mutable `~Copyable` types cannot have multiple references, I think I'd find it rare that a `~Copyable` type weren't `Sendable`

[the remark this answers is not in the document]
FranzBusch: You are right that those three things are not exclusive and we should think here about what we expect people to do. Do we expect them to share a Subprocess over multiple isolation regions? With the new region isolation proposals a lot has changed here.

I personally am not yet convinced that this should be Sendable since we have the standard output and error types which must not be called from multiple concurrent contexts in parallel.

Sources/FoundationEssentials/Subprocess/Subprocess+API.swift
iCharlesHu: It's the other way around. These are intended to be the "main" APIs. I only exposed the `configuration` based one for the use case where someone might want to "store" a configuration somewhere and use it later. Otherwise they should always use theses APIs.

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                return try await withThrowingTaskGroup(of: RunState<R>.self) { group in
+                    group.addTask {
+                        let status = monitorProcessTermination(
```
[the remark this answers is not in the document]
iCharlesHu: Could you elaborate on this? Are you saying all blocking methods need to be in a separate executor? That would make Swift's concurrency system really hard to use...

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
```diff
@@ -0,0 +1,615 @@
...
+                            forProcessWithIdentifier: process.processIdentifier)
+                        return .monitorChildProcess(status)
+                    }
+                    group.addTask {
+                        do {
+                            let result = try await body(process, .init(fileDescriptor: writeFd))
```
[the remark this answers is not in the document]
iCharlesHu: > If we cancel their body closure 
We are not currently canceling the body closure when the subprocess finishes. They run on separate tasks and `run` waits for both to finish.

Sources/FoundationEssentials/AsyncLineSequence.swift
```diff
@@ -0,0 +1,168 @@
...
+// Copyright (c) 2019 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+public struct AsyncLineSequence<Base: AsyncSequence>: AsyncSequence where Base.Element == UInt8 {
```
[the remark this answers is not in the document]
iCharlesHu: This is already in Foundation today. I'm simply moving it to SwiftFoundation.

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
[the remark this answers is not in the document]
FranzBusch: You are right this is very subtle since the monitor task is not throwing. My other point still stands that we should avoid running the users `body` closure in a child task here since it forces the closure to become `@Sendable` and `@escaping` which makes it hard to use in some context e.g. inside an actor.

We should run the closure in the calling task and the monitoring of the child process can happen in a child task but in the end it must surface through the `read/write` operations of the `Subprocess`. From what I can tell this is already happening and you can just move the call to `body` to the body of the `withTaskGroup` here.

Sources/FoundationEssentials/Subprocess/Subprocess+Configuration.swift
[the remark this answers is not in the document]
FranzBusch: I know we discussed an alternative solution for this offline and we should use kqueue,epoll,io_uring to monitor to process instead. However, to answer your general question for anyone that's reading along: Yes you can never make a blocking sys call in neither a sync nor async method really since those might be called from one of the Concurrency threads which will get blocked. Any blocking call must run on its own thread and the canonical spelling starting is going to be `withTaskExecutor` once that feature has landed in a released Swift version.

Sources/FoundationEssentials/Subprocess/Subprocess+AsyncBytes.swift
FranzBusch: This NIO PR https://github.com/apple/swift-nio/pull/2648 shows how a non blocking read/write can be done via kqueue. We just need to extend that to epoll and it will work on Linux as well. While I agree this is an implementation detail we **should not** release this without handling cancellation; otherwise, this becomes a big foot gun when using the new APIs.

Sources/_CShims/process_shims.c
weissi: @MaxDesiatov  I think `POSIX_SPAWN_SETEXEC` makes `posix_spawn` essentially `execve` on steroids.

Sources/FoundationEssentials/Subprocess/Subprocess+IO.swift
```diff
@@ -0,0 +1,397 @@
...
+        internal let method: Storage
+
+        internal init(method: Storage) {
+            self.method = method
+        }
+
+        internal func createExecutionInput() throws -> ExecutionInput {
+            switch self.method {
+            case .noInput:
+                let devnull: FileDescriptor = try .open("/dev/null", .readOnly)
```
jakepetroules, line 32: Should this be using NUL on Windows?

iCharlesHu:
Implement in swift-subprocess

PR #440, opened 2024-02-27
[Gregorian Calendar] Guard against potential infinite loop caused by iterative search

itingliu:
Guard against infinite loop by validating each intermittent result. This is similar to what we're doing in Calendar enumeration functions.

Resolves 123660148


itingliu:
@swift-ci please test

parkera (approved):
Looks good to me.

PR #442, opened 2024-02-28
[Gregorian Calendar] Guard against julian dates that are too far away to be representable by `Int`

itingliu:
We triggered a Swift runtime error when converting a `Double` to an `Int` because the value exceeds the value representable by an `Int`. This patch works-around this issue with the following:
- Return a capped value if the date exceeds the bound before conversion.
- Ensure compatibility among various `_CalendarProtocol` observers by capping the input date upfront.

Resolves 123664232


itingliu:
@swift-ci please test

parkera (approved):
We'll have to figure out how to keep from passing an uncapped date through in other API.

itingliu:
> We'll have to figure out how to keep from passing an uncapped date through in other API.

I don't have a great answer either. When we first introduced the capped date idea, we had to audit every API. I suspect that's just what we'll have to do in the future. 

At least we have guards in `_CalendarProtocol`'s internal implementation too, like the fix for `julianDay: Int` in this PR and ICU's internal cap. That means uncapped date shouldn't cause detrimental harm but only gives you wrong results. Still bad, but I can't think of any ways to handle that unless we change the API to `throw`

PR #443, opened 2024-02-29
foundation_sysdir_start_search_path_enumeration shouldn't crash with extra option bits set

jmschonfeld:
Some clients of the search paths `FileManager` API may provide values greater than `UInt32.max` that still fit inside the API defined requirement of `UInt`. These values don't hold meaning at runtime so we should simply truncate the high order bits instead of crashing here.


parkera (approved):
Looks good to me.

jmschonfeld:
@swift-ci please test

jmschonfeld:
The test failures are unrelated to this change, will be investigating them separately

PR #446, opened 2024-02-29
Update 9998 availability to 10000 availability

jmschonfeld:
We use `9998` as a version number to signal a "future" version because it does not have the same compiler magic as being available everywhere like `9999`. However, there are APIs in the standard library we use only on a future OS that are marked as available in the magic `9999` version. This PR switches uses of `9998` to `10000` to still avoid the compiler magic around `9999`, but ensure the version is greater than that of the standard library future APIs we use, such as `_literalPattern`. This should resolve the compilation errors we're seeing in CI


jmschonfeld:
@swift-ci please test

itingliu (approved):
LGTM as long as it passes CI

parkera (approved):

PR #447, opened 2024-03-01
Unexpected behavior when getting a `Date` after setting `day` on a date components

itingliu:
This is due to ICU's handling of ambiguous/conflicting components. If both "day of year" and "day of month" fields are set via `ucal_set`, the newer one takes priority. Therefore it breaks existing behavior when we set "day of year" after having set "day of month".

Fix this by moving "day of year" before "day of month" to maintain compatibility.

Resolves 123703256

### DIFF
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1188,14 +1188,16 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
             if let value = components.weekdayOrdinal { ucal_set(ucalendar, UCAL_DAY_OF_WEEK_IN_MONTH, Int32(truncatingIfNeeded: value)) }
             // DateComponents month field is +1 from ICU
             if let value = components.month { ucal_set(ucalendar, UCAL_MONTH, Int32(truncatingIfNeeded: value - 1)) }
+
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1188,14 +1188,16 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
             if let value = components.weekdayOrdinal { ucal_set(ucalendar, UCAL_DAY_OF_WEEK_IN_MONTH, Int32(truncatingIfNeeded: value)) }
             // DateComponents month field is +1 from ICU
             if let value = components.month { ucal_set(ucalendar, UCAL_MONTH, Int32(truncatingIfNeeded: value - 1)) }
+
+            // The later the value is set via `ucal_set` the higher priority it takes when ICU resolves ambiguous components. For compatibility, always set `day of year` before `day (of month)`
```
itingliu, line 1192: Alternatively we can also only set "day of year" if "day of month" isn't set, but I'm not sure how that'd work with other fields. ICU has its unique way of resolving combinations of fields, so I'd like to still set it always and let ICU handle it

parkera (approved):
This approach sounds fine to me.

itingliu:
This is handled in https://github.com/apple/swift-foundation/pull/455

parkera:
Oops, sorry about duplicating this.

PR #450, opened 2024-03-01
Data.write fails with EINVAL when writing 2GiB or more

parkera:
On Darwin, at least, we need to limit the size of each chunk to 2GB. Other platforms may have higher requirements, which this is compatible with. If any have lower, we will need to account for that in the future.

I also fixed a bit of code where we are passing a `0` to `POSIXError` instead of errno.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #451, opened 2024-03-04
Prevent data writing integer underflow on 32 bit platforms

jmschonfeld:
Previously on 32 bit platforms, taking `1 << 31` and subtracting `1` here would cause an integer underflow (which causes an abort) as `1 << 31 == Int.min` on 32 bit platforms. Instead, we now use `Int32.max` here (which is equivalent to `(1 << 31) - 1` but without possibly causing issues due to different integer sizes)


jmschonfeld:
@swift-ci please test

parkera (approved):
Thanks!

PR #452, opened 2024-03-05
FileManager should standardize home directory paths

jmschonfeld:
The previous ObjC implementation of `FileManager` standardized the home directory path, but the new Swift implementation did not. This could lead to home directory paths (vended by APIs such as `urls(for:in:)`) beginning with prefixes such as `/private/var` instead of `/var`. Restore the old behavior of standardizing home directory paths for behavioral compatibility.


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #453, opened 2024-03-05
Round towards zero, matching behavior of previous ICU implementation

parkera:
Round towards zero, matching previous ICU implementation and also avoiding a situation where we round from 999 to 1000 for milliseconds.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #454, opened 2024-03-05
Fix lexical sorting for SortComparator

parkera:
SortComparator’s StandardComparator always uses `Locale.current` instead of checking the `isLocalized` property.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #455, opened 2024-03-05
Reorder day of year and day of month for Calendar calculations

parkera:
This bug is due to ICU's handling of ambiguous/conflicting components. If both "day of year" and "day of month" fields are set via `ucal_set`, the newer one takes priority. Therefore it breaks existing behavior when we set "day of year" after having set "day of month".

Fix this by moving "day of year" before "day of month" to maintain compatibility.

123703256


parkera:
@swift-ci test

jmschonfeld (approved):

PR #460, opened 2024-03-06
ProcessInfo Bincompat: always use full executable path for arg0

iCharlesHu:
Swift's `CommandLine.arguments` doesn't always report the full executable path, most noticeably when the app is launched via `PATH` instead of full path name (`ls` vs `/bin/ls`). This patch adds a fallback to retrieve the full executable path for arg0

resolves rdar://122930912


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #461, opened 2024-03-06
ProcessInfo.environment should create String values outside of lock

iCharlesHu:
`_platform_shims_lock_environ` locks the global environment, which is very disruptive. We should limit the amount of call outs as much as possible. This patch copies the environment values before attempting to create strings out of it. 

resolves: rdar://123704736


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #462, opened 2024-03-06
Calendar should return the first day of week of its locale

itingliu:
"First day of week" and "minimum days in first week" are locale specific. Move the logic of fetching the locale's preferred value into `LocaleProtocol`, so that calendar can return the value from its enclosed locale.

Resolves 123630636


itingliu:
@swift-ci please test

parkera (approved):

PR #463, opened 2024-03-06
Standardize on using internal import

jmschonfeld:
We've previously had to use `@_implementationOnly` imports in Foundation itself for internal imports, while we used `package` imports in the package to achieve the same goal in swift-foundation of hiding imports from clients. The `internal import` feature has been completed for some time now and provides us the same fundamental behaviors in both the resilient Foundation module as well as the non-resilient swift-foundation modules. This PR changes all uses of `@_implementationOnly`/`package` imports to `internal import` to standardize on one spelling of these imports. This also resolves a slew of build warnings produced by `package import`s that were not used in any `package` declarations.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
@@ -11,9 +11,9 @@
 //===----------------------------------------------------------------------===//
 
 #if FOUNDATION_FRAMEWORK
-@_implementationOnly @_spi(Unstable) import CollectionsInternal
+@_spi(Unstable) internal import CollectionsInternal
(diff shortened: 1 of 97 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
On linux this PR reduces the number of warnings produced while building from 503 build warnings down to 79 build warnings, a huge improvement 🙂

jmschonfeld:
The test failure on macOS is unrelated (@itingliu is looking into it) but this change passed tests on linux and still built successfully on macOS

parkera (approved):

parkera:
Awesome!

PR #464, opened 2024-03-06
Implement Calendar.RecurrenceRule.recurrences(of: )

hristost:
This commit implements the `Calendar.RecurrenceRule` type first pitched in https://github.com/apple/swift-foundation/pull/422.

This type models a subset of RRULE as specified in RFC-5545, section 3.3.10. One
notable difference is that it doesn't support a frequency of "secondly", as that
was not part of the original proposal. It also implements RFC-7529 Non-Gregorian
Recurrence Rules and works with any instance of `Calendar`.

Recurrences are calculated according to our interpretation of the RFCs. As such,
the resulting dates may differ in certain edge cases when compared to other open
source implementations.

Resolves rdar://120559017&123337748

### DIFF
--- a/Benchmarks/Benchmarks/Calendar/Calendar.swift
+++ b/Benchmarks/Benchmarks/Calendar/Calendar.swift
@@ -26,11 +26,22 @@ let benchmarks = {
     Benchmark.defaultConfiguration.scalingFactor = .kilo
     Benchmark.defaultConfiguration.metrics = [.cpuTotal, .wallClock, .mallocCountTotal, .throughput]
     
-    let thanksgivingComponents = DateComponents(month: 11, weekday: 5, weekOfMonth: 4)
+    let thanksgivingComponents = DateComponents(month: 11, weekday: 5, weekdayOrdinal: 4)
     let cal = Calendar(identifier: .gregorian)
     let currentCalendar = Calendar.current
     let thanksgivingStart = Date(timeIntervalSinceReferenceDate: 496359355.795410) //2016-09-23T14:35:55-0700
     
+    Benchmark("nextThousandThursdaysInTheFourthWeekOfNovember") { benchmark in
+        // This benchmark used to be nextThousandThanksgivings, but the name was deceiving since it does not compute the next thousand thanksgivings
+        let components = DateComponents(month: 11, weekday: 5, weekOfMonth: 4)
+        var count = 1000
+        cal.enumerateDates(startingAfter: thanksgivingStart, matching: components, matchingPolicy: .nextTime) { result, exactMatch, stop in
+            count -= 1
+            if count == 0 {
+                stop = true
+            }
+        }
+    }
     Benchmark("nextThousandThanksgivings") { benchmark in
         var count = 1000
         cal.enumerateDates(startingAfter: thanksgivingStart, matching: thanksgivingComponents, matchingPolicy: .nextTime) { result, exactMatch, stop in
@@ -40,7 +51,26 @@ let benchmarks = {
             }
         }
     }
-
+    Benchmark("nextThousandThanksgivingsSequence") { benchmark in
+        var count = 1000
+        for _ in cal.dates(byMatching: thanksgivingComponents, startingAt: thanksgivingStart, matchingPolicy: .nextTime) {
+            count -= 1
+            if count == 0 {
+                break
+            }
+        }
+    }
+    Benchmark("nextThousandThanksgivingsUsingRecurrenceRule") { benchmark in
+        var rule = Calendar.RecurrenceRule(calendar: cal, frequency: .yearly, end: .afterOccurrences(1000))
+        rule.months = [11]
+        rule.weekdays = [.nth(4, .thursday)]
+        rule.matchingPolicy = .nextTime
+        var count = 0
+        for _ in rule.recurrences(of: thanksgivingStart) {
+            count += 1
+        }
+        assert(count == 1000)
+    }
     Benchmark("CurrentDateComponentsFromThanksgivings") { benchmark in
         var count = 1000
         currentCalendar.enumerateDates(startingAfter: thanksgivingStart, matching: thanksgivingComponents, matchingPolicy: .nextTime) { result, exactMatch, stop in
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -1,0 +1,820 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// Recurrence enumeration
+//
+// This file implements enumerating occurences according to a recurrence rule as
+// specified in RFC5545 and RFC7529
+
+extension Calendar.RecurrenceRule.Frequency {
+    /// The calendar component corresponding to the frequency interval
+    var component: Calendar.Component {
+        switch self {
+        case .minutely: .minute
+        case .hourly:   .hour
+        case .daily:    .day
+        case .weekly:   .weekOfMonth
+        case .monthly:  .month
+        case .yearly:   .year
+        }
+    }
+}
+
+/// The action of a component of the recurrence rule.
+///
+/// Each field in the recurrence rule has a different effect on the output which
+/// is also different for different frequencies. The effect is documented in the
+/// RFC, and copied below for convenience:
+///
+/// | Component\Freq |.minutely|.hourly |.daily  |.weekly |.monthly|.yearly |
+/// |----------------|---------|--------|--------|--------|--------|--------|
+/// |months          |.limit   |.limit  |.limit  |.limit  |.limit  |.expand |
+/// |weeks           |nil      |nil     |nil     |nil     |nil     |.expand |
+/// |daysOfTheYear   |.limit   |.limit  |nil     |nil     |nil     |.expand |
+/// |daysOfTheMonth  |.limit   |.limit  |.limit  |nil     |.expand |.expand |
+/// |weekdays        |.limit   |.limit  |.limit  |.expand |note 1  |note 2  |
+/// |hours           |.limit   |.limit  |.expand |.expand |.expand |.expand |
+/// |minutes         |.limit   |.expand |.expand |.expand |.expand |.expand |
+/// |setPositions    |.limit   |.limit  |.limit  |.limit  |.limit  |.limit  |
+///
+/// - Note 1: Limit if `daysOfTheMonth` is present; otherwise expand
+/// - Note 2: Limit if `daysOfTheYear` or `daysOfTheMonth` is present; otherwise
+///           expand
+enum ComponentAction {
+    /// Replace each date found so far with a list of dates
+    case expand
+    /// Remove some of the dates which have been found so far
+    case limit
+}
+
+extension Calendar {
+    /// A `Sequence` of `Date`s produced by a given recurrence rule
+    struct DatesByRecurring : Sendable, Sequence {
+        typealias Element = Date
+        
+        /// The starting point for the recurrence rule
+        let start: Date
+        /// The recurrenece rule
+        let recurrence: RecurrenceRule
+        /// Range in which the search should occur. If `nil`, return all results
+        let range: Range<Date>?
+        
+        init(start: Date, recurrence: RecurrenceRule, range: Range<Date>?) {
+            self.start = start
+            self.recurrence = recurrence
+            self.range = range
+        }
+        
+        struct Iterator: Sendable, IteratorProtocol {
+            /// The starting date for the recurrence
+            let start: Date
+            /// The recurrence rule that should be used for enumeration
+            let recurrence: RecurrenceRule
+            /// The range in which the sequence should produce results
+            let range: Range<Date>?
+            
+            /// The lower bound of `range`, adjusted so that date expansions may
+            /// still fit in range even if this value is outside the range. This
+            /// value is used as a lower bound for ``nextBaseRecurrenceDate()``.
+            let rangeLowerBound: Date?
+            
+            /// How many occurrences have been found so far
+            var resultsFound = 0
+            
+            let monthAction, weekAction, dayOfYearAction, dayOfMonthAction, weekdayAction, hourAction, minuteAction, secondAction: ComponentAction?
+           
+            /// An iterator for a sequence of dates spaced evenly from the start
+            /// date, by the interval specified by the recurrence rule frequency
+            /// This does not include the start date itself.
+            var baseRecurrence: Calendar.DatesByMatching.Iterator
+            
+            /// How many elements we have consumed from `baseRecurrence` 
+            var iterations: Int = 0
+            
+            /// Whether we are finished enumerating the sequence, either because
+            /// we're past the end of the search range, or because we've had too
+            /// many iterations without finding matches
+            var finished: Bool = false
+            
+            /// How many times `nextGroup()` can be executed without any results
+            /// before we abort the sequence
+            let searchLimit: Int
+            /// How many times `nextGroup()` has been called without returning a
+            /// result
+            var iterationsSinceLastResult: Int = 0
+            
+            internal init(start: Date, 
+                          matching recurrence: RecurrenceRule,
+                          range: Range<Date>?) {
+                // Copy the calendar if it's autoupdating
+                var recurrence = recurrence
+                if recurrence.calendar == .autoupdatingCurrent {
+                    recurrence.calendar = .current
+                }
+                self.recurrence = recurrence
+                
+                self.start = start
+                self.range = range
+                
+                let frequency = recurrence.frequency
+                
+                // Find the appropriate action for every field - expand or limit
+                if recurrence.months.isEmpty {
+                    monthAction = nil
+                } else {
+                    monthAction = switch frequency {
+                    case .yearly: .expand
+                    default: .limit
+                    }
+                }
+                
+                if recurrence.weeks.isEmpty { 
+                    weekAction = nil
+                } else {
+                    weekAction = switch frequency {
+                    case .yearly: .expand
+                    default: nil
+                    }
+                }
+                
+                if recurrence.daysOfTheYear.isEmpty { 
+                    dayOfYearAction = nil
+                } else {
+                    dayOfYearAction = switch frequency {
+                    case .yearly: .expand
+                    case .minutely, .hourly: .limit
+                    default: nil
+                    }
+                }
+                
+                if recurrence.daysOfTheMonth.isEmpty { 
+                    dayOfMonthAction = nil
+                } else {
+                    dayOfMonthAction = switch frequency {
+                    case .yearly, .monthly: .expand
+                    case .minutely, .hourly, .daily: .limit
+                    default: nil
+                    }
+                }
+                
+                if recurrence.weekdays.isEmpty { 
+                    weekdayAction = nil
+                } else {
+                    weekdayAction = switch frequency {
+                    case .minutely, .hourly, .daily: .limit
+                    case .weekly: .expand
+                    case .monthly: 
+                        if recurrence.daysOfTheMonth.isEmpty { 
+                            .expand
+                        } else {
+                            .limit
+                        }
+                    case .yearly: if (!recurrence.daysOfTheMonth.isEmpty || !recurrence.daysOfTheYear.isEmpty || !recurrence.weeks.isEmpty) { .limit } else { .expand }
+                    }
+                }
+                
+                if recurrence.hours.isEmpty { 
+                    hourAction = nil
+                } else {
+                    hourAction = switch frequency {
+                    case .minutely, .hourly: .limit
+                    case .yearly, .monthly, .weekly, .daily: .expand
+                    }
+                }
+                
+                if recurrence.minutes.isEmpty { 
+                    minuteAction = nil
+                } else {
+                    minuteAction = switch frequency {
+                    case .minutely: .limit
+                    case .yearly, .monthly, .weekly, .daily, .hourly: .expand
+                    }
+                }
+                
+                if recurrence.seconds.isEmpty { 
+                    secondAction = nil
+                } else {
+                    secondAction = .expand
+                }
+                
+                if let range {
+                    rangeLowerBound = recurrence.calendar.dateInterval(of: frequency.component, for: range.lowerBound)?.start
+                } else {
+                    rangeLowerBound = nil
+                }
+                
+                // Create date components that enumerate recurrences without any
+                // rules applied. Retrieve the date components of the start date
+                // but leave the field for the recurrence frequency empty, so it
+                // is the only component that increases
+                let components: Calendar.ComponentSet = switch recurrence.frequency {
+                    case .minutely: [.second]
+                    case .hourly:   [.second, .minute]
+                    case .daily:    [.second, .minute, .hour]
+                    case .weekly:   [.second, .minute, .hour, .weekday]
+                    case .monthly:  [.second, .minute, .hour, .day]
+                    case .yearly:   [.second, .minute, .hour, .day, .month]
+                }
+                let componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
+                
+                let rangeForBaseRecurrence: Range<Date>? = nil
+                baseRecurrence = Calendar.DatesByMatching(calendar: recurrence.calendar,
+                                                          start: start,
+                                                          range: rangeForBaseRecurrence,
+                                                          matchingComponents: componentsForEnumerating,
+                                                          matchingPolicy: recurrence.matchingPolicy,
+                                                          repeatedTimePolicy: recurrence.repeatedTimePolicy,
+                                                          direction: .forward).makeIterator()
+                                                          
+                searchLimit = switch recurrence.frequency {
+                    // We want to find a search limit that will stop an infinite
+                    // loop, but that also enumerates through results that users
+                    // might be looking for. Assume that the recurrence requests
+                    // that a date falls on a leap day that occurs on a specific
+                    // weekday, that would yield a result at least once every 40
+                    // years.
+                    case .yearly:   40
+                    case .monthly:  40 * 12
+                    case .weekly:   40 * 53
+                    case .daily:    40 * 366
+                    case .hourly:   40 * 366 * 24
+                    case .minutely: 40 * 366 * 24 * 60
+                }
+            }
+            
+            /// A group of dates within the interval specified by the recurrence
+            /// frequency, sorted so the earliest date is last.
+            /// 
+            /// The result of the sequence consists of dates that are popped off
+            /// `currentGroup`. Once the array is empty, `nextGroup()` is called
+            /// to recompute the array with new dates.
+            /// 
+            /// For example, consider a recurrence with a monthly frequency, and
+            /// `weekdays = [.every(.tuesday)]`. If the initial date is March 1,
+            /// `currentGroup` initially contains all Tuesdays in March. When it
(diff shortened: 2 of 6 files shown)


sarimabbas:
LGTM 👍

sarimabbas (approved):
LGTM 👍

iCharlesHu (approved):

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                    // might be looking for. Assume that the recurrence requests
+                    // that a date falls on a leap day that occurs on a specific
+                    // weekday, that would yield a result at least once every 40
+                    // years.
+                    case .yearly:   40
+                    case .monthly:  40 * 12
+                    case .weekly:   40 * 53
+                    case .daily:    40 * 366
+                    case .hourly:   40 * 366 * 24
+                    case .minutely: 40 * 366 * 24 * 60
```
hristost, line 266: These limits are somewhat arbitrary, but ultimately we wouldn't want `RecurrenceRule` to be used as a predicate for finding rare dates

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+            let start: Date
+            /// The recurrence rule that should be used for enumeration
+            let recurrence: RecurrenceRule
+            /// The range in which the sequence should produce results
+            let range: Range<Date>?
+            
+            /// Interval containing `range.lowerBound`, that has the same length
+            /// as the frequency interval. This can be used as a lower bound for
+            /// ``nextBaseRecurrenceDate()``
+            let rangeLowerBoundInterval: DateInterval?
```
parkera, line 90: I sort of wish we didn't mix `DateInterval` and `Range<Date>`, but it's not a showstopper.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+            
+            /// How many occurrences have been found so far
+            var resultsFound = 0
+            
+            let monthAction, weekAction, dayOfYearAction, dayOfMonthAction, weekdayAction, hourAction, minuteAction, secondAction: ComponentAction?
+           
+            /// An iterator for a sequence of dates spaced evenly from the start
+            /// date, by the interval specified by the recurrence rule frequency
+            /// This does not include the start date itself.
```
parkera, line 99: Style nit, but elsewhere in the project we do not manually wrap these comments.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                
+                // Create date components that enumerate recurrences without any
+                // rules applied. Retrieve the date components of the start date
+                // but leave the field for the recurrence frequency empty, so it
+                // is the only component that increases
+#if FIXED_124640028
+                // rdar://124640028
+                // https://github.com/apple/swift-foundation/issues/483
+                // Calendar.dates(byMatching: ...) does not respect repeatedTimePolicy when .nanosecond is set #483
+                let components: Calendar.ComponentSet = switch recurrence.frequency {
```
parkera, line 225: Does it even make sense to ever include nanoseconds in these results? It's so sparsely supported in the first place. Maybe it should really just always be set to 0.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                    case .daily:    [.second, .minute, .hour]
+                    case .weekly:   [.second, .minute, .hour, .weekday]
+                    case .monthly:  [.second, .minute, .hour, .day]
+                    case .yearly:   [.second, .minute, .hour, .day, .month]
+                }
+#endif
+                let componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
+                
+                let rangeForBaseRecurrence: Range<Date>? = nil
+                baseRecurrence = Calendar.DatesByMatching(calendar: recurrence.calendar,
```
parkera, line 246: Why create this directly instead of calling the `dates(byMatching:...)` API on `Calendar`? There doesn't seem to be much reason to break the abstraction.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                        nextDate = self.start
+                    } else {
+                        nextDate = self.baseRecurrence.next()
+                    }
+                    iterations += 1
+                    return nextDate
+                }
+                while let nextDate = next() {
+                    // Skip every few iterations when an interval has been given
+                    if (iterations - 1) % recurrence.interval != 0 {
```
parkera, line 289: I don't quite understand what this logic is about. Can you explain a bit more?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                        recurrence._expandOrLimitTimeComponent(.minute, dates: &dates, anchor: date, action: action)
+                    }
+                    if secondAction == action {
+                        recurrence._expandOrLimitTimeComponent(.second, dates: &dates, anchor: date, action: action)
+                    }
+                }
+                
+                dates = dates.filter { $0 >= self.start }
+                
+                if case let .afterDate(limit) = recurrence.end._guts {
```
parkera, line 344: The usage of `_guts` seems unfortunately leaked here. Perhaps just put some function on the recurrence to do this calculation and keep the abstraction layer tighter.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                    let hadDates = !dates.isEmpty
+                    dates = dates.filter { $0 <= limit }
```
parkera, line 346: I'm not sure if this is actionable, but it seems like a reason to specify a limit is to avoid the cost of finding those dates in the first place - not just to filter them out at the end. Is it possible to do that?

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -0,0 +1,671 @@
...
+@testable import FoundationInternationalization
+@testable import FoundationEssentials
+#endif // FOUNDATION_FRAMEWORK
+
+@available(FoundationPreview 0.4, *)
+final class CalendarRecurrenceRuleTests: XCTestCase {
+    /// A Gregorian calendar with a time zone set to California
+    var gregorian: Calendar = {
+        var gregorian = Calendar(identifier: .gregorian)
+        gregorian.timeZone = .init(identifier: "US/Pacific")!
```
parkera, line 29: Naming the time zone "US/Pacific" won't work in an Essentials-only test, which may not have time zone data. You can use a fixed GMT-8 if you like; or, if your test requires checking daylight saving transitions, move this up to the internationalization tests.

I don't have an issue with Essentials functionality being tested in the Internationalization tests, if that gives us better test coverage. They really all run together anyway.

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -0,0 +1,671 @@
...
+//===----------------------------------------------------------------------===//
+
+#if canImport(TestSupport)
+import TestSupport
+#endif
+
+#if FOUNDATION_FRAMEWORK
+@testable import Foundation
+#else
+@testable import FoundationInternationalization
```
parkera, line 20: This is in the essentials tests but uses Internationalization. More on this below.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
hristost: We need to store the iterator, since it is used throughout the enumeration and we don't want to reset. We can't use `dates(byMatching: ...).makeIterator()` because we need to save this in a var the structure, and thus it'd need to be `Sendable`. `any (IteratorProtocol<Date> & Sendable)` does not currently compile (rdar://104888113), so we just represent it using the internal type which is known to be `Sendable`

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
hristost: Moved tests that depend on this timezone to Internalization tests, all others are in GMT now.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                guard let date = self.nextBaseRecurrenceDate() else {
+                    finished = true
+                    return
+                }
+                
+                var dates: [Date] = [date]
+                 
+                // First expand the set of dates, and then filter it
+                for action in [ComponentAction.expand, ComponentAction.limit] {
+                    if monthAction == action {
```
itingliu, line 316: Can we add a comment there that the order of expanding or limiting each calendar component needs to stay exactly as this?

Benchmarks/Benchmarks/Calendar/Calendar.swift
```diff
@@ -40,7 +40,26 @@ let benchmarks = {
...
+    Benchmark("nextThousandThanksgivingsUsingRecurrenceRule") { benchmark in
+        var rule = Calendar.RecurrenceRule(calendar: cal, frequency: .yearly, end: .afterOccurrences(1000))
+        rule.months = [11]
+        rule.weekdays = [.nth(4, .thursday)]
+        rule.matchingPolicy = .nextTime
+        var count = 0
+        for _ in rule.recurrences(of: thanksgivingStart) {
+            count += 1
+        }
+        assert(count == 1000)
```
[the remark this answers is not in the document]
hristost: I fixed an obvious performance pitfall (we were enumerating over a wider range than needed), and now we are within <4x runtime of the benchmark that uses enumeration by date components. Since we also own the implementation for matching date components, the next step would be to bypass the public API and use some of the internals to get the runtime even closer.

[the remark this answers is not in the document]
hristost: The original benchmark (`nextThousandThanksgivings`) does not actually compute Thanksgivings as the name implies. It finds the Thursday in the fourth week of November, which is not necessarily the fourth Thursday in November. Is it permissible to update the benchmark given that the runtime will be slightly different?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                    case .weekly:   40 * 53
+                    case .daily:    40 * 366
+                    case .hourly:   40 * 366 * 24
+                    case .minutely: 40 * 366 * 24 * 60
+                }
+            }
+            
+            /// A group of dates within the interval specified by the recurrence
+            /// frequency
+            var currentGroup: [Date] = []
```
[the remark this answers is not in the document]
hristost: Yes, that is correct. I can perhaps include this an example in the comment

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                        }
+                    }
+                    return nextDate
+                }
+                return nil
+            }
+            
+            /// Compute `currentGroup` by advancing by a frequency interval from
+            /// the most recent group
+            mutating func nextGroup() {
```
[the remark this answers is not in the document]
hristost: No, it's only called when `finished` is `false`. I'll add that as a precondition.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                let offset = weekIdx - week
+                return calendar.date(byAdding: .weekOfYear, value: offset, to: date)
+            }
+        }
+    }
+    
+    internal func _expandOrLimitMonths(dates: inout [Date], anchor: Date, action: ComponentAction) {
+        let monthRange = calendar.range(of: .month, in: .year, for: anchor)
+        let months = months.map { month in
+            if month.index > 0 {
```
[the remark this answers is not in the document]
hristost: I think `0` is more legible here.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                return month
+            } else {
+                let newIndex = monthRange!.upperBound + month.index
```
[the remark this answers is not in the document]
hristost: I wouldn't call that normalization, it's simply so we can address months which are usually counted backwards. We can normalize by filtering the resulting array later, but also we achieve the same results now by leaving invalid months in, and having Calendar APIs return `nil`

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+                let components: Calendar.ComponentSet = switch recurrence.frequency {
+                    case .minutely: [.second]
+                    case .hourly:   [.second, .minute]
+                    case .daily:    [.second, .minute, .hour]
+                    case .weekly:   [.second, .minute, .hour, .weekday]
+                    case .monthly:  [.second, .minute, .hour, .day]
+                    case .yearly:   [.second, .minute, .hour, .day, .month]
+                }
+#endif
+                let componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start)
```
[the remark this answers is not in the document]
hristost: This is a good catch. I shadowed `recurrence` right where this function begins so it's not possible to get confused again.

Benchmarks/Benchmarks/Calendar/Calendar.swift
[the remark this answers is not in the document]
hristost: > Since we also own the implementation for matching date components, the next step would be to bypass the public API and use some of the internals to get the runtime even closer.

I got started on this, but the potential performance improvement does not seem to justify the amount of work it takes to get there. I'd like to merge the implementation as it is now, and revisit performance with a later PR. The benchmarks taken on my machine are as follows:

```
nextThousandThanksgivings
╒═══════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                    │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *          │     426 │     426 │     426 │     426 │     426 │     426 │     426 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Throughput (# / s) (K)    │       2 │       2 │       2 │       2 │       2 │       2 │       2 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (μs) *   │     580 │     581 │     581 │     583 │     583 │     583 │     583 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (wall clock) (μs) *  │     582 │     584 │     584 │     586 │     586 │     586 │     586 │       6 │
╘═══════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛

nextThousandThanksgivingsSequence
╒═══════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                    │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *          │     426 │     426 │     426 │     426 │     426 │     426 │     426 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Throughput (# / s) (K)    │       2 │       2 │       2 │       2 │       2 │       2 │       2 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (μs) *   │     580 │     581 │     581 │     581 │     581 │     581 │     581 │       6 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (wall clock) (μs) *  │     582 │     584 │     584 │     585 │     586 │     586 │     586 │       6 │
╘═══════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛

nextThousandThanksgivingsUsingRecurrenceRule
╒═══════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                    │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *          │     861 │     861 │     861 │     861 │     861 │     861 │     861 │       2 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (μs) *   │    2121 │    2121 │    2121 │    2129 │    2129 │    2129 │    2129 │       2 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (wall clock) (μs) *  │    2135 │    2136 │    2136 │    2140 │    2140 │    2140 │    2140 │       2 │
╘═══════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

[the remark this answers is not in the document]
parkera: Ok, that seems reasonable for now. Let's file a follow-up bug just so we don't lose track of the task.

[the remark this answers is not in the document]
hristost: Tracking that in #555 (rdar://126761035)

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,835 @@
...
+enum ComponentAction {
+    /// Replace each date found so far with a list of dates
+    case expand
+    /// Remove some of the dates which have been found so far
+    case limit
+}
+
+extension Calendar {
+    /// A `Sequence` of `Date`s produced by a given recurrence rule
+    @available(FoundationPreview 0.4, *)
```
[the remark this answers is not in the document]
hristost: Possibly not, is there any downside to annotating availability?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
itingliu: @hristost What do you think?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+    }
+    
+    internal func _expandOrLimitMonths(dates: inout [Date], anchor: Date, action: ComponentAction) {
+        lazy var monthRange = calendar.range(of: .month, in: .year, for: anchor)!
+        let months = months.map { month in
+            if month.index > 0 {
+                return month
+            } else {
+                let newIndex = monthRange.upperBound + month.index
+                // The upper bound is the last month plus one. Subtracting 1 we get the last month
```
itingliu, line 489: Is this comment still correct?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+            dates = dates.filter { date in
+                let day = calendar.component(.day, from: date)
+                let dayRange = calendar.range(of: .day, in: .month, for: date)
+                return daysOfTheMonth.map {
+                    if $0 > 0 {
+                        $0
+                    } else {
+                        dayRange!.upperBound + $0
+                    }
+                }.contains(day)
```
itingliu, line 527: We can short path this by flattening the `map` and `contains` into one for loop

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+                .day,
+                .hour,
+                .minute,
+                .second,
+                .nanosecond
+            ]
+            let anchorComponents = calendar._dateComponents(components, from: anchor)
+            
+            var componentsForEnumerating: [DateComponents] = []
+            let enumerationDateInterval = calendar.dateInterval(of: frequency.component, for: anchor)!
```
itingliu, line 540: This can be moved down to line 544 since it's only used `if frequency == .yearly`, right?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+            
+            if frequency == .yearly {
+                let monthRange = calendar.range(of: .month, in: .year, for: anchor)!
+                let firstDayOfYear = enumerationDateInterval.start
+                let monthsToDaysInMonth = monthRange.reduce(into: [Int: Int]()) {
```
itingliu, line 545: We don't need this if every `day` in `daysOfTheMonth` is greater than 0 (if we enter the `if day > 0 {` closure below at line 552). 
Can we calculate this more lazily?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+            dates = dates.filter { date in
+                let day = calendar.component(.dayOfYear, from: date)
+                let dayRange = calendar.range(of: .dayOfYear, in: .year, for: date)
+                return daysOfTheYear.map {
+                    if $0 > 0 {
+                        $0
+                    } else {
+                        dayRange!.upperBound + $0
+                    }
+                }.contains(day)
```
itingliu, line 609: Ditto about flattening `map + contains` into a for loop so we can early return

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+                .second,
+                .nanosecond
+            ]
+            let anchorComponents = calendar._dateComponents(components, from: anchor)
+            
+            var componentsForEnumerating: [DateComponents] = []
+            let enumerationDateInterval = calendar.dateInterval(of: frequency.component, for: anchor)!
+            
+            
+            let daysInYear = calendar.range(of: .dayOfYear, in: .year, for: anchor)!.upperBound - 1
```
itingliu, line 624: Ditto about lazily calculate this

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,846 @@
...
+            .weekOfMonth
+        } else {
+            .weekOfYear
+        }
+        /// The components we return for matching and enumeration
+        let componentSet: Calendar.ComponentSet = [.weekday, .hour, .minute, .second, .nanosecond]
+        
+                
+        guard
+            let weekRange = calendar.range(of: weekComponent, in: parent, for: anchor),
```
itingliu, line 807: Ditto about lazily compute this

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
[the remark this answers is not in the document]
parkera: At the very least, it's confusing. I would like to know what the correct answer is, though.

itingliu (commented):
Super cool to see this in action

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
hristost: yeah! the upper bound would be 13 in the Gregorian calendar. So the last month is 13 + (-1) = 12

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,853 @@
...
+    internal func _expandOrLimitDaysOfTheMonth(dates: inout [Date], anchor: Date, action: ComponentAction) {
+        if action == .limit {
+            dates = dates.filter { date in
+                let day = calendar.component(.day, from: date)
+                var dayRange: Range<Int>? = nil
+                for dayOfMonth in daysOfTheMonth {
+                    if dayOfMonth > 0 {
+                        if dayOfMonth == day { return true }
+                    } else {
+                        if dayRange == nil {
```
hristost, line 525: We can't have `lazy var` inside a closure for some reason, so we store the day range in an optional and calculate it the first time it is assigned.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,836 @@
...
+    /// Replace each date found so far with a list of dates
+    case expand
+    /// Remove some of the dates which have been found so far
+    case limit
+}
+
+extension Calendar {
+    /// A `Sequence` of `Date`s produced by a given recurrence rule
+    @available(FoundationPreview 0.4, *)
+    struct DatesByRecurring : Sendable, Sequence {
```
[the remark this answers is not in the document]
hristost: We could. Is there a precedent for other sequences being lazy by default? If we don't add conformance, the user can always use `.lazy` to lazily map and filter.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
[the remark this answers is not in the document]
itingliu: @hristost Were you able to remove this? If not, what kind of error did the compiler give?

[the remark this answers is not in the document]
itingliu: @hristost Were you able to remove this? If not, what error did the compiler give?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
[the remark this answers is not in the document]
itingliu: There are some in other modules, but not in Swift-foundation. I'm not proposing for doing that here of course since that would require API proposal, just something we can consider in the future since we were talking about performance

itingliu (changes requested):
Just one more pending question about availability annotation

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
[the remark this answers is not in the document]
hristost: My bad, I missed this message. I removed the annotation -- we don't need it.

hristost:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -0,0 +1,822 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+//
+// Recurrence enumeration
+//
+// This file implements enumerating occurences according to a recurrence rule as
+// specified in RFC5545 and RFC7529
+
+@available(FoundationPreview 0.4, *)
```
parkera, line 18: Let's see if we can do without this.

parkera (approved):

hristost:
@swift-ci please test

hristost:
@swift-ci please test

PR #466, opened 2024-03-06
Update Data.swift framework side builds; add base64 benchmark

parkera:
Update Data with some requirements for the Framework side builds of base64 and others. For now we will continue to use the NSData implementation there, and come back later to unify once we have validated the performance requirements. Added a base64 benchmark.


parkera:
@swift-ci test

glessard (approved):

PR #467, opened 2024-03-06
Update swift-collections dependency to the 1.1 tag

jmschonfeld:
Now that [swift-collections 1.1](https://github.com/apple/swift-collections/releases/tag/1.1.0) is released, we can update our dependency to point at that tag. Our previous dependency was on a particular commit on the `release/1.1` branch, but now we'll depend on the official tag instead of a particular commit. This tag includes everything that swift-foundation needs.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -48,7 +48,7 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/apple/swift-collections",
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #469, opened 2024-03-07
FileManager should always return values for immutable/appendOnly attributes

jmschonfeld:
Currently, `FileManager`'s `attributesOfItem(atPath:)` returns a dictionary which only contains values for the `immutable` and `appendOnly` keys if the value is true, otherwise the dictionary does not contain the key. Instead, we should always add the value to the dictionary even if the value is false to match the previous behavior of the ObjC implementation.

This also adds a drive-by fix to ensure that we can also read `Bool` values for these keys in `setAttributes(_:ofItemAtPath:)` instead of just numeric values like `Int` outside of `FOUNDATION_FRAMEWORK`.

Resolves rdar://123713046

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -78,7 +78,9 @@ func _readFileAttributePrimitive(_ value: Any?, as type: Bool.Type) -> Bool? {
     }
     #endif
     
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

jmschonfeld:
@swift-ci please test

PR #470, opened 2024-03-07
Fix FileManager macOS trash folder truncation

jmschonfeld:
This code is only called for specially handled cases in search path enumeration. In these specially handled cases, the previous code to strip trailing slashes from the string accidentally truncated one too many characters for strings ending in a slash, resulting in paths ending like `<some path>/.Tras` instead of `<some path>/.Trash`. This replaces the buffer truncation logic with a string initialization and string index logic (which is now correct). It's ok to not call to FileManager here since we use the default file manager, so calling `String(cString:)` directly is much more efficient.


jmschonfeld:
@swift-ci please test

PR #472, opened 2024-03-08
Duplicated DiscreteFormatStyle in tests

itingliu:
We started getting errors about duplicated symbols now that `DiscreteFormatStyle` has landed in SDK. Workaround this by adding a typealias like other types.


itingliu:
@swift-ci please test

parkera (approved):

PR #477, opened 2024-03-12
FileManager should not crash on malformed modificationDate attribute value

jmschonfeld:
Previously, the ObjC implementation of `FileManager` did not crash on modification dates provided that were malformed (for example, having a time interval of +/-Inf or NaN), but the new swift implementation crashes because that new value cannot fit inside of the integer sized type defined by the system module. Instead, we now ignore this attribute value if the date cannot fit into the appropriately sized integer type.


jmschonfeld:
@swift-ci please test

PR #478, opened 2024-03-13
(124302607) Path-based CocoaErrors should use both NSFilePathErrorKey and NSURLErrorKey

jmschonfeld:
Our recent rewrites of `Data`/`FileManager` have switched around whether we use `NSFilePathErrorKey` or `NSURLErrorKey` in some implementations. While we don't make any guarantees about which will be used in a given API, we can update the implementation to provide values for both keys for compatibility.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #480, opened 2024-03-14
Return early if the component value being added is zero

itingliu:
When adding month or year to a `Date`, we convert a `Date` to a `DateComponent`, modify the component, and convert the modified `DateComponent` back to `Date`. However for dates falling into the "repeated" time frame on DST end date, we do not have the information of which time the `DateComponent` represents when converting it back to `Date`.
    
This has been the design, and we have always assumed the second occurrence. The only implementation difference is that in the ICU implementation, we return early when the value is 0 and do not proceed onto adding. We should do that for GregorianCalendar too.
    
Resolves 124231654


itingliu:
@swift-ci please test

parkera (approved):

PR #481, opened 2024-03-14
Add bounds check for CommandLine.arguments since it could be empty

iCharlesHu:
`ProcessInfo.arguments` assumed `CommandLine.arguments` is always non-empty, and that turns out not to be the case 100%. Added bounds check to avoid crashing.


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #482, opened 2024-03-14
Add components in the order of granularity in `date(byAdding:to:)`

itingliu:
We compute the differences in Calendar component values from the largest to the smallest in `dateComponents(_:from:to:)`, but we are not following the same order in `date(byAdding:to:)`. Change the order of the latter so we add the largest components first.

The results are the same for most cases regardless of the order. The only exception is when the addition moves the date across DST transition. The reason for this is that the implementation aims to maintain the clock time when adding units that are larger than `day` to a date, so that the time in the day remains unchanged even after time zone offset changes. However, we cannot hold this promise if the result lands in the "skipped hour" on the DST start date as that time does not actually exist. To adjust for this case, we adjust the time of the day to the correct timezone.

For example, the DST start of year 2024 in the Los Angeles time zone is 2024-03-10, where 02:00:00-0800 becomes 03:00:00-0700.

Here's what happens when we add 1 day first, then add 1 week, to 2024-03-09T02:34:36-0800, the day before DST start:

We start by adding 1 day:
2024-03-09T02:34:36-0800 + 1 day --> 2024-03-10T02:34:36-0800

This is not a valid date in this time zone, so we adjust it to -0700:
2024-03-10T02:34:36-0800 --> 2024-03-10T03:34:36-0700

Then we add a week, which is 7 days:
2024-03-10T03:34:36-0700 + 7 days -> 2024-03-17T03:34:36-0700

Notice that the local time is different in the result from that of the original, which changes from 02:34 to 03:34.

If we start by adding 1 week:
2024-03-09T02:34:36-0800 + 7 days -> 2024-03-16T02:34:36-0700

Then add a day:
2024-03-16T02:34:36-0700 + 1 day -> 2024-03-17T02:34:36-0700

The local time is the same as that of the original.

We go for the latter to allow round-tripping a date to the date by adding their differences back.

Resolves 124414807

### DIFF
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1263,7 +1263,9 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 
             // No leap month support needed here, since these are quantities, not values
 
-            // Add from the largest component to the smallest
+            // Add from the largest component to the smallest to match the order used in `dateComponents(_:from:to:)` to allow round tripping
+            // The results are the same for most cases regardless of the order except for when the addition moves the date across DST transition.
+            // We aim to maintain the clock time when adding larger-than-day units, so that the time in the day remains unchanged even after time zone changes. However, we cannot hold this promise if the result lands in the "skipped hour" on the DST start date as that time does not actually exist. In this case we adjust the time of the day to the correct timezone. There is always some ambiguity, but matching the order used in `dateComponents(_:from:to:)` at least allows round tripping.
             if let amount = components.era { _ = _locked_add(UCAL_ERA, amount: amount, wrap: wrappingComponents, status: &status) }
             if let amount = components.year { _ = _locked_add(UCAL_YEAR, amount: amount, wrap: wrappingComponents, status: &status) }
             if let amount = components.yearForWeekOfYear { _ = _locked_add(UCAL_YEAR_WOY, amount: amount, wrap: wrappingComponents, status: &status) }
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1262,20 +1262,27 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
             var nanosecond = 0
 
             // No leap month support needed here, since these are quantities, not values
+
+            // Add from the largest component to the smallest
             if let amount = components.era { _ = _locked_add(UCAL_ERA, amount: amount, wrap: wrappingComponents, status: &status) }
             if let amount = components.year { _ = _locked_add(UCAL_YEAR, amount: amount, wrap: wrappingComponents, status: &status) }
             if let amount = components.yearForWeekOfYear { _ = _locked_add(UCAL_YEAR_WOY, amount: amount, wrap: wrappingComponents, status: &status) }
             // TODO: Support quarter
             // if let _ = components.quarter {  }
             if let amount = components.month { _ = _locked_add(UCAL_MONTH, amount: amount, wrap: wrappingComponents, status: &status) }
-            if let amount = components.day { _ = _locked_add(UCAL_DAY_OF_MONTH, amount: amount, wrap: wrappingComponents, status: &status) }
-            if let amount = components.dayOfYear { _ = _locked_add(UCAL_DAY_OF_YEAR, amount: amount, wrap: wrappingComponents, status: &status) }
+
+            // Weeks
             if let amount = components.weekOfYear { _ = _locked_add(UCAL_WEEK_OF_YEAR, amount: amount, wrap: wrappingComponents, status: &status) }
+            if let amount = components.weekOfMonth { _ = _locked_add(UCAL_WEEK_OF_MONTH, amount: amount, wrap: wrappingComponents, status: &status) }
+            if let amount = components.weekdayOrdinal { _ = _locked_add(UCAL_DAY_OF_WEEK_IN_MONTH, amount: amount, wrap: wrappingComponents, status: &status) }
             // `week` is for backward compatibility only, and is only used if weekOfYear is missing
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1833,6 +1840,7 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
     }
 
     private func _locked_add(_ field: UCalendarDateFields, amount: Int, wrap: Bool, status: inout UErrorCode) -> UDate {
+        let original = ucal_getMillis(ucalendar, &status)
```
hristost, line 1843: where is this being used?

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -987,6 +987,51 @@ final class CalendarTests : XCTestCase {
...
+            (Date(timeIntervalSinceReferenceDate: 1962440826), Date(timeIntervalSinceReferenceDate: 1963132026)),
+
+            // 2024-03-08T02:34:36-0800, 2063-03-18T11:27:06-0700
+            (Date(timeIntervalSinceReferenceDate: 731586876.690495), Date(timeIntervalSinceReferenceDate: 1963160826.550588)),
+            
+            // 2024-03-03T02:34:36-0800, 2024-03-11T02:34:36-0700
+            (Date(timeIntervalSinceReferenceDate: 731154876), Date(timeIntervalSinceReferenceDate: 731842476)),
+        ]
+
+        for test in tests {
```
hristost, line 1026:
```suggestion
        for (a, b) in tests {
```

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
itingliu: It's not used. Thanks

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1883,6 +1892,7 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
                 hour = ucal_get(ucalendar, UCAL_HOUR_OF_DAY, &status)
             }
 
+
```
theMomax, line 1895: In case that wasn't intentional, double newline.

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -554,32 +555,37 @@ final class GregorianCalendarTests : XCTestCase {
         testAdding(.init(month: -1, day: 30 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:823304100.0))
         testAdding(.init(year: 4, day: -1 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:951867300.0))
         testAdding(.init(day: -1, hour: 24 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:825636900.0))
-        testAdding(.init(day: -1, weekday: 1 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:825723300.0))
-        testAdding(.init(day: -7, weekOfYear: 1 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:828401700.0))
-        testAdding(.init(day: -7, weekOfMonth: 1 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:825982500.0))
-        testAdding(.init(day: -7, weekOfMonth: 1, weekOfYear: 1 ), to: march1_1996, wrap: true, expected: Date(timeIntervalSince1970:829006500.0))
```
theMomax, line 560: Looks like you removed this test case. Was it invalid?

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -1083,6 +1081,48 @@ final class CalendarTests : XCTestCase {
...
+            (Date(timeIntervalSinceReferenceDate: 1962440826), Date(timeIntervalSinceReferenceDate: 1963132026)),
+
+            // 2024-03-08T02:34:36-0800, 2063-03-18T11:27:06-0700
+            (Date(timeIntervalSinceReferenceDate: 731586876.690495), Date(timeIntervalSinceReferenceDate: 1963160826.550588)),
+            
+            // 2024-03-03T02:34:36-0800, 2024-03-11T02:34:36-0700
+            (Date(timeIntervalSinceReferenceDate: 731154876), Date(timeIntervalSinceReferenceDate: 731842476)),
+        ]
+
+        for (start, end) in tests {
```
theMomax, line 1120: I prefer tests to not loop over samples as it makes it hard to tell which sample caused the failure, especially in CI...but your choice!

theMomax (approved):
Thank you, looks great!

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
itingliu: It's an ambiguous test case in my opinion since the expected wrapping result depends on the assumed order of the addition, but maybe it's not invalid. 

Let me add this back with an updated expectation.

Tests/FoundationInternationalizationTests/CalendarTests.swift
itingliu: I absolutely agree with you. I was just being lazy.

itingliu:
@swift-ci please test

PR #492, opened 2024-03-15
Avoid copying mutable NSData subclasses when writing

parkera:
A case I missed when bringing Data writing into Swift -- `NSMutableData` is copied when bridged into `struct Data`. There is no need to do that.

Here I just lower the stack of writing functions to the lowest common denominator, which we used eventually when we call the bare `write` function -- `UnsafeRawBufferPointer`. Explicitly listing the input type of the ObjC function as `NSData` instead of `Data` avoids the bridge itself and allows the Swift code to grab a pointer/size pair.


parkera:
@swift-ci test

parkera:
@swift-ci test

itingliu (approved):
🙌

PR #494, opened 2024-03-18
[Gregorian Calendar] Correctly handle last week of year and first week of year when calculating a `Date` from `DateComponents`

itingliu:
When deciding the component values when calculating `Date` given a `DateComponent`, we incorrectly rewind to the previous year if `weekOfYear == 1` regardless of the passed-in `month`. This adjustment is incorrect and unncessary when `month == 1` -- being in the first month already presumes the start of the year, so doing so results in over-adjustment.

Fix this by skipping this adjustment if we're already in the first month.

Resolves 124661497

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -67,7 +67,7 @@ enum ResolvedDateComponents {
     case weekOfMonth(year: Int, month: Int, weekOfMonth: Int, weekday: Int?)
 
     // Pick the year field between yearForWeekOfYear and year and resovles era
-    static func yearMonth(forDateComponent components: DateComponents) -> (year: Int, month: Int) {
+    static func yearOrYearForWOYAdjustingEra(from components: DateComponents) -> (year: Int, month: Int) {
         var rawYear: Int
         // Don't adjust for era if week is also specified
         var adjustEra = true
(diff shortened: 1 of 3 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -115,19 +115,26 @@ enum ResolvedDateComponents {
...
         let minWeekdayOrdinal = 1
 
         // TODO: Check day of year value here
         if let d = components.day {
-            if components.yearForWeekOfYear != nil, let weekOfYear = components.weekOfYear {
-                if components.month == nil && weekOfYear >= 52 {
-                    year += 1
-                } else if weekOfYear == 1 {
-                    year -= 1
+            let extendedYear: Int
```
parkera, line 123: Is this really 'yearForWeekOfYear'?

itingliu: Are you referring to the line below? 
```swift
if components.yearForWeekOfYear != nil, let weekOfYear = components.weekOfYear,
```

Here we're checking if `weekOfYear` is set. And if it is, we would also expect `yearForWeekOfYear` to be set instead of `year`, don't we?

parkera: All I meant here was that the concept of an "extended year" is the same as a "year for week of year", right? If so I was wondering if the new terminology may be confusing to readers.

itingliu: I borrowed that term from ICU. It actually means "either `year` or `yearForWeekOfYear`, depending on if you want to calculate a date from given `month`+`day`,  or `week`+ `weekday`"

This isn't very clear admittedly, so I'm still thinking about refactoring this to make it more obvious.

itingliu:
@swift-ci please test

parkera (approved):

PR #495, opened 2024-03-18
Use a CShim to avoid an unescapable warning about using mktemp

parkera:
We work around the security issue with `mktemp` by checking for the existence of a conflict and retrying if necessary. The comment already in the source code explains why we can't use `mkstemp`. To avoid this warning appearing for every client of the package, work around it by adding a C shim before we call it.

### DIFF
--- a/Sources/_CShims/include/data_shims.h
+++ b/Sources/_CShims/include/data_shims.h
@@ -1,0 +1,20 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CSHIMS_DATA_H
+#define CSHIMS_DATA_H
+
+#include "_CShimsMacros.h"
+
+INTERNAL char *_datashims_mktemp(char *arg);
+
+#endif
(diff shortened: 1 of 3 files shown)


parkera:
@swift-ci test

Sources/_CShims/include/data_shims.h
```diff
@@ -0,0 +1,20 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef CSHIMS_DATA_H
+#define CSHIMS_DATA_H
+
+#include "_CShimsMacros.h"
+
+INTERNAL char *_data_mktemp(char *arg);
```
jmschonfeld, line 18: nit: I think the pattern used in some of the other shim files like `filemanager_shims` and `string_shims` would be `_datashims_mktemp`. I'm not sure if we've really decided on that pattern, but should we be as consistent as we can be?

parkera: They are actually sort of all over the place. I'd like to do a follow up commit to address the naming and styles.

DougGregor:
@swift-ci test

DougGregor:
Huh, the Linux one picked up yesterday's toolchain

DougGregor:
@swift-ci please test Linux

DougGregor:
@swift-ci please clean test Linux

jmschonfeld (approved):

DougGregor:
@swift-ci please test Linux

DougGregor:
Well, we got the new toolchain on Linux, but `gold` is failing to link on Linux

parkera:
@swift-ci test

iCharlesHu (approved):

parkera:
@swift-ci test

parkera:
We should be able to use the new warning control flags to handle this now.

PR #497, opened 2024-03-20
Ensure stable home directory for processes calling seteuid

jmschonfeld:
Historically, Foundation has ensured a stable home directory for processes that call `seteuid` by falling back to `getuid()` when the EUID is the root user. This ensures that processes calling `seteuid(0)` maintain a stable home directory. In order to preserve compatibility with clients expecting this behavior, restore the behavior to the swift implementation.

### DIFF
--- a/Sources/FoundationEssentials/Platform.swift
+++ b/Sources/FoundationEssentials/Platform.swift
@@ -109,16 +109,22 @@ private var _cachedUGIDs: (uid_t, gid_t) = {
 #endif
 
 extension Platform {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
Linux CI failure is a known compiler crash that has been fixed on ToT Swift, merging this now since it passed on Linux previously

PR #498, opened 2024-03-20
Do not include _CShims from the test project

parkera:
The module is internal to Foundation's implementation only. In addition, the call site for it was a disabled test.


parkera:
@swift-ci test

parkera:
I suspect, although have not fully proved yet, that this import was leaking into downstream projects use of testing with swift-foundation. I was seeing errors there about not finding the "_CShims" module.

jmschonfeld (approved):

PR #501, opened 2024-03-21
attributesForItem(atPath:) uses FileAttributeType instead of String

jmschonfeld:
Previously when `attributesForItem(atPath:)` was written in Objective-C, the effective type of the value for the `.type` key was an `NSString`. In Swift, we added it to the dictionary as a `FileAttributeType` but to preserve compatibility with clients relying on `NSString` to `String` casting (ex. `attributesForItem(atPath: ...)[.type] as? String`) we will store the value as an `NSString` instead where `NSString` is available (in non-`FOUNDATION_FRAMEWORK` we continue storing it as a `FileAttributeType` since this promotes the best Swift interface rather than storing the raw value as a `String`)

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -43,7 +43,7 @@ private func _nameFor(gid: gid_t) -> String? {
 }
 
 extension mode_t {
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:

iCharlesHu (approved):

PR #502, opened 2024-03-21
Remove `_FoundationInternals` symlink workaround in favor of `package` modifier

jmschonfeld:
We currently have a `_FoundationInternals` folder and a symlink in both FoundationEssentials + FoundationInternationalization pointing to this folder to share implementations. However, this adds unnecessary complexity and can be replaced by using the `package` modifier. Now `LockedState.swift` lives in FoundationEssentials and is declared as `package`-visible. The `BinaryFloatingPoint.swift` file was actually only used by FoundationInternationalization, so I've moved that file to the internationalization module.


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #504, opened 2024-03-24
Remove an unnecessary sort

koher:
Because it is unnecessary to sort values to get the smallest one, replaced `sorted` with `min`.

(Since a smaller `rawValue` represents a larger unit, an inequality operator is inverted.)


itingliu:
@swift-ci please test

parkera (approved):

PR #507, opened 2024-03-25
`_SearchPaths` doesn't correctly mask invalid domains

jmschonfeld:
The previous ObjC implementation masked the provided domains to a list of known, valid, public domains. For clients providing the "all domains" value, the new swift implementation doesn't appropriately drop bit values for domains not in this masking list resulting in unexpected behavior. This PR adds the masking behavior to the new Swift implementation


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #512, opened 2024-03-27
Calendar.dateComponents(:) performance issues

itingliu:
CalendarGregorian currently decomposes `Locale` into its identifer plus preferences, stores the two properties separately, and recreates everytime a `Locale` when needed. This results in us always creating new locales, and calling ICU whenever requesting its properties. While we do have cache in place for `Locale.init(identifier:preferences:)` for nil `preferences`, we are not hitting the cache because `preferences` is always non-nil when calendar is `.current`.

Fix this by storing a `Locale` inside calendar so we always hit the cache. This also allows `Locale.autoupdatingCurrent` to work properly -- previously you would not get the right locale back if you set `calendar.locale = Locale.autoupdatingCurrent`. This change fixes that as the identity of the `Locale` is now preserved.

Resolves 125169335

### DIFF
--- a/Benchmarks/Benchmarks/Calendar/Calendar.swift
+++ b/Benchmarks/Benchmarks/Calendar/Calendar.swift
@@ -28,6 +28,7 @@ let benchmarks = {
     
     let thanksgivingComponents = DateComponents(month: 11, weekday: 5, weekOfMonth: 4)
     let cal = Calendar(identifier: .gregorian)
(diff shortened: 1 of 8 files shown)


itingliu:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu (approved):

PR #515, opened 2024-03-28
Support data writing to special files on linux

jmschonfeld:
Calling `Data.write(to:)` invokes a call to `fsync` on the file descriptor before returning. On Linux, `fsync` will return `-1` and set `errno` to `EINVAL` for special files (such as `/dev/stdout`) but this is safe to ignore. This change pulls in a test case originally in swift-corelibs-foundation to test this and suppresses this error only on Linux (the `fsync` call on macOS succeeds)

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -237,10 +237,15 @@ private func write(buffer: UnsafeRawBufferPointer, toFileDescriptor fd: Int32, p
     if !buffer.isEmpty {
         if fsync(fd) < 0 {
             let savedErrno = errno
+            let error = CocoaError.errorWithFilePath(path, errno: savedErrno, reading: false)
+            #if os(Linux)
+            // Linux returns -1 and errno == EINVAL if trying to sync a special file, eg a fifo, character device etc which can be ignored.
             if savedErrno != EINVAL {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test macOS

parkera (approved):

jmschonfeld:
This still appears to be failing on macOS CI, I'll push some commits to try to debug

jmschonfeld:
@swift-ci please test macOS

jmschonfeld:

jmschonfeld:
It seems that the macOS version that CI is running returns `ENOTSUP` for this `fsync` call. The SCF implementation has the note

> macOS can also return ENOTSUP for pipes but dont ignore it, so that the behaviour matches Darwin's Foundation.

so I don't think we want to change the behavior of that here. Instead, I've modified the implementation and behavior to just linux and windows, excluding macOS while I do a more in-depth investigation into why some versions of macOS return `ENOTSUP` and others return `EINVAL`. This provides consistency in the linux/windows implementation between SCF and swift-foundation while also keeping the behavior of the Darwin implementation consistent with what Foundation.framework has historically done.

jmschonfeld:
@swift-ci please test

jmschonfeld:

parkera (approved):

PR #520, opened 2024-03-29
FoundationEssentials: add some missing imports for Windows

compnerd:
This allows us to get further into building FoundationEssentials once again on Windows. Much of the file system work has resulted in this module no longer being viable on Windows and will need to be replaced to allow building on Windows which does not have the `fts` APIs.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -14,9 +14,7 @@
 internal import os
 #elseif canImport(Glibc)
 import Glibc
-#endif
-
-#if canImport(CRT)
+#elseif canImport(CRT)
 import CRT
 #endif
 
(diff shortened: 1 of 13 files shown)


compnerd:
@swift-ci please test

parkera (approved):

jmschonfeld (approved):
Thanks for these updates! Yeah at some point it'd be great to take some of the `FileManager` implementations for non-Darwin/non-Linux in swift-corelibs-foundation and copy them over to here to maintain that support (without changing the Darwin behavior), I'm just not quite set up to do that yet but I think that's the next step for `FileManager` support

compnerd:
I think that FileManager and POSIXError are the two cases currently holding back FoundationEssentials from building for Windows. I think that @shahmishal is pretty close to getting Windows CI setup with Docker, so it should be possible test this on Windows soon as well.

PR #521, opened 2024-03-29
Add an autoreleasepool around callout to Swift data write

parkera:
`data.bytes` will add a retain/autorelease on the argument `NSData` here (because it needs to extend the lifetime of the object). This can be surprising to clients calling in from ObjC who do not have their own autoreleasepools in place.


parkera:
@swift-ci test

Catfish-Man (approved):
lgtm

PR #522, opened 2024-03-29
FoundationEssentials: perform explicit conversion of `mode_t`

compnerd:
Windows uses `int` instead of a `mode_t` type. Create a shim alias in `_CShims` and use explicit type conversions to allow sharing of code paths across platforms. This helps reduce some of the build errors enabling further work to re-enable building FoundationEssentials on Windows.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -209,7 +209,7 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
     }
     
     let fileSize = min(Int(clamping: filestat.st_size), maxLength ?? Int.max)
-    let fileType = filestat.st_mode & S_IFMT
+    let fileType = mode_t(filestat.st_mode) & mode_t(S_IFMT)
     let preferredChunkSize = filestat.st_blksize
 #if !NO_FILESYSTEM
     let shouldMap = shouldMapFileDescriptor(fd, path: inPath, options: options)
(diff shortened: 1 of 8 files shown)


compnerd:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -209,17 +209,17 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
...
+    let fileType = mode_t(filestat.st_mode) & mode_t(S_IFMT)
     let preferredChunkSize = filestat.st_blksize
 #if !NO_FILESYSTEM
     let shouldMap = shouldMapFileDescriptor(fd, path: inPath, options: options)
 #else
     let shouldMap = false
 #endif
         
-    if fileType != S_IFREG {
+    if fileType != mode_t(S_IFREG) {
```
jmschonfeld, line 220: For my reference, could you elaborate a bit more on why these casts are all necessary? What type is `S_IFREG` on Windows? Does it use some other integer type, or does it use `int`? If so I think I'd expect that declaring a `typealias mode_t = Int` (or the appropriate integer type) would eliminate the need for these casts because `S_IFREG` would be considered the same type was `fileType` here. Or is there a detail I might be missing here?

shahmishal:
@swift-ci test windows

Sources/FoundationEssentials/Data/Data+Reading.swift
compnerd: There is no `mode_t` on Windows, there is an alias that I have created which is `int`. However, the type system often sees the values as `UInt16` which would then be incompatible with `int`. The explicit casting avoids the mismatched types on the various sides.

Sources/_CShims/include/platform_shims.h
```diff
@@ -24,6 +24,10 @@
 #include <libkern/OSThermalNotification.h>
 #endif
 
+#if defined(_WIN32)
+typedef int mode_t;
```
lorentey, line 28: What does `mode_t` mean on Windows? Why would we want Foundation to use UNIX/POSIX APIs on Windows?

compnerd:
Let's hold off on this. I think that the recent `FileManager` work has been different than the original swift-corelibs-foundation support work. There has been more replacement here with native Windows APIs rather than the compatible APIs which allows us to reduce the locations where we reference `mode_t` and thus we might end up in a place where we do not need this after all.

compnerd:
Going to close this off. Instead of addressing the `mode_t` issues, it is better to use Win32 APIs directly and avoid the POSIX type matching.

PR #524, opened 2024-03-29
Update platform_shims' import condition to __has_include

iCharlesHu:
Update import condition to `__has_include` so it's more permissive (we only really care about being able to import `<unistd.h>` here anyway).

### DIFF
--- a/Sources/_CShims/platform_shims.c
+++ b/Sources/_CShims/platform_shims.c
@@ -16,7 +16,7 @@
 #include <crt_externs.h>
 #elif defined(_WIN32)
 #include <stdlib.h>
-#elif defined(__unix__)
+#elif __has_include(<unistd.h>)
 #include <unistd.h>
 extern char **environ;
 #endif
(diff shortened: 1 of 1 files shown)


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

compnerd:
I'm not sure that this is better. Is there a concrete reason to prefer `__has_include` over the proper `__unix__` macro which *requires* that `unistd.h` be available?

compnerd:
(Note that this change actively impairs portability of the code; I think that this is something that the Platform Steering Group may have thoughts on - @al45tair)

al45tair:
Honestly I'm not sure why we're even including `<unistd.h>` at all, since we explicitly declare `environ` anyway. As for why not `__unix__`, my guess is that someone has something that doesn't define `__unix__` but that does have `<unistd.h>`. I think I'm fairly relaxed about it overall, since I doubt anyone who has `<unistd.h>` doesn't have `environ`, and it seems unlikely that anyone would use that name for anything other than UNIX APIs.

PR #525, opened 2024-04-03
FoundationEssentials: squelch some warnings

compnerd:
When trying to revive the Windows port, these warnings clutter the current set of things to resolve still. Silence the warnings by explicitly ignoring the results.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

parkera (approved):
Thanks for looking at these. I wasn't sure if the return value was platform-specific or not.

compnerd:
`memcpy` and `memset` are defined by the C standard to return the destination which we do not need, so blackholing them here makes sense I think.

PR #527, opened 2024-04-03
FoundationEssentials: support mmap'ed reading on Windows

compnerd:
This adjusts the data reading path for supporting Windows which does not have the `mmap` function but does provide memory-mapped IO. The difference in the API is sufficient to create a separate path for the wrapping function rather than trying to have a single shared implementation that simply uses different APIs for the system integration.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -22,6 +22,7 @@ import Darwin
 import Glibc
 #elseif os(Windows)
 import CRT
+import WinSDK
 #endif
 
 private func readExtendedAttributesFromFileDescriptor(_ fd: Int32, attrsToRead: [String]) -> [String : Data] {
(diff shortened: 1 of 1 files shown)


compnerd:
Note that this doesn't currently work as I don't know how to handle the errors. The error information is not really an `errno` (POSIX error code) but rather a Win32 Error (`GetLastError`). The build failure on Windows for this file is now just the error reporting.

jmschonfeld:
> Note that this doesn't currently work as I don't know how to handle the errors

`CocoaError.errorWithFilePath` with an `errno` just creates a `CocoaError` by mapping the `errno` to an applicable `CocoaError.Code` and then creates a `CocoaError` with that code and a user info dictionary with the underlying `POSIXError`. Would we be able to create another overload of `errorWithFilePath` but instead of an `errno:` parameter perhaps with a `win32Error:` parameter that similarly maps error codes to `CocoaError.Code` and then creates a user info dictionary with just the underlying code instead of using a `POSIXError`?

compnerd:
Thanks @jmschonfeld that sounds like it should be possible to do!

compnerd:
I've created #528 with the new error mapping that can hopefully be used to build this out.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -167,12 +168,126 @@ struct ReadBytesResult {
...
 internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, maxLength: Int?, options: Data.ReadingOptions, attributesToRead: [String], attributes: inout [String: Data]) throws -> ReadBytesResult {
     if inPath.isEmpty {
         // For compatibility, throw a different error than the perhaps-expected 'file not found' here (41646641)
         throw CocoaError(.fileReadInvalidFileName)
     }
-    
+
+#if os(Windows)
+    let hFile: HANDLE = inPath.withFileSystemRepresentation {
+        String(cString: $0!).withCString(encodedAs: UTF16.self) {
```
jmschonfeld, line 210: Given that Windows uses UTF-16 for the file system representation, I think we should probably just make `withFileSystemRepresentation` provide a UTF-16 pointer on Windows so that callers don't have to do this every time. Thoughts? The linux code path in `String+Path.swift` already just basically equates to a call to `withCString` so I think we could just make a new path for Windows that calls `withCString(encodedAs: UTF16.self)` and then just ensure the function signatures use the right types

compnerd: Ah ... so, we now run into the critical juncture of - how do we make code portable? If the signature is different between windows and !windows - the body will need to be replaced. For internal uses, I think that I want to never use `withFileSystemRepresentation` in Windows specific paths and prefer to use `withNTPathRepresentation`. Let me upload that as a splintered change so that we can start using it and setup context.

compnerd: Put up #531 so that we can discuss the windows path representation first.

compnerd: Okay, we can now convert a `String` type to the proper spelling for a NT path, however, we still do not handle `URL` properly.

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #528, opened 2024-04-04
FoundationEssentials: introduce a new error domain for Windows

compnerd:
Introduce the new Win32 error domain and error type to allow us to represent the underlying OS error for file system operations. This prepares us for the necessary support in `FileManager`.

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -17,6 +17,9 @@ internal import _ForSwiftFoundation
 import Darwin
 #elseif canImport(Glibc)
 import Glibc
+#elseif os(Windows)
+import CRT
+import WinSDK
 #endif
 
 extension CocoaError.Code {
@@ -34,7 +37,10 @@ extension CocoaError.Code {
             case ENOENT: .fileNoSuchFile
             case EPERM, EACCES: .fileWriteNoPermission
             case ENAMETOOLONG: .fileWriteInvalidFileName
-            case EDQUOT, ENOSPC: .fileWriteOutOfSpace
+#if !os(Windows)
+            case EDQUOT: .fileWriteOutOfSpace
+#endif
+            case ENOSPC: .fileWriteOutOfSpace
             case EROFS: .fileWriteVolumeReadOnly
             case EEXIST: .fileWriteFileExists
             default: .fileWriteUnknown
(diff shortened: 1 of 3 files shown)


Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -34,7 +37,10 @@ extension CocoaError.Code {
             case ENOENT: .fileNoSuchFile
             case EPERM, EACCES: .fileWriteNoPermission
             case ENAMETOOLONG: .fileWriteInvalidFileName
-            case EDQUOT, ENOSPC: .fileWriteOutOfSpace
+#if !os(Windows)
+            case EDQUOT: .fileWriteOutOfSpace
```
jmschonfeld, line 41: Out of curiosity, what makes this case different than the others? I was surprised to see that this entire function wouldn't be `#if !os(Windows)`

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -121,7 +127,68 @@ extension CocoaError {
...
     }
-    
+
+#if os(Windows)
+    static func errorWithFilePath(_ path: PathOrURL, win32 dwError: DWORD, reading: Bool, variant: String? = nil, userInfo: [String : AnyHashable] = [:]) -> CocoaError {
+        switch path {
+        case let .path(path):
+            return Self.errorWithFilePath(path, win32: dwError, reading: reading, variant: variant, userInfo: userInfo.addingUserInfo(forPath: path))
+        case let .url(url):
+            return Self.errorWithFilePath(url.withUnsafeFileSystemRepresentation { String(cString: $0!) }, win32: dwError, reading: reading, variant: variant, userInfo: userInfo.addingUserInfo(forURL: url))
```
jmschonfeld, line 137: Is it worth just using `url.path` here? Or is routing through the file system representation important?

Sources/FoundationEssentials/Error/ErrorCodes+Win32.swift
```diff
@@ -0,0 +1,33 @@
...
+ // Licensed under Apache License v2.0 with Runtime Library Exception
+ //
+ // See https://swift.org/LICENSE.txt for license information
+ //
+ //===----------------------------------------------------------------------===//
+
+#if os(Windows)
+import WinSDK
+
+public struct Win32Error: Error {
```
jmschonfeld, line 15: Since this is a new `public` type, we might want to run this through review. @itingliu thoughts?

jmschonfeld (commented):
Thanks for writing this up, super excited to see this!

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
compnerd: Windows is POSIX compliant, so there are cases where we want to return POSIX errors. However, `EDQUOT` is not defined on Windows.

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
compnerd: I think that the routing through the FSR is important as the error message would give you a path that is not usable otherwise.

Sources/FoundationEssentials/Error/ErrorCodes+Win32.swift
jmschonfeld: That sounds like a good plan to me

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #533, opened 2024-04-08
FoundationEssentials: repair `FileManager.currentDirectoryPath` on Win32

compnerd:
Use the Windows APIs properly to compute the current working directory. Query the required storage before performing the allocation and then use temporary storage to acquire a buffer. This repairs some of the build errors encountered when building on Windows.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -384,18 +384,21 @@ extension _FileManagerImpl {
     }
     
     var currentDirectoryPath: String? {
+#if os(Windows)
+        let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
+            if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {
+                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+            }
+            return nil
+        }
+#else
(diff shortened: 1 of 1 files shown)


Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -384,18 +384,21 @@ extension _FileManagerImpl {
     }
     
     var currentDirectoryPath: String? {
-        withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
-#if !os(Windows)
-            guard getcwd(buffer.baseAddress!, FileManager.MAX_PATH_SIZE) != nil else {
-                return nil
+#if os(Windows)
+        let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength + 1)) {
```
jmschonfeld, line 389: Do we need to use `dwLength + 1` here instead of just `dwLength`? I was reading through [these docs](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcurrentdirectory) for context and saw:

> If the buffer that is pointed to by `lpBuffer` is not large enough, the return value specifies the required size of the buffer, in characters, including the null-terminating character.

So my take was that `dwLength` should include space for the null byte - does `GetCurrentDirectoryW` behave the same way?

compnerd: Ugh, I got tripped up between the return value in the success and the query - one is _with_ the trailing NULL, and one is _without_ 😮‍💨.

compnerd:
@swift-ci please test

PR #535, opened 2024-04-09
Avoid filling file system representation with partial scalars on failure

jmschonfeld:
Filling a file system representation into a buffer currently partially fills the buffer in the failure case, and the buffer can be filled with a partial unicode scalar (i.e. only the first UTF-8 scalar out of 3 in a unicode scalar) before failing. Previously the ObjC implementation ensured that while we partially fill the buffer during failure we never partially fill it with incomplete UTF-8 so this adjusts the swift implementation to behave the same way.

This doesn't impact the `withFileSystemRepresentation` function our swift implementations use as we never use the partial buffer in the failure case.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #537, opened 2024-04-09
FoundationEssentials: tweak FileManager utilities for Windows

compnerd:
Add a helper for time conversion and remove some functions which do not make sense on Windows - users are not identified by integral IDs but rather by SIDs. This helps reduce the error diagnostics when building FoundationEssentials on Windows.

### DIFF
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -35,6 +35,9 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
     /// The number of seconds from 1 January 1970 to the reference date, 1 January 2001.
     public static let timeIntervalBetween1970AndReferenceDate : TimeInterval = 978307200.0
 
+    /// The number of seconds from 1 January 1601 to the reference date, 1 January 2001.
+    internal static let timeIntervalBetween1601AndReferenceDate: TimeInterval = 12622780800.0
+
     /// The interval between 00:00:00 UTC on 1 January 2001 and the current date and time.
     public static var timeIntervalSinceReferenceDate : TimeInterval {
         return Self.getCurrentAbsoluteTime()
@@ -118,6 +121,10 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
     */
     public var timeIntervalSince1970: TimeInterval {
         return self.timeIntervalSinceReferenceDate + Date.timeIntervalBetween1970AndReferenceDate
+    }
+
+    internal var timeIntervalSince1601: TimeInterval {
+        return self.timeIntervalSinceReferenceDate + Date.timeIntervalBetween1601AndReferenceDate
     }
 
     /// Return a new `Date` by adding a `TimeInterval` to this `Date`.
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -22,6 +22,7 @@ import Glibc
 internal import _CShims
 #elseif os(Windows)
 import CRT
+import WinSDK
 #endif
 
 extension Date {
@@ -30,6 +31,7 @@ extension Date {
     }
 }
 
+#if !os(Windows)
 private func _nameFor(uid: uid_t) -> String? {
     guard let pwd = getpwuid(uid), let name = pwd.pointee.pw_name else {
         return nil
@@ -66,6 +68,7 @@ extension mode_t {
     fileprivate var fileType: FileAttributeType { _fileType }
     #endif
 }
+#endif
 
 func _readFileAttributePrimitive<T: BinaryInteger>(_ value: Any?, as type: T.Type) -> T? {
     guard let value else { return nil }
@@ -121,6 +124,7 @@ func _writeFileAttributePrimitive(_ value: Bool) -> Any {
     #endif
 }
 
+#if !os(Windows)
 extension stat {
     var modificationDate: Date {
         #if canImport(Darwin)
@@ -203,6 +207,7 @@ extension FileProtectionType {
         }
     }
 }
+#endif
 #endif
 
 extension FileAttributeKey {
@@ -335,11 +340,21 @@ extension _FileManagerImpl {
     }
     
     private func _fileExists(_ path: String) -> (exists: Bool, isDirectory: Bool) {
+#if os(Windows)
+        guard !path.isEmpty else { return (false, false) }
+        return try! path.withNTPathRepresentation {
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            guard GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) else {
+                return (false, false)
+            }
+            return (true, faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_DIRECTORY) == DWORD(FILE_ATTRIBUTE_DIRECTORY))
+        }
+#else
         path.withFileSystemRepresentation { rep -> (Bool, Bool) in
             guard let rep else {
                 return (false, false)
             }
-            
+
             var fileInfo = stat()
             guard stat(rep, &fileInfo) == 0 else {
                 return (false, false)
@@ -347,6 +362,7 @@ extension _FileManagerImpl {
             let isDir = (fileInfo.st_mode & S_IFMT) == S_IFDIR
             return (true, isDir)
         }
+#endif
     }
     
     func fileExists(atPath path: String) -> Bool {
@@ -362,24 +378,50 @@ extension _FileManagerImpl {
         isDirectory = result.isDirectory
         return true
     }
-    
+
+#if !os(Windows)
     private func _fileAccessibleForMode(_ path: String, _ mode: Int32) -> Bool {
         path.withFileSystemRepresentation { ptr in
             guard let ptr else { return false }
             return access(ptr, mode) == 0
         }
     }
-    
+#endif
+
     func isReadableFile(atPath path: String) -> Bool {
+#if os(Windows)
+        return (try? path.withNTPathRepresentation {
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            return GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes)
+        }) ?? false
+#else
         _fileAccessibleForMode(path, R_OK)
+#endif
     }
     
     func isWritableFile(atPath path: String) -> Bool {
+#if os(Windows)
+        return (try? path.withNTPathRepresentation {
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            guard GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) else {
+                return false
+            }
+            return faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_READONLY) != DWORD(FILE_ATTRIBUTE_READONLY)
+        }) ?? false
+#else
         _fileAccessibleForMode(path, W_OK)
+#endif
     }
     
     func isExecutableFile(atPath path: String) -> Bool {
+#if os(Windows)
+        return (try? path.withNTPathRepresentation {
+            var dwBinaryType: DWORD = 0
+            return GetBinaryTypeW($0, &dwBinaryType)
+        }) ?? false
+#else
         _fileAccessibleForMode(path, X_OK)
+#endif
     }
     
     func isDeletableFile(atPath path: String) -> Bool {
@@ -387,7 +429,10 @@ extension _FileManagerImpl {
         if parent.isEmpty {
             parent = fileManager.currentDirectoryPath
         }
-        
+
+#if os(Windows)
+        return fileManager.isWritableFile(atPath: parent) && fileManager.isWritableFile(atPath: path)
+#else
         guard fileManager.isWritableFile(atPath: parent),
               let dirInfo = fileManager._fileStat(parent) else {
             return false
@@ -405,8 +450,10 @@ extension _FileManagerImpl {
         } else {
             return true
         }
+#endif
     }
-    
+
+#if !os(Windows)
     private func _extendedAttribute(_ key: UnsafePointer<CChar>, at path: UnsafePointer<CChar>, followSymlinks: Bool) throws -> Data? {
         #if canImport(Darwin)
         var size = getxattr(path, key, nil, 0, 0, followSymlinks ? 0 : XATTR_NOFOLLOW)
@@ -472,8 +519,38 @@ extension _FileManagerImpl {
         }
         return extendedAttrs
     }
-    
+#endif
+
     func attributesOfItem(atPath path: String) throws -> [FileAttributeKey : Any] {
+#if os(Windows)
+        return try path.withNTPathRepresentation { pwszPath in
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            guard GetFileAttributesExW(pwszPath, GetFileExInfoStandard, &faAttributes) else {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+
+            let size: UInt64 = (UInt64(faAttributes.nFileSizeHigh) << 32) | UInt64(faAttributes.nFileSizeLow)
+            let creation: Date = Date(timeIntervalSince1970: faAttributes.ftCreationTime.interval)
+            let modification: Date = Date(timeIntervalSince1970: faAttributes.ftLastWriteTime.interval)
+            return [
+                .size: size,
+                .modificationDate: modification,
+                .creationDate: creation,
+
+                // TODO(compnerd) support these attributes, remapping the Windows semantics...
+                // .posixPermissions: ...,
+                // .referenceCount: ...,
+                // .systemNumber: ...,
+                // .systemFileNumber: ...,
+                // .type: ...,
+                // .ownerAccountID: ...,
+                // .groupownerAccountID: ...,
+                // .ownerAccountName: ...,
+                // .groupOwnerAccountName: ...,
+                // .deviceIdentifier: ...,
+            ]
+        }
+#else
         try fileManager.withFileSystemRepresentation(for: path) { fsRep in
             guard let fsRep else {
                 throw CocoaError.errorWithFilePath(.fileReadUnknown, path)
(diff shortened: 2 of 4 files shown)


compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -28,8 +28,26 @@ import Glibc
...
+package var NSTimeIntervalSince1970: Double {
+    978307200.0
+}
+
+package var NSTimeIntervalSince1601: Double {
+    12622780800.0
+}
+
+extension FILETIME {
+    package var interval: TimeInterval {
```
jmschonfeld, line 44: I think it might make more sense to wait to include these as part of a PR that uses them. `NSTimeIntervalSince1970` is already available as `Date.timeIntervalBetween1970AndReferenceDate` so I think we can just use that instead of `NSTimeIntervalSince1970`, and it doesn't look like we use `NSTimeIntervalSince1601` here right?

Depending on the usages, it might make sense to rename this to `timeIntervalSinceReferenceDate` to indicate exactly what this represents and I suspect it can likely just be `internal`, but being able to see where it's used would be quite helpful

compnerd: I can try to pull out parts of the changes that I have that use this. I am using these two constants for converting between `FILETIME` and `TimeInterval`. I don't see any reason to not use `internal` for the second one. However, I am not sure `timeIntervalSinceReferenceDate` is a great name for the constant - perhaps `Date.timeIntervalBetween1601AndReferenceDate` would work?

jmschonfeld: Ah sorry, I was suggesting `timeIntervalSinceReferenceDate` as the property name on `FILETIME` instead of just `interval`, I think having an internal `Date.timeIntervalBetween1601AndReferenceDate` for the constant seems fine if it's needed for Windows

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -362,32 +378,86 @@ extension _FileManagerImpl {
...
+        }
+#else
         _fileAccessibleForMode(path, R_OK)
+#endif
     }
     
     func isWritableFile(atPath path: String) -> Bool {
+#if os(Windows)
+        do {
+            return try path.withNTPathRepresentation {
```
jmschonfeld, line 409: nit: do you think it's better to write it like this or maybe as `try? path.withNTPathRepresentation { ... } ?? false` if you want, just to avoid some extra nesting - thoughts?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -362,32 +378,86 @@ extension _FileManagerImpl {
...
     }
-    
+#endif
+
     func isReadableFile(atPath path: String) -> Bool {
+#if os(Windows)
+        do {
+            return try path.withNTPathRepresentation {
+                var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+                return GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) == true
```
jmschonfeld, line 396:
```suggestion
                return GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes)
```

Does that work, or does this return an optional?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -362,32 +378,86 @@ extension _FileManagerImpl {
...
+        if faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_READONLY) == DWORD(FILE_ATTRIBUTE_READONLY) {
+            return false
+        }
+        guard (try? path.withNTPathRepresentation({
+            GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes)
+        })) == true else { return false }
+        if faAttributes.dwFileAttributes & DWORD(FILE_ATTRIBUTE_READONLY) == DWORD(FILE_ATTRIBUTE_READONLY) {
+            return false
+        }
+        return true
```
jmschonfeld, line 459: To reuse code, could all of this just be

```swift
fileManager.isWritableFile(atPath: parent) && fileManager.isWritableFile(atPath: path)
```

That'd be a lot simpler / more readable if I'm write that it's the same behavior

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -472,8 +544,36 @@ extension _FileManagerImpl {
...
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            guard GetFileAttributesExW(pwszPath, GetFileExInfoStandard, &faAttributes) else {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+
+            let size: UInt64 = (UInt64(faAttributes.nFileSizeHigh) << 32) | UInt64(faAttributes.nFileSizeLow)
+            let creation: Date = Date(timeIntervalSince1970: faAttributes.ftCreationTime.interval)
+            let modification: Date = Date(timeIntervalSince1970: faAttributes.ftLastWriteTime.interval)
+            return [
+                .size: size,
```
jmschonfeld, line 561: Maybe we could still call `_writeFileAttributePrimitive(_:as:)` here just so we're consistent even if it doesn't really do anything (since we do that on Linux too)?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -472,8 +544,36 @@ extension _FileManagerImpl {
...
+                .modificationDate: modification,
+                .creationDate: creation,
+                // .posixPermissions: ...,
```
jmschonfeld, line 564: Are these a TODO for coming back later (if so, we could leave a TODO comment) or are they just not relevant for Windows (if so, we could leave out the comment and wrap the keys in `#if !os(Windows)`

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -503,9 +603,47 @@ extension _FileManagerImpl {
             #endif
             return attributes
         }
+#endif
     }
     
     func attributesOfFileSystem(forPath path: String) throws -> [FileAttributeKey : Any] {
+#if os(Windows)
+        return try path.withNTPathRepresentation { pwszPath in
+            let dwLength: DWORD = GetFullPathNameW(pwszPath, 0, nil, nil)
```
jmschonfeld, line 612: `withNTPathRepresentation` already calls `GetFullPathNameW` right? If so, is there an equivalent of `strlen` we could use here instead that might be faster?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -503,9 +603,47 @@ extension _FileManagerImpl {
...
+            guard dwLength > 0 else {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+
+            var szVolumeName: [WCHAR] = Array<WCHAR>(repeating: 0, count: Int(dwLength + 1))
```
jmschonfeld, line 617: `withUnsafeTemporaryAllocation` might be a better choice here than an array - also do we need the `+1`?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -503,9 +603,47 @@ extension _FileManagerImpl {
...
+                throw CocoaError.errorWithFilePath(path, win32: DWORD(hr & 0xffff), reading: true)
+            }
+
+            var dwVolumeSerialNumber: DWORD = 0
+            guard GetVolumeInformationW(&szVolumeName, nil, 0, &dwVolumeSerialNumber, nil, nil, nil, 0) else {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+
+            return [
+                .systemSize: UInt64(liTotal.QuadPart),
```
jmschonfeld, line 639: Same comment here about using `_writeFileAttributePrimitive(_:as:)` and the commented out lined below

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -576,20 +714,51 @@ extension _FileManagerImpl {
...
     ) throws {
+        let mode = _readFileAttributePrimitive(attributes[.posixPermissions], as: UInt.self)
+        let immutable = _readFileAttributePrimitive(attributes[.immutable], as: Bool.self)
+        let appendOnly = _readFileAttributePrimitive(attributes[.appendOnly], as: Bool.self)
+
+#if os(Windows)
+        try path.withNTPathRepresentation {
+            if immutable != nil || appendOnly != nil {
+                // Setting these flags is not supported on this platform
+                throw CocoaError.errorWithFilePath(.featureUnsupported, path)
```
jmschonfeld, line 732: If these attributes aren't supported on windows, I think we should wrap their declarations in `#if !os(Windows)` so that it's clear to clients at compile time

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -576,20 +714,51 @@ extension _FileManagerImpl {
...
+            }
+
+            if let modification = attributes[.modificationDate] as? Date {
+                let seconds = modification.timeIntervalSince1970 + (Date.timeIntervalBetween1601AndReferenceDate - Date.timeIntervalBetween1970AndReferenceDate)
```
jmschonfeld, line 736: Should we just have a `timeIntervalSince1601` property on `Date` for Windows that is `internal`?

jmschonfeld (commented):
Thank you so much for your work on this! Left some comments inline but I really appreciate you iterating on this to help us fix up Windows support!

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: As long as I can get it to compile, I think I prefer the `try? ... ?? false` variant.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: There in lies the question - I would like to believe that we should define them, even on Windows. I would rather that we mark these as TODO - yes, some attributes are unavailable, but, hopefully the combination of unit tests and real world usage would expose those. This allows us to reduce the noise in the build and get to the point where we might be able to draw a line in the sand to avoid backsliding the platform support.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: Not in the case that you have a NT path:

```swift
return try path.withCString(encodedAs: UTF16.self) { pwszPath in
  guard !path.hasPrefix(#"\\"#) else { return try body(pwszPath) }
  ...
}
```

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: Aye and nay.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: TBH, I don't know about these. I'm just replicating the non-Darwin path.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: I like that!

Sources/FoundationEssentials/Date.swift
```diff
@@ -120,6 +123,10 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
         return self.timeIntervalSinceReferenceDate + Date.timeIntervalBetween1970AndReferenceDate
     }
 
+    public var timeIntervalSince1601: TimeInterval {
```
parkera, line 126: Can this be internal or package?

compnerd: That is meant to be `internal`, copy-pasta :(.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -335,18 +340,29 @@ extension _FileManagerImpl {
     }
     
     private func _fileExists(_ path: String) -> (exists: Bool, isDirectory: Bool) {
+#if os(Windows)
+        guard !path.isEmpty else { return (false, false) }
+        return try! path.withNTPathRepresentation {
```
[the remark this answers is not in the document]
compnerd: Yeah, I think that `(false, false)` is the right behaviour. If you have a path that either doesn't conform to the requirements (and thus cannot exist), it makes sense to report that as non-existent. Additionally, the API call failing is expected for the cases where the file doesn't exist yet.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: Ah gotcha ok yeah I see now. Marking it as TODO looks good then we can always decide which are applicable later once we get it building

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: I suppose that should be harmless, but I find that it does hurt readability somewhat.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test Windows

compnerd:
Let's see where we stand with this PR - I still have some backlog to work through so I cannot determine this very well locally.

compnerd:
@parkera, @jmschonfeld I might need help to get this merged - I wanted to see how we were doing on Windows, and didn't realise that it would block the merge for the PR :(

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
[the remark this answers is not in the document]
jmschonfeld: Looks like this one might still need to be resolved - we should swap out the `try!` for returning `(false, false)` right?

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
jmschonfeld: Do we want to leave this as just `interval`? What interval does it represent (1601 until the time, or 1970 until the time, or other?)

jmschonfeld:
@compnerd I left two comments about things the look like they might still be unresolved, once we square those away we can help make sure this gets merged. Does that Windows run show us what you were looking for (that we're at least heading in the right direction w.r.t. the number of build failures)?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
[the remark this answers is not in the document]
compnerd: Oh, missed the `try!`

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
compnerd: I think so. The `interval` property converts the `FILETIME` to a `TimeInterval`. The `TimeInterval` initialiser uses the UNIX epoch, which is what this converts it to.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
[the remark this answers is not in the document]
compnerd: Done

compnerd:
Yeah, it seems that we are headed in the right direction. We are getting further, but still have a long road.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
jmschonfeld: Gotcha, in that case do you think we should name this `timeIntervalSince1970` instead of just `interval` to make that clearer? My main worry is that `interval` or even `TimeInterval` on its own don't convey the "start" of the interval and it might make this more approachable/readable to specify the start of the interval (or perhaps just outright convert to a `Date` here if we always use this to represent an instance in time).

Also, am I correct in understanding that `FILETIME` represents the time since 1601? If so, shouldn't this be the following:

```swift
Double((self.dwHighDateTime << 32) | self.dwLowDateTime) - Date.timeIntervalBetween1601AndReferenceDate + Date.timeIntervalBetween1970AndReferenceDate
```

I'm not sure subtracting the interval between 1970 and the reference date from the interval between 1601 and the date in question get's us to the right value

compnerd: I suppose that I can name it `timeIntervalSince1970`. It is meant more as a type conversion. I guess the question is, do you think that this is better:

```swift
TimeInterval(ftTime)
```

or is this more obvious:

```
ftTime.interval
```

The calculation itself is correct, the following is from swift-corelibs-foundation:

```swift
extension TimeInterval {
  init(_ ftTime: FILETIME) {
    self = Double((ftTime.dwHighDateTime << 32) | ftTime.dwLowDateTime) - NSTimeIntervalSince1970;
  }
}
```

compnerd: BTW, a FileTime represents the number of 100-nanosecond intervals since the NT Epoch (1/1/1601).

parkera: Keep in mind `TimeInterval` is simply a typealias of `Double`, so any extensions to it will also appear on any Double as well.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -28,8 +28,20 @@ import Glibc
...
 import CRT
+import WinSDK
 #endif
 
+#if os(Windows)
+extension FILETIME {
+    package var timeIntervalSince1970: TimeInterval {
+        var count: Double = Double((UInt64(self.dwHighDateTime) << 32) | UInt64(self.dwLowDateTime))
+        count /= 1e7 // 100 nanoseconds to seconds
+        return count - Date.timeIntervalBetween1601AndReferenceDate
```
jmschonfeld, line 39: Thanks for updating this, I like the `timeIntervalSince1970` spelling a lot better (and it mirrors the `Date` API nicely). I think this is almost correct: if `count` is the interval between 1601 and the time in question, then subtracting `timeIntervalBetween1601AndReferenceDate` gets us the interval between the reference date (2001) and the time in question. I think we need to add `Date.timeIntervalBetween1970AndReferenceDate` here if we want this to be 1970-aligned (or we should rename this to `timeIntervalSinceReferenceDate` and use that `Date` initializer instead of we want this to be 2001-aligned)

compnerd: Ah, okay - that does end up being exactly what I had initially - except I don't remember why we ended up reworking this 🤔

compnerd: FWIW, I find it to be a bit verbose of a spelling and not very helpful, but 🤷‍♂️

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
jmschonfeld: Cool thanks for updating. Yeah it's a bit more verbose, but I think it'll be quite helpful for other developers reading this since it makes it very clear whether you should use this with `Date(timeIntervalSince1970:)` vs. `Date(timeIntervalSinceReferenceDate:)` for example. But this all looks good to me now

jmschonfeld (approved):
LGTM pending macOS/Linux CI just to make sure we didn't break anything there

PR #543, opened 2024-04-11
FoundationEssentials: add missing import for numerics

compnerd:
Add a missing import for the math routines that are required for the binary integer representation to be processed properly.


compnerd:
@swift-ci please test

iCharlesHu (approved):

PR #544, opened 2024-04-11
FoundationEssentials: support symbolic link operations on Windows

compnerd:
This adjusts the symbolic link paths to be supported on Windows. The symbolic link destination computation is particularly gnarly as it requires re-defining the device driver interface in a shim header to access that and then performing aliasing contortions to gain access to the trailing VLA to get the value.


Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -40,14 +42,35 @@ extension _FileManagerImpl {
         guard !destPath.isEmpty else {
             throw CocoaError.errorWithFilePath(.fileNoSuchFile, destURL)
         }
-        
-        try createSymbolicLink(atPath: path, withDestinationPath: destPath)
+
+#if os(Windows)
+        let bIsDirectory = try path.withNTPathRepresentation {
```
jmschonfeld, line 47: Instead of calling `GetFileAttributesExW` here should we just do

```swift
var bIsDirectory = false
_ = self.fileExists(atPath: path, isDirectory: bIsDirectory)
```

So that we don't duplicate that logic?

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -40,14 +42,35 @@ extension _FileManagerImpl {
...
+        let bIsDirectory = false
+#endif
+        try createSymbolicLink(atPath: path, withDestinationPath: destPath, isDirectory: bIsDirectory)
     }
     
     func createSymbolicLink(
         atPath path: String,
-        withDestinationPath destPath: String
+        withDestinationPath destPath: String,
+        isDirectory: Bool
```
jmschonfeld, line 63: This function gets called from the URL function above, the API in `SwiftFileManager.swift`, and the API in `FileManager+Bridge.swift`. Instead of duplicating the `isDirectory` check to each of those places, could we just calculate it in this function before we call `CreateSymbolicLinkW`?

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -97,6 +121,79 @@ extension _FileManagerImpl {
...
+                throw CocoaError.errorWithFilePath(path, win32: ERROR_BAD_ARGUMENTS, reading: true)
+            }
+
+            let hFile: HANDLE = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS, nil)
+            if hFile == INVALID_HANDLE_VALUE {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+            defer { CloseHandle(hFile) }
+
+            let buffer: UnsafeMutableRawBufferPointer =
```
jmschonfeld, line 141: It looks like this buffer is never deallocated. Would it be better to use `withUnsafeTemporaryAllocation` here so that it's stack allocated if possible? If not, we should add a deferred call to `deallocate` to make sure the memory is free'd

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -97,6 +121,79 @@ extension _FileManagerImpl {
...
+            guard let pBuffer = buffer.baseAddress else {
+                throw CocoaError.errorWithFilePath(path, win32: ERROR_INVALID_DATA, reading: false)
+            }
+
+            var dwBytesWritten: DWORD = 0
+            guard DeviceIoControl(hFile, FSCTL_GET_REPARSE_POINT, nil, 0, pBuffer, DWORD(buffer.count), &dwBytesWritten, nil) else {
+                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
+            // Ensure that we have enough data.
+            guard dwBytesWritten >= MAXIMUM_REPARSE_DATA_BUFFER_SIZE else {
```
jmschonfeld, line 152: Should this be a `guard` that checks if the number of bytes written is less than or equal to the size of the buffer instead of greater than or equal to? If `dwBytesWritten > MAXIMUM_REPARSE_DATA_BUFFER_SIZE` that means our buffer wasn't big enough right?

Sources/_CShims/include/io_shims.h
```diff
@@ -43,4 +43,37 @@ typedef struct FullPathAttributes {
 #pragma pack(pop)
 
 #endif // TARGET_OS_EXCLAVEKIT
+
+#if TARGET_OS_WINDOWS
+
+// Replicated from ntifs.h
```
jmschonfeld, line 49: Are we able to just import this header here to expose it directly to swift so that we don't have to copy the definition? (that's what I did with some of the Glibc headers that we could import but weren't part of the modulemap in the toolchain

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
compnerd: Yeah, that probably makes sense to share the implementation. We only care about the directory type flag.

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
compnerd: Yeah, I think that we should be able to sink it here.

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
compnerd: Yeah, I think that `withUnsafeTemporaryAllocation` makes sense.

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
compnerd: Yeah, I suppose that the right check is `==`. The comment above is because I tripped over this myself for a moment.

Sources/_CShims/include/io_shims.h
compnerd: No, I don't think that we should be vending the header, and getting the header path is really complicated.

jmschonfeld:
@swift-ci please test

compnerd:
@jmschonfeld anything else that needs to be addressed here?

jmschonfeld (approved):
LGTM now! Thanks!

PR #546, opened 2024-04-12
FoundationEssentials: improve Windows coverage for `Platform.swift`

compnerd:
Adjust the codepaths to improve Windows support in the file. This is simply blindly trying to match semantics to reduce the errors reported during building `FoundationEssentials`.


Sources/FoundationEssentials/Platform.swift
```diff
@@ -220,6 +246,15 @@ extension Platform {
...
         return try? FileManager.default.destinationOfSymbolicLink(
             atPath: "/proc/self/exe").standardizingPath
+#elseif os(Windows)
+        let hFile = GetModuleHandleW(nil)
+        let dwLength: DWORD = GetFinalPathNameByHandleW(hFile, nil, 0, FILE_NAME_NORMALIZED)
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { lpBuffer in
+            guard GetFinalPathNameByHandleW(hFile, lpBuffer.baseAddress, dwLength, FILE_NAME_NORMALIZED) == dwLength - 1 else {
+                return nil
+            }
+            return String(decodingCString: lpBuffer.baseAddress!, as: UTF16.self)
```
jmschonfeld, line 256: The other platforms call `.standardizingPath` on the result, should we do the same here for consistency?

compnerd:
@swift-ci please test

Sources/FoundationEssentials/Platform.swift
compnerd: Technically that is already done - `GetFinalPathNameByHandleW` will resolve the symlink and standardize the path in the process. We get the fully normalized location that the file is at.

jmschonfeld: Ah I see, ok if it's already fully standardized the same way that `standardizePath` works then that seems fine to leave it as-is here. Could we add a comment to note that just so it's clear why it's different than the other OS code paths?

jmschonfeld:
It looks like the `CInt(Platform.MAX_HOSTNAME_LENGTH)` change broke the linux/macOS build:

```
/build/swift-foundation/Sources/FoundationEssentials/Platform.swift:206:48: error: cannot convert value of type 'CInt' (aka 'Int32') to expected argument type 'Int'
204 |     static func getHostname() -> String {
205 |         return withUnsafeTemporaryAllocation(of: CChar.self, capacity: Platform.MAX_HOSTNAME_LENGTH + 1) {
206 |             guard gethostname($0.baseAddress!, CInt(Platform.MAX_HOSTNAME_LENGTH)) == 0 else {
    |                                                `- error: cannot convert value of type 'CInt' (aka 'Int32') to expected argument type 'Int'
207 |                 return ""
208 |             }
```

Sources/FoundationEssentials/Platform.swift
compnerd: Sure! That sounds reasonable.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

jmschonfeld (approved):
Looks great now, thanks! (and TIL about `numericCast(x)` 🙂)

PR #547, opened 2024-04-15
FoundationEssentials: implement some path utilities on Windows

compnerd:
Add an implementation for querying the user's home directory and for resolving symbolic links on Windows.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -137,6 +137,81 @@ extension String {
     
     #if !NO_FILESYSTEM
     internal static func homeDirectoryPath(forUser user: String? = nil) -> String {
+#if os(Windows)
+        func GetUserProfile() -> String? {
+            return "USERPROFILE".withCString(encodedAs: UTF16.self) { pwszVariable in
+                let dwLength: DWORD = GetEnvironmentVariableW(pwszVariable, nil, 0)
+                // Ensure that `USERPROFILE` is defined.
+                if dwLength == 0 { return nil }
+                return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
+                    guard GetEnvironmentVariableW(pwszVariable, $0.baseAddress, dwLength) == dwLength - 1 else {
+                        return nil
+                    }
+                    return String(decoding: $0, as: UTF16.self)
+                }
+            }
+        }
+
+        if let user {
+            return user.withCString(encodedAs: UTF16.self) { pwszUserName in
+                var cbSID: DWORD = 0
+                var cchReferencedDomainName: DWORD = 0
+                var eUse: SID_NAME_USE = SidTypeUnknown
+                guard LookupAccountNameW(nil, pwszUserName, nil, &cbSID, nil, &cchReferencedDomainName, &eUse) else {
+                    fatalError("unable to lookup SID for user \(user)")
+                }
+
+                return withUnsafeTemporaryAllocation(of: CChar.self, capacity: Int(cbSID)) { pSID in
+                    return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(cchReferencedDomainName)) { pwszReferencedDomainName in
+                        guard LookupAccountNameW(nil, pwszUserName, pSID.baseAddress, &cbSID, pwszReferencedDomainName.baseAddress, &cchReferencedDomainName, &eUse) else {
+                            fatalError("unable to lookup SID for user \(user)")
+                        }
+
+                        var pwszSID: LPWSTR? = nil
+                        guard ConvertSidToStringSidW(pSID.baseAddress, &pwszSID) else {
+                            fatalError("unable to convert SID to string for user \(user)")
+                        }
+
+                        return #"SOFTWARE\Microsoft\Windows NT\CurrentVersion\ProfileList\#\(String(decodingCString: pwszSID!, as: UTF16.self))"#.withCString(encodedAs: UTF16.self) { pwszKeyPath in
+                            return "ProfileImagePath".withCString(encodedAs: UTF16.self) { pwszKey in
+                                var cbData: DWORD = 0
+                                guard RegGetValueW(HKEY_LOCAL_MACHINE, pwszKeyPath, pwszKey, RRF_RT_REG_SZ, nil, nil, &cbData) == ERROR_SUCCESS else {
+                                    fatalError("unable to query ProfileImagePath for user \(user)")
+                                }
+                                return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(cbData)) { pwszData in
+                                    guard RegGetValueW(HKEY_LOCAL_MACHINE, pwszKeyPath, pwszKey, RRF_RT_REG_SZ, nil, pwszData.baseAddress, &cbData) == ERROR_SUCCESS else {
+                                        fatalError("unable to query ProfileImagePath for user \(user)")
+                                    }
+                                    return String(decodingCString: pwszData.baseAddress!, as: UTF16.self)
+                                }
+                            }
+                        }
+
+                    }
+                }
+            }
+        }
+
+        var hToken: HANDLE? = nil
+        guard OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken) else {
+            guard let UserProfile = GetUserProfile() else {
+                fatalError("unable to evaluate `%UserProfile%`")
+            }
+            return UserProfile
+        }
+        defer { CloseHandle(hToken) }
+
+        var dwcchSize: DWORD = 0
+        _ = GetUserProfileDirectoryW(hToken, nil, &dwcchSize)
+
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwcchSize)) {
+            var dwcchSize: DWORD = DWORD($0.count)
+            guard GetUserProfileDirectoryW(hToken, $0.baseAddress, &dwcchSize) else {
+                fatalError("unable to query user profile directory")
+            }
+            return String(decoding: $0, as: UTF16.self)
+        }
+#else
         #if targetEnvironment(simulator)
         if user == nil, let envValue = getenv("CFFIXED_USER_HOME") ?? getenv("HOME") {
             return String(cString: envValue).standardizingPath
@@ -169,6 +244,7 @@ extension String {
         
         // If all else fails, log and fall back to /var/empty
         return "/var/empty"
+#endif
     }
     
     // From swift-corelibs-foundation's NSTemporaryDirectory. Internal for now, pending a better public API.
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -373,6 +392,24 @@ extension String {
     
     func _resolvingSymlinksInPath() -> String? {
         guard !isEmpty else { return nil }
+
+#if os(Windows)
+        return self.withNTPathRepresentation {
+            let hFile: HANDLE = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, nil)
```
jmschonfeld, line 398: Just to make sure I have the right understanding, will `CreateFileW` actually create the file on disk if it does not exist, or is it just creating a handle to a file that may or may not exist?

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -137,6 +137,24 @@ extension String {
...
+            return ""
+        }
+        defer { CloseHandle(hToken) }
+
+        var dwcchSize: DWORD = 0
+        _ = GetUserProfileDirectoryW(hToken, nil, &dwcchSize)
+
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwcchSize)) {
+            var dwcchSize: DWORD = DWORD($0.count)
+            guard GetUserProfileDirectoryW(hToken, $0.baseAddress, &dwcchSize) else {
```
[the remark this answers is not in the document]
compnerd: Yes, I would expect it to be. It is not just reading the environment.

Sources/FoundationEssentials/String/String+Path.swift
compnerd: This particular call will _not_ create the file. The disposition (`OPEN_EXISTING`) controls the behaviour here.

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -137,6 +137,24 @@ extension String {
...
     internal static func homeDirectoryPath(forUser user: String? = nil) -> String {
+#if os(Windows)
+        var hToken: HANDLE? = nil
+        guard OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken) else {
+            return ""
+        }
+        defer { CloseHandle(hToken) }
+
+        var dwcchSize: DWORD = 0
+        _ = GetUserProfileDirectoryW(hToken, nil, &dwcchSize)
```
jmschonfeld, line 148: It looks like we're ignoring the provided `user` value here - do we want to support that? I see that the CF source within swift-corelibs-Foundation on windows has a length C implementation to lookup another user's home directory at https://github.com/apple/swift-corelibs-foundation/blob/77af9c5a984f40e186c0ba127cff88f2a2588c15/CoreFoundation/Base.subproj/CFPlatform.c#L619, maybe we should port that to Swift here?

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -137,6 +137,24 @@ extension String {
...
+
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwcchSize)) {
+            var dwcchSize: DWORD = DWORD($0.count)
+            guard GetUserProfileDirectoryW(hToken, $0.baseAddress, &dwcchSize) else {
+                return ""
```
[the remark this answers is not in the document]
jmschonfeld: Gotcha, I see in SCF we fallback to the process' current working directory (https://github.com/apple/swift-corelibs-foundation/blob/77af9c5a984f40e186c0ba127cff88f2a2588c15/CoreFoundation/Base.subproj/CFPlatform.c#L542) - should we do the same here? We could just return `FileManager.default.currentDirectoryPath` here.

As for the environment variables, gotcha is `USERPROFILE` a commonly used environment variable for this purpose on Windows? It looks like the SCF implementation in C use the value of `HOMEPATH` appended to `HOMEDRIVE` as a fallback if set, is it worth doing the same here?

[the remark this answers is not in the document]
compnerd: Hmm, that is a good fallback option. We could also consider a temporary directory.

`USERPROFILE` is the equivalent to `HOME` really. `HOMEDRIVE` is _just_ the drive component and `HOMEPATH` is the drive relative path to the user profile. I would prefer to use `USERPROFILE` as that is more "modern" (`HOMEDRIVE`, `HOMEPATH` are NT4-era variables which are no longer considered "standard") and should account for local and roaming profiles if the user is an Active Directory environment. If you reference https://learn.microsoft.com/en-us/windows/win32/shell/knownfolderid?redirectedfrom=MSDN - the known folders also reference `%UserProfile%`.

Sources/FoundationEssentials/String/String+Path.swift
compnerd: Oh, yeah, if we need to be able to query another user's home directory, that would need to be done. I forgot that I had written that code in C 😂

Sources/FoundationEssentials/String/String+Path.swift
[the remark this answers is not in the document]
jmschonfeld: Thanks for the explanation, in that case yeah I think using `USERPROFILE` as a potential fallback if set makes sense to me

Sources/FoundationEssentials/String/String+Path.swift
jmschonfeld: Yeah it's not super common (it's really used to support the `FileManager.homeDirectory(forUser:)` API + standardizing paths that look like `~userName/foo`) but something we'll want to support here in addition to the current user (when the parameter is `nil`).

Haha yeah I assumed it was probably you who wrote that in C back in the day, hopefully it can be ported to Swift without too much extra leg work

Sources/FoundationEssentials/String/String+Path.swift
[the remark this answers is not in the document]
compnerd: I believe that I have done this the best I can do.

Sources/FoundationEssentials/String/String+Path.swift
compnerd: It is patently ugly, but I think it should work?

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -137,6 +137,72 @@ extension String {
...
+                        }
+
+                    }
+                }
+            }
+        }
+
+        var hToken: HANDLE? = nil
+        guard OpenProcessToken(GetCurrentProcess(), TOKEN_QUERY, &hToken) else {
+            return "USERPROFILE".withCString(encodedAs: UTF16.self) { pwszVariable in
```
jmschonfeld, line 181: I think we probably want to take the `USERPROFILE` path as a potential fallback in all of the failure cases and not just this one, right? Should we factor this out into a function that we can call from each failure case?

Sources/FoundationEssentials/String/String+Path.swift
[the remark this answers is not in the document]
jmschonfeld: Thanks! Overall it looks good, but I added a comment above about using the fallback in more places. Do we want to use the current directory as the final fallback instead of the empty string too?

Sources/FoundationEssentials/String/String+Path.swift
compnerd: Well, the failure case itself can fail too. So we can only go so far with that.

jmschonfeld: Agreed, if looking up this environment value fails then we need to have something to fallback to (like the empty string), but it feels somewhat arbitrary to only fallback to this `USERPROFILE` value in this one case. In the non-Windows path, we always check `HOME` before falling back to `/var/empty` if `HOME` isn't set, so it seems like if we want consistent behavior we should always check `USERPROFILE` before falling back to something like the empty string, thoughts?

Sources/FoundationEssentials/String/String+Path.swift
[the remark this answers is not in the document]
jmschonfeld: That's fair, I mainly suggested it because that's what we do today in swift-corelibs-foundation - even if the program exists in a privileged location it's still entirely possible that the current working directory is a reasonable path, but I agree if we don't think that's satisfactory we don't need to replicate it here. I am concerned however about using the empty string as I feel this is prone to fail later down the line. If we have no reasonable fallback path to use, should we instead just `fatalError` here to make it clear what the problem is? Otherwise, my suspicion is that this scenario will just turn into some other function throwing because it doesn't know how to operate on an empty path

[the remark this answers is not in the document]
compnerd: I'm totally fine with a `fatalError`. We can always try to do alternative fallbacks if we hit these in practice.

Sources/FoundationEssentials/String/String+Path.swift
compnerd: Yeah, always falling back to `USERPROFILE` if no user is specified seems reasonable.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@jmschonfeld I think that this will unblock some additional further work towards FileManager support for Windows.

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -137,6 +137,81 @@ extension String {
...
+        }
+        defer { CloseHandle(hToken) }
+
+        var dwcchSize: DWORD = 0
+        _ = GetUserProfileDirectoryW(hToken, nil, &dwcchSize)
+
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwcchSize)) {
+            var dwcchSize: DWORD = DWORD($0.count)
+            guard GetUserProfileDirectoryW(hToken, $0.baseAddress, &dwcchSize) else {
+                fatalError("unable to query user profile directory")
```
jmschonfeld, line 210: Should we fallback to `USERPROFILE` here too?

jmschonfeld (commented):
One small comment, but otherwise I think this looks fine. If we end up hitting the `fatalError`s in practice we can come back and decide what a better behavior would be in that concrete scenario

Sources/FoundationEssentials/String/String+Path.swift
compnerd: I didn't fallback here because we just called it above to get the required buffer size. If that succeeded and this failed, I think we are in some weird state and a `fatalError` feels safer.

jmschonfeld: Ah good point thanks for the explanation

jmschonfeld (approved):

PR #548, opened 2024-04-16
FoundationEssentials: implement file diffing support for Windows

compnerd:
Windows does not have the same Unix file system APIs (e.g. FTS). Implement a platform specific implementation to compare files on Windows.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
@@ -16,6 +16,25 @@ import Darwin
 import Glibc
 #elseif os(Windows)
 import CRT
+import WinSDK
+#endif
+
+#if os(Windows)
+extension _FILE_ID_128 /* : @retroactive Equatable */ {
+    internal static func _equals(_ lhs: _FILE_ID_128, _ rhs: _FILE_ID_128) -> Bool {
+        return withUnsafeBytes(of: lhs.Identifier) { pLHS in
+            return withUnsafeBytes(of: rhs.Identifier) { pRHS in
+                return memcmp(pLHS.baseAddress, pRHS.baseAddress, MemoryLayout.size(ofValue: lhs.Identifier)) == 0
+            }
+        }
+    }
+}
+
+extension LARGE_INTEGER /* : @retroactive Equatable */ {
+    internal static func _equals(_ lhs: LARGE_INTEGER, _ rhs: LARGE_INTEGER) -> Bool {
+        return lhs.QuadPart == rhs.QuadPart
+    }
+}
 #endif
 
 internal struct _FileManagerImpl {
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@jmschonfeld anything holding this up?

compnerd:
@parkera @jmschonfeld - would be nice to get this going. There are a few pending changes for `FileManager` and one that I am still working on. The sum of those would however seem to be sufficient to build `FileManager` on Windows. That would allow me to focus on the next few items blocking building `FoundationEssentials` on Windows. Although that would still not allow the full package to build, it would be a great milestone.

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -16,6 +16,28 @@ import Darwin
 import Glibc
 #elseif os(Windows)
 import CRT
+import WinSDK
+#endif
+
+#if os(Windows)
+extension _FILE_ID_128: Equatable {
+    public static func == (_ lhs: _FILE_ID_128, _ rhs: _FILE_ID_128) -> Bool {
+        return withUnsafePointer(to: lhs.Identifier.0) { pLHS in
```
jmschonfeld, line 25: Is `lhs.Identifier` a tuple of size 16? If so I think you'll likely need `withUnsafePointer(to: lhs.Identifier)` instead of the pointer to `.0` - `withUnsafePointer` only guarantees you a pointer to what you provide it with so getting a pointer to the first element in the tuple doesn't guarantee that the remaining elements will follow, but getting a pointer to the tuple itself should work fine

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -16,6 +16,28 @@ import Darwin
...
+            return withUnsafePointer(to: rhs.Identifier.0) { pRHS in
+                let pRHSBuffer = UnsafeBufferPointer(start: pRHS, count: 16)
+
+                return pLHSBuffer.elementsEqual(pRHSBuffer)
+            }
+        }
+    }
+}
+
+extension LARGE_INTEGER: Equatable {
```
jmschonfeld, line 36: These will add public conformances to these types for all clients of `FoundationEssentials` - it seems like this might not be the right place for such an API. Should we downgrade these to `internal static` functions for now until we determine where the best place for such an API should be?

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -60,6 +82,134 @@ internal struct _FileManagerImpl {
...
+                }
+
+                var fbiRHS: FILE_BASIC_INFO = .init()
+                guard GetFileInformationByHandleEx(hRHS, FileBasicInfo, &fbiRHS, DWORD(MemoryLayout.size(ofValue: fbiRHS))) else {
+                    return false
+                }
+
+                if fbiLHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT,
+                   fbiRHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT {
+                    return (try? destinationOfSymbolicLink(atPath: path) == destinationOfSymbolicLink(atPath: other)) ?? false
```
jmschonfeld, line 133:
```suggestion
                    return (try? fileManager.destinationOfSymbolicLink(atPath: path) == fileManager.destinationOfSymbolicLink(atPath: other)) ?? false
```

This will allow deferring to subclasses when present

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
compnerd: Yes, its a tuple of size 16. Okay, I'll change that to `lhs.Identifier` then.

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
compnerd: Sure we can do that. I don't mind pushing this into the WinSDK overlay - the conformances are kinda nice. The problem is that we won't be able to use them until we get an updated snapshot.

jmschonfeld: Sounds good, yeah if there's a better home for these we could add it there and add internal only functions for now, and then just switch over to the conformances once we get a newer snapshot

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -16,6 +16,28 @@ import Darwin
 import Glibc
 #elseif os(Windows)
 import CRT
+import WinSDK
+#endif
+
+#if os(Windows)
+extension _FILE_ID_128 /* : @retroactive Equatable */ {
+    internal static func == (_ lhs: _FILE_ID_128, _ rhs: _FILE_ID_128) -> Bool {
```
jmschonfeld, line 24: Will this cause any ambiguity if in the future something from below Foundation adds `_FILE_ID_128.==` as part of an equatable conformance? Just want to double check that we don't need to name this something like `_equals` just so that we don't conflict in that probable future

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
jmschonfeld: Looks like this one still needs to be accounted for

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
compnerd: Sure, I can rename this to `_equals`.

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
jmschonfeld: I think this change is still missing, but other than that this looks good to me!

compnerd: :sigh: I think that I might've applied it and then lost it. Thanks for the diligence in the review!

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -60,6 +79,134 @@ internal struct _FileManagerImpl {
...
+                    return (try? destinationOfSymbolicLink(atPath: path) == destinationOfSymbolicLink(atPath: other)) ?? false
+                } else if fbiLHS.FileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY,
+                          fbiRHS.FileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY {
+                    // We immediately convert the contents to a set to allow us
+                    // to compare the two sets in a single pass. This is
+                    // computationally less expensive (O(n) vs O(n^2)) than
+                    // comparing each item in the array.
+                    guard let aLHSItems = try? Set(fileManager.contentsOfDirectory(atPath: path)),
+                          let aRHSItems = try? Set(fileManager.contentsOfDirectory(atPath: other)),
+                          aLHSItems.count == aRHSItems.count, aRHSItems.intersection(aLHSItems) == aLHSItems else {
```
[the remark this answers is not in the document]
jmschonfeld: Actually, I think the full `aLHSItems.count == aRHSItems.count, aRHSItems.intersection(aLHSItems) == aLHSItems` is just `aLHSItems == aRHSItems`, right?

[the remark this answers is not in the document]
compnerd: Oh, right, the counts being identical means that yes, this is equality!

jmschonfeld (approved):
Thanks for iterating on this, this looks good to me!

PR #549, opened 2024-04-16
Remove SPI annotations from Expression

jmschonfeld:
Now that the previous build issues are resolved, we can remove the `@_spi` annotations from `Expression` and its related types


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #551, opened 2024-04-17
Fix AttributedString.UnicodeScalarView.distance(from:to:) availability

jmschonfeld:
In macOS 14 aligned releases we added a "back deployed" `AttributedString.UnicodeScalarView.distance(from:to:)` function. However, the availability check in the emitted implementation attempts to call the entrypoint within Foundation on macOS 13 aligned releases due to an incorrect availability annotation. This updates the annotation so that building against an SDK that contains this change and back deploying to macOS 13 no longer causes a crash at runtime.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #554, opened 2024-04-19
FoundationEssentials: support directory enumeration operations on Windows

compnerd:
Implement the directory enumeration operations in FileManager on Windows. The implementation uses the Win32 API surface to ensure that we can use the full features of the platform.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -51,7 +51,7 @@ extension _FileManagerImpl {
         URL(filePath: String.temporaryDirectoryPath, directoryHint: .isDirectory)
     }
     
-    #if FOUNDATION_FRAMEWORK
+#if FOUNDATION_FRAMEWORK
     func url(
         for directory: FileManager.SearchPathDirectory,
         in domain: FileManager.SearchPathDomainMask,
@@ -137,7 +137,11 @@ extension _FileManagerImpl {
         }
         #endif
         var result: [String] = []
+#if os(Windows)
+        let iterator = _Win32DirectoryContentsSequence(path: path, appendSlashForDirectory: false).makeIterator()
+#else
         let iterator = _POSIXDirectoryContentsSequence(path: path, appendSlashForDirectory: false).makeIterator()
+#endif
         if let error = iterator.error {
             throw error
         } else {
@@ -149,6 +153,31 @@ extension _FileManagerImpl {
     }
     
     func subpathsOfDirectory(atPath path: String) throws -> [String] {
+#if os(Windows)
+        var results: [String] = []
+        let iterator = _Win32DirectoryContentsSequence(path: path, appendSlashForDirectory: true).makeIterator()
+        while let item = iterator.next() {
+            results.append(item.fileName)
+            if item.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY &&
+                item.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT != FILE_ATTRIBUTE_REPARSE_POINT {
+
+                var pwszSubPath: PWSTR? = nil
+                let hr = PathAllocCombine(path, item.fileName, PATHCCH_ALLOW_LONG_PATHS, &pwszSubPath)
+                guard hr == S_OK else {
+                    throw CocoaError.errorWithFilePath(path, win32: WIN32_FROM_HRESULT(hr), reading: true)
+                }
+                defer { LocalFree(pwszSubPath) }
+
+                results.append(contentsOf: try subpathsOfDirectory(atPath: item.fileName).map {
+                    var pwszFullPath: PWSTR? = nil
+                    _ = PathAllocCombine(item.fileName, $0, PATHCCH_ALLOW_LONG_PATHS, &pwszFullPath)
+                    defer { LocalFree(pwszFullPath) }
+                    return String(decodingCString: pwszFullPath!, as: UTF16.self).standardizingPath
+                 })
+            }
+        }
+        return results
+#else
         return try path.withFileSystemRepresentation { fileSystemRep in
             guard let fileSystemRep else {
                 throw CocoaError.errorWithFilePath(.fileNoSuchFile, path)
@@ -181,6 +210,7 @@ extension _FileManagerImpl {
             }
             return results
         }
+#endif
     }
     
     #if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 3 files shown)


Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
```diff
@@ -10,7 +10,92 @@
...
+        private var ffdData: WIN32_FIND_DATAW = .init()
+        private var prefix: String = ""
+        private var slash: Bool
+
+        var error: CocoaError?
+
+        init(path: String, appendSlashForDirectory: Bool, prefix: [String]) {
+            self.slash = appendSlashForDirectory
+
+            hFind = try? "\(path)\\*".withNTPathRepresentation {
```
jmschonfeld, line 35: Should we just catch the error from here and set it to `error` so that we produce an error with the most detailed/correct info?

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -157,4 +161,38 @@ package var VOLUME_NAME_DOS: DWORD {
...
 
+package func PathAllocCombine(_ pszPathIn: String, _ pszMore: String, _ dwFlags: ULONG, _ ppszPathOut: UnsafeMutablePointer<PWSTR?>?) -> HRESULT {
+    pszPathIn.withCString(encodedAs: UTF16.self) { pszPathIn in
+        pszMore.withCString(encodedAs: UTF16.self) { pszMore in
+            WinSDK.PathAllocCombine(pszPathIn, pszMore, dwFlags, ppszPathOut)
+        }
+    }
+}
+
+@_transparent
```
jmschonfeld, line 172: Is there a reason to mark these as `@_transparent` rather than letting the compiler make the inlining optimization decisions on its own?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -207,6 +234,35 @@ extension _FileManagerImpl {
         withIntermediateDirectories createIntermediates: Bool,
         attributes: [FileAttributeKey : Any]? = nil
     ) throws {
+#if os(Windows)
+        try path.withNTPathRepresentation { pwszPath in
+            if createIntermediates {
+                var isDirectory: Bool = false
+                if fileExists(atPath: path, isDirectory: &isDirectory) {
```
jmschonfeld, line 241:
```suggestion
                if fileManager.fileExists(atPath: path, isDirectory: &isDirectory) {
```

This will allow subclasses to override the behavior of this `fileExists` call

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -207,6 +234,35 @@ extension _FileManagerImpl {
...
+                    guard isDirectory else {
+                        throw CocoaError.errorWithFilePath(path, win32: ERROR_FILE_EXISTS, reading: false)
+                    }
+                    return
+                }
+
+                let parent = path.deletingLastPathComponent()
+                if !parent.isEmpty {
+                    try createDirectory(atPath: parent, withIntermediateDirectories: true, attributes: attributes)
```
jmschonfeld, line 250: This pattern is really similar to the non-`FOUNDATION_FRAMEWORK` pattern below. Do you think there's any way to coalesce the two together into a portable solution (I think the only difference is really `withFileSystemRepresentation` vs. `withNTPathRepresentation` and `mkdir` vs. `CreateDirectoryW`). It might not be doable, but if for this one we can just if-def around those calls maybe we can make this a single implementation rather than an effective "fork"?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -149,6 +153,28 @@ extension _FileManagerImpl {
...
+                    throw CocoaError.errorWithFilePath(path, win32: WIN32_FROM_HRESULT(hr), reading: true)
+                }
+                results.append(contentsOf: try subpathsOfDirectory(atPath: item.fileName).map {
+                    var pwszFullPath: PWSTR? = nil
+                    _ = PathAllocCombine(item.fileName, $0, PATHCCH_ALLOW_LONG_PATHS, &pwszFullPath)
+                    return String(decodingCString: pwszFullPath!, as: UTF16.self).standardizingPath
+                 })
+            }
+        }
+        return results
```
jmschonfeld, line 176: Similar comment as below, is there any way for us to implement a `.subpaths` var on `_Win32DirectoryContentsSequence` so that the implementation of dealing with the subpaths themselves could be a single implementation?

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
compnerd: Sure.

Sources/FoundationEssentials/WinSDK+Extensions.swift
compnerd: These are really supposed to be transparent - they should be imported through the clang importer, but it doesn't understand complex macros.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
compnerd: I tried to do that but found myself quickly annoyed with the `#if` conditions.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
compnerd: I could of course move the iteration into a `subpaths` property and then use that to "simplify" `subpathsOfDirectory(atPath:)` to `_Win32DirectoryContentsSequence(...).subpaths`. But that would then also suggest that we should sink the non-Windows path into `_FTSSquence` no?

compnerd:
@swift-ci please test

Sources/FoundationEssentials/WinSDK+Extensions.swift
jmschonfeld: I spoke with @lorentey about this and after touching base with him I think we should use `@inline(__always)` instead of `@_transparent` here. I agree that we shouldn't have to incur a function call for these here, but `@_transparent` also inherently adds `@inlinable` to these declarations which means their implementations could accidentally leak into our interface if we accidentally included them in other inlinable code. `@inline(__always)` should get us the same performance wins here given that these are `internal` declarations without carrying the risk of accidentally letting this implementation detail leak into the interface without an intentional decision around that

compnerd: Okay, seems reasonable enough.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jmschonfeld: `_FTSSequence` has a `subpaths` property that returns a `Collection<String>` which we then apply logic to here (for surfacing errors and trimming the prefix). I was wondering if it was possible to implement a similar `subpaths` property on `_Win32DirectoryContentsSequence` so that the logic below (lines 190-210) could be shared between the two implementations to ensure equivalent behavior. Is that possible? Or are the APIs too different to do this

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jmschonfeld: Ok that's fair if the `#if` conditions become too unruly that way then the separate block is probably cleaner - it's just nice to maintain one implementation wherever we can so that the behavior is the same

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
compnerd: Oh, the APIs are completely different. It isn't a full enumerated sequence, instead we have to manually do the BFS to get the tree.

jmschonfeld: The non-`FOUNDATION_FRAMEWORK` path below also takes a BFS-style approach since we don't have a full enumeration there either - is that one compatible enough to adapt here?

jmschonfeld: Ah I see now gotcha yeah the semantics are a bit different then. Ok no worries

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
compnerd: Absolutely, I prefer to have a single path whenever possible.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

jmschonfeld:
CI is currently broken due to an SPM issue that will be resolved in tonight's toolchain, but I've validated that this doesn't break the macOS/linux build myself. Merging this to keep making forward progress on the Windows updates

PR #556, opened 2024-04-22
FoundationEssentials: enable unprivilleged symlinks if possible

compnerd:
If developer mode is enabled or the user is assigned the SeCreateSymbolicLinkPrivilege privilege, permit the unelevated symbolic link creation.


parkera (commented):
Is this a behavior most clients would expect, or does it need to be propagated out as some kind of option from Foundation?

compnerd:
I think that anyone coming from the Unix would would expect this. On the Windows world, I think that anything who enabled developer mode would expect this. However, since it would not impact anyone on Windows who does not have developer mode or symlink creation privilege explicitly assigned, it would be pretty expected I would say.

compnerd:

compnerd:

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -145,6 +145,18 @@ package var PAGE_READONLY: DWORD {
     DWORD(WinSDK.PAGE_READONLY)
 }
 
+package var PATHCCH_ALLOW_LONG_PATHS: ULONG {
+    ULONG(WinSDK.PATHCCH_ALLOW_LONG_PATHS.rawValue)
+}
+
+package var RRF_RT_REG_SZ: DWORD {
```
jmschonfeld, line 152: It looks like this declaration and the one above aren't used in this change. Are they necessary?

compnerd: They are used in subsequent PRs that I have in my tree. The PRs take a while to get merged and the overloads here are just difficult to keep separated.

jmschonfeld:

jmschonfeld (approved):

PR #557, opened 2024-04-23
Implement ASCII encoding for String.data(using:)

jmschonfeld:
This PR implements `String` to `Data` conversion for the `.ascii` and `.nonLossyAscii` encodings in Swift

Resolves rdar://123921236

### DIFF
--- a/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
+++ b/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
@@ -43,7 +43,7 @@ extension String {
                 var data = Data(count: self.utf8.count)
                 let allASCII = data.withUnsafeMutableBytes {
                     $0.withMemoryRebound(to: UInt8.self) { buffer in
-                        buffer.initializeAll(fromContentsOf: self.utf8)
+                        _ = buffer.initialize(fromContentsOf: self.utf8)
                         if let earlyCheckAllASCII {
                             return earlyCheckAllASCII
                         } else {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
```diff
@@ -21,6 +21,28 @@ extension String {
...
+                            $0.appendElement(lossyReplacement)
+                        }
+                    }
+                }
+            } else {
+                var data = Data(count: self.utf8.count)
+                let allASCII = data.withUnsafeMutableBytes {
+                    $0.withMemoryRebound(to: UInt8.self) { buffer in
+                        buffer.initializeAll(fromContentsOf: self.utf8)
+                        return _allASCII(UnsafeBufferPointer(buffer))
```
parkera, line 41: I'm not sure about this, but since we have to do this check every time - does it make sense to do it first? Then, if it fails, we do no allocation or copy.

jmschonfeld: I can move this check beforehand if the `String` has a contiguous buffer available to perform the check on. In the case where the `String` doesn't have a contiguous storage to perform this check (i.e. it's a lazily bridged string from ObjC) - it's probably best to perform the check afterwards since performing the check beforehand will need to transcode everything anyways

jmschonfeld: Added the early check (when we have a contiguous buffer) in e45250b

parkera: Makes sense, thanks.

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

parkera (approved):

parkera:
Thanks!

jmschonfeld:
@swift-ci please test

PR #558, opened 2024-04-23
Correct non-Darwin PropertyListFormat raw values

jmschonfeld:
This aligns the raw values of this enum with those on Darwin and those in swift-corelibs-foundation which will make swift-corelibs-foundation adoption of `PropertyListDecoder` easier


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
CI is broken due to an SPM issue, but I've validated that this builds/passes tests locally and this is a minimal change that shouldn't affect behavior

PR #559, opened 2024-04-25
Add ProcessInfo implementations for other platforms

jmschonfeld:
This adds the implementations of `ProcessInfo` APIs for other platforms such as Windows and WASI (copied from swift-corelibs-foundation) so that this type can fully replace the `ProcessInfo` present in swift-corelibs-foundation

### DIFF
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -262,23 +262,24 @@ extension _ProcessInfo {
         // do our best here to construct something consistent. Unfortunately, to provide a useful result, this requires
         // hardcoding several of the somewhat ambiguous values in the table provided here:
         //  https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_osversioninfoexw#remarks
-        switch (osVersionInfo.dwMajorVersion, osVersionInfo.dwMinorVersion) {
-            case (5, 0): versionString += " 2000"
-            case (5, 1): versionString += " XP"
-            case (5, 2) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " XP Professional x64"
-            case (5, 2) where osVersionInfo.wSuiteMask == VER_SUITE_WH_SERVER: versionString += " Home Server"
-            case (5, 2): versionString += " Server 2003"
-            case (6, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " Vista"
-            case (6, 0): versionString += " Server 2008"
-            case (6, 1) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " 7"
-            case (6, 1): versionString += " Server 2008 R2"
-            case (6, 2) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " 8"
-            case (6, 2): versionString += " Server 2012"
-            case (6, 3) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " 8.1"
-            case (6, 3): versionString += " Server 2012 R2" // We assume the "10,0" numbers in the table for this are a typo
-            case (10, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: versionString += " 10"
-            case (10, 0): versionString += " Server 2019" // The table gives identical values for 2016 and 2019, so we just assume 2019 here
-            case let (maj, min): versionString += " \(maj).\(min)" // If all else fails, just give the raw version number
+        versionString += " "
+        versionString += switch (osVersionInfo.dwMajorVersion, osVersionInfo.dwMinorVersion) {
+            case (5, 0): "2000"
+            case (5, 1): "XP"
+            case (5, 2) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "XP Professional x64"
+            case (5, 2) where osVersionInfo.wSuiteMask == VER_SUITE_WH_SERVER: "Home Server"
+            case (5, 2): "Server 2003"
+            case (6, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "Vista"
+            case (6, 0): "Server 2008"
+            case (6, 1) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "7"
+            case (6, 1): "Server 2008 R2"
+            case (6, 2) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "8"
+            case (6, 2): "Server 2012"
+            case (6, 3) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "8.1"
+            case (6, 3): "Server 2012 R2" // We assume the "10,0" numbers in the table for this are a typo
+            case (10, 0) where osVersionInfo.wProductType == VER_NT_WORKSTATION: "10"
+            case (10, 0): "Server 2019" // The table gives identical values for 2016 and 2019, so we just assume 2019 here
+            case let (maj, min): "\(maj).\(min)" // If all else fails, just give the raw version number
         }
         versionString += " (build \(osVersionInfo.dwBuildNumber))"
         // For now we ignore the `szCSDVersion`, `wServicePackMajor`, and `wServicePackMinor` values.
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -103,7 +103,7 @@ final class _ProcessInfo: Sendable {
 
     var globallyUniqueString: String {
         let uuid = UUID().uuidString
-        let pid = UInt64(getpid())
+        let pid = processIdentifier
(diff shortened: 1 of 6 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -140,9 +144,28 @@ final class _ProcessInfo: Sendable {
             return username
         }
         return ""
-#else
-        // TODO: Windows
+#elseif os(WASI)
+        // WASI does not have user concept
         return ""
+#elseif os(Windows)
+        return withUnsafeTemporaryAllocation(of: wchar_t.self, capacity: 1040) { usernameBuffer in
```
compnerd, line 151: Is this meant to be the user that the process is running as or whatever user is logged in? I wonder if we should query the information from the process.

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -182,17 +213,117 @@ extension _ProcessInfo {
...
+        
+        guard let osVersionInfo = self._rawOperatingSystemVersionInfo else {
+            return versionString
+        }
+
+        // Windows has no canonical way to turn the fairly complex `RTL_OSVERSIONINFOW` version info into a string. We
+        // do our best here to construct something consistent. Unfortunately, to provide a useful result, this requires
+        // hardcoding several of the somewhat ambiguous values in the table provided here:
+        //  https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/wdm/ns-wdm-_osversioninfoexw#remarks
+        switch (osVersionInfo.dwMajorVersion, osVersionInfo.dwMinorVersion) {
```
compnerd, line 265: Might be nice to use the switch expressions that were recently introduced to convert the tuple to a string and append once rather than in each case.

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -276,8 +424,20 @@ extension _ProcessInfo {
...
+#elseif os(Windows)
+        var sysInfo = SYSTEM_INFO()
+        GetSystemInfo(&sysInfo)
+        let activeProcessorMask = sysInfo.dwActiveProcessorMask
+        // assumes sizeof(DWORD_PTR) is 64 bits or less
+        var v = activeProcessorMask
+        v = v - ((v >> 1) & 0x5555555555555555)
+        v = (v & 0x3333333333333333) + ((v >> 2) & 0x3333333333333333)
+        v = (v + (v >> 4)) & 0xf0f0f0f0f0f0f0f
+        return (v * 0x0101010101010101) >> ((sizeof(v) - 1) * 8)
```
compnerd, line 438: This is just popcount isn't it? We are in swift, so why not just do `sysInfo.dwActiveProcessMask.nonzeroBitCount`? That should lower to popcount.

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -117,7 +117,11 @@ final class _ProcessInfo: Sendable {
     }
 
     var processIdentifier: Int32 {
```
compnerd, line 119: Is there a reason to make this `Int32` when it is almost universally `UInt32`?

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
jmschonfeld: I believe it should be the "current user" (i.e. whichever user is logged in) based on what I've heard from @iCharlesHu - this code was mostly just copied from swift-corelibs-foundation as-is so I think it should behave the same as what SCF currently does

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
jmschonfeld: The public `ProcessInfo.processIdentifier` API today is of the type `Int32` so we need to maintain that type here for source compatibility (and ABI compatibility on Darwin)

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -117,7 +117,11 @@ final class _ProcessInfo: Sendable {
...
-        return getpid()
+#if os(Windows)
+        return Int32(GetProcessId(GetCurrentProcess()))
```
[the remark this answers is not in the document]
jmschonfeld: Yes it will trap if high bits are set. Is it possible for Windows to return a process ID with high bits set? (on Darwin `pid_t` is defined as a 32-bit signed integer). If so, is there a behavior here that you think would be better? (I believe in SCF it will trap today if the high bits are set)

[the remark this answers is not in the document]
compnerd: Yeah, it is. The ProcessId is a DWORD (UInt32). I guess we could do a `Int32(bitPattern:)` initializer?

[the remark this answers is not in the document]
jmschonfeld: We could, but that would be a behavioral change from SCF today. Do you think that change / this new behavior would be expected for Windows clients of Foundation?

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
jmschonfeld: It is! Good call, updated 👍🏻

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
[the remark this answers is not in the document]
compnerd: I think that it would be more expected than a trap which would not make sense - I didn't create an `Int32`, why is this trapping?

compnerd:
BTW, I do have some changes related to this area that I've put up as a draft at https://github.com/apple/swift-foundation/pull/561 I'll see if I can pull them over here as your change is more comprehensive.

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -261,8 +408,10 @@ extension _ProcessInfo {
             return 0
         }
         return Int(count)
-#else
+#elseif os(Linux) || os(FreeBSD)
```
compnerd, line 411:
```suggestion
#elseif os(Windows)
        var siInfo = SYSTEM_INFO()
        GetSystemInfo(&siInfo)
        return Int(siInfo.dwNumberOfProcessors)
#elseif os(Linux) || os(FreeBSD)
```

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -293,10 +514,12 @@ extension _ProcessInfo {
             }
             return 0
         }
-#else
+#elseif os(Linux) || os(FreeBSD)
```
compnerd, line 517:
```suggestion
#elseif os(Windows)
        var TotalMemoryKB: ULONGLONG = 0
        guard GetPhysicallyInstalledSystemMemory(&TotalMemoryKB) else {
            return 0
        }
        return TotalMemoryKB * 1024
#elseif os(Linux) || os(FreeBSD)
```

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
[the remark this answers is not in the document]
jmschonfeld: Ok I can update that to use a bit pattern then if you feel that'd be the better, more expected behavior for our windows clients

jmschonfeld:
> BTW, I do have some changes related to this area that I've put up as a draft at #561 I'll see if I can pull them over here as your change is more comprehensive.

Sounds good! I accepted your two changes above and the other changes make sense to pull in - let me know if you want me to pull them in myself or if you're planning to post suggestions to do it

compnerd:
> > BTW, I do have some changes related to this area that I've put up as a draft at #561 I'll see if I can pull them over here as your change is more comprehensive.
> 
> Sounds good! I accepted your two changes above and the other changes make sense to pull in - let me know if you want me to pull them in myself or if you're planning to post suggestions to do it

If you could pull them in, that would be wonderful!

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows

compnerd (approved):
The windows side looks good to me!

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows

iCharlesHu (approved):
Thanks for fixing the tests!!

iCharlesHu:
@swift-ci please test Windows

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #560, opened 2024-04-25
FileManager doesn't defer to subclasses when creating symbolic links

jmschonfeld:
The `URL`-based symlink functions called directly to the `String`-based implementations without deferring to subclasses when present. This simple change defers to a possible subclass when calling to the `String`-based implementations to ensure correct behavior for subclasses that only implement the `String`-based functions and not the `URL`-based ones


jmschonfeld:
@swift-ci please test

hristost (approved):

jmschonfeld:
CI failed due to an ongoing SPM issue that should be resolved soon, but I've validated that this builds on macOS and linux myself

PR #561, opened 2024-04-26
WIP: FoundationEssentials: add Windows ProcessInfo changes

compnerd:


Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -183,9 +193,61 @@ extension _ProcessInfo {
         }
     }
 
+#if os(Windows)
+    internal var _rawOSVersion: RTL_OSVERSIONINFOEXW? {
+        typealias RtlGetVersionTy = @convention(c) (UnsafeMutablePointer<RTL_OSVERSIONINFOEXW>) -> NTSTATUS
+
+        guard let NTDLL = "ntdll.dll".withCString(encodedAs: UTF16.self, LoadLibraryW) else {
```
compnerd, line 200: This is less secure than the version that @jmschonfeld ported over from SCF. The primary difference being that DLL paths could impact the resolved library here vs the original version I had used in CF which would ensure that we pick up the version in the system. However, that also has a downside that it may not be the version of the DLL that is actually loaded which makes this a slightly better option.

PR #564, opened 2024-04-26
Use ar_SA locale for tests expecting Arab-Indic numerals

jmschonfeld:
To account for underlying changes in ICU - we now use the `ar_SA` locale here to ensure that we get the expected arabic-indic numerals in the formatted output


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #569, opened 2024-04-29
Change ProcessInfo.procssName to be lazily loaded

iCharlesHu:
Moved process name resolution to be "on demand" instead of always eagerly resolving  them in `init()` to improved app launch time.


iCharlesHu:
@swift-ci please test

PR #571, opened 2024-04-30
FoundationEssentials: clean up some Windows warnings

compnerd:
Use the Windows POSIX API spellings to avoid the warnings. Take the opportunity to change the environment handling to ensure that we are able to handle weird unicode environment variables which may not be rendered properly in the ASCII environment representation. Adjust some types to better match the implementation on Windows which is a minor uop to avoid unnecessary `trunc` and `sext` or `zext`.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -306,21 +306,25 @@ private func readBytesFromFileDescriptor(_ fd: Int32, path: PathOrURL, buffer in
         }
         
         // We will only request a max of Int32.max bytes
-        var numBytesRequested = Int32(clamping: preferredChunkSize)
+        var numBytesRequested = CUnsignedInt(clamping: preferredChunkSize)
         
         // Furthermore, don't request more than the number of bytes remaining
         if numBytesRequested > numBytesRemaining {
-            numBytesRequested = Int32(clamping: numBytesRemaining)
+            numBytesRequested = CUnsignedInt(clamping: numBytesRemaining)
         }
-        
-        var numBytesRead: Int
+
+        var numBytesRead: CInt
         repeat {
             if let localProgress, localProgress.isCancelled {
                 throw CocoaError(.userCancelled)
             }
             
             // read takes an Int-sized argument, which will always be at least the size of Int32.
-            numBytesRead = read(fd, buffer, Int(numBytesRequested))
+#if os(Windows)
+            numBytesRead = _read(fd, buffer, numBytesRequested)
+#else
+            numBytesRead = CInt(read(fd, buffer, Int(numBytesRequested)))
+#endif
         } while numBytesRead < 0 && errno == EINTR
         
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

iCharlesHu:
It looks like the Linux failure is related to this change.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@iCharlesHu - addressed

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -306,21 +306,37 @@ private func readBytesFromFileDescriptor(_ fd: Int32, path: PathOrURL, buffer in
...
+#if os(Windows)
+            numBytesRequested = CUnsignedInt(clamping: numBytesRemaining)
+#else
             numBytesRequested = Int32(clamping: numBytesRemaining)
+#endif
         }
-        
+
+#if os(Windows)
+        var numBytesRead: CInt
```
jmschonfeld, line 325: Are these necessary because `_read`'s type signature on Windows differs in integer types from the signature for `read`? If so, I wonder if it'd be better to just add a cast at the call to `_read` below to minimize the number of ifdefs and reduce the chance that we accidentally break something here due to the differing types on different platforms. What are your thoughts?

compnerd: Yeah, I was thinking that we might want to merge towards the narrower width, wdyt?

jmschonfeld: For `numBytesRequested` `CUnsignedInt` is `UInt32` so we could probably just use `Int32` everywhere but do a quick `CUnsignedInt(numBytesRequested)` in the call to `_read` below to avoid the two ifdefs above.

For `numBytesRead` here - ah yeah you're right that `CInt` is narrower than `Int`. In general I'd agree that yeah the narrower is probably better but since we never mutate the value ourselves and we end up clamping it down to `Int` on line 351 below anyways (which is probably fine since I don't think `numBytesRequested` will ever be bigger than `Int.max` (in fact it can't be since it's an `Int32`) I think just leaving it as an `Int` and doing `Int(_read(...))` below might help us make the code less likely to break the windows build if we change it in the future. But if you think that the `CInt` here would be better semantically then I'm not strongly opposed to it

compnerd: BTW, should we truncate on both sides? `size_t read(int, void *, size_t)` vs `int _read(int, void *, int)` would mean that we truncate on the in and out.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Reading.swift
compnerd: Truncated to the narrower widths to share the codepath.

jmschonfeld (approved):

PR #573, opened 2024-04-30
Introduce Cmake support for SwiftFoundation

iCharlesHu:
Add Cmake support for SwiftFoundation which includes
- Adding CMakeLists.txt in major directories;
- Adding `module.modulemap` for `_CShims`

Note: I had to change `uuid.h`'s header guard because it's somehow conflicting with system ones while using Cmake to build.

---
# Testing Notes

In order to build `SwiftFoundation` with Cmake you will need to install `cmake` and `ninja`: `brew install cmake && brew install ninja`

Currently the Cmake build for `SwiftFoundationICU` (`SwiftFoundation`'s dependency) is hitting two compiler crashes. To work around these issues, first download the nightly Swift snapshot **newer than 4/25's build**. Then add the following lines in the main `CMakeLists.txt` file, right after `cmake_minimum_required`:

```cmake
set(CMAKE_BUILD_TYPE "RelWithDebInfo")
set(CMAKE_Swift_COMPILER "/Library/Developer/Toolchains/<Toolchain you downloaded>/usr/bin/swiftc")
```

As of now, the main `CMakeLists.txt` does not fetch dependencies for you (yet). You'll have to manually checkout `SwiftFoundation`'s two dependencies:
- https://github.com/iCharlesHu/swift-foundation-icu.git (branch `charles/cmake-support`)
- https://github.com/apple/swift-collections.git (branch `release/1.1`)

Next, go to each directory and manually build them: `cmake -S . -B ./build -G Ninja && cd ./build && ninja`

Finally, add these two lines to `SwiftFoundation`'s main `CMakeLists.txt`, right below the two lines you just added:
```cmake
set(SwiftCollections_DIR "/path/to/SwiftCollections/build/cmake/modules")
set(SwiftFoundationICU_DIR "/path/to/SwiftFoundationICU/build/cmake/modules")
```
**Make sure to use the `build` folder above**. This will instruct Cmake to use the dependencies you just checked out.

Now you can finally build SwiftFoundation with Cmake:
```
cmake -S . -B ./build -G Ninja && cd ./build && ninja
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,90 @@
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift open source project
+##
+## Copyright (c) 2024 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.md for the list of Swift project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+cmake_minimum_required(VERSION 3.24)
+
+if(POLICY CMP0156)
+    # Deduplicate linked libraries where appropriate
+    cmake_policy(SET CMP0156 NEW)
+endif()
+
+if(POLICY CMP0157)
+    # New Swift build model: improved incremental build performance and LSP support
+    cmake_policy(SET CMP0157 NEW)
+endif()
+
+project(SwiftFoundation
+    LANGUAGES C Swift)
+
+if(NOT SWIFT_SYSTEM_NAME)
+  if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+    set(SWIFT_SYSTEM_NAME macosx)
+  else()
+    set(SWIFT_SYSTEM_NAME "$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>")
+  endif()
+endif()
+
+set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
+set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
+
+set(BUILD_TESTING NO)
+
+set(COLLECTIONS_SINGLE_MODULE YES)
+set(COLLECTIONS_FOUNDATION_TOOLCHAIN_MODULE YES)
+
+# Make sure our dependencies exists
+include(FetchContent)
+if (_SwiftFoundationICU_SourceDIR)
+    FetchContent_Declare(SwiftFoundationICU
+        SOURCE_DIR ${_SwiftFoundationICU_SourceDIR})
+else()
+    FetchContent_Declare(SwiftFoundationICU
+        GIT_REPOSITORY https://github.com/apple/swift-foundation-icu.git
+        GIT_TAG 0.0.8)
+endif()
+
+if (_SwiftCollections_SourceDIR)
+    FetchContent_Declare(SwiftCollections
+        SOURCE_DIR ${_SwiftCollections_SourceDIR})
+else()
+    FetchContent_Declare(SwiftCollections
+        GIT_REPOSITORY https://github.com/apple/swift-collections.git
+        GIT_TAG 1.1.1)
+endif()
+FetchContent_MakeAvailable(SwiftFoundationICU SwiftCollections)
+
+list(APPEND CMAKE_MODULE_PATH ${SwiftFoundation_SOURCE_DIR}/cmake/modules)
+
+# Availability Macros (only applies to FoundationEssentials and FoundationInternationalization)
+set(_SwiftFoundation_availability_macros)
+list(APPEND _SwiftFoundation_availability_macros
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPreview 0.1:macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPreview 0.2:macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPreview 0.3:macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPreview 0.4:macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPredicate 0.1:macOS 14, iOS 17, tvOS 17, watchOS 10\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPredicate 0.2:macOS 14, iOS 17, tvOS 17, watchOS 10\">"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend \"AvailabilityMacro=FoundationPredicate 0.3:macOS 14, iOS 17, tvOS 17, watchOS 10\">"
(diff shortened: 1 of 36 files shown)


iCharlesHu:
@swift-ci please build toolchian

CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the SwiftFoundation open source project
```
parkera, line 3: I think we just use "Swift" not "SwiftFoundation" in our other copyright headers.

Sources/Foundation/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.md for the list of SwiftFoundation project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+file(GLOB_RECURSE _FoundationPreviewSources "*.swift")
```
parkera, line 15: Let's not build the "FoundationPreview" package in this mode.

compnerd: You are aware of the limitation here that if you add a new file or remove a file you will need to do a clean build as the added/removed source will not be reflected in the build.

Sources/_CShims/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+file(GLOB_RECURSE _CShimsSources "*.c")
+add_library(_CShims STATIC ${_CShimsSources})
+
+target_include_directories(_CShims PUBLIC include)
+
+if(CMAKE_SYSTEM_NAME STREQUAL Windows)
+    target_compile_options(_CShims PUBLIC
+        "$<$<COMPILE_LANGUAGE:C,CXX>:SHELL:-D_CRT_SECURE_NO_WARNINGS>")
+endif()
```
compnerd, line 23: Do we actually need this? I think that most of the code in Foundation is clean wrt CRT security.

Sources/_CShims/include/module.modulemap
```diff
@@ -0,0 +1,15 @@
...
+    header "CFUniCharBitmapData.h"
+    header "CFUniCharBitmapData.inc.h"
+    header "filemanager_shims.h"
+    header "io_shims.h"
+    header "platform_shims.h"
+    header "string_shims.h"
+    header "uuid.h"
+
+    export *
+}
```
compnerd, line 15: Nit: Missing newline.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
+
+# Returns the current architecture name in a variable
+#
+# Usage:
+#   get_swift_host_arch(result_var_name)
+#
+# If the current architecture is supported by Swift, sets ${result_var_name}
+# with the sanitized host architecture name derived from CMAKE_SYSTEM_PROCESSOR.
+function(get_swift_host_arch result_var_name)
```
etcwilde, line 9: Rather than hard-code each architecture, we can ask the compiler what the module triple is based on either the default target triple for the toolchain or if someone is trying to cross-compile, based on `CMAKE_Swift_COMPILER_TARGET`:

```cmake
if(NOT SwiftFoundation_MODULE_TRIPLE)
  # TODO: This logic should migrate to CMake once CMake supports installing swiftmodules
  set(module_triple_command "${CMAKE_Swift_COMPILER}" -print-target-info)
  if(CMAKE_Swift_COMPILER_TARGET)
    list(APPEND module_triple_command -target ${CMAKE_Swift_COMPILER_TARGET})
  endif()
  execute_process(COMMAND ${module_triple_command} OUTPUT_VARIABLE target_info_json)
  string(JSON module_triple GET "${target_info_json}" "target" "moduleTriple")
  set(SwiftFoundation_MODULE_TRIPLE "${module_triple}" CACHE STRING "swift module triple used for installed swiftmodule and swiftinterface files")
  mark_as_advanced(SwiftFoundation_MODULE_TRIPLE)
endif()
```

https://github.com/etcwilde/swift/blob/e10563e336f2fdd7bde5b4b5f1c2a181aa2fe024/stdlib/cmake/modules/PlatformInfo.cmake#L16-L26

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+    install(FILES $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${module_name}.swiftmodule
+      RENAME ${swift_arch}.swiftmodule)
+  else()
+    install(FILES
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${swift_arch})
+  endif()
+endfunction()
```
etcwilde, line 94: missing trailing newline

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+  get_swift_host_arch(swift_arch)
+  get_target_property(module_name ${module} Swift_MODULE_NAME)
+  if(NOT module_name)
+    set(module_name ${module})
+  endif()
+
+  if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+    install(FILES $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
+      DESTINATION lib/${swift}/${swift_os}/${module_name}.swiftmodule
```
stmontgomery, line 83: Is this a copy/paste oversight, should its extension be `.swiftdoc` to match the lines above and below?
```suggestion
      DESTINATION lib/${swift}/${swift_os}/${module_name}.swiftdoc
```

etcwilde: No, this looks right. The swiftdoc/interface/module are generated in the build directory in the flat form with everything exposed at the top-level. That's fine for the build directory since you're only supposed to be building for one triple at a time, but this is moving it to the nested form where the interface, binary module, and swiftdoc file are all going into the `Foundation.swiftmodule/` directory.

stmontgomery: Ah right of course, I forgot about the "outer" .swiftmodule wrapper directory, thank you!

iCharlesHu:
Merged all commits. Now this PR is ready for review

iCharlesHu:
@swift-ci please test

Sources/FoundationMacros/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+##===----------------------------------------------------------------------===##
+
+# SwiftSyntax Dependency
+include(FetchContent)
+find_package(SwiftSyntax)
+if(NOT SwiftSyntax_FOUND)
+    # If building at desk, check out and link against the SwiftSyntax repo's targets
+    FetchContent_Declare(SwiftSyntax
+        GIT_REPOSITORY https://github.com/apple/swift-syntax.git
+        GIT_TAG 510.0.0)
```
etcwilde, line 22: Can't do much about it at the moment, but it's unfortunate that this tag doesn't have the changes for picking up the `SWIFT_HOST_MODULE_TRIPLE` automatically. So just as a heads up, you're likely to get folks forgetting to set that and asking about an the error message:

```
CMake Error at build/_deps/swiftsyntax-src/CMakeLists.txt:80 (message):
  Host module triple required
```

The correct response will be to run `swiftc -print-target-info` manually and pull out the triple under `target` > `moduleTriple` and pass that to their CMake invocation e.g. `cmake -G 'Ninja' -DSWIFT_HOST_MODULE_TRIPLE=x86_64-unknown-linux-gnu ...`.

Sources/FoundationMacros/CMakeLists.txt
```diff
@@ -0,0 +1,55 @@
...
+
+set_target_properties(FoundationMacros
+    PROPERTIES
+        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+        INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/lib
+)
+
+target_link_libraries(FoundationMacros PUBLIC
+    SwiftSyntax::SwiftSyntax
```
etcwilde, line 40: The 510.x tag of SwiftSyntax doesn't namespace its targets like this.
Newer versions will though.

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+      RENAME ${swift_arch}.swiftdoc)
```
iCharlesHu, line 84: SwiftFoundation_MODULE_TRIPLE.swiftdock

cmake/modules/SwiftSupport.cmake
```diff
@@ -0,0 +1,94 @@
...
+    install(FILES $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${module_name}.swiftmodule
+      RENAME ${swift_arch}.swiftmodule)
+  else()
+    install(FILES
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftdoc
+      $<TARGET_PROPERTY:${module},Swift_MODULE_DIRECTORY>/${module_name}.swiftmodule
+      DESTINATION lib/${swift}/${swift_os}/${swift_arch})
+  endif()
```
iCharlesHu, line 93: Remove and only use if for all platforms.

iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

etcwilde (approved):
I think this will work. I was able to successfully build and run code against this build locally.

In the `string_shims.h`, it might be worth figuring out where we're pulling that header in in C++ mode and wrapping the declarations in `extern "C" {}` to ensure that they get mangled correctly everywhere. If they don't, we'll get a linker error saying that some `_Z...` C++ symbol doesn't exist because it should actually be looking for the C mangle.

iCharlesHu:
@swift-ci please test

PR #574, opened 2024-04-30
Add autorelease pool to __swiftFillFileSystemRepresentation

jmschonfeld:
The previous ObjC implementation of this function did not add objects to the autorelease pool, but the calls to `_fastCharacterContents` and `_fastCStringContents` implicitly retain+autorelease `self` here since those functions are `NS_RETURNS_INNER_POINTER`. Adding this `autoreleasepool` invocation ensures that we drain these items from the pool before returning so that we don't add to the pool


jmschonfeld:
@swift-ci please test

PR #576, opened 2024-05-01
FoundationEssentials: tweak TimeZone handling for Windows

compnerd:
Use the Win32 APIs for environment access to ensure that we handle unicode values properly, adjust the logic to deal with encoding differences.


compnerd:
@swift-ci please test

Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
```diff
@@ -121,14 +121,50 @@ struct TimeZoneCache : Sendable {
         /// Reads from environment variables `TZFILE`, `TZ` and finally the symlink pointed at by the C macro `TZDEFAULT` to figure out what the current (aka "system") time zone is.
         mutating func findCurrentTimeZone() -> TimeZone {
 #if !NO_TZFILE
+#if os(Windows)
```
jmschonfeld, line 124: Should we just change this whole section to use `ProcessInfo.processInfo.environment` so that it is portable for both platforms and we don't have to copy this code around?

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #578, opened 2024-05-02
FoundationEssentials: repair `ProcessInfo` queries for Windows

compnerd:
Use a dynamically allocated buffer to avoid truncation/underruns that are not possible to recover from. Fix some parameter passing to match the type signature of the C function.

### DIFF
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -170,13 +170,11 @@ final class _ProcessInfo: Sendable {
         // WASI does not have user concept
         return ""
 #elseif os(Windows)
-        return withUnsafeTemporaryAllocation(of: wchar_t.self, capacity: 1040) { usernameBuffer in
-            usernameBuffer[0] = 0
-            var size: DWORD = 1040
-            if GetUserNameW(usernameBuffer.baseAddress!, &size) {
-                // discount the extra NULL by decrementing the size
-                return String(decoding: usernameBuffer.prefix(size - 1), as: UTF16.self)
-            } else {
+        var dwSize: DWORD = 0
+        _ = GetUserNameW(nil, &dwSize)
+
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwSize)) {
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -203,10 +202,12 @@ final class _ProcessInfo: Sendable {
         return ""
 #elseif os(Windows)
         var ulLength: ULONG = 0
-        GetUserNameExW(NameDisplay, NULL, &ulLength)
+        _ = GetUserNameExW(NameDisplay, nil, &ulLength)
 
-        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: ulLength + 1) { wszBuffer in
-            GetUserNameExW(NameDisplay, wszBuffer.baseAddress!, &ulLength)
+        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(ulLength + 1)) { wszBuffer in
```
jmschonfeld, line 207: From the [docs](https://learn.microsoft.com/en-us/windows/win32/api/secext/nf-secext-getusernameexw) I see:

> If lpNameBuffer is too small, the function fails and [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror) returns ERROR_MORE_DATA. This parameter receives the required buffer size, in Unicode characters (whether or not Unicode is being used), including the terminating null character.

I think that means we can allocate `length` instead of `ulLength + 1` right? (and we probably also need to use `ulLength - 1` below)

compnerd: Ah, right! I didn't cross-reference that.

jmschonfeld:
@swift-ci please test

compnerd:
I have a small further change that I am pushing up that fixes the rest of the issues with `ProcessInfo.swift`. At least locally, I am now able to build FoundationEssentials (more patches incoming).

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #581, opened 2024-05-03
FoundationEssentials: support data writing on Windows

compnerd:
This implements the IO aspects of data writing for Windows to ensure that we can support the full API surface.


compnerd:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -128,6 +128,11 @@ private func writeToFileDescriptorWithProgress(_ fd: Int32, buffer: UnsafeRawBuf
 
 private func cleanupTemporaryDirectory(at inPath: String?) {
     guard let inPath else { return }
+#if os(Windows)
```
jmschonfeld, line 131: I wonder if we should just change the whole function to use `FileManager.default.removeItem(atPath: inPath)` to remove all of the conditional logic. @parkera is there any benefit to calling `rmdir` directly here instead of letting FileManager "do the right thing"?

parkera: I dropped a comment from the original ObjC when I ported this:

```
We try to use rmdir() because it's definitely just one fs_usage and the temporary directory is supposed to be empty at this point anyway. Only if that fails do we turn the awesome power of a fully operational -[NSFileManager removeItemAtURL:error:] on it.
```

Is that even still true?

jmschonfeld: Hmm, on non-darwin the first thing `FileManager.removeItem` does is it tries `rmdir` and it only does other things if that fails, but on Darwin we do call `removefile` instead. However, I'm not really sure if `removefile` on an empty directory is any slower than `rmdir` on an empty directory. I can try to validate that?

parkera: I guess we should just removefile, and if that can be optimized then it should do it on its own. So here, just use file manager.

compnerd: Okay, I will replace this with `fileManager.removeItem(atPath:)` then @parkera.

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #582, opened 2024-05-03
FoundationInternationalization: adjust types for enumerations

compnerd:
Windows uses a signed integer for enumerations where as other platforms use an unsigned integer. Account for this in the types for the bridged values in Swift.

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
@@ -1269,8 +1269,13 @@ internal final class _LocaleICU: _LocaleProtocol, Sendable {
 
             var weekdayTypes : [UCalendarWeekdayType] = [UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY]
 
-            var onset: UInt32?
-            var cease: UInt32?
+#if os(Windows)
+            var onset: CInt?
+            var cease: CInt?
+#else
+            var onset: CUnsignedInt?
+            var cease: CUnsignedInt?
+#endif
 
             var status = U_ZERO_ERROR
             let cal = ucal_open(nil, 0, identifier, UCAL_DEFAULT, &status)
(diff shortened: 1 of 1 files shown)


Sources/FoundationInternationalization/Locale/Locale_ICU.swift
```diff
@@ -1269,8 +1269,13 @@ internal final class _LocaleICU: _LocaleProtocol, Sendable {
 
             var weekdayTypes : [UCalendarWeekdayType] = [UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY, UCAL_WEEKDAY]
 
-            var onset: UInt32?
-            var cease: UInt32?
+#if os(Windows)
+            var onset: CInt?
+            var cease: CInt?
+#else
+            var onset: CUnsignedInt?
```
parkera, line 1276: I'm a bit confused on this one - where does the difference in type come from?

compnerd: The platform ABI dictates the underlying type of a C enumeration. The underlying type differs in signededness between the platforms and because we use the C types here, the type difference will propagate.

parkera: I'm wondering if there might be a way to factor this so the type doesn't virally spread into the places where we use this structure.

parkera: In other words, the instant we get it out of ICU I don't think it's valuable to preserve the fact that it is a "platform int" vs a standard Swift integer type.

compnerd: It is the signededness that is the issue, not the fact that it is a platform int vs a Swift Int I think. The use of the platform type names is just a readability aid (at least for me). Did you have an approach in mind for abstracting over the representation of the enumeration? What comes to my mind is a replicated enumeration that stores the raw value and exposes a typed enumerator.

parkera: On further thought, I guess this is ok. I was thinking it was in the `package struct` but it's just local to this file.

parkera (approved):

PR #583, opened 2024-05-03
FoundationEssentials: properly handle env duplication

compnerd:
The terminator is a valid string and will give a valid pointer. However, the string itself is empty (`\0\0`) and indicates the termination of the process environment block. Handle this properly to avoid a buffer overrun.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #584, opened 2024-05-03
Use the #Expression macro instead of initializing Expression directly

jmschonfeld:
Now that blocking issues have been resolved, we can update our unit tests to invoke the `#Expression` macro directly instead of initializing an `Expression` without the macro


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):
It's finally all coming together 🥳

PR #586, opened 2024-05-04
Port URL and URLComponents to swift-foundation

jrflat:
This PR adds Swift-native implementations of `URL`, `URLComponents`, `URLQueryItem`, and the underlying RFC 3986 parsing to `FoundationEssentials`. These will also serve as the implementations of their Swift and ObjC counterparts on Darwin. Most of the `URL` APIs are addressed in this PR, but some file system APIs, such as those involving `URLResourceKey/Values`, are yet to be implemented.

The Swift implementation contains some bug fixes that make it behave slightly differently than previously on Darwin. Notably, `.path()` now matches its documentation and resolves against the base URL like the deprecated `.path` did. Because `.path` always returned a file system path without a trailing slash, and `.path` was widely used by other `NSURL` APIs (such as `.pathComponents`), I've added an internal `.fileSystemPath` var and used this in many cases for compatibility. `URL` now also properly parses rootless paths and passes all of the RFC 3986 resolution examples (including allowing empty string for a relative path).

IDNA-encoding of hostnames is supported if `FoundationInternationalization` is imported. Otherwise, we aim to maintain a smaller footprint in `FoundationEssentials` by not importing the Unicode library, which can be quite large.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #587, opened 2024-05-06
FoundationEssentials: correct an off-by-one error on Windows

compnerd:
The returned value in the success case does not account for the terminating nul character, resulting in the value being 1 less than the allocation required. This was causing a spurious failure and thus unable to compute the location of the process.


jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #588, opened 2024-05-06
Use URL in JSON/Data APIs

jmschonfeld:
Now that we have the `URL` type ported to swift-foundation, we can begin adopting it in our other APIs. This PR removes the `FOUNDATION_FRAMEWORK` blocks around usages of `URL` in `JSONEncoder`/`JSONDecoder` and the `Data` reading/writing APIs. I'll followup with a change for `FileManager`, but this should bring Data/JSON APIs on-par with what Foundation provides for URL-related behavior and APIs (and removes some of the temporary `String`-based APIs we added

### DIFF
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -466,12 +466,12 @@ extension _ProcessInfo {
 #if os(Linux)
     // Support for CFS quotas for cpu count as used by Docker.
     // Based on swift-nio code, https://github.com/apple/swift-nio/pull/1518
-    private static let cfsQuotaPath = URL(filePath: "/sys/fs/cgroup/cpu/cpu.cfs_quota_us", directoryHint: .notDirectory)
-    private static let cfsPeriodPath = URL(filePath: "/sys/fs/cgroup/cpu/cpu.cfs_period_us", directoryHint: .notDirectory)
-    private static let cpuSetPath = URL(filePath: "/sys/fs/cgroup/cpuset/cpuset.cpus", directoryHint: .notDirectory)
+    private static let cfsQuotaURL = URL(filePath: "/sys/fs/cgroup/cpu/cpu.cfs_quota_us", directoryHint: .notDirectory)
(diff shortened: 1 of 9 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows

jmschonfeld:
Windows failure is just because the toolchain is outdated and it doesn't have the fix for the SPM bug

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -466,12 +466,11 @@ extension _ProcessInfo {
...
-    private static let cfsQuotaPath = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
-    private static let cfsPeriodPath = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
-    private static let cpuSetPath = "/sys/fs/cgroup/cpuset/cpuset.cpus"
+    private static let cfsQuotaPath = URL(filePath: "/sys/fs/cgroup/cpu/cpu.cfs_quota_us", directoryHint: .notDirectory)
+    private static let cfsPeriodPath = URL(filePath: "/sys/fs/cgroup/cpu/cpu.cfs_period_us", directoryHint: .notDirectory)
+    private static let cpuSetPath = URL(filePath: "/sys/fs/cgroup/cpuset/cpuset.cpus", directoryHint: .notDirectory)
 
-    private static func firstLineOfFile(path: String) throws -> Substring {
-        // TODO: Replace with URL version once that is available in FoundationEssentials
+    private static func firstLineOfFile(path: URL) throws -> Substring {
```
itingliu, line 473: nit: I think we prefer calling these `URL`-type arguments and variables something like `url` rather than `path`. I understand that's probably going to clutter this diff though so I'm ok with leaving them as-is for now

itingliu (approved):

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
jmschonfeld: That's a fair point, I went ahead and updated it

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #589, opened 2024-05-06
Utilize URL in FileManager

jmschonfeld:
Now that we have `URL` in `FileManager`, we can adopt it throughout the APIs. Many of the APIs already existed using the `URL` stub previously available, but now the delegate methods and other APIs have been updated to use the full `URL` implementation.

Some APIs are currently still not available/ported because they rely on the `URLResourceValues` implementation which hasn't been ported yet. Those will be added in a followup PR. I also made 2 minor warning fixes while I was here.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -164,7 +164,7 @@ extension FileManager {
                 delegateResponse = delegate.fileManager?(self, shouldProceedAfterError: error, removingItemAtPath: path)
             }
             #else
-            delegateResponse = delegate.fileManager(self, shouldProceedAfterError: error, removingItemAtPath: path)
+            delegateResponse = delegate.fileManager(self, shouldProceedAfterError: error, removingItemAt: URL(fileURLWithPath: path))
             #endif
         }
         
@@ -181,7 +181,7 @@ extension FileManager {
                 delegateResponse = delegate.fileManager?(self, shouldRemoveItemAtPath: path)
             }
             #else
-            delegateResponse = delegate.fileManager(self, shouldRemoveItemAtPath: path)
+            delegateResponse = delegate.fileManager(self, shouldRemoveItemAt: URL(fileURLWithPath: path))
             #endif
         }
         return delegateResponse ?? true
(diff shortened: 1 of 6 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Sources/FoundationInternationalization/Locale/Locale_ICU.swift
```diff
@@ -1221,7 +1221,7 @@ internal final class _LocaleICU: _LocaleProtocol, Sendable {
                 }
 
                 // Check prefs
-                if let firstWeekdayPref = prefs?.firstWeekday {
+                if prefs?.firstWeekday != nil {
```
parkera, line 1224: I looked at this one a while ago and couldn't decide if the error was not comparing with nil or not using the value from prefs. It might be best to split it out.

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -164,7 +164,7 @@ extension FileManager {
                 delegateResponse = delegate.fileManager?(self, shouldProceedAfterError: error, removingItemAtPath: path)
             }
             #else
-            delegateResponse = delegate.fileManager(self, shouldProceedAfterError: error, removingItemAtPath: path)
```
itingliu, line 167: In the FOUNDATION_FRAMEWORK branch above, we fall back to the `String`-based delegate functions after failing the `URL`-based delegate functions. Do we want to do the same here?

jmschonfeld: In the `FOUNDATION_FRAMEWORK` branch, we fallback when the delegate does not implement the URL-based functions and we can detect this because it's an ObjC `@optional` protocol requirement. Since we don't have that same construct here, we can't differentiate between the URL-based function returning an explicit value or the default value. Instead, this patch updates the URL-based function's default implementation to call the String-based function, so if the client only implements the String-based function, calling the URL-based function here will call through to the String based one (which will return the default value if neither are implemented)

itingliu: Didn't realize these are `optional` in obj-c, and yeah having the default implementation in Swift for URL-based ones definitely helps. Thanks that makes sense

itingliu (approved):

parkera (approved):

PR #590, opened 2024-05-07
Use ICU-equivalent \b behavior with Regex for String.range(of:)

jmschonfeld:
Previously the `String.range(of:)` function's regex matching used ICU word boundaries rather than the more advanced Unicode boundaries. However, when we adopted `Regex` for the implementation we changed the behavior because `Regex` defaults to more advanced unicode word boundaries. This changes `String.range(of:)` back to ICU word boundaries for this API


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #591, opened 2024-05-07
FoundationEssentials: correct memory mapping on Windows

compnerd:
We were using the wrong buffer when creating the result, forming the wrong deallocation for the buffer. Remove the unnecessary mapping and correct the buffer usage to prevent a cross-domain free of the data.


compnerd:
@swift-ci please test

parkera (approved):
Glad you found this!

PR #592, opened 2024-05-07
Ensure Data Writing throws errors with correct errnos

jmschonfeld:
In some cases we were throwing a `CocoaError` with an `errno` that was 0. This PR updates all of the places in Data writing where we throw an error with an `errno` that could be outdated (since we make calls between when we capture `errno` and when the function fails). I've updated all of the places where we have this pattern to correctly save the right `errno` value


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #593, opened 2024-05-07
FoundationEssentials: recover from backslide on Windows

compnerd:
This new helper was added but did not support windows as `stat` is a Unix operation and the Windows equivalent is more nuanced when you are trying to minimise IO overheads. Add a Windows specific path to repair the build on Windows.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1991,7 +1991,15 @@ extension URL {
 
 #if !NO_FILESYSTEM
     private static func isDirectory(_ path: String) -> Bool {
-        return FileManager.default._fileStat(path)?.isDirectory ?? false
+#if !FOUNDATION_FRAMEWORK
+        var isDirectory: Bool = false
+        _ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
+        return isDirectory
+#else
+        var isDirectory: ObjCBool = false
+        _ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
+        return isDirectory.boolValue
+#endif
     }
 #endif // !NO_FILESYSTEM
 


compnerd:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1991,7 +1995,17 @@ extension URL {
...
+#if os(Windows)
+        return (try? path.withNTPathRepresentation {
+            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
+            guard GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) else {
+                return false
+            }
+            return faAttributes.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY
+        }) ?? false
+#else
         return FileManager.default._fileStat(path)?.isDirectory ?? false
```
jmschonfeld, line 2007: I wonder if we should just use `FileManager.default.fileExists(atPath:isDirectory:)` which is already cross platform. Do you think that'd be ok here? @jrflat do you see any issues with that?

compnerd: Yeah, that does seem nicer.

compnerd: @jmschonfeld should I just revise this PR to do that?

jrflat: Using `fileExists(atPath:isDirectory:)` sounds good to me!

compnerd:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1991,7 +1991,9 @@ extension URL {
 
 #if !NO_FILESYSTEM
     private static func isDirectory(_ path: String) -> Bool {
-        return FileManager.default._fileStat(path)?.isDirectory ?? false
+        var isDirectory: Bool = false
+        FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
```
jmschonfeld, line 1995: I'm seeing a warning that states:

```
Result of call to 'fileExists(atPath:isDirectory:)' is unused
```

here, any chance we can add a `_ =` to silence that warning?

Also, when building this change as part of the internal Foundation framework, we hit the error

```
Sources/FoundationEssentials/URL/URL.swift:1995:67: error: cannot convert value of type 'UnsafeMutablePointer<Bool>' to expected argument type 'UnsafeMutablePointer<ObjCBool>'
```

Because on Darwin this API uses an `ObjCBool` instead of a `Bool`. Could we update it to account for that, maybe something like

```swift
#if !FOUNDATION_FRAMEWORK
var isDirectory: Bool = false
_ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
return isDirectory
#else
var isDirectory: ObjCBool = false
_ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
return isDirectory.boolValue
#endif
```

We probably need to rethink the portability of this function but for now that should let us merge this change

compnerd: Trying a small tweak to see if we can just rely on the type inference to figure out the types.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
jmschonfeld: Sadly it's not smart enough to infer the type of `isDirectory` from the usage on the line below 😢:

```
Sources/FoundationEssentials/URL/URL.swift:1995:71: error: cannot convert value of type 'UnsafeMutablePointer<Bool>' to expected argument type 'UnsafeMutablePointer<ObjCBool>'
Sources/FoundationEssentials/URL/URL.swift:1999:28: error: value of type 'Bool' has no member 'boolValue'
```

compnerd:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1991,7 +1991,17 @@ extension URL {
 
 #if !NO_FILESYSTEM
     private static func isDirectory(_ path: String) -> Bool {
-        return FileManager.default._fileStat(path)?.isDirectory ?? false
+#if !FOUNDATION_FRAMEWORK
+        var isDirectory: Bool = false
+#else
+        var isDirectory: ObjCBool = false
+#endif
+        _ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
```
jrflat, line 1999: I think we could maybe just group it for clarity
```
#if FOUNDATION_FRAMEWORK
var isDirectory: ObjCBool = false
_ = FileManager.default.fileExists(atPath: path, isDirectory: &isDirectory)
return isDirectory.boolValue
#else
...
#endif
```
but that's really a minor nit. Thanks for fixing this!

jrflat (approved):

compnerd:
@swift-ci please test

PR #594, opened 2024-05-07
testRangeRegexB can use System Foundation instead of FoundationEssentials

jmschonfeld:
This unit test started failing in some configurations on macOS. This is because the `range(of:)` function is not defined in FoundationEssentials (just in Foundation) so this test was inadvertently calling to the system Foundation which does not have the fixed behavior on all OS's. While the test now looks a bit more clunky, it explicitly calls the internal entry point in FoundationEssentials rather than the system Foundation to ensure we're testing the right version of this function


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #596, opened 2024-05-08
build: adjust package manifest to allow building on Windows

compnerd:
SPM does not permit building macro tests on Windows due to mishandling of build types, resulting in multiply defined `main` which is a strong error on Windows. As a workaround, remove the macro tests for the time being.


jmschonfeld (approved):
Makes sense to me since it looks like swift-testing does the same thing currently

PR #599, opened 2024-05-09
FoundationEssentialTests: `nanosleep` is not portable, avoid it

compnerd:
Use `Thread.sleep(forTimeInterval:)` instead of `nanosleep` which is not a portable sleep primitive. This allows us to build the package tests on Windows.

### DIFF
--- a/Tests/FoundationEssentialsTests/ProcessInfoTests.swift
+++ b/Tests/FoundationEssentialsTests/ProcessInfoTests.swift
@@ -94,13 +94,12 @@ final class ProcessInfoTests : XCTestCase {
         XCTAssertTrue(memory > 0, "ProcessInfo doesn't think we have any memory")
     }
 
-    func testSystemUpTime() {
+    func testSystemUpTime() async throws {
         let now = ProcessInfo.processInfo.systemUptime
         XCTAssertTrue(
             now > 1, "ProcessInfo returned an unrealistically low system uptime")
         // Sleep for 0.1s
-        var ts: timespec = timespec(tv_sec: 0, tv_nsec: 100000000)
-        nanosleep(&ts, nil)
+        try await Task.sleep(for: .milliseconds(100))
         XCTAssertTrue(
             ProcessInfo.processInfo.systemUptime > now,
             "ProcessInfo returned the same system uptime with 400")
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/FoundationEssentialsTests/ProcessInfoTests.swift
```diff
@@ -99,8 +99,7 @@ final class ProcessInfoTests : XCTestCase {
         XCTAssertTrue(
             now > 1, "ProcessInfo returned an unrealistically low system uptime")
         // Sleep for 0.1s
-        var ts: timespec = timespec(tv_sec: 0, tv_nsec: 100000000)
-        nanosleep(&ts, nil)
+        Thread.sleep(forTimeInterval: 0.1)
```
jmschonfeld, line 102: `Thread` comes from the `Foundation` module which we'd like to try to avoid in this project (even in the tests). Is there something else we could use here, maybe some `async`/`await` or `Task` related API to accomplish the same goal in a portable manner?

compnerd: https://developer.apple.com/documentation/swift/task/sleep(for:tolerance:clock:) should suffice I suppose for that.

compnerd:
@swift-ci please test

Tests/FoundationEssentialsTests/ProcessInfoTests.swift
```diff
@@ -94,13 +94,12 @@ final class ProcessInfoTests : XCTestCase {
...
 
-    func testSystemUpTime() {
+    func testSystemUpTime() async throws {
         let now = ProcessInfo.processInfo.systemUptime
         XCTAssertTrue(
             now > 1, "ProcessInfo returned an unrealistically low system uptime")
         // Sleep for 0.1s
-        var ts: timespec = timespec(tv_sec: 0, tv_nsec: 100000000)
-        nanosleep(&ts, nil)
+        try await Task.sleep(for: .milliseconds(1000))
```
jmschonfeld, line 102: nit (sorry): the comment says "0.1s" but the code sleeps for 1.0s, should we update one or the other to be consistent?

compnerd: No, that should've been 100, not 1000.

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #600, opened 2024-05-09
FoundationEssentialTests: elide some tests on Windows

compnerd:
Windows does not have the concept of a "root" user, instead all users have rights assigned to them with a capabilities system. As there is no equivalent for `euid`, simply elide this suite on Windows for now.

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -523,8 +523,11 @@ final class FileManagerTests : XCTestCase {
             XCTAssertFalse($0.fileExists(atPath: "does_not_exist"))
         }
     }
-    
+
     func testFileAccessAtPath() throws {
+#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -523,7 +523,8 @@ final class FileManagerTests : XCTestCase {
             XCTAssertFalse($0.fileExists(atPath: "does_not_exist"))
         }
     }
-    
+
+#if !os(Windows)
```
jmschonfeld, line 527: Could we leave a comment about why this test isn't supported on Windows? (sometimes I find it helpful to put the if-def within the function and in the `os(Windows)` path `throw XCTSkip("Some explanation")` just so it's clear in the future that this is intentionally not supported (rather than the API just not being finished yet, for example) - not a requirement here but I think we should indicate the reasoning in some form)

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #601, opened 2024-05-09
Minor fixes to support using URL in swift-corelibs-foundation

jmschonfeld:
This PR includes two minor bug fixes identified while adopting `URL` in the re-cored package branch of SCF. This ensures that we use the proper error user info key declaration and also ensures that standardizing a URL or resolving symlinks in a URL will respect the directory designation when that `URL` was initialized even if it doesn't match what's on the filesystem.


jmschonfeld:
@swift-ci please test

jrflat (approved):

PR #603, opened 2024-05-10
FoundationEssentials: simplify path normalization

compnerd:
We would previously conditionally call `GetFullPathNameW` and do string manipulations for normalising the path. Instead, always call `GetFullPathNameW` to normalise the path as per Windows' rules. This more importantly will collapse runs of the arc separator, which is crucial when using extended paths as the NT path form must always use the normalised paths or the access will fail.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -28,51 +28,23 @@ extension String {
             throw CocoaError.errorWithFilePath(.fileReadInvalidFileName, "")
         }
 
-        // 1. Normalize the path first.
-
-        var path = self
+        var iter = self.utf8.makeIterator()
+        let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
 
         // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
-        var iter = path.makeIterator()
-        if iter.next() == "/", iter.next()?.isLetter ?? false, iter.next() == ":" {
-            path.removeFirst()
-        }
-
-        // Win32 APIs can support `/` for the arc separator. However,
-        // symlinks created with `/` do not resolve properly, so normalize
-        // the path.
-        path.replace("/", with: "\\")
-
-        // Drop trailing slashes unless it follows a drive specification.  The
-        // trailing arc separator after a drive specifier indicates the root as
-        // opposed to a drive relative path.
-        while path.count > 1, path.last == "\\" {
-            let first = path.startIndex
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -28,51 +28,23 @@ extension String {
...
-        // 2. Perform the operation on the normalized path.
-
-        return try path.withCString(encodedAs: UTF16.self) { pwszPath in
-            guard !path.hasPrefix(#"\\"#) else { return try body(pwszPath) }
-
-            let dwLength = GetFullPathNameW(pwszPath, 0, nil, nil)
-            let path = try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-                guard GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil) == dwLength - 1 else {
-                    throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+        return try self.dropFirst(bLeadingSlash ? 1 : 0).withCString(encodedAs: UTF16.self) { pwszPath in
```
jmschonfeld, line 37: One minor thing that I didn't think about when we first added this method: should we use the UTF-8 view instead of the `Character` view when performing these lookups/manipulations? That might be decently faster and minimizes the risk that grapheme extending/joining characters cause issues here. If we want to do that, we'd just want to use the UTF-8 iterator above and then this line would become `Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString(...)`. I think we have mixed usages but lately have tried to stick to the UTF-8 view when doing these types of operations. Thoughts?

compnerd: Yeah, I can see the value in the UTF-8 view. Let me test it and if there aren't any regressions, lets switch over. The string itself should be UTF-8 encoded prior to the re-encoding.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -28,51 +28,23 @@ extension String {
             throw CocoaError.errorWithFilePath(.fileReadInvalidFileName, "")
         }
 
-        // 1. Normalize the path first.
-
-        var path = self
+        var iter = self.makeIterator()
+        let bLeadingSlash = if iter.next() == "/", iter.next()?.isLetter ?? false, iter.next() == ":" { true } else { false }
```
jmschonfeld, line 32: Should we use the UTF-8 view here too? (`self.utf.makeIterator()`, `iter.net() == ._slash`, etc.)

compnerd: Done

compnerd:
@swift-ci please test

jmschonfeld (approved):

jrflat (approved):

PR #604, opened 2024-05-10
Remove extraneous CocoaError stub

jmschonfeld:
There's a leftover `CocoaError` stub inside `FoundationInternationalization` from before we ported it over. This PR removes that stub and ensures that things in `FoundationInternationalization` throw the "real" `CocoaError` type


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #605, opened 2024-05-13
FoundationEssentials: protect against a failed query on Windows

compnerd:
In the case that the query fails due to a reason other than an insufficient buffer, the returned value is 0. However, a 0 sized allocation can trigger an issue, so ensure that we always have a non-zero buffer size before we attempt an allocation.


jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #602, opened 2024-05-09
Support Windows URL paths in FoundationEssentials

jrflat:
Supports initializing file URLs with Windows paths. Allow `"\"` as the path separator, allow UNC paths (`"\\"`) with host, and allow/standardize drive letter paths (`"C:"`). However, we always store the URL with forward slashes to stay consistent for path parsing operations. I think `String.withNTPathRepresentation(_:)` should then be used to convert any `"/"` back to `"\"` before consulting the file system on Windows. @compnerd thoughts on this approach?

The file URL initializers also now resolve against the base URL (or current directory) for checking if the file is a directory.

PR #606, opened 2024-05-13
URL.init(filePath:) should resolve against the base URL before checking if the file is a directory

jrflat:
The fix I added in #602 to address this wasn't correct since `.mergedPath(for:)` should be called on an URL that has already been initialized with a base URL. I separated this logic out into a String extension that handles the merging generally and use this method when determining the absolute file path to directory check in `.init(filePath:)`.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #607, opened 2024-05-14
FoundationEssentials: correct the use of `open` on Windows

compnerd:
`open` is a variatic function and does not take the same permission modes on Windows. Prefer to use `_sopen_s` to allow us to have a non-variadic signature and repair the runtime on Windows.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #610, opened 2024-05-14
FoundationEssentials: correct error handling for item removal

compnerd:
The condition of the error check was inverted as it was meant to be a `guard` statement. This was resulting in an infinite loop. With this we now pass a few additional tests and the test suite no longer hangs.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #611, opened 2024-05-15
FoundationEssentials: avoid precondition failure on Windows

compnerd:
Repair the error reporting on Windows in the case of a failure in `SHGetKnownFolder` - the returned value is a `HRESULT`, which is the extended error code unlike the Win32 APIs which return a `BOOL`.

The "Recyle Bin" is a virtual folder and does not have an associated path with the folder that we can query. Undo the implementation here for the time being to avoid the assertion failure in the runtime.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/SearchPaths/FileManager+WindowsSearchPaths.swift
+++ b/Sources/FoundationEssentials/FileManager/SearchPaths/FileManager+WindowsSearchPaths.swift
@@ -16,10 +16,10 @@ import WinSDK
 
 private func _url(for id: KNOWNFOLDERID) -> URL {
     var pszPath: PWSTR?
-    let hrResult: HRESULT = withUnsafePointer(to: id) { id in
+    let hr: HRESULT = withUnsafePointer(to: id) { id in
         SHGetKnownFolderPath(id, KF_FLAG_DEFAULT, nil, &pszPath)
     }
-    precondition(SUCCEEDED(hrResult), "SHGetKnownFolderpath failed \(GetLastError())")
+    precondition(SUCCEEDED(hr), "SHGetKnownFolderPath failed \(String(hr, radix: 16))")
     defer { CoTaskMemFree(pszPath) }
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/FileManager/SearchPaths/FileManager+WindowsSearchPaths.swift
```diff
@@ -60,7 +60,9 @@ func _WindowsSearchPathURL(for directory: FileManager.SearchPathDirectory, in do
         _url(for: FOLDERID_Public)
 
     case (.trashDirectory, .userDomainMask):
-        _url(for: FOLDERID_RecycleBinFolder)
+        // The "Recycle Bin" is a virtual folder and we cannot get a path from
+        // it directly using `SHGetKnownFolderPath`.
+        nil
```
jmschonfeld, line 65: Is this a permanent change (i.e. expected behavior) for Windows or something we want to come back and fix later? If it's "permanent" we could just remove this case altogether and let it fallback to the `default`.

Additionally, `FileManagerTests.testSearchPaths` currently asserts that `(.trashDirectory, .userDomainMask)` produces a directory on Windows. Should we update the test expectations for this change? I think to resolve it we'd just need to add a

```swift
#if os(Windows)
let isWindows = true
#else
let isWindows = false
#endif
```

to the test and then change the `(isMacOS && isFramework) || !isDarwin` conditional to `(isMacOS && isFramework) || (!isDarwin && !isWindows)` to ensure it validates that we don't produce a directory on Windows

jmschonfeld: Ok sounds good then leaving the explicit case / `return nil` with the comment makes sense to me (maybe we should add a "TODO:" to the comment just so it's a bit more visible as something we want to come back to?

compnerd: Sure, I'll add a comment.

compnerd:
@swift-ci please test

jrflat (approved):

PR #612, opened 2024-05-15
Improve String+Path performance by removing Regex

iCharlesHu:
Reimplement `String._transmutingCompressingSlashes` and `pathHasDotDot` without using Regex. Reimplementing this method without Regex allows us to:
- Reduce the additional frames at launch since the Swift rewrite (Regex parsing brings in tons of frames)
- Not having to load additional dylib (`libswift_StringProcessing)


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #614, opened 2024-05-16
FoundationEssentials: correct error reporting for `_mktemp_s`

compnerd:
On failure, the error information is stored in `errno` not the Win32 error domain. Use the correct initialiser to ensure that we get the details of the failure.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #615, opened 2024-05-16
FoundationEssentials: avoid a sharing violation on Windows

compnerd:
When trying to perform a move, we would encounter a sharing violation due to the file descriptor being open preventing the file from being moved. Perform an early close of the file descriptor before performing the move. This adds complexity to the defer, however, avoiding the defer complicates the code flow analysis for error propagation.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -336,7 +336,7 @@ private func writeToFileAux(path inPath: PathOrURL, buffer: UnsafeRawBufferPoint
             throw CocoaError.errorWithFilePath(inPath, errno: errno, reading: false)
         }
 
-        defer { _close(fd) }
+        defer { if fd >= 0 { _close(fd) } }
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -361,6 +361,9 @@ private func writeToFileAux(path inPath: PathOrURL, buffer: UnsafeRawBufferPoint
         // We're done now
         guard let auxPath else { return }
 
+        _close(fd)
```
jmschonfeld, line 364: I'm curious what the difference between windows and non-windows is here, since the setup is the same for both right. Is this just a subtle behavioral difference between `rename` on POSIX systems and `MoveFileExW` on windows? Or is there another difference in the code that leads to this?

compnerd: Yes, it's the use of the windows APIs and the requirement to close the handle before the move.

jmschonfeld (approved):

PR #616, opened 2024-05-16
Gracefully handle cases where the isDirectory url resource might not be set

jmschonfeld:
In some cases, the `isDirectory` resource value may not be set even though it was requested. We should gracefully handle this to avoid a crash in this situation


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #617, opened 2024-05-17
FoundationEssentials: use the correct flags on Windows

compnerd:
Windows spells their `open` flags differently. Ensure that we use the Windows spellings on Windows to get the correct value.


parkera:
@swift-ci test

parkera (approved):

PR #618, opened 2024-05-18
Rename benchmark files to avoid duplicating names from the project sources

parkera:
When we have the benchmarks and the project sources open in the same Xcode or VSCode environment, open quickly can get confused about which files to open - benchmark or project. Let's standardize on prefixing these with the name Benchmark to make it clearer what they are for.


parkera:
@swift-ci test

PR #619, opened 2024-05-18
Unify ComparisonResult codable conformance

parkera:
Ensure that the framework and package are using the same implementation of `Codable` for `ComparisonResult`.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #621, opened 2024-05-20
FoundationEssentials: do not strip non-absolute paths

compnerd:
We would strip the first characters of a path if it were non-absolute. For special files like `CON` we would end up stripping the name to `N`.


compnerd:
@swift-ci please test

PR #622, opened 2024-05-20
Implement String(contentsOf:) family in Swift

parkera:
Now that we have all of the underlying pieces in place for data reading/writing and string encoding - finish the work to unify the implementations of `String(contentsOf:)`.


parkera:
@swift-ci test

parkera:
@swift-ci test windows

Benchmarks/Benchmarks/String/BenchmarkString.swift
```diff
@@ -106,4 +106,84 @@ let benchmarks = {
         
     // MARK: - UTF32
 
+    Benchmark("utf32-encode", configuration: .init(warmupIterations: 1, scalingFactor: .kilo)) { benchmark in
```
parkera, line 109: These new benchmarks round out the UTF16/8 ones from the previous String work.

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -25,24 +25,34 @@ import CRT
...
 private func readExtendedAttributesFromFileDescriptor(_ fd: Int32, attrsToRead: [String]) -> [String : Data] {
-#if canImport(Darwin) && !NO_FILESYSTEM
+#if !NO_FILESYSTEM
     var output: [String : Data] = [:]
     for key in attrsToRead {
         key.withCString { keyStr in
             let maxXAttrLength = 1000
             withUnsafeTemporaryAllocation(of: CUnsignedChar.self, capacity: maxXAttrLength) { buf in
-                let result = fgetxattr(fd, keyStr, buf.baseAddress, maxXAttrLength, 0, 0)
+                let result = _fgetxattr(fd, keyStr, buf.baseAddress, maxXAttrLength, 0, 0)
```
parkera, line 45: @compnerd - could use your advice here on the Windows version of this code.

Sources/FoundationEssentials/String/String+Essentials.swift
```diff
@@ -80,169 +80,4 @@ extension String {
...
-    /// Creates a new string equivalent to the given bytes interpreted in the
-    /// specified encoding.
-    ///
-    /// - Parameters:
-    ///   - bytes: A sequence of bytes to interpret using `encoding`.
-    ///   - encoding: The encoding to use to interpret `bytes`.
-    public init?<S: Sequence>(bytes: __shared S, encoding: Encoding)
-        where S.Iterator.Element == UInt8
-    {
-        // TODO: Move init?(bytes:encoding) to Swift
```
parkera, line 95: Resolved

Sources/FoundationEssentials/Data/Data+Reading.swift
compnerd: Windows doesn't really have "extended attributes". The permission model on Windows is pretty rich and allows for per-identity ACLing of different forms. I believe that we currently do no really handle that per se.

parkera: Ok - the purpose of this xattr is to store some metadata about the encoding of the file. We can't guarantee its presence on non-Darwin platforms, either, so we could just modify this to do nothing on Windows if you think that's correct.

parkera: Actually, I already did that above in the stub. It looks like we're good to go for now.

jmschonfeld (approved):

PR #623, opened 2024-05-20
Reduce number of read calls when not asking for progress reporting.

parkera:
Also, when asking for progress reporting, cap the number of updates at approximately 100 instead of fileSize/4k.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -388,8 +388,10 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
 #endif
 }
 
-// Takes an `Int` size and returns an `Int` to match `Data`'s count. If we are going to read more than Int.max, throws - because we won't be able to store it in `Data`.
-private func readBytesFromFileDescriptor(_ fd: Int32, path: PathOrURL, buffer inBuffer: UnsafeMutableRawPointer, length: Int, reportProgress: Bool) throws -> Int {
+/// Read data from a file descriptor.
+/// Takes an `Int` size and returns an `Int` to match `Data`'s count. If we are going to read more than Int.max, throws - because we won't be able to store it in `Data`.
+/// If `readUntilEOF` is `false`, this can be used to read from something like a socket.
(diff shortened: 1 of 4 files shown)


Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -370,7 +369,7 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
         
         localProgress?.becomeCurrent(withPendingUnitCount: Int64(fileSize))
         do {
-            let length = try readBytesFromFileDescriptor(fd, path: inPath, buffer: bytes, length: fileSize, chunkSize: size_t(preferredChunkSize), reportProgress: reportProgress)
```
parkera, line 373: The block size is just too small (4k on macOS, e.g.), and results in a very large number of `read` calls.

kperryua (approved):

jmschonfeld:
The linux test failure looks related:

```
Test Case 'DataIOTests.test_largeFile' started at 2024-05-20 18:29:52.910
/build/swift-foundation/Tests/FoundationEssentialsTests/DataIOTests.swift:215: error: DataIOTests.test_largeFile : XCTAssertEqual failed: ("2147549184") is not equal to ("2147479552") - 
Test Case 'DataIOTests.test_largeFile' failed (17.749 seconds)
```

parkera:
@swift-ci test

parkera:
Updated to provide a "read until EOF" option, true by default. On Darwin we were always reading the requested `0x7ffffff`, but on Linux we were reading `0x7fff000`, then failing to loop again. This mirrors the implementation from swift-corelibs-foundation.

itingliu:
LGTM! Can we merge this?

parkera:
Yah, I wanted @kperryua to take a second look if possible.

kperryua:
> Yah, I wanted @kperryua to take a second look if possible.

Took me a bit to grok what the readUntilEOF change is about. I take a little issue with the name, since it presumes that the `length` parameter is expected to be EOF of a regular file. That's only technically true in this particular case of Data reading, since we're first querying the length of the file on disk and then passing that parameter for length. In theory one could call this function with a length less than the full length of the file, if one has a limited buffer.

kperryua:
> it's more like `readUntilLength`

I like that better.

kperryua (approved):

PR #624, opened 2024-05-20
Adjust errors thrown from FileManager's copyItem function

jmschonfeld:
Due to an incorrect check of the `lstat` return value, the copy item functions can return incorrect errors when the source file/directory does not exist. This corrects that logic to ensure the appropriate error codes are thrown (and match the error codes in the old ObjC implementation)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #625, opened 2024-05-20
Handle ICU failures better

itingliu:
Instead of assuming ICU calls always succeed, surface the error back to callsites and fallback to some default value if we can't get ICU formatters.

Resovles 125918488


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #627, opened 2024-05-21
FoundationEssentials: strip leading `/`

compnerd:
When creating a file system representation from the path, we need to strip the leading `/` which is added for the RFC representation of the Windows path. Without this, the C runtime will fail with an invalid argument error. Unfortunately this regresses the test suite by causing a hang.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -138,7 +138,17 @@ extension String {
             return try block(buffer.baseAddress!)
         }
         #else
-        try self.withCString {
+        var iter = self.utf8.makeIterator()
+#if os(Windows)
+        let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
+#else
+        let bLeadingSlash = false
+#endif
+
+        // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
+        // leading slash indicates a rooted path on the drive for the current
+        // working directory.
+        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString {
             try block($0)
         }
         #endif
@@ -153,7 +163,16 @@ extension String {
             return try block(buffer.baseAddress!)
         }
         #else
-        var mut = self
+        var iter = self.utf8.makeIterator()
+#if os(Windows)
+        let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
+#else
+        let bLeadingSlash = false
+#endif
+
+        var mut: Substring =
+            Substring(self.utf8[self.utf8.index(self.utf8.startIndex, offsetBy: bLeadingSlash ? 1 : 0)...])
+
         return try mut.withUTF8 { utf8Buffer in
             // Leave space for a null byte at the end
             try withUnsafeTemporaryAllocation(of: CChar.self, capacity: utf8Buffer.count + 1) { temporaryBuffer in


Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -138,7 +138,13 @@ extension String {
             return try block(buffer.baseAddress!)
         }
         #else
-        try self.withCString {
+        var iter = self.utf8.makeIterator()
+        let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
```
jmschonfeld, line 142: do we need to exclude this behavior on linux? I think `/C:/foo` is technically a valid linux path, right?

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -138,7 +138,13 @@ extension String {
...
+
+        // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
+        // leading slash indicates a rooted path on the drive for the current
+        // working directory.
+        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString {
             try block($0)
         }
         #endif
```
jmschonfeld, line 150: Do we need to apply this to the `withMutableFileSystemRepresentation` function below as well?

Sources/FoundationEssentials/String/String+Internals.swift
compnerd: Technically, as per RFC8089, it is a URI, it shouldn't care about the host that is being accessed. The URI should work in either case. I don't think that we have a way to differentiate between a Windows path on Linux vs a Linux path on Windows (which would be treated as a relative path).

karwa: FWIW I wouldn't pay attention to RFC8089.

The real important principle for URL->FilePath conversion is that it should **never** produce a relative path. File URLs cannot encode relative paths at all.

jrflat: I do think we'd need to keep this change Windows-only since `/C:/foo` is a valid path on Linux and `C:/foo` would otherwise not represent that path

compnerd: This does mean that you cannot handle Windows paths on non-Windows platforms, which is unfortunate.

compnerd: What do we want to do about the cross-platform representation? It would be nice to be able to make progress on the remaining test failures.

jmschonfeld: For my reference, do you have an example of a situation in which we'd end up needing to deal with a windows path with a drive letter on Linux or other platforms? I'm not sure I'm familiar with a situation when needing to handle that cross platform representation comes up in practice

compnerd: Well, with cross-platform software, it is very helpful. As an example - I am running the test suite on Linux but would like to verify that the Windows path is not being broken for syntactic operations.

compnerd: What do we want to do here? If we use `#if os(Windows)`, and were to verify this path, we could only do so on Windows which makes it feel more fragile.

jmschonfeld: I thought it over for a bit and my opinion is that for now we should conditionalize this behavior behind `os(Windows)`. I can definitely see the value in being able to take this path on non-Windows with windows-style paths, but I think that ideally we'll gate that behind a different API rather than changing the behavior of the existing APIs. In the future, I could see a new option on the file system representation functions (or maybe on the `URL`/`FilePath` initializers) that take a path style allowing the client to pick between the windows or POSIX style like the `CFURL` APIs have on Darwin. But I think until we have a way for the client to specify the style we should match the style to the current platform to prevent changing Darwin/linux behavior while still doing the right thing on Windows

compnerd: Okay, I've conditionalised this for now.

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #628, opened 2024-05-21
FoundationEssentials: drop invalid modification dates

compnerd:
Follow the expected behaviour to drop invalid modification dates, e.g. `+NaN`. This matches the semantics that are expected on Unix and avoids a crash due to the initialiser.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #629, opened 2024-05-22
FoundationEssentials: correct directory enumeration on Windows

compnerd:
When enumerating a directory on Windows, we would fail to increment the iterator as we would increment _after_ the return. Insert a `defer`'ed increment on the iterator to allow us to track the state properly. Because the iteration was may have completed on the previous run, we need to track an additional bit of state as `bValid` which provides us an indication if iteration has completed.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
@@ -23,6 +23,7 @@ struct _Win32DirectoryContentsSequence: Sequence {
         }
 
         private var hFind: HANDLE?
+        private var bValid: Bool = true
         private var ffdData: WIN32_FIND_DATAW = .init()
         private var prefix: String = ""
         private var slash: Bool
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
```diff
@@ -80,6 +82,7 @@ struct _Win32DirectoryContentsSequence: Sequence {
 
                 let prefixed: String =
                     prefix + name + ((slash && ffdData.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY) ? #"\"# : "")
+                defer { bValid = FindNextFileW(hFind, &ffdData) }
                 return Element(fileName: name, fileNameWithPrefix: prefixed, dwFileAttributes: ffdData.dwFileAttributes)
             } while FindNextFileW(hFind, &ffdData)
             return nil
```
jmschonfeld, line 88: Should we set `bValid = false` here so that if the `FindNextFileW` from the `while` loop fails and we return `nil` that we don't produce values after that `nil` value if called again?

compnerd: The `repeat` is to skip over the `.` and `..` entries and re-run the iteration. Perhaps we should just add a `bValid = false` before the `return nil`.

jmschonfeld: Yeah adding that before the `return nil` would make sense to me, just to make sure that after one `next()` invocation returns `nil` that we don't accidentally try to read from `ffdData` on the following `next()` call

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #630, opened 2024-05-22
FoundationEssentials: correct the template path for Windows

compnerd:
We need to use the FSR for the template path to ensure that we are giving the underlying system a valid path. This was causing a spurious failure with the MSVC runtime due to the errant leading `/`.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -225,7 +225,7 @@ private func createProtectedTemporaryFile(at destinationPath: String, inPath: Pa
     }
 #endif
     
-    let temporaryDirectoryPath = destinationPath.deletingLastPathComponent()
+    let temporaryDirectoryPath = destinationPath.deletingLastPathComponent().withFileSystemRepresentation { String(cString: $0!) }
     let (fd, auxFile) = try createTemporaryFile(at: temporaryDirectoryPath, inPath: inPath, prefix: ".dat.nosync", options: options)
     return (fd, auxFile, nil)
 }


Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -225,7 +225,7 @@ private func createProtectedTemporaryFile(at destinationPath: String, inPath: Pa
     }
 #endif
     
-    let temporaryDirectoryPath = destinationPath.deletingLastPathComponent()
+    let temporaryDirectoryPath = destinationPath.deletingLastPathComponent().withFileSystemRepresentation { String(cString: $0!) }
```
jmschonfeld, line 228: Is the goal here just to "standardize" the path into something that doesn't have superfluous path separators? If so, we have some functions in `String+Path.swift` that do this - I think the FSR is a bit of a heavy weight here because it requires allocating a new buffer and (on Darwin) decomposing the text, so the String path functions for standardization might be better suited for this.

Additionally, this makes me wonder if the usage of `temporaryDirectoryPath` is wrong - are we just forgetting to get its FSR at the point where we use it? If possible, it'd be best to correct this at the usage site rather than pushing the standardization onto the callers

compnerd: It is less "standardize" and more "use the FSR". We are using `.path` which is not the FSR, so you end up with `/C:/...` which is not a valid template. If we can ensure that `.deletingLastPathComponent()` will not re-introduce the leading `/`, I think that might be fine to hoist up at a higher level.

jmschonfeld: Inside of `createTemporaryFile(at:inPath:prefix:options:)` I see we have

```swift
let template = directoryPath + prefix + pidString + ".XXXXXX"
// ...
try template.withMutableFileSystemRepresentation { templateFileSystemRep -> (Int32, String)? in
// ...
```

Shouldn't calling `withMutableFileSystemRepresentation` on the provided `template` take care of this issue already? I think I'd rather let `createTemporaryFile(at:)` do the right thing for all of its callers if possible, and I think your other PR would fix this since it calls wMFSR on the template - if not it sounds like maybe the use of wMFSR has a bug there?

compnerd: No, it appears that it is insufficient on its own, and this is still needed.

jmschonfeld: Do we know why? Something seems off to me - if this path is something like `/C:/foo/` and calling `withFileSystemRepresentation` corrects it to `C:/foo/`, but later when it becomes `/C:/foo/bar1234.XXXXXX` the FSR doesn't correct it to `C:/foo/bar1234.XXXXXX`, that seems like a bug right? This might address the problem here but if `withMutableFileSystemRepresentation` can't handle it, are we going to end up needing to do this elsewhere?

compnerd:
~~~
(lldb) p template
(String) "/C:/Users/abdulras/AppData/Local/Temp/420B3F9A-7B14-41E3-B884-FCB6A754977A/.dat.nosyncDD0.XXXXXX"
(lldb) memory read -a 0xea9cf1c800
0xea9cf1c800: 43 3a 2f 55 73 65 72 73 2f 61 62 64 75 6c 72 61  C:/Users/abdulra
0xea9cf1c810: 73 2f 41 70 70 44 61 74 61 2f 4c 6f 63 61 6c 2f  s/AppData/Local/
~~~

We don't have a FSR when we strip the leading `/` which means that we are passing an invalid path to the system.

jmschonfeld: I'm not sure I'm following, is that `0xea9cf1c800` address the address of `templateFileSystemRep` in `createTemporaryFile(at:)`? If so, it looks like that file system representation does have the leading slash dropped, and the return value is the `String` created from that pointer, so `auxFile` should be the string with the leading slash dropped. Am I missing the spot where we pass `template` directly to the system instead of using its FSR?

compnerd: Right, that is the memory address of the temporary string buffer (its really annoying that lldb doesn't have WinDBG's `da` command). The path here is passed along to the system and note that the slashes are not the system separator but rather the Unix separator.

jmschonfeld: Gotcha, so maybe what I'm not grasping is how the added call to `withFileSystemRepresentation` helps here - doesn't that have the same issue? Or if not, I think that means there's a behavioral difference between `withFileSystemRepresentation` and `withMutableFileSystemRepresentation` which would be a bug we should fix right? Sorry perhaps I'm still lost on the crux of the issue here

compnerd: Lets come back to this once we fix the `withFileSystemRepresentation` and `withMutableFileSystemRepresentation` (#627).

compnerd:
I think that a different change allows this to work - going to close this off for now.

PR #632, opened 2024-05-23
Use percentEncode(path, component: .path) instead of percentEncodePath(path)

nnabeyang:
In RFC3986Parser, while constructing the percent-encoded urlString, .user, .password, .host, .query, and .fragment all call `percentEncode(_:component:)`, but only .path calls `percentEncodePath(_:)`. 
This PR replaces `percentEncodePath(_:)` with `percentEncode(_:component:)` for consistency.


jmschonfeld:
@swift-ci please test

jrflat (approved):
LGTM thanks!

PR #633, opened 2024-05-23
Use Data+Reading for loading TZ files

jmschonfeld:
This code previously had platform-specific loading code for reading the TZ file, but now that we have Data reading APIs we can simply call `Data(contentsOfFile:)` instead which is a lot shorter and avoids a lot of platform if-defs


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test Windows

PR #635, opened 2024-05-24
FoundationEssentials: account for relative paths in iteration

compnerd:
We are not changing the working directory in between recursive calls. This is a problem when trying to enumerate the subpaths of a directory in the recursive call where the parent path was lost. Concatenate the parent path before recursing.


compnerd:
@swift-ci please test

PR #637, opened 2024-05-28
FoundationEssentials: limit autosave to user domain

compnerd:
During the review for the changes to support Windows, we realized that the location that is being used for the autosave information is only available within the user domain mask. The Windows implementation was corrected as part of the implementation, but this corrects the XDG compliant path to reflect this reality.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #639, opened 2024-05-29
FoundationEssentials: correct path canonicalisation handling on Windows

compnerd:
Strip the extended path prefix for bypassing the Win32 API layer. This is guaranteed to be prefixed on strings as per the documentation.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -665,7 +665,9 @@ extension String {
                 guard GetFinalPathNameByHandleW(hFile, $0.baseAddress, dwLength, VOLUME_NAME_DOS) == dwLength - 1 else {
                     return nil
                 }
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+
+                // When using `VOLUME_NAME_DOS`, the returned path uses `\\?\`.
+                return String(String(decodingCString: $0.baseAddress!.advanced(by: 4), as: UTF16.self))
             }
         }
 #else
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -667,7 +667,7 @@ extension String {
                 }
 
                 // When using `VOLUME_NAME_DOS`, the returned path uses `\\?\`.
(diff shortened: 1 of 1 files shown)


Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -665,7 +665,9 @@ extension String {
                 guard GetFinalPathNameByHandleW(hFile, $0.baseAddress, dwLength, VOLUME_NAME_DOS) == dwLength - 1 else {
                     return nil
                 }
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+
+                // When using `VOLUME_NAME_DOS`, the returned path uses `\\?\`.
+                return String(String(decodingCString: $0.baseAddress!, as: UTF16.self).dropFirst(4))
```
jmschonfeld, line 670:
```suggestion
                return String(decodingCString: $0.baseAddress!.advanced(by: 4), as: UTF16.self)
```

Are we able to advance the pointer here instead of calling `dropFirst` to avoid needing to create an intermediate `String` and copy to a new `String` (just a minor performance optimization)

compnerd: Yes, that is a good idea!

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -665,7 +665,9 @@ extension String {
                 guard GetFinalPathNameByHandleW(hFile, $0.baseAddress, dwLength, VOLUME_NAME_DOS) == dwLength - 1 else {
                     return nil
                 }
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+
+                // When using `VOLUME_NAME_DOS`, the returned path uses `\\?\`.
+                return String(String(decodingCString: $0.baseAddress!.advanced(by: 4), as: UTF16.self))
```
jmschonfeld, line 670:
```suggestion
                return String(decodingCString: $0.baseAddress!.advanced(by: 4), as: UTF16.self)
```


I think there's one extra wrapping `String` initializer here 🙂

jmschonfeld (approved):

PR #641, opened 2024-05-29
Return file attributes as NSNumbers when SCL-F is imported

jmschonfeld:
Clients of the existing SCL-F expect to be able to cast file attribute values to any numeric type via `NSNumber` implicit bridging. This works in Foundation.framework today and this PR enables this functionality when SCL-F is imported on non-Darwin platforms by dynamically checking to see if we can initialize an `NSNumber` (falling back to the existing concrete numeric type when `Foundation` is not loaded.


jmschonfeld:
@swift-ci please test

jmschonfeld:
The `NSNumber` initializer is added to SCL-F in https://github.com/apple/swift-corelibs-foundation/pull/4966

itingliu (approved):

iCharlesHu (approved):

PR #642, opened 2024-05-30
FoundationEssentials: repair `readlink` implementation on Windows

compnerd:
The existing code did not properly read the structure (possibly due to copying). This adjustst he path to use `UnsafePointer` to use that to read the data from the response of the `ioctl` call.


compnerd:
@swift-ci please test

Sources/_CShims/include/io_shims.h
```diff
@@ -75,5 +77,13 @@ typedef struct _REPARSE_DATA_BUFFER {
     };
 } REPARSE_DATA_BUFFER;
 
+static inline intptr_t GetSymbolicLinkReparseBufferPathBufferOffset(void) {
```
jmschonfeld, line 80: For the other shims, we've been prefixing functions with something like `_ioshims_` should we follow suit here (@parkera I can't remember if that's required for some reason or just styling. Since these are `static inline` I don't think they'd have symbols so perhaps it doesn't matter for these?)

parkera: A styling choice to make it obvious where we are using them (and hopefully can get rid of it in the future).

compnerd: I don't see these as being possible to remove anytime soon - there is no equivalent of `offsetof`.

jmschonfeld: Does `MemoryLayout<REPARSE_DATA_BUFFER>.offset(of: \.SymbolicLinkReparseBuffer.PathBuffer)` work here instead of the c shim?

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -55,9 +55,11 @@ extension _FileManagerImpl {
         _ = fileManager.fileExists(atPath: destPath, isDirectory: &bIsDirectory)
 
         try path.withNTPathRepresentation { lpSymlinkFileName in
-            try destPath.withNTPathRepresentation { lpTargetFileName in
-                if CreateSymbolicLinkW(lpSymlinkFileName, lpTargetFileName, SYMBOLIC_LINK_FLAG_ALLOW_UNPRIVILEGED_CREATE | (bIsDirectory ? SYMBOLIC_LINK_FLAG_DIRECTORY : 0)) == 0 {
-                    throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: false)
+            try destPath.withFileSystemRepresentation {
+                try String(cString: $0!).withCString(encodedAs: UTF16.self) { lpTargetFileName in
```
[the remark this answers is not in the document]
compnerd: Yes, there is a big difference: the `withNTPathRepresentation` will convert the path to absolute where as the `withFileSystemRepresentation` will not.

Sources/_CShims/include/io_shims.h
compnerd: It does not, we end up with `nil` for the offset (the type is not exactly the easiest on the clang importer).

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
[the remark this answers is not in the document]
jmschonfeld: Gotcha. This seems fine for now, but in the future we should come back to this and add a version of `withFileSystemRepresentation` that creates a UTF-16 string for this purpose. The underlying mechanisms of `withFileSystemRepresentation` already support output to an arbitrary encoding, we'd just need a new API that provides a buffer of the right element type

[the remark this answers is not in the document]
compnerd: Seems reasonable - I think that having a unicode version of that would be very convenient.

Sources/_CShims/include/io_shims.h
jmschonfeld: Gotcha, in that case could you update the naming of these functions to follow the naming convention of the other cshim functions to try to keep them relatively consistent? Something along the lines of `_ioshims_symboliclinkreparsebuffer_pathbufferoffset` maybe?

compnerd:
@swift-ci please test

PR #644, opened 2024-05-31
Prevent throwing when applying file attributes retrieved via `attributesOfItem(atPath:)`

jmschonfeld:
There are some scenarios in which you want to copy attributes from one file to another (for example, from within `copyItem(atPath:toPath:)`. However currently, calling `setAttributes(_:forItemAtPath:)` with attributes retrieved from `attributesForItem(atPath:)` will `throw` because we can fetch `.creationDate` but we cannot set it (yet). Until this functionality is implemented in swift-foundation, we exclude `.creationDate` from the `featureUnsupported` check to ensure that applying attributes that Foundation provides will succeed.

This prevents throwing an error from `copyItem` calls made by SwiftPM when using the new re-cored Foundation

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -112,15 +112,19 @@ extension _FileManagerImpl {
         #endif
     }
     
-    // Exclude .creationDate from this check because it is unconditionally returned from attributesOfItem(atPath:)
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
I updated the logic slightly to ensure that `creationDate` is still a relevant key for `FOUNDATION_FRAMEWORK` callers

PR #646, opened 2024-05-31
Avoid crashing due to malformed paths creating URLs

jmschonfeld:
In some cases, file manager APIs may be called with malformed string paths. We correctly `throw` an error in those cases, but when throwing an error we try to initialize a `URL` from the path. In certain `FOUNDATION_FRAMEWORK` scenarios this leads to calling into an `NSURL` initializer which crashes. This code works around that crash by creating an internal failable path we can take instead when we know we're in a scenario where we may have malformed paths.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #648, opened 2024-06-03
FoundationEssentials: correct file overwriting behaviour

compnerd:
We would previously always overwrite the file which would not match the expectation. Correct the flags to allow us to properly create or overwrite the file.


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #651, opened 2024-06-04
Add compatibility path for String(bytes:encoding:)

parkera:
Add a compatibility path for `String(bytes:encoding:)`. Some clients expected this to correctly handle null-terminated strings. This patch falls back on `NSString`'s behavior of producing a string with embedded nulls in some cases.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #654, opened 2024-06-04
FoundationEssentialsTests: we convert to FSR when checking

compnerd:
The internal function returns a FSR value where as the value being returned from the `appendingPathComponent` is not the FSR. Explicitly convert to the FSR for the check.


compnerd:
I'm not sure if there is a more elegant manner to handle this scenario. This value is being compared to an internal API which returns a FSR representation so it seems reasonable that the check would convert to the FSR when trying to check the result.

compnerd:
@swift-ci please test

jmschonfeld:
> I'm not sure if there is a more elegant manner to handle this scenario. This value is being compared to an internal API which returns a FSR representation so it seems reasonable that the check would convert to the FSR when trying to check the result.

Yeah normally I'd want to avoid this and use a different API instead, but given we're testing an internal function I think it's probably fine. Presumably converting to the file system representation is normalizing something about the path?

compnerd:
Correct - is the path separator being canonicalized.

jmschonfeld (approved):

PR #657, opened 2024-06-04
FoundationEssentialsTests: repair the test suite on Windows

compnerd:
We would use the Unix (non-portable) arc separator (`/`). On Windows, we need to ensure that we account for the proper arc separator (`\`) when checking the FSR of the string.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #658, opened 2024-06-05
Prevent `IntegerParseStrategy` from crashing when encountering OOB numbers

jmschonfeld:
Internally, `IntegerParseStrategy` currently parses as `Int64` or `Double` (since that's what ICU supports) and then casts to the format output type. However, if the format output type is an unsigned integer or an integer of width less than 64 bits, this causes a crash. We should instead detect this and throw an error during parsing if the parsed number does not fit within the bounds of the specified output type.


jmschonfeld:
@swift-ci please test

parkera (approved):

itingliu (approved):
Thanks!!

PR #659, opened 2024-06-05
FoundationEssentials: tweak `removeFile(_:with:)` on Windows

compnerd:
Ensure that we call the callbacks at the appropriate time. Change the indentation and encapsulating scope to avoid a double callback.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -317,6 +317,12 @@ enum _FileOperations {
         try path.withNTPathRepresentation {
             var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
             guard GetFileAttributesExW($0, GetFileExInfoStandard, &faAttributes) else {
+                // NOTE: in the case that the 'stat' failed, we want to ensure
+                // that we query if the item should be removed. If the item
+                // should not be removed, we can continue, else we should check
+                // if we should proceed after the error.
+                guard filemanager?._shouldRemoveItemAtPath(path) ?? true else { return }
+
                 let error = CocoaError.removeFileError(GetLastError(), path)
                 guard (filemanager?._shouldProceedAfter(error: error, removingItemAtPath: path) ?? false) else {
                     throw error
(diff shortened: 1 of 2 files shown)


compnerd:
I recommend that you look at this with the whitespace diffs ignored as there was an indentation change that causes the diff to look more complicated than it is.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -342,52 +348,69 @@ enum _FileOperations {
...
-                if RemoveDirectoryW($0) { return }
-                let dwError: DWORD = GetLastError()
-                guard dwError == ERROR_DIR_NOT_EMPTY else {
-                    let error = CocoaError.removeFileError(dwError, directory)
-                    guard (filemanager?._shouldProceedAfter(error: error, removingItemAtPath: directory) ?? false) else {
-                        throw error
+            var stack = [path]
+            while let directory = stack.popLast() {
+                try directory.withNTPathRepresentation {
+                    let path = String(decodingCString: $0, as: UTF16.self)
```
jmschonfeld, line 355: Would we be able to pick a new name for this variable? It shadows the `path` from the function scope and it was a bit confusing to me seeing this here because I thought that `path` below was the one provided to the function instead of the decoded directory, maybe `standardizedDirectory` or something to denote this is the same "path" as `directory`?

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -342,52 +348,69 @@ enum _FileOperations {
...
-                            if entry.dwFileAttributes & FILE_ATTRIBUTE_READONLY == FILE_ATTRIBUTE_READONLY {
-                                guard SetFileAttributesW($0, entry.dwFileAttributes & ~FILE_ATTRIBUTE_READONLY) else {
-                                    throw CocoaError.removeFileError(GetLastError(), path)
+                    for entry in _Win32DirectoryContentsSequence(path: directory, appendSlashForDirectory: false, prefix: [directory]) {
+                        try entry.fileNameWithPrefix.withNTPathRepresentation {
+                            let path = String(decodingCString: $0, as: UTF16.self)
+
+                            if entry.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY,
+                                    entry.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT != FILE_ATTRIBUTE_REPARSE_POINT {
+                                if filemanager?._shouldRemoveItemAtPath(path) ?? true {
```
jmschonfeld, line 375: I think we check this whenever we pop something off of the stack which means we'd ask about this directory twice. Is that intentional?

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -342,52 +348,69 @@ enum _FileOperations {
...
+                                        guard (filemanager?._shouldProceedAfter(error: error, removingItemAtPath: entry.fileNameWithPrefix) ?? false) else {
+                                            throw error
+                                        }
                                     }
                                 }
                             }
                         }
                     }
+
+                    if RemoveDirectoryW($0) { return }
```
jmschonfeld, line 405: Which directory does this call remove? Since we put the original `path` on the stack, shouldn't iterating the stack take care of everything?

Sources/FoundationEssentials/FileManager/FileOperations.swift
compnerd: Sure, renamed to `ntpath`.

Sources/FoundationEssentials/FileManager/FileOperations.swift
compnerd: Right, that was part of the problem. We now will do a clean up post-traversal which ensures that the question is only asked once. Fortunately, the test coverage ensures that we get this correct.

Sources/FoundationEssentials/FileManager/FileOperations.swift
compnerd: This is related to the previous point - we need to do this to ensure that we do not prompt twice for the same directory. The Unix side takes care of this through FTS having a post-order traversal callback.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
jmschonfeld: I'm not certain the unit test actually tests this case, something like:

- foo (DIR)
    - bar (DIR)
        - baz (FILE)
 
In that case, I think we:

- Pop the first thing off the stack (foo) and ask if it should be removed
- Since it's a directory we enumerate its contents
    - While doing this, we see bar is a directory and ask if it should be removed, then put it on the stack
- Then we hit the remove call at the end which fails because foo isn't empty
- Then we pop the next thing off the stack (bar) and ask if it should be removed (a second time)
- It's not empty so we iterate it's contents
    - When we see baz we see it's a file and immediately ask and remove it
- Then we fall down to the remove call at the end which removes bar because it is now empty

I think that leaves us with not removing foo and asking about bar twice right? I think the case that would hit this is removing a directory with a nested directory specifically

compnerd: Added that as a test case and tweaked the logic to avoid the duplicate check.

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #660, opened 2024-06-05
Mark Bundle stub as package instead of public

jmschonfeld:
This is only an implementation detail and we don't want it to become ambiguous with the `Bundle` from SCL-F currently


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #662, opened 2024-06-06
FoundationEssentials: silence a warning

compnerd:
Convert a `var` to `let` as it was originally. This variable is no longer mutated and causes a warning.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #663, opened 2024-06-06
String(bytes:encoding:) should not skip BOM on UTF16 to UTF8 conversion

parkera:
When converting from UTF16 BE/LE or UTF32 BE/LE, we should leave the BOM in place when converting to UTF8. This enables strings round-tripped through `PropertyListEncoder` (which uses explicit UTF16-BE for file format compatibility reasons) to compare equal.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #665, opened 2024-06-06
FoundationEssentials: ensure that the path exists before iteration

compnerd:
When performing a `subpathsOfDirectory`, ensure that the path exists. On the non-Windows path, `_FTSSequence` will perform the initial `statvfs` during construction which obscured this requirement. We instead explicitly inline this call in the implementation. This repairs further test failures.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #666, opened 2024-06-06
FoundationEssentials: handle directory diff properly

compnerd:
We need to use `FILE_FLAG_BACKUP_SEMANTICS` to ensure that we are able to diff directories as well. Previously we would mishandle this case and was caught by the test suite.

Thanks to Jeremy Day for helping resolve this!


compnerd:
@swift-ci please test

jmschonfeld (approved):

compnerd:

Issue #613, opened 2024-05-16
Unexpected JSON decoding behavior of Int64


kcatigbe-cbs:
### Description
(Forum thread discussing this issue:  https://forums.swift.org/t/unexpected-decoding-behavior-of-int64/71762)

When decoding an explicit signed 64 bit value (Int64), if that value is smaller than `Int64.min`, I am seeing the value being clamped to `Int64.min`.

I am running this in Xcode 15.3 / M1Pro / Sonoma, but the same behavior happens in 15.2. Not sure when this was introduced, but we have some unit tests now failing due to this.

For example, if my JSON contains `"value" = -9223372036854775809`, the decoder will assign my struct property the value `-9223372036854775808`

I believe this might be unexpected because decoding an Int32 with a value less than Int32.min throws a JSON Decoding error:
`"Number -2147483649 is not representable in Swift."`

Performing a similar decode with JSONSerialization doesn't suffer from this, although it exhibits the classic integer overflow (circling back to the positive value)

This behavior isn't present when testing against `Int64.max`, as that will throw an error as expected.

### Reproduction
```swift
do {
    print("-----  MIN 64-Bit Int: \(Int64.min)")
    let json = "{ \"value\": -9223372036854775809 }"
    struct Tester: Codable {
        let value: Int64
    }

    // JSONSerialize
    let serial = try JSONSerialization.jsonObject(with: Data(json.utf8)) as! [String: Any]
    
    let value: NSDecimalNumber = serial["value"] as! NSDecimalNumber
    let v: Int64 = value.int64Value
    print("Type: \(type(of: value)) - value: \(value) - Int64 rep: \(v)")
    
    // Codable
    let tester = try JSONDecoder().decode(Tester.self, from: Data(json.utf8))
    print("Clamped Value: \(tester.value) - Int64.min: \(Int64.min)")
    
} catch {
    print("Error: \(error)")
}
```

### Expected behavior
Expected to throw a JSON Decoding error: `"Number -9223372036854775809 is not representable in Swift."`

### Environment
Swift version 5.10 (swift-5.10-RELEASE)

### Additional information
Based on the discussion in the forums and with my own local experimentation, it seems to be related to 
`func _slowpath_unwrapFixedWidthInteger<T: FixedWidthInteger>(...)` where the value is instantiated with:
`T(exactly: double)`, where using `Int64(exactly:)` has a range of values where it's returning `Int64.min`.
[...]

PR #667, opened 2024-06-07
Limit IntegerParseStrategy's fallback Double path to lossless integer range

oscbyspro:
Hello friends, we meet again!

I've come to propose some changes to IntegerParseStrategy's fallback Double path. Ideally, it should never need to go through Double, but that is out-of-scope for this contribution. Today, I'll only suggest that you reject magnitudes >=(2^53).

In short, Double is a 52-bit window that slides around. The greater the value, the greater the distance between adjacent values. This distance is <=(1) for values in [-2^53, 2^53]. Outside this range, the parseAsDouble(_:) rounds its input String to the nearest Double before the  Double is passed to BinaryInteger.init(exactly:). The lossless range is, therefore, 52 bits since ±(2^53) may represent ±(2^53+1) rounded towards zero. The rounding error is ±(1024) when parsing UInt64 values through Double. Here's a brief illustration:

```swift
Double("9223372036854775808") // 2^63
Double("9223372036854776832") // 2^63 + 1024 becomes 2^63
Double("9223372036854776833") // 2^63 + 1025 becomes 2^63 + 2048
Double("9223372036854777856") // 2^63 + 2048
```

### Proposed solution

Reject magnitudes >=(2^53) in IntegerParseStrategy's fallback Double path.

### Additional comments

1. This rounding issue is similar to the one mentioned in (#613) but w.r.t. IntegerParseStrategy rather than JSONDecoder.
2. This path may or may not do something when limited to values that fit in an Int64. Somebody more familiar with ICU may know. In either case, this contribution adds tests that exemplify the rounding errors that Double may cause.

### DIFF
--- a/Tests/FoundationInternationalizationTests/Formatting/NumberParseStrategyTests.swift
+++ b/Tests/FoundationInternationalizationTests/Formatting/NumberParseStrategyTests.swift
@@ -222,17 +222,17 @@ final class NumberParseStrategyTests : XCTestCase {
             // TODO: Parse integers greater than Int64
             let format: IntegerFormatStyle<UInt64> = .init(locale: locale)
             let parseStrategy = IntegerParseStrategy(format: format, lenient: true)
-            XCTAssertEqual(try parseStrategy.parse(UInt64.min.formatted(format)), UInt64.min)
-            //XCTAssertEqual(try parseStrategy.parse(UInt64.max.formatted(format)), UInt64.max)
+            XCTAssertEqual(      try parseStrategy.parse(UInt64.min.formatted(format)), UInt64.min)
+            XCTAssertThrowsError(try parseStrategy.parse(UInt64.max.formatted(format)))
             XCTAssertThrowsError(try parseStrategy.parse("-1"))
             XCTAssertThrowsError(try parseStrategy.parse("18446744073709551616"))
             
             // TODO: Parse integers greater than Int64
             let maxInt64 = UInt64(Int64.max)
-            XCTAssertEqual(try parseStrategy.parse((maxInt64 + 0).formatted(format)), maxInt64 + 0) // not a Double
-            //XCTAssertEqual(try parseStrategy.parse((maxInt64 + 1).formatted(format)), maxInt64 + 1) // exact Double
-            //XCTAssertEqual(try parseStrategy.parse((maxInt64 + 2).formatted(format)), maxInt64 + 2) // not a Double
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/NumberParseStrategyTests.swift
```diff
@@ -192,22 +192,68 @@ final class NumberParseStrategyTests : XCTestCase {
...
+        
+        do {
+            // TODO: Parse integers greater than Int64
             let format: IntegerFormatStyle<UInt64> = .init(locale: locale)
             let parseStrategy = IntegerParseStrategy(format: format, lenient: true)
-            XCTAssertEqual(try parseStrategy.parse(0.formatted(format)), 0)
-            let aboveInt64Max = UInt64(Int64.max) + 1
-            XCTAssertEqual(try parseStrategy.parse(aboveInt64Max.formatted(format)), aboveInt64Max)
+            XCTAssertEqual(try parseStrategy.parse(UInt64.min.formatted(format)), UInt64.min)
+            //XCTAssertEqual(try parseStrategy.parse(UInt64.max.formatted(format)), UInt64.max)
```
itingliu, line 226: What's the issue with this? Does this currently crash?

Anyways if we are going to leave the tests commented out, a comment with an issue number would be useful

oscbyspro: Values greater than Int64.max go through Double before this patch, but such values are considered out-of-bounds after this patch. 

Int64.max + 1 is an exact Double, so the current fallback path formats it correctly. The problem is that the subsequent 2047 values are not Double(s) so they don't get formatted as expected. I saw no tests covering integers between Double(s), so I've assumed the current Int64.max + x behavior is an oversight. I could assert that it throws an error, perhaps with some comments about ICU limitations, but that seems like a temporary solution.

I suppose this matters more when decoding JSON. I thought about amending the JSON stuff first since there's a code comment about losslessness there, but it is much easier to discuss the formatting case since it has a straightforward test suite.

itingliu: Thanks yeah that's reasonable. Can we add the test back and assert that it would now throw? Same applies to other commented-out tests below

itingliu:
@swift-ci please test

itingliu (approved):
Looks great. Thanks!

PR #668, opened 2024-06-07
FoundationEssentials: handle canonical and non-canonical paths

compnerd:
In the case that the path is canonicalised, the `/` would be replaced with `\`. The leading `\` would interfere with our ability to recognise the absolute path representation and thus get the wrong response.


compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #669, opened 2024-06-07
FoundationEssentials: canonicalise the arc separator more aggressively

compnerd:
When trying to get the file system representation of a string, we may have previously used `appendingPathComponent(_:)` which will have used the incorrect separator. Ensure that all the separators are properly converted prior to returning the path.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -138,19 +138,20 @@ extension String {
             return try block(buffer.baseAddress!)
         }
         #else
-        var iter = self.utf8.makeIterator()
 #if os(Windows)
+        var iter = self.utf8.makeIterator()
         let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
-#else
-        let bLeadingSlash = false
-#endif
-
         // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
-        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString {
+        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).replacing(._slash, with: ._backslash).withCString {
             try block($0)
         }
+#else
+        return try withCString {
+            try block($0)
+        }
+#endif
         #endif
     }
     
@@ -163,15 +164,15 @@ extension String {
             return try block(buffer.baseAddress!)
         }
         #else
-        var iter = self.utf8.makeIterator()
 #if os(Windows)
+        var iter = self.utf8.makeIterator()
         let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
-#else
-        let bLeadingSlash = false
-#endif
-
-        var mut: Substring =
+        var mut: String =
             Substring(self.utf8[self.utf8.index(self.utf8.startIndex, offsetBy: bLeadingSlash ? 1 : 0)...])
+                .replacing(._slash, with: ._backslash)
+#else
+        var mut: String = self
+#endif
 
         return try mut.withUTF8 { utf8Buffer in
             // Leave space for a null byte at the end


compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -138,20 +138,20 @@ extension String {
...
         let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
-#else
-        let bLeadingSlash = false
-#endif
-
         // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
-        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString {
+        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).replacing("/", with: "\\").withCString {
```
jmschonfeld, line 147: Would it be better to adjust `appendingPathComponent` to use the `\` separator instead of the `/` separator, or do you think it's better that we always work with `/` separators until converting to a file system representation like we do here? (cc @jrflat since the former might update `URL`'s behavior I think)

jrflat: I think we do want to use `/` in `URL` up until converting to FSR, where we then convert to `\` on Windows. The `URL` APIs such as `.appending(path:)` handle inputs containing either `/` or `\` for Windows and standardizes them to `/`.

jmschonfeld: Ok cool thanks for the context - yeah then this change makes sense to me to only swap the separators when going to a file system representation. Looks like CI hit some build failures but the approach to the change seems fine to me then

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -138,19 +138,20 @@ extension String {
...
```
jrflat, line 147: There's an internal `StringProtocol` extension that's much faster if you're replacing single ascii characters: `internal func replacing(_ a: UInt8, with b: UInt8)`

Sources/FoundationEssentials/String/String+Internals.swift
jrflat, line 1: Do we also want to make this change for `withMutableFileSystemRepresentation`?

compnerd: Probably would be a good idea to do that, but not sure how to do that given the buffer.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
jmschonfeld: I think we can take a similar approach, just using

```swift
var mut: String = Substring(/* ... */).replacing(._slash, with: ._backslash)
```

and then using `mut.withUTF8` as it does now to create the mutable buffer based on the replaced string - right?

compnerd: Ah, do it early rather than later. That is fair. I was trying to do it after the duplication, but I suppose that doesn't really matter much in the grand scheme of things.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -170,8 +171,9 @@ extension String {
         let bLeadingSlash = false
 #endif
 
-        var mut: Substring =
+        var mut: String =
             Substring(self.utf8[self.utf8.index(self.utf8.startIndex, offsetBy: bLeadingSlash ? 1 : 0)...])
+                .replacing(._slash, with: ._backslash)
```
jrflat, line 176: We probably need to guard this with `os(Windows)`, too

compnerd: Yeah, that is a good point.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -138,19 +138,20 @@ extension String {
...
         let bLeadingSlash = if iter.next() == ._slash, iter.next()?.isLetter ?? false, iter.next() == ._colon { true } else { false }
-#else
-        let bLeadingSlash = false
-#endif
-
         // Strip the leading `/` on a RFC8089 path (`/[drive-letter]:/...` ).  A
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
-        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString {
+        return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).replacing(._slash, with: ._backslash).withCString {
```
jrflat, line 147: Ah sorry one last thing. I think getting the substring directly from `self` using `try self[utf8.index(startIndex, offsetBy: bLeadingSlash ? 1 : 0)...].replacing(...` is more efficient because it doesn't create and mutate a UTF8 array and create a new substring from that. (I could be wrong or it might be compiled out, though.) This would apply for `mut` below as well

jmschonfeld: I think it should effectively be the same because `self.utf8.dropFirst(_:)` returns a `Substring.UTF8View` (so it's just a substring's view into the original storage - this is O(1)), and then I believe the `Substring(Substring.UTF8View)` initializer just "adopts" the UTF8 view as its backing store (also O(1)). So performance wise I'd expect these to be equivalent since they should all effectively be constant time (and won't trigger a copy of the UTF-8 buffer until the call to `replacing(_:with:)`

jrflat: Oh oops you're totally right that makes sense. I just tested it and the performance is the exact same :)

jrflat (approved):

jmschonfeld (approved):

PR #671, opened 2024-06-09
FoundationEssentials: adjust error code constants for Windows

compnerd:
The Windows API calls return `ERROR_TOO_MANY_OPEN_FILES` (4) which is "The system cannot open the file.". One of the tests ensures that the return code is 260 which does not match reality.

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -129,6 +129,14 @@ extension CocoaError {
         )
     }
 
+    static func errorWithFilePath(_ code: CocoaError.Code, _ path: String, variant: String? = nil, userInfo: [String : AnyHashable] = [:]) -> CocoaError {
+        var info: [String:AnyHashable] = userInfo.addingUserInfo(forPath: path)
+        if let variant {
+            info[NSUserStringVariantErrorKey] = [variant]
+        }
+        return CocoaError(code, userInfo: info)
+    }
+
 #if os(Windows)
     static func errorWithFilePath(_ path: PathOrURL, win32 dwError: DWORD, reading: Bool, variant: String? = nil, userInfo: [String : AnyHashable] = [:]) -> CocoaError {
         switch path {
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -31,6 +31,14 @@ internal import QuarantinePrivate
 internal import _CShims
 
 extension CocoaError {
+    fileprivate static func fileOperationError(_ code: CocoaError.Code,  _ sourcePath: String, _ destinationPath: String? = nil, variant: String? = nil) -> CocoaError {
+        var info: [String : AnyHashable] = [NSSourceFilePathErrorKey:sourcePath]
+        if let destinationPath {
+            info[NSDestinationFilePathErrorKey] = destinationPath
+        }
+        return CocoaError.errorWithFilePath(code, sourcePath, variant: variant, userInfo: info)
+    }
+
 #if os(Windows)
     private static func fileOperationError(_ dwError: DWORD, _ suspectedErroneousPath: String, sourcePath: String? = nil, destinationPath: String? = nil, variant: String? = nil) -> CocoaError {
         var path = suspectedErroneousPath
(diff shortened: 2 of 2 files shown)


Sources/FoundationEssentials/Error/ErrorCodes.swift
```diff
@@ -30,7 +30,11 @@ extension CocoaError.Code {
         return CocoaError.Code(rawValue: 259)
     }
     public static var fileReadNoSuchFile: CocoaError.Code {
+#if os(Windows)
+        return CocoaError.Code(rawValue: 4)
```
jmschonfeld, line 34: This feels like the wrong place to adjust this, this would cause `.fileNoSuchFile == .fileReadNoSuchFile` to be true on Windows but false on other platforms - I suspect this could lead to unexpected behavior in cross-platform clients.

Should we instead update the mapping from Windows error codes to `CocoaError.Code` in order to account for this, or should we update the location where the error is thrown to throw a different error?

compnerd: I don't think that mapping is the right answer - the error codes may be used in different ways in different APIs. Updating the location where the error is thrown might be reasonable though.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -788,7 +788,7 @@ enum _FileOperations {
         try src.withNTPathRepresentation { pwszSource in
             var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
             guard GetFileAttributesExW(pwszSource, GetFileExInfoStandard, &faAttributes) else {
-                throw CocoaError.copyFileError(GetLastError(), src, dst)
+                throw CocoaError.errorWithFilePath(.fileReadNoSuchFile, src)
```
jmschonfeld, line 791: Thanks for updating, this looks like a better location to address the issue. As for which error to throw, `copyFileError` attaches a handful of extra keys to the user info. The unit tests might not test for them specifically but for consistency, should we make a variant of `copyFileError` that takes a `CocoaError.Code` instead of an errno which adds the extra keys like the variant / destination?

compnerd: I guess I'm missing what the consistency would be with. I don't really understand what the expectation is here, so I'm just relying on the tests to tell me if the error object is properly constructed.

jmschonfeld: It'd be consistency with the behavior across other platforms. You're right today we don't have a unit test for that. We have

```
XCTAssertThrowsError(try $0.copyItem(atPath: "does_not_exist", toPath: "dir3")) {
    XCTAssertEqual(($0 as? CocoaError)?.code, .fileReadNoSuchFile)
}
```

But that only tests the code and not the other keys like the source/destination file and the variant. We could probably update the unit test to account for those but I think the expectation here in my mind is that it throws the same error as Darwin/Linux which both currently set the variant to "Copy" and provide source/destination paths in the user info. It might not be critical or load bearing, but since we're already working here to try to maintain consistent behavior do you think we should update this to throw the same error contents in entirety rather than losing the other contents while getting the code right?

compnerd: Hmm, that type of uniformity is definitely preferable. I can look into adding the extra initialiser for that.

compnerd:
@swift-ci please test

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -129,6 +129,18 @@ extension CocoaError {
         )
     }
 
+    static func errorWithFilePath(_ code: CocoaError.Code, _ path: String? = nil, variant: String? = nil, userInfo: [String : AnyHashable] = [:]) -> CocoaError {
+        var info: [String : AnyHashable] = userInfo
+        if let path, info[NSFilePathErrorKey] == nil {
```
jmschonfeld, line 134: Sorry to nitpick on this, but instead of setting `NSFilePathErrorKey` directly here, could we use `userInfo.addingUserInfo(forPath: path)`? This will also take care of adding `NSURLErrorKey` and will help minimize duplicated code. Otherwise this changes looks good to me

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@jmschonfeld - anything else left for this? I have one more change after this for `FileManager`, and I think that we are nearly done!

jmschonfeld (approved):
Nope this looks good now, thanks! And that's great!

PR #672, opened 2024-06-10
Potential stack buffer overflow in `_withFixedCharBuffer`

itingliu:
In `_withFixedCharBuffer` helper, we allocate a buffer with the requested size + 1 (for null-termination). With the assumption that the closure's return value `len <= the requested size`, the null-terminated character should be on index `len`, not `len + 1`.

Fixes rdar://129154192


itingliu:
@swift-ci please test

PR #674, opened 2024-06-12
Re-enable tests for Duration based DiscreteFormatStyle conformances on 32 bit platforms

theMomax:
This test was unnecessary disabled for 32 bit platforms.


theMomax:
@swift-ci please test

itingliu (approved):

PR #676, opened 2024-06-12
FoundationEssentials: relax error handling for `SHGetFolderPath`

compnerd:
When running in CI, we would see failures due to the environment variables being changed. This relaxes the precondition to allow the test failure to propagate.

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -726,7 +726,6 @@ final class FileManagerTests : XCTestCase {
             .downloadsDirectory,
             .moviesDirectory,
             .musicDirectory,
-            .picturesDirectory,
             .sharedPublicDirectory
         ], exists: true)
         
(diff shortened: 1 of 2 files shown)


compnerd:

jmschonfeld:
```
Test Case 'FileManagerTests.testSearchPaths' started at 2024-06-12 17:11:56.556

C:\source\swift-foundation\Tests\FoundationEssentialsTests\FileManager\FileManagerTests.swift:718: error: FileManagerTests.testSearchPaths : XCTAssertEqual failed: ("false") is not equal to ("true") - Directory picturesDirectory produced an unexpected number of paths (expected to exist: true, produced: [])

Test Case 'FileManagerTests.testSearchPaths' failed (0.015 seconds)
```

Looks like we should update the unit test to account for this - looks like the pictures directory doesn't exist on CI

compnerd:
@swift-ci please test

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -775,6 +774,10 @@ final class FileManagerTests : XCTestCase {
         #if !os(watchOS) && !os(tvOS)
         assertSearchPaths([.trashDirectory], exists: (isMacOS && isFramework) || (!isDarwin && !isWindows))
         #endif
+
+#if !os(Windows)
+        assertSearchPaths([.picturesDirectory], exists: true)
```
jmschonfeld, line 779: Could we add a comment here saying that it's not tested on Windows because the directory doesn't exist in CI but likely exists on desktop machines? Just to make it clearer why it's not written as `assertSearchPaths([.picturesDirectory], exists: !isWindows)`

compnerd: Ugh, I failed to save the changes for that. I had the following comment:

```
        // .picturesDirectory does not exist in CI, though it does exist in user
        // desktop scenarios.
```

compnerd:
@swift-ci please test

jmschonfeld (approved):

PR #677, opened 2024-06-13
FoundationEssentialsTests: skip distant{Future,Past} tests on Windows

compnerd:
The MSVC libc does not properly handle dates that are far out from the Unix epoch. For the purposes of allowing the test suite to pass on CI, skip these tests on Windows.

### DIFF
--- a/Tests/FoundationEssentialsTests/DateTests.swift
+++ b/Tests/FoundationEssentialsTests/DateTests.swift
@@ -90,16 +90,24 @@ final class DateTests : XCTestCase {
         XCTAssertEqual("1970-01-01 00:00:00 +0000", date.description)
     }
 
-    func testDescriptionDistantPast() {
+    func testDescriptionDistantPast() throws {
+#if os(Windows)
+        throw XCTSkip("ucrt does not support distant past")
+#else
 #if FOUNDATION_FRAMEWORK
         XCTAssertEqual("0001-01-01 00:00:00 +0000", Date.distantPast.description)
 #else
         XCTAssertEqual("0000-12-30 00:00:00 +0000", Date.distantPast.description)
 #endif
+#endif
     }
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

Tests/FoundationEssentialsTests/DateTests.swift
```diff
@@ -90,15 +98,18 @@ final class DateTests : XCTestCase {
         XCTAssertEqual("1970-01-01 00:00:00 +0000", date.description)
     }
 
-    func testDescriptionDistantPast() {
+    func testDescriptionDistantPast() throws {
+        try XCTSkipIf(libcIsUCRT(), "ucrt does not support distant past")
```
parkera, line 102: I think I'd prefer a style here where we simply put the `if os(Windows)` in these two places directly instead of behind the function. The reasoning is that it's a lot more clear when looking at individual tests, and I don't think we expect to have many Date tests skipped for Windows.

compnerd: These two are the ones that I expect to have this temporarily for. However, that does cause a warning, which is why I was preferring to do this. If you feel strongly, I don't mind doing something like:

```swift
#if os(Windows)
try XCTSkipIf(true, "...")
#else
...
#endif
```

jmschonfeld: The style that we've used in other tests like `FileManagerTests` is

```swift
func testFoo() throws {
    #if os(Windows)
    throw XCTSkip("This test is not applicable because ...")
    #else
    /* ... test ... */
    #endif
}
```

Maybe we could use that here similar to your suggest above but a little more concise? I don't think that should produce any warnings

compnerd:
@swift-ci please test

jmschonfeld:

PR #680, opened 2024-06-14
Fix calloc build failure on macOS

jmschonfeld:
This resolves a build failure on macOS that was accidentally introduced by https://github.com/apple/swift-foundation/pull/670


jmschonfeld:
@swift-ci please test

itingliu (approved):
I ran into this error yesterday and I thought it was me that missed something. Thanks!

PR #682, opened 2024-06-15
Limit JSONDecoder's fallback Double path to lossless integer range

oscbyspro:
Hello again, friends!

This patch is the sequel to (#667) that fixes JSONDecoder's Int64 clamping-esque behavior (#613). 

It also fixes a bug in \_parseJSON5Integer(\_:isHex:), which forgot to drop the sign for decimal inputs.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONDecoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONDecoder.swift
@@ -981,10 +981,12 @@ extension JSONDecoderImpl: Decoder {
     static private func _slowpath_unwrapFixedWidthInteger<T: FixedWidthInteger>(as type: T.Type, json5: Bool, numberBuffer: BufferView<UInt8>, fullSource: BufferView<UInt8>, digitBeginning: BufferViewIndex<UInt8>, for codingPathNode: _CodingPathNode, _ additionalKey: (some CodingKey)?) throws -> T {
         // This is the slow path... If the fast path has failed. For example for "34.0" as an integer, we try to parse as either a Decimal or a Double and then convert back, losslessly.
         if let double = Double(prevalidatedBuffer: numberBuffer) {
-            guard let value = T(exactly: double) else {
-                throw JSONError.numberIsNotRepresentableInSwift(parsed: String(decoding: numberBuffer, as: UTF8.self))
+            if double.magnitude < Double(sign: .plus, exponent: Double.significandBitCount + 1, significand: 1) {
+                guard let value = T(exactly: double) else {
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/JSON/JSONDecoder.swift
```diff
@@ -981,10 +981,12 @@ extension JSONDecoderImpl: Decoder {
     static private func _slowpath_unwrapFixedWidthInteger<T: FixedWidthInteger>(as type: T.Type, json5: Bool, numberBuffer: BufferView<UInt8>, fullSource: BufferView<UInt8>, digitBeginning: BufferViewIndex<UInt8>, for codingPathNode: _CodingPathNode, _ additionalKey: (some CodingKey)?) throws -> T {
         // This is the slow path... If the fast path has failed. For example for "34.0" as an integer, we try to parse as either a Decimal or a Double and then convert back, losslessly.
         if let double = Double(prevalidatedBuffer: numberBuffer) {
-            guard let value = T(exactly: double) else {
-                throw JSONError.numberIsNotRepresentableInSwift(parsed: String(decoding: numberBuffer, as: UTF8.self))
+            if double.magnitude < Double(sign: .plus, exponent: Double.significandBitCount + 1, significand: 1) {
```
kperryua, line 984: A brief comment here explaining how this check works would be great.

kperryua (approved):

oscbyspro:
Just some clarification: this solution :tm: only prevents floating-point errors >=(±1).

The fallback Double path still introduces lossy behavior for smaller JSON values that cannot be represented by Double. I don't think such errors can be caught after the fact. A proper solution would validate the input String in the parsing step. I suppose ICU doesn't provide integer validation APIs?
 
```swift
print(try! JSONDecoder().decode(Int64.self, from: "9007199254740991.0".data(using: .utf8)!)) // 2^53 - 1, OK
print(try! JSONDecoder().decode(Int64.self, from: "9007199254740990.5".data(using: .utf8)!)) // 2^53 - 2, :(
print(try! JSONDecoder().decode(Int64.self, from: "9007199254740990.0".data(using: .utf8)!)) // 2^53 - 2, OK
```

Sources/FoundationEssentials/JSON/JSONDecoder.swift
oscbyspro: Right. I've added some comments about it and the guard statement below it. It's not a perfect solution but it makes it so you don't get integer A from integer B.

itingliu:
LGTM. Thanks!

Issue #638, opened 2024-05-28
Foundation.Data.copyBytes broken for any non-0-start InlineSlice-based instance


jkoritzinsky:
The implementation of `Foundation.Data.copyBytes<DestinationType>(to: UnsafeMutableBufferPointer<DestinationType>, from range: Range<Data.Index>? = nil)` when the underlying representation is an `InlineSlice` does not correctly handle recalculating the offsets for the underlying slice.

For example, the following program results in SIGILL on x64 due to a precondition violation:

```swift
import Foundation

let sourceData = Data([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20])
let slice = sourceData[3..<20]
let unmanaged = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: sourceData.count);
unmanaged.initialize(repeating: 0)
_ = slice.copyBytes(to: unmanaged) // SIGILL here.
print(unmanaged)
unmanaged.deallocate()
```

Whereas the following program succeeds:

```swift

import Foundation

let sourceData = Data([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20])
let slice = sourceData[3..<20]
let unmanaged = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: sourceData.count);
unmanaged.initialize(repeating: 0)
slice.copyBytes(to: unmanaged, from: 3..<20)
print(unmanaged)
unmanaged.deallocate()
```

Whereas this program is successful:

```swift

import Foundation

let sourceData = Data([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20])
let slice = sourceData[3..<20]
let unmanaged = UnsafeMutableBufferPointer<UInt8>.allocate(capacity: sourceData.count);
unmanaged.initialize(repeating: 0)
slice.copyBytes(to: unmanaged, from: ) // SIGILL here.
print(unmanaged)
unmanaged.deallocate()
```

I would expect that for any `Foundation.Data` instance, that calling `data.copyBytes(to: buffer)` would successfully fill the buffer irrespective of the underlying storage mechanism, as the documentation specifies.

vcsjones:
It looks like this was fixed by https://github.com/swiftlang/swift-foundation/pull/683. Should this issue be closed?

PR #683, opened 2024-06-16
Data/copyBytes(to:from:) slice offset fix

oscbyspro:
This patch fixes (#638) and adds a regression test.


itingliu:
@swift-ci please test

parkera:
Wow, thanks. Looks like this bug has been around since 2016, the introduction of the struct Data type.

parkera (approved):

PR #684, opened 2024-06-17
Follow up fix for buffer overflow

itingliu:
Start the outer buffer with the capacity of `initialSize + 1` so later we can null terminate on the index of `initialSize`, assuming the return length is smaller or equal to `initialSize`. We don't bother to verify the assumption since that'd be a user error.

Also fix a terrible copy-pasta error: We should null-terminate the inner buffer, not the outer buffer.

Fixed 129806508


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #690, opened 2024-06-18
Overflow when calling `Calendar.date(from:)`

itingliu:
Validate the given `DateComponents` values up front to align with the supported Calendar calculation date range, defined as `Date.validCalendarRange`.

We could alternatively guard all arithmetic operations with `...reportingOverflow`, but there are too many operations, and so it seems untenable. I opted for a more realistic approach instead. `_CalendarICU` unconditionally truncates values to `Int32`, so the results for `Calendar.date(from:)` have always been incorrect for distant dates anyways.

Fixed 129782208


itingliu:
@swift-ci please test

parkera (approved):

PR #692, opened 2024-06-20
(129922624) Don't convert already-converted NSURLs

jrflat:
I had previously used `.init(reference:)` so that an initializer was called in all code paths, but this caused the underlying `NSURL` to be `_converted` twice in those cases. This causes an issue when force-casting `as! NSURL` if the first conversion returns `nil`. This PR adds and uses a `.init(convertedReference:)` that assigns an already-converted `NSURL` to `_url` directly.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #693, opened 2024-06-20
Decimal.init(string:) should return nil for invalid Decimal strings

iCharlesHu:
For binary compatibility, `Decimal.init(string:)` should return `nil` instead of `0` for totally invalid Decimal strings. 

resolves: rdar://130209835


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #694, opened 2024-06-20
Add SWIFTCI_USE_LOCAL_DEPS build mode

jmschonfeld:
When building/running Foundation tests in swift-ci as part of a full toolchain test (and not just an isolated test of only Foundation), we want to build this package against the locally checked out versions of our dependencies. This checks a new environment variable `SWIFTCI_USE_LOCAL_DEPS` and, when set, will look in sibling directories for our dependencies rather than checking out dependencies from GitHub. This is similar to the setup that some other packages such as [swift-format](https://github.com/apple/swift-format/blob/d46e30fe67b722c950424e09840dfc391651dbfa/Package.swift#L155-L169) use.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,6 +41,34 @@ let concurrencyChecking: [SwiftSetting] = [
     .enableUpcomingFeature("InferSendableFromCaptures")
 ]
 
+var dependencies: [Package.Dependency] {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #695, opened 2024-06-21
Ensure EOPNOTSUPP thrown via Data reading from a socket does not crash

jmschonfeld:
On some platforms like Darwin, attempting to read `Data` from a path representing a socket can cause the `EOPNOTSUPP` error to be thrown, but `POSIXError.Code` is missing a case for this errno so we hit an assertion failure in our code. This patch works around this case to avoid the crash by dropping the posix info from the produced error while still asserting for other invalid errno's to catch cases where we incorrectly capture `errno`


jmschonfeld:
@swift-ci please test

jrflat (approved):

PR #696, opened 2024-06-22
[CMake] Update CMake build to account for _CShims --> _FoundationCShims rename

jmschonfeld:
Now that `_CShims` has been renamed to `_FoundationCShims`, we need to update the cmake files to account for this rename


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #698, opened 2024-06-24
Slices of no copy Data bridged to and back incorrectly account for the start index

itingliu:
when bridging back from objc the start index was not correctly re-constructed into the slices of no copy datas. This adds functionality for that.

resolves rdar://130106808


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #699, opened 2024-06-24
[CMake] Copy headers to known directory for direct client test builds

jmschonfeld:
This copies the public headers to a known location in the build tree so that direct clients (before installation) can specify that directory as an include path when not building via a CMake dependency


jmschonfeld:
@swift-ci please test

parkera (approved):

iCharlesHu (approved):

PR #700, opened 2024-06-24
Base64 must relinquish use of OutputBuffer on error

parkera:
Make sure to always call `relinquishBorrowedMemory`, even on error.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #703, opened 2024-06-26
_FoundationCShims: enable building on Android

compnerd:
The `strncasecmp_l` was added at API level 23. For now, simply abort on older releases, though we should consider shimming the path.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #704, opened 2024-06-26
FoundationEssentials: initial pass to add Android support

compnerd:
This adds the necessary guards and includes for the Android modules. While the module does not compile currently due to nullability differences (and in some cases missing declarations), this at least brings the module to a point where we can start working on the errors and differences to create a maintainable codebase for Android.


parkera:
@swift-ci test

parkera (approved):

compnerd:
@swift-ci please test

PR #709, opened 2024-06-28
[CMake] Include GNUInstallDirs

jmschonfeld:
The GNUInstallDirs module sets CMAKE_INSTALL_BINDIR which is used on Windows as the destination for installed .dll files. This addresses issues that I saw when building a Windows toolchain where the .dll files were not installed into the toolchain previously, but with this change are correctly installed into the bin directory.


jmschonfeld:
@swift-ci please test

itingliu (approved):
For my information, cmake's documentation says this defines `MAKE_INSTALL_<dir>` where `dir` can be one of many options, `BINDIR` being one of those. Does the includer automatically choose `BINDIR` on Windows or was there another config you had to set to get that to be set?

jmschonfeld:
In our case, we use it inside of `SwiftFoundationSwiftSupport.cmake` which has a `_swift_foundation_install_target` function that installs a swift target (it installs the binary and copies over swift module files). When it installs the target, it installs via

```cmake
install(TARGETS ${module}
  ARCHIVE DESTINATION lib/${swift}/${swift_os}
  LIBRARY DESTINATION lib/${swift}/${swift_os}
  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
```

and in this case without this include, `CMAKE_INSTALL_BINDIR` would be empty. On linux this didn't matter since libraries get installed into the library destination path. However on Windows, `.dll` files get installed into the runtime destination path while `.lib` files get installed into the library destination path, so with this runtime destination left empty the installed toolchain was missing the `.dll` files

PR #712, opened 2024-07-02
Revert "Slices of no copy Data bridged to and back incorrectly account for the start index (#698)"

phausler:
This reverts commit 7fa9d1f4a948da4eeed3928946d0b458d2f3bcf6. It was found that the bridging of already bridged slices behavioral change causes regressions so this had to be backed out.


jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #713, opened 2024-07-02
FoundationInternationalization: add `Android` imports

compnerd:
This adds the missing imports for Android and enables building for Android.


compnerd:
@swift-ci please test

compnerd:
@parkera @jmschonfeld - okay, with this patch and a local toolchain build off of main, I am now able to build swift-foundation for Android with CMake. Considering an option to do the cross-compilation testing would be nice. We still need to fix the macros build though.

jmschonfeld (approved):

PR #714, opened 2024-07-03
FoundationMacros: use cross-compilation to build for host

compnerd:
Use `ExternalProject` to switch FoundationMacros to cross-compilation. This allows us to build the macros for the right OS/architecture when cross-compiling Foundation for other environments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,8 @@ set(BUILD_TESTING NO)
 set(COLLECTIONS_SINGLE_MODULE YES)
 set(COLLECTIONS_FOUNDATION_TOOLCHAIN_MODULE YES)
 
+set(SwiftFoundation_MACRO "" CACHE STRING "Path to Foundation macro plugin")
+
 # Make sure our dependencies exists
 include(FetchContent)
 if (_SwiftFoundationICU_SourceDIR)
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -13,11 +13,5 @@
 ##===----------------------------------------------------------------------===##
 
 add_subdirectory(_FoundationCShims)
-
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-	add_subdirectory(FoundationMacros)
-endif()
-
 add_subdirectory(FoundationEssentials)
 add_subdirectory(FoundationInternationalization)
--- a/Sources/FoundationMacros/CMakeLists.txt
+++ b/Sources/FoundationMacros/CMakeLists.txt
@@ -12,29 +12,45 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+cmake_minimum_required(VERSION 3.22)
+
+if(POLICY CMP0156)
+    # Deduplicate linked libraries where appropriate
+    cmake_policy(SET CMP0156 NEW)
+endif()
+if(POLICY CMP0157)
+    # New Swift build model: improved incremental build performance and LSP support
+  cmake_policy(SET CMP0157 NEW)
+endif()
+
+project(FoundationMacros
+  LANGUAGES Swift)
+
 # SwiftSyntax Dependency
-include(FetchContent)
-find_package(SwiftSyntax)
+find_package(SwiftSyntax QUIET)
 if(NOT SwiftSyntax_FOUND)
+    include(FetchContent)
+
     # If building at desk, check out and link against the SwiftSyntax repo's targets
     FetchContent_Declare(SwiftSyntax
         GIT_REPOSITORY https://github.com/swiftlang/swift-syntax.git
         GIT_TAG 4c6cc0a3b9e8f14b3ae2307c5ccae4de6167ac2c) # 600.0.0-prerelease-2024-06-12
     FetchContent_MakeAvailable(SwiftSyntax)
+else()
+  message(STATUS "Using swift-syntax from ${SwiftSyntax_DIR}")
 endif()
 
 add_library(FoundationMacros SHARED
     FoundationMacros.swift
     PredicateMacro.swift)
-    
+
 target_compile_definitions(FoundationMacros PRIVATE FOUNDATION_MACROS_LIBRARY)
 
-set_target_properties(FoundationMacros
-    PROPERTIES
-        ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
-        LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
-        INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/lib
-)
+target_compile_options(FoundationMacros PRIVATE -parse-as-library)
+target_compile_options(FoundationMacros PRIVATE
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend AccessLevelOnImport>"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend StrictConcurrency>"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-upcoming-feature -Xfrontend InferSendableFromCaptures>")
 
 target_link_libraries(FoundationMacros PUBLIC
     SwiftSyntax::SwiftSyntax
@@ -43,18 +59,19 @@ target_link_libraries(FoundationMacros PUBLIC
     SwiftSyntax::SwiftSyntaxBuilder
 )
 
-# The macro is installed into lib/swift/host/plugins, but needs to load libraries from lib/swift/host and lib/swift/${SWIFT_SYSTEM_NAME}
 set_target_properties(FoundationMacros PROPERTIES
+    ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+    LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib
+    PDB_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin
+    RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin
+
+    INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_BINARY_DIR}/lib
+
+    # The macro is installed into lib/swift/host/plugins, but needs to load
+    # libraries from lib/swift/host and lib/swift/${SWIFT_SYSTEM_NAME}
     INSTALL_RPATH "$ORIGIN/../../../swift/${SWIFT_SYSTEM_NAME}:$ORIGIN/.."
     INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
 
-target_compile_options(FoundationMacros PRIVATE -parse-as-library)
-target_compile_options(FoundationMacros PRIVATE
-    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend AccessLevelOnImport>"
-    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend StrictConcurrency>"
-    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-upcoming-feature -Xfrontend InferSendableFromCaptures>")
-
 install(TARGETS FoundationMacros
-    ARCHIVE DESTINATION lib/swift/host/plugins
     LIBRARY DESTINATION lib/swift/host/plugins
-    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+    RUNTIME DESTINATION bin)
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -49,10 +49,9 @@ add_subdirectory(String)
 add_subdirectory(TimeZone)
 add_subdirectory(URL)
 
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-    # Depend on FoundationMacros
-    add_dependencies(FoundationEssentials FoundationMacros)
-    target_compile_options(FoundationEssentials PRIVATE -plugin-path ${CMAKE_BINARY_DIR}/lib)
+if(SwiftFoundation_MACRO)
+  target_compile_options(FoundationEssentials PRIVATE
+    "SHELL:-plugin-path ${SwiftFoundation_MACRO}")
 endif()
 
 if(CMAKE_SYSTEM_NAME STREQUAL Linux OR CMAKE_SYSTEM_NAME STREQUAL Android)


Sources/FoundationMacros/CMakeLists.txt
```diff
@@ -12,10 +12,19 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+cmake_minimum_required(VERSION 3.22)
+if(POLICY CMP0157)
+  cmake_policy(SET CMP0157 NEW)
+endif()
```
etcwilde, line 18: While we're at it, can we also enable `CMP0156` as well? Thanks.

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,29 @@
...
-add_subdirectory(_FoundationCShims)
-
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-	add_subdirectory(FoundationMacros)
+include(ExternalProject)
+if(CMAKE_HOST_WIN32)
+  set(_SwiftFlags -DCMAKE_Swift_FLAGS="-use-ld=lld")
+else()
+  set(_SwiftFlags)
```
etcwilde, line 19: This shouldn't be necessary.

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,29 @@
...
 ##===----------------------------------------------------------------------===##
 
-add_subdirectory(_FoundationCShims)
-
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-	add_subdirectory(FoundationMacros)
+include(ExternalProject)
+if(CMAKE_HOST_WIN32)
+  set(_SwiftFlags -DCMAKE_Swift_FLAGS="-use-ld=lld")
```
etcwilde, line 17: Since it's flags for the macro build, would you mind making that clearer in the variable name?
e.g. `_SwiftMacroFlags` or something like that? Thanks.

etcwilde (commented):
Few nits, but otherwise LGTM.

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,29 @@
...
+  CMAKE_ARGS
+    -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    ${_SwiftFlags}
+  INSTALL_COMMAND "")
+ExternalProject_Get_Property(FoundationMacros BINARY_DIR)
+if(CMAKE_HOST_WIN32)
+  set(_SwiftFoundation_PredicateMacro "${BINARY_DIR}/FoundationMacros.exe#PredicateMacro")
+  set(_SwiftFoundation_ExpressionMacro "${BINARY_DIR}/FoundationMacros.exe#ExpressionMacro")
+else()
+  set(_SwiftFoundation_PredicateMacro "${BINARY_DIR}/FoundationMacros#PredicateMacro")
```
jmschonfeld, line 34: It doesn't look like these values are used anywhere, is that intentional? Also it looks like it's listing these as executables but the macros build as a library (like the Observation / other macros produced by the stdlib) so I don't think this are quite accurate - I'd expect these to look like `libFoundationMacros.so` and `FoundationMacros.dll`

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,29 @@
...
+endif()
+ExternalProject_Add(FoundationMacros
+  SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/FoundationMacros"
+  PREFIX "${CMAKE_BINARY_DIR}/_deps"
+  BINARY_DIR "macros"
+  CMAKE_ARGS
+    -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    ${_SwiftFlags}
+  INSTALL_COMMAND "")
+ExternalProject_Get_Property(FoundationMacros BINARY_DIR)
```
jmschonfeld, line 29: Will this impact any of the non-macro builds below by changing the value of `BINARY_DIR`?

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,29 @@
...
+  set(_SwiftFlags)
+endif()
+ExternalProject_Add(FoundationMacros
+  SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/FoundationMacros"
+  PREFIX "${CMAKE_BINARY_DIR}/_deps"
+  BINARY_DIR "macros"
+  CMAKE_ARGS
+    -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    ${_SwiftFlags}
+  INSTALL_COMMAND "")
```
jmschonfeld, line 28: Could you elaborate on what this setting does specifically? Does this disable the install step for the macros?

Sources/CMakeLists.txt
compnerd: Yes, this is intentional - the macros are not currently used in the CMake build AFAICT. The default behaviour for the macros is to build as executables, the in-process (library) model is something that is still being implemented. Supporting that as an option would add more complexity.

Sources/CMakeLists.txt
compnerd: No, `BINARY_DIR` is not the same as `CMAKE_BINARY_DIR` or `CMAKE_CURRENT_BINARY_DIR`.

Sources/CMakeLists.txt
compnerd: Yes, this is disabling the install step of the external build.

Sources/FoundationMacros/CMakeLists.txt
compnerd: Done

Sources/CMakeLists.txt
compnerd: Done

Sources/CMakeLists.txt
compnerd: Done

compnerd:
@swift-ci please test

Sources/CMakeLists.txt
jmschonfeld: @bnbarham remind me - I think we discussed that the macro should be a library rather than an executable on linux (am I remembering that right?). Did we have a reason for that? If so, should we update this to account for that difference (an executable on Windows but still building as an `.so` on linux)?

Sources/CMakeLists.txt
jmschonfeld: Will this also prevent installing the shared library on linux? That currently works as expected and we should try not to regress that if possible

compnerd: There is no guarantee that the library would be usable - if I built the library on macOS, the MachO binary wouldn't run on Linux.

jmschonfeld: I think I might be missing some details here so I apologize if I don't have the right interpretation - but to me it seems like this will in effect disable the macro build on all platforms since the install command is empty. It makes sense that building the macro on macOS and deploying to linux wouldn't work, but do we support building as that configuration today? Right now we have the macro successfully building for linux and being installed into the linux toolchain in swift-ci via our cmake build and I want to make sure we don't regress that with this change.

Sources/CMakeLists.txt
bnbarham: > The default behaviour for the macros is to build as executables

For external macros that's true, but for macros included in the toolchain it's better if they're a library linking to the `lib/swift/host` swift-syntax which is then loaded into `swift-plugin-server`. They could theoretically be executables, though I'm not sure we have any search paths for executable macros in the toolchain by default (CC @rintaro in case I'm wrong)

> the in-process (library) model is something that is still being implemented

In-process already exists, the work @rintaro has been doing is to allow a separate swift-syntax to be used for those libraries compared to the compiler. In any case, that isn't what would be being used here anyway.

Sources/CMakeLists.txt
compnerd: I think that we do support that, e.g. for the static SDKs that we are building. The android work also is done as Windows -> Android, and I've done Windows -> Linux in the past as well. This means that we would be installing Windows executables into a Linux installation. In order to install the macros prebuilt into the toolchain, I think that we should be building just the macros manually for Linux and installing them into the toolchain.

etcwilde: It will still run the build and produce the macros into the build directory. It won't copy them into the final install location of the library itself. We don't have a great distribution story for macros and cross-compiling at the moment. Let's say, for example, that I'm on a Mac trying to cross-compile to a Linux. The macros will build for macOS so that I can run them locally while building, while the library will build for Linux. It doesn't really make sense to include the macOS-built macros in the Linux SDK. If there are macros that we need to ship, we will need to build them separately for all platforms that have a toolchain that can use the SDK. Long-term, WASM might be interesting in this space so that we only build the macros for WASM and then ship a WASM runtime in the toolchain , but I don't think that's in any of the current cards.

An alternative is to set the install prefix of the EP_Add to add `-DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>` to the cmake args and have it install the executables to a temporary location in the Foundation build directory, but I don't see too much point to copying temporaries around unless we have to.

compnerd: Note that failure to pass `-D CMAKE_INSTALL_PREFIX=...` after such a change would also break the build as it will attempt to install to the root where the user may not have permission (or if running as root, could install over the system copy).

jmschonfeld: How do the Stdlib macros that are installed into the toolchain already today work? IMO we should follow the same pattern here, do they support cross compiling in this manner?

compnerd: The stdlib macros are built multiply (for each standard library host). They are not used to build the standard library or run its tests. We currently have references of the macros in the tests, which require that the macros are built for build. We would need to do a second build of the macros for the host, which will also require building swift-syntax for the host and increase the overall build times. However, that seems like the right thing to do if we need to ship the macros as prebuilt binaries. This is likely going to be problematic as the SDK and toolchains can go out of sync and there is no ABI stability outside of Apple platforms though and so would be better to avoid IMO.

compnerd: @etcwilde @jmschonfeld - what do we want to do here? Introduce an option to build a second copy of the macros?

etcwilde: If I knew how to install them without them conflicting with each other, I'd suggest a loop over a list of target triples passed into the ExternalProject invocations so that someone can list off what toolchains their SDK will work with, sort of like how the LLVM build allows you to specify which triples you want to build the runtimes for.

compnerd: That seems like a good deal of infrastructure that we will need to build up and share - much like `SwiftSupport.cmake`, especially while we don't have `install(...)` working for Swift targets.

jmschonfeld: For clarity - do we officially support cross-compiling a linux toolchain or is that just something that you _can_ do but isn't something that's load bearing at the moment? As of today, we're able to successfully produce linux toolchains that contain `libFoundationMacros.so` and the macro can be used when using that toolchain. I think this change would remove that library from the toolchain which would be a step backwards on Linux, and I'd like to avoid making that regression unless we deem that producing this library in the current manner is breaking a load bearing, officially supported path.

As for the windows side, yeah I'm not sure whether it'd be best to build the macro for just the host (to use when building Foundation itself) or just the destination (to package into the created toolchain). I suppose starting with the former and hopefully supporting the latter in the future is a path forward that we can take here, it's just unfortunate that Windows toolchains would be missing out on these APIs like `#Predicate`. This sounds like a general problem that we'll likely need to solve quickly given that swift-testing will likely run into the same issue shortly, but their macros are much more critical to their API surface (cc @grynspan / @stmontgomery)

compnerd: On Windows, at least this is "easier": the only supported hosts are windows x86_64, arm64. Because we build the SDK for both of these platforms in CI, we could build the macros for both of the hosts.

The problem is when we build for Android, what are the supported hosts? Windows x86_64, arm64, Linux x86_64, arm64, Darwin arm64, FreeBSD arm64, x86_64? What about Haiku? We would need to have the SDKs for all of those platforms available to build the macros. This quickly leads to an exponential growth problem.

grynspan: That's a good argument in favour of Wasm macro plugins.

compnerd: @jmschonfeld btw, AFAIK, cross-compilation of the toolchain for non-Darwin targets is not officially supported. While there are some intrepid users (e.g. Facebook and at times myself) who do this, it is simply on them to figure out how to piece that together. I do try to make it easy to cross-compile, but this might be one of those things where incremental improvements are the way to address this.

shahmishal:
@swift-ci please test

shahmishal:
@swift-ci please test

compnerd:
CC: @bnbarham there was a conversation on swiftlang/swift#75206 which seems to have converged around moving ahead with this change.

compnerd:
@jmschonfeld @bnbarham - any idea if this is something we should be moving forward with? This is going to break the windows builds at least, but I want to ensure that we only are investing in this if we are going to move forward with this.

jmschonfeld:
> @jmschonfeld @bnbarham - any idea if this is something we should be moving forward with? This is going to break the windows builds at least, but I want to ensure that we only are investing in this if we are going to move forward with this.

I think I'm still a little fuzzy on the details for this, so just to lay it out to confirm:

Currently we have:

(builder platform = host building the toolchain, toolchain platform = host where the toolchain will be used, sdk platform = platform where things the toolchain builds will be run)

Linux:
- Macro builds as a library and links the early swift syntax build
- Macro builds for builder platform (platform that is building the toolchain)
- Macro gets used for FoundationEssentials build
- Macro gets installed into the toolchain since in the supported case, builder platform == toolchain platform

Windows: Macro does not build

And with this PR, can you clarify what changes? In specific

- Does this still install the macro on linux where builder platform == toolchain platform
- Do we build the macro on Windows for the builder platform to use to build FoundationEssentials (this is yes right?)
- Do we build the macro on Windows for the toolchain platform? And it is not built for the SDK platform right?
- Is it still built as a library?

I think I'm still unclear on those points and I want to make sure that we do not regress what Linux already has working in the process of getting it right on Windows (but I agree we need to make sure we get it right on Windows too)

etcwilde:
So the goal is to split where the SDK library that actually goes in the runtime environment and the macros that are only useful if you're compiling things against that library.
Right now, the build is set up such that the macro will always build for whatever you're building swift-foundation to run on, but not necessarily where you're running the compiler to build things against that SDK, which isn't particularly helpful if you don't want/need the macro to run there. It also means that we'll need to build and include Swift-Syntax in the built SDK, even though it will just be dead weight in situations where you're never running a compiler.

What this PR does is, when building Swift-Foundation, build the macro for the platform that is doing the build, rather than what swift-foundation is being cross-compiled for. This is just like what we're doing for Swift-Testing (https://github.com/apple/swift-testing/blob/b19c9af0750e219fc972217af3f061b1b0b60897/Sources/CMakeLists.txt#L22-L30).

It's still not perfect, but it's better because it means that we can use these macros in the build of swift-foundation.
In a future iteration, we might be able to teach the build to produce macros for each of the toolchain platforms that we support so that we can have an SDK that works universally with each of the supported toolchains (good for cross-compiling story) instead of only the toolchain that is on the system that you're deploying for (not great if you're compiling for a low-powered phone or something).

compnerd:
Just to clear up some of the confusion on what is happening today:

Windows:
- Macros build as executables, link the early swift syntax build
- Macros build for the platform that foundation are being built for (e.g. Android)
- Macros get used for the tests

Note that this is identical to Linux where the macros are built for the platform that foundation is being built for (e.g. macOS).

bnbarham:
> Macros build as executables, link the early swift syntax build

In the toolchain build or standalone? If the former, how are they being picked up at runtime? IIRC we only add toolchain search paths for the libraries, not executables.

compnerd:
> > Macros build as executables, link the early swift syntax build
> 
> In the toolchain build or standalone? If the former, how are they being picked up at runtime? IIRC we only add toolchain search paths for the libraries, not executables.

Ugh, wait, I mixed up two scenarios. The reason that it is building with the toolchain early swift-syntax is because I'm building locally with a reference to the toolchain build. That change is dependent on the other change that I have pending.

The CI builds currently use SPM based builds in a docker container, so they are building their own copy of swift-syntax.

For the toolchain distribution, we _should_ be using the toolchain build of swift-syntax though to avoid bloat (the toolchain is already large enough that it prevents pre-installation on GHA agents).

bnbarham:
> For the toolchain distribution, we should be using the toolchain build of swift-syntax though to avoid bloat

Sure, that should be the case here - we should be passing through `SwiftSyntax_DIR` and thus going down the FindPackage path.

compnerd:
> > For the toolchain distribution, we should be using the toolchain build of swift-syntax though to avoid bloat
> 
> Sure, that should be the case here - we should be passing through `SwiftSyntax_DIR` and thus going down the FindPackage path.

I don't think that was wired up in the swift-corelibs-foundation build which is now re-cored to swift-foundation. We don't build swift-foundation on its own yet, so this particular path is more future direction and I don't think that we are setup to test it in CI.

jmschonfeld:
> > > For the toolchain distribution, we should be using the toolchain build of swift-syntax though to avoid bloat
> > 
> > 
> > Sure, that should be the case here - we should be passing through `SwiftSyntax_DIR` and thus going down the FindPackage path.
> 
> I don't think that was wired up in the swift-corelibs-foundation build which is now re-cored to swift-foundation. We don't build swift-foundation on its own yet, so this particular path is more future direction and I don't think that we are setup to test it in CI.

AFAIK it should be - when invoking the swift-corelibs-foundation build from both `build.ps1` and the `build-script`, we provide `SwiftSyntax_DIR` (and swift-corelibs-foundation pulls swift-foundation in via `FetchContent` and the goal is to go down the `find_package` path, if that's not happening it's likely a bug we need to fix in the build script or CMake file

compnerd:
For the installation, I don't think that we have a good way to detect when it is being built for a toolchain host. There is `CMAKE_CROSSCOMPILING` that tells us whether it is cross-compiling, but I could be cross-compiling for a toolchain host (which is, in fact, what we do for ARM64). So, really, the best way to handle this is to explicitly require that you build the `install` target if you want to install the macros.

The plugin (DLL) approach is not really viable. The entire compiler codebase needs to be annotated with `__declspec(dllimport)` and `__declspec(dllexport)` to get that to work (including the LLVM one). Otherwise we run out of space for exporting everything (not to mention, there is a load time and runtime memory penalty for doing that). Executables are the way forward for the foreseeable future (we _could_ try to invasively fork LLVM for Swift and annotate LLVM and split it up of course).

bnbarham:
> The plugin (DLL) approach is not really viable. The entire compiler codebase needs to be annotated with __declspec(dllimport) and __declspec(dllexport) to get that to work (including the LLVM one). Otherwise we run out of space for exporting everything (not to mention, there is a load time and runtime memory penalty for doing that). Executables are the way forward for the foreseeable future (we could try to invasively fork LLVM for Swift and annotate LLVM and split it up of course).

We currently use libraries for the stdlib plugins. The driver isn't setup to load executables from the toolchain at all, so if we did end up going down that route we'd need updates there. But I'm not sure I understand why we have to. Why does LLVM matter at all here?

compnerd:
> We currently use libraries for the stdlib plugins. The driver isn't setup to load executables from the toolchain at all, so if we did end up going down that route we'd need updates there. But I'm not sure I understand why we have to. Why does LLVM matter at all here?

The plugin host needs to expose interfaces to the plugin (i.e. `-rdynamic`). If the plugins have no assumed library interface, then this should be possible.

bnbarham:
> The plugin host needs to expose interfaces to the plugin (i.e. -rdynamic). If the plugins have no assumed library interface, then this should be possible.

There's an interface, but it uses swift-syntax APIs and they are in SwiftSyntax.dll.

Just to re-iterate, the stdlib macro plugins are both libraries today and as far as I'm aware they don't have any issues. This seems like an identical case to me.

hyp:
Added additional changes to make this also work for the android swift-foundation building from build.ps1 windows host:
https://github.com/compnerd/swift-foundation/pull/1
https://github.com/swiftlang/swift/pull/75549

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,28 @@
...
+    -DCMAKE_SYSTEM_NAME=${CMAKE_HOST_SYSTEM_NAME}
+    "-DCMAKE_Swift_FLAGS=${_FoundationMacrosSwiftFlags} ${CMAKE_HOST_Swift_FLAGS}"
+  INSTALL_COMMAND "")
+ExternalProject_Get_Property(FoundationMacros BINARY_DIR)
+if(CMAKE_HOST_WIN32)
+  set(_SwiftFoundation_PredicateMacro "${BINARY_DIR}/FoundationMacros.dll#PredicateMacro")
+  set(_SwiftFoundation_ExpressionMacro "${BINARY_DIR}/FoundationMacros.dll#ExpressionMacro")
+else()
+  set(_SwiftFoundation_PredicateMacro "${BINARY_DIR}/FoundationMacros#PredicateMacro")
+  set(_SwiftFoundation_ExpressionMacro "${BINARY_DIR}/FoundationMacros#ExpressionMacro")
```
jmschonfeld, line 34: I think I'm still confused on what the intention of these settings are? Are they supposed to be paths that we provide as `-plugin-path` to the FoundationEssentials build? If so, I think I'd expect them to be in the library directory on non-windows and also look like `libFoundationMacros.so` rather than `FoundationMacros#ExpressionMacro`.

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,28 @@
...
-add_subdirectory(_FoundationCShims)
-
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-	add_subdirectory(FoundationMacros)
+include(ExternalProject)
+if(CMAKE_HOST_WIN32)
+  set(_FoundationMacrosSwiftFlags "-use-ld=lld")
+endif()
+ExternalProject_Add(FoundationMacros
```
jmschonfeld, line 19: Just to clarify with this approach, I know that the build script will need to invoke the macro build separately. Is the goal that it needs to invoke the macro build once for the current host (and pass that path to the swift-foundation build to use when building itself) and once for each target host (and that it will actually install instead of just building)? In `FoundationEssentials/CMakeLists.txt` we have a `FoundationEssentials` --> `FoundationMacros` dependency added on non-Windows, I think we'll need to adjust that accordingly if there's no longer a target built in this same project

Sources/CMakeLists.txt
```diff
@@ -12,12 +12,28 @@
...
+  SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/FoundationMacros"
+  PREFIX "${CMAKE_BINARY_DIR}/_deps"
+  BINARY_DIR "macros"
+  CMAKE_ARGS
+    -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+    -DCMAKE_SYSTEM_NAME=${CMAKE_HOST_SYSTEM_NAME}
```
kateinoigakukun, line 25:
```suggestion
    -DCMAKE_SYSTEM_NAME=${CMAKE_HOST_SYSTEM_NAME}
    -DCMAKE_Swift_COMPILER=${CMAKE_Swift_COMPILER}
```
To ensure using a consistent compiler between libraries and macros, I think it might be better to propagate `CMAKE_Swift_COMPILER` instead of assuming swiftc is available in PATH.

Sources/CMakeLists.txt
compnerd: Yes, this was the expected parameter for the macro. Although, after the discussion with @rintaro yesterday, I don't think it matters. Once we have the custom option for the macro, this will no longer be needed.

CMakeLists.txt
```diff
@@ -45,6 +45,8 @@ set(BUILD_TESTING NO)
 set(COLLECTIONS_SINGLE_MODULE YES)
 set(COLLECTIONS_FOUNDATION_TOOLCHAIN_MODULE YES)
 
+set(SwiftFoundation_MACRO "" CACHE STRING "Path to Foundation macro plugin")
```
jmschonfeld, line 48: Thanks for updating to add this - could we update this PR to remove the `FoundationEssentials` --> `FoundationMacros` target dependency and update the `-plugin-path` argument passed to the build?

Sources/FoundationMacros/CMakeLists.txt
```diff
@@ -12,29 +12,47 @@
...
+if(POLICY CMP0157)
+    # New Swift build model: improved incremental build performance and LSP support
+  cmake_policy(SET CMP0157 NEW)
+endif()
+
+project(FoundationMacros
+  LANGUAGES Swift)
+
+include(GNUInstallDirs)
```
jmschonfeld, line 29: When testing this out for linux, I'm seeing:

```
CMake Warning (dev) at /Repos/swift-dev/cmake/Modules/GNUInstallDirs.cmake:243 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
Call Stack (most recent call first):
  CMakeLists.txt:29 (include)
```

from this include, are we missing something here?

CMakeLists.txt
compnerd: Yeap!

Sources/FoundationMacros/CMakeLists.txt
compnerd: It seems that Swift does not count as a "language" for this module for some reason. That is a bug in CMake that we should address. For the time being, I think that this should be possible to workaround by explicitly passing in `-D CMAKE_INSTALL_LIBDIR=lib`. I think that we want to keep the include for the install paths.

Sources/CMakeLists.txt
compnerd: This should no longer be a concern given that we do not do the internal build.

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):
This LGTM now - approved to merge, but let's try to merge it with both of our swiftlang/swift changes and your swift-installer-script change just so that everything switches to the new build style at once

Sources/FoundationMacros/CMakeLists.txt
etcwilde: Yeah, we'll need to teach Swift to set up the `CMAKE_SYSTEM_NAME` and `CMAKE_SIZEOF_VOID_P`. In theory the frontend should know the size of a pointer (and endianness info), so I wonder if we couldn't stick that in the target info and pull it from there instead of trying to generate the tables in Swift? 🤔

PR #718, opened 2024-07-05
PropertyListEncoder should encode whole-value reals in XML without the .0 suffix for backward compatibility

iCharlesHu:
Historically `PropertyListEncoder` encodes whole-value reals _without_ the `.0` suffix. We should maintain this behavior.


iCharlesHu:
@swift-ci please test

kperryua (approved):

PR #720, opened 2024-07-08
Use .mappedIfSafe in DataIOTests.test_largeFile

iCharlesHu:
DataIOTests.test_largeFile reads a larger than 2GB file twice into the memory. Adding `.mappedIfSafe` option when reading Data to prevent the test getting killed due to memory usage.

resolves: rdar://130769047


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #723, opened 2024-07-09
FoundationEssentials: fix corner case of flushing file descriptor

compnerd:
On Windows, `_commit` can only be used for file descriptors backing files that are on disks. We would previously flush all file types which would cause an error.

Thanks to @z2oh for tracking down the cause of the error!


compnerd:
@swift-ci please test

compnerd:

jmschonfeld (approved):

PR #725, opened 2024-07-10
Remove the .0 suffix when encoding whole-value reals.

parkera:
This was a behavioral change in PropertyListEncoder when switching from the ObjC/C property list format to the new Swift one.


parkera:
@swift-ci test

kperryua (approved):

parkera:
Oops, @iCharlesHu posted this as #718

PR #727, opened 2024-07-11
Fix AttributedString slicing behavior on range expressions

lorentey:
`AttributedString` goes out of its way to emulate the stdlib’s behavior for `String` slicing. Unfortunately, it went a little overboard:

```swift
import Foundation

let str1 = "F\u{301}a\u{308}n\u{303}c\u{327}y\u{30a}" // "F́äñçẙ"
let str2 = AttributedString(str1)

let substr1 = str1[...str1.startIndex] // “F\u{301}”
let substr2 = str2[...str2.startIndex] // “F”
```

We expect slicing on range expressions to always produce  consistent results between `String` and `AttributedString`.

When slicing character-based string views on `…str.startIndex`, we naturally expect the range expression to cover the entire first `Character`, and that is precisely what `String` does.

`AttributedString`’s default view is not as obviously `Character`-based as `String` is, but emulating the stdlib’s behavior is much preferable than to allow such simple, character-aligned range expressions to slice the attributed string at random sub-character positions.

rdar://129804049

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString.swift
@@ -292,8 +292,9 @@ extension AttributedString {
 
     public mutating func replaceSubrange(_ range: some RangeExpression<Index>, with s: some AttributedStringProtocol) {
         ensureUniqueReference()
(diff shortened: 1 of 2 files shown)


lorentey:
@swift-ci test

PR #728, opened 2024-07-11
Gracefully handle confstr failures

jmschonfeld:
There appear to be some instances where `confstr` can fail to create a temporary directory that are not the direct result of a programmer error. To accommodate this, we should handle this gracefully instead of aborting so that we fall through to either `TMPDIR` or `/tmp`.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #730, opened 2024-07-12
[Windows] Implement additional file attributes

jmschonfeld:
This implements the remaining attributes for `attributesOfItem(atPath:)` for windows. This is required for the Windows toolchain build because `TSCBasic` requires these attributes and `swiftc.exe` ends up calling down to this API while compiling SwiftPM manifests.

The implementation of all of these attributes comes from the existing sources in swift-corelibs-foundation (roughly).

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -560,7 +560,9 @@ extension _FileManagerImpl {
             defer { CloseHandle(hFile) }
 
             var info: BY_HANDLE_FILE_INFORMATION = BY_HANDLE_FILE_INFORMATION()
-            GetFileInformationByHandle(hFile, &info)
+            guard GetFileInformationByHandle(hFile, &info) else {
+              throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+            }
 
             let dwFileType = GetFileType(hFile)
             let fatType: FileAttributeType = switch (dwFileType) {
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -581,9 +581,17 @@ extension _FileManagerImpl {
             let referenceCount = UInt64(info.nNumberOfLinks)
 
             let isReadOnly = info.dwFileAttributes & FILE_ATTRIBUTE_READONLY != 0
-            var posixPermissions = UInt16(_S_IEXEC | _S_IREAD)
+            let isExecutable = if fatType == .typeDirectory {
+                true // Directories are always considered executable
+            } else {
+                SaferiIsExecutableFileType(pwszPath, 0)
+            }
+            var posixPermissions = UInt16(_S_IREAD)
             if !isReadOnly {
                 posixPermissions |= UInt16(_S_IWRITE)
+            }
+            if isExecutable {
+                posixPermissions |= UInt16(_S_IEXEC)
             }
 
             let size: UInt64 = (UInt64(info.nFileSizeHigh) << 32) | UInt64(info.nFileSizeLow)
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -553,22 +553,20 @@ extension _FileManagerImpl {
     func attributesOfItem(atPath path: String) throws -> [FileAttributeKey : Any] {
 #if os(Windows)
         return try path.withNTPathRepresentation { pwszPath in
-            var faAttributes: WIN32_FILE_ATTRIBUTE_DATA = .init()
-            guard GetFileAttributesExW(pwszPath, GetFileExInfoStandard, &faAttributes) else {
-                throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
-            }
-
             let hFile = CreateFileW(pwszPath, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, 0, nil)
             if hFile == INVALID_HANDLE_VALUE {
                 throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
             }
             defer { CloseHandle(hFile) }
 
+            var info: BY_HANDLE_FILE_INFORMATION = BY_HANDLE_FILE_INFORMATION()
+            GetFileInformationByHandle(hFile, &info)
+
             let dwFileType = GetFileType(hFile)
             let fatType: FileAttributeType = switch (dwFileType) {
                 case FILE_TYPE_CHAR: FileAttributeType.typeCharacterSpecial
                 case FILE_TYPE_DISK:
-                    faAttributes.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY
+                    info.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY
                             ? FileAttributeType.typeDirectory
                             : FileAttributeType.typeRegular
                 case FILE_TYPE_PIPE: FileAttributeType.typeSocket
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -553,49 +553,61 @@ extension _FileManagerImpl {
...
-            }
-
             let hFile = CreateFileW(pwszPath, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, 0, nil)
             if hFile == INVALID_HANDLE_VALUE {
                 throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
             }
             defer { CloseHandle(hFile) }
 
+            var info: BY_HANDLE_FILE_INFORMATION = BY_HANDLE_FILE_INFORMATION()
+            GetFileInformationByHandle(hFile, &info)
```
compnerd, line 563: Can we ensure that this succeeded?

```suggestion
            guard GetFileInformationByHandle(hFile, &info) else {
              throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
            }
```

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -553,49 +553,61 @@ extension _FileManagerImpl {
...
 
-            let size: UInt64 = (UInt64(faAttributes.nFileSizeHigh) << 32) | UInt64(faAttributes.nFileSizeLow)
-            let creation: Date = Date(timeIntervalSince1970: faAttributes.ftCreationTime.timeIntervalSince1970)
-            let modification: Date = Date(timeIntervalSince1970: faAttributes.ftLastWriteTime.timeIntervalSince1970)
-            return [
+            let systemNumber = UInt64(info.dwVolumeSerialNumber)
+            let systemFileNumber = UInt64(info.nFileIndexHigh << 32) | UInt64(info.nFileIndexLow)
+            let referenceCount = UInt64(info.nNumberOfLinks)
+
+            let isReadOnly = info.dwFileAttributes & FILE_ATTRIBUTE_READONLY != 0
```
compnerd, line 581: I think that it is safer to ensure that the bit is set ...

```suggestion
            let isReadOnly = info.dwFileAttributes & FILE_ATTRIBUTE_READONLY == FILE_ATTRIBUTE_READONLY
```

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -553,49 +553,61 @@ extension _FileManagerImpl {
...
+            var posixPermissions = UInt16(_S_IEXEC | _S_IREAD)
```
compnerd, line 582: This is technically not correct. Not everything is implicitly executable, and there is a test for this in SPM even. I think that using something like [SaferIsExecutableFileType](https://learn.microsoft.com/en-us/windows/win32/api/winsafer/nf-winsafer-saferiisexecutablefiletype) would be good to see if we should add `_S_IEXEC`.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -553,49 +553,61 @@ extension _FileManagerImpl {
...
-                // .groupOwnerAccountName: ...,
-                // .deviceIdentifier: ...,
+                // Uid is always 0 on Windows systems
+                .ownerAccountID: _writeFileAttributePrimitive(0, as: UInt.self),
+
+                // Group id is always 0 on Windows
+                .groupOwnerAccountID: _writeFileAttributePrimitive(0, as: UInt.self)
             ]
+
+            if fatType == .typeCharacterSpecial || fatType == .typeBlockSpecial {
```
compnerd, line 607: This seems ... inverted? If it is a special file then we add the device identifier? Shouldn't this be if the file is a disk file?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
al45tair: Personally I prefer the `!= 0` version. If we're doing a bitwise AND, we _know_ the answer is either `0` or `FILE_ATTRIBUTE_READONLY`. The only way we could trip up is if `FILE_ATTRIBUTE_READONLY` is not really a bit mask, but that would be wrong for _either_ formulation (and we already know it's a bit mask anyway).

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
al45tair: `SaferIsExecutableFileType` is a good idea here, but don't forget that if this is a directory, it should also show the `x` bit as set.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
al45tair: I thought `st_rdev` (which is the field we're emulating here) was only relevant if it's a device node. I'm not actually sure what `dwVolumeSerialNumber` would be for e.g. `LPT` or `COM1`, mind; the documentation doesn't really say what it'll give for those. I'm not sure `dwVolumeSerialNumber` is the right value here, mind.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: Sure - this was copied over from SCL-F effectively but if we can do something better here I'm all for it. So in this case, we should mark it as executable if `SaferIsExecutableFileType` provides a `true` value or if the `type` set above is `.typeDirectory` right?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: Oh interesting, so this comes from what we have in SCL-F today and I missed a detail. SCL-F has:

```swift
statInfo.st_rdev = _dev_t(info.dwVolumeSerialNumber)
```

and then

```swift
#if os(Windows)
        result[.deviceIdentifier] = NSNumber(value: UInt64(s.st_rdev))
#endif

if type == .typeBlockSpecial || type == .typeCharacterSpecial {
    result[.deviceIdentifier] = NSNumber(value: UInt64(s.st_rdev))
}
```

So it seems that we only set `deviceIdentifier` for blockSpecial/characterSpecial on other platforms but on Windows we unconditionally set it. Should I go with that behavior here to mirror what SCL-F did?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: Yes, that is correct.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: It looks like the tools-support-core repo doesn't use that value, so that might be a viable option

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: Sounds good, fixed!

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -581,10 +581,18 @@ extension _FileManagerImpl {
             let referenceCount = UInt64(info.nNumberOfLinks)
 
             let isReadOnly = info.dwFileAttributes & FILE_ATTRIBUTE_READONLY != 0
-            var posixPermissions = UInt16(_S_IEXEC | _S_IREAD)
+            let isExecutable = if fatType == .typeDirectory {
+                true // Directories are always considered executable
+            } else {
+                SaferiIsExecutableFileType(pwszPath, 0)
```
compnerd:
```suggestion
                SaferiIsExecutableFileType(pwszPath, false)
```

The second parameter is a boolean, might as well as match that :)

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -553,49 +553,68 @@ extension _FileManagerImpl {
...
-            let creation: Date = Date(timeIntervalSince1970: faAttributes.ftCreationTime.timeIntervalSince1970)
-            let modification: Date = Date(timeIntervalSince1970: faAttributes.ftLastWriteTime.timeIntervalSince1970)
-            return [
+            let systemNumber = UInt64(info.dwVolumeSerialNumber)
+            let systemFileNumber = UInt64(info.nFileIndexHigh << 32) | UInt64(info.nFileIndexLow)
+            let referenceCount = UInt64(info.nNumberOfLinks)
+
+            let isReadOnly = info.dwFileAttributes & FILE_ATTRIBUTE_READONLY != 0
+            // Directories are always considered executable, but we check for other types
+            let isExecutable = fatType == .typeDirectory || SaferiIsExecutableFileType(pwszPath, 0)
```
compnerd, line 585:
```suggestion
            let isExecutable = fatType == .typeDirectory || SaferiIsExecutableFileType(pwszPath, false)
```

jmschonfeld: Unfortunately that didn't compile for me:

```
C:\Users\jmschonfeld\Documents\Repos\swift-foundation\Sources\FoundationEssentials\FileManager\FileManager+Files.swift:587:54: error: cannot convert value of type 'Bool' to expected argument type 'BOOLEAN' (aka 'UInt8')
585 |                 true // Directories are always considered executable
586 |             } else {
587 |                 SaferiIsExecutableFileType(pwszPath, false)
    |                                                      `- error: cannot convert value of type 'Bool' to expected argument type 'BOOLEAN' (aka 'UInt8')
588 |             }
589 |             var posixPermissions = UInt16(_S_IREAD)
```

It looks like `BOOLEAN` is `UInt8` and not expressible by a boolean literal?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jmschonfeld: I went ahead and updated this to a TODO, based on some GitHub searching I don't think this should affect anything in the toolchain build

jmschonfeld:
@swift-ci please test

al45tair (approved):

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
compnerd: That is good to know. We can deal with that later if we want.

compnerd (approved):

PR #731, opened 2024-07-12
[Windows] Standardize slashes before path processing

jmschonfeld:
All of our `String`-based path processing assumes that the paths are standardized to forward slashes, however there are many sources (such as `FileManager`) where these paths may come in on Windows with backslashes resulting in incorrect path manipulation. This updates all of these functions to standardize to forward slashes where needed when performing path-related functions. On windows, these slashes will already be automatically changed back to backslashes if necessary by the file system representation functions.

_Note: this currently doesn't handle drive letters any differently than other path components. We can probably do better there, but this at least unblocks the toolchain work and gets us on-par with what SCL-F used to do_

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -29,7 +29,11 @@ extension StringProtocol {
         // Standardize the path to use forward slashes before processing for consistency
         return self.replacing(._backslash, with: ._slash)
         #else
-        return String(self)
+        if let str = _specializingCast(self, to: String.self) {
+            return str
+        } else {
+            return String(self)
+        }
         #endif
     }
 }
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -473,15 +473,11 @@ extension String {
     // From swift-corelibs-foundation's NSTemporaryDirectory. Internal for now, pending a better public API.
     internal static var temporaryDirectoryPath: String {
         func normalizedPath(with path: String) -> String {
-            guard path.utf8.last != ._slash else {
-                return path
+            var result = path._standardizingSlashes()
+            guard result.utf8.last != ._slash else {
+                return result
             }
-            #if os(Windows)
-            guard path.utf8.last != ._backslash else {
-                return path
-            }
-            #endif
-            return path + "/"
+            return result + "/"
         }
 #if os(Windows)
         let cchLength: DWORD = GetTempPathW(0, nil)
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -22,8 +22,24 @@ import WinSDK
 
 internal import _FoundationCShims
 
+extension StringProtocol {
+    fileprivate func _standardizingSlashes() -> String {
+        #if os(Windows)
+        // The string functions below all assume that the path separator is a forward slash
+        // Standardize the path to use forward slashes before processing for consistency
+        return self.replacing(._backslash, with: ._slash)
+        #else
+        return String(self)
+        #endif
+    }
+}
+
 extension String {
     internal func deletingLastPathComponent() -> String {
+        _standardizingSlashes()._deletingLastPathComponent()
+    }
+    
+    private func _deletingLastPathComponent() -> String {
         let lastSlash = self.lastIndex { $0 == "/" }
         guard let lastSlash else {
             // No slash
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -282,10 +282,16 @@ final class FileManagerTests : XCTestCase {
             try $0.createDirectory(atPath: "create_dir_test2/nested2", withIntermediateDirectories: true)
             XCTAssertEqual(try $0.contentsOfDirectory(atPath: "create_dir_test2").sorted(), ["nested", "nested2"])
             XCTAssertNoThrow(try $0.createDirectory(atPath: "create_dir_test2/nested2", withIntermediateDirectories: true))
+            
+            #if os(Windows)
+            try $0.createDirectory(atPath: "create_dir_test3\\nested", withIntermediateDirectories: true)
```
jmschonfeld, line 287: This is an example of a previous API that used to fail without this change. Previously, this API called `deletingLastPathComponent` on the provided path, which would return an empty string and therefore fail to create the parent `create_dir_test3` directory. Now it correctly creates `create_dir_test3` followed by `create_dir_test3/nested`

parkera (approved):

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -170,11 +194,11 @@ extension String {
             return false
         }
         if let lastDot = pathExtension.utf8.lastIndex(of: UInt8(ascii: ".")) {
-            let beforeDot = pathExtension[..<lastDot].unicodeScalars
-            let afterDot = pathExtension[pathExtension.index(after: lastDot)...].unicodeScalars
+            let beforeDot = pathExtension[..<lastDot]._standardizingSlashes().unicodeScalars
+            let afterDot = pathExtension[pathExtension.index(after: lastDot)...]._standardizingSlashes().unicodeScalars
```
compnerd, line 198: Where do we expect an arc separator _after_ the extension separator? That is, the extension should be on the final arc, and so there would be no slashes right?

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -440,18 +472,16 @@ extension String {
...
+                return path
+            }
+            #if os(Windows)
+            guard path.utf8.last != ._backslash else {
                 return path
-            } else {
-                return path + String(validPathSeps.last!)
             }
+            #endif
+            return path + "/"
```
compnerd, line 484: Should this be `path._standardizingSlashes() + "/"`?

compnerd (approved):
Seems like there are two small improvements that can bee made, but seems fine otherwise.

Sources/FoundationEssentials/String/String+Path.swift
jmschonfeld: We don't expect it to occur, but this is to ensure that we validate that it doesn't. I can't come up with any scenarios where we would currently end up in this situation, but `invalidExtensionScalars` contains a `/` and so to ensure that POSIX and Windows behaviors are consistent, I added this standardization here so that if we do somehow end up with a `\` in the path extension that we mark it as invalid on Windows

Sources/FoundationEssentials/String/String+Path.swift
jmschonfeld: That seems like a good addition - today on Windows this only gets called with the result of `standardizingPath` which will already have its slashes standardized (so the `if` statement above seems like it might always be a no-op) but adding the extra check here is probably a good measure for future-proofing this code

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -22,8 +22,24 @@ import WinSDK
...
 internal import _FoundationCShims
 
+extension StringProtocol {
+    fileprivate func _standardizingSlashes() -> String {
+        #if os(Windows)
+        // The string functions below all assume that the path separator is a forward slash
+        // Standardize the path to use forward slashes before processing for consistency
+        return self.replacing(._backslash, with: ._slash)
+        #else
+        return String(self)
```
iCharlesHu, line 32: (nit) is it worth checking whether self is String and just directly return, since most of the extensions below are actually on String?

iCharlesHu (approved):

iCharlesHu (approved):

Sources/FoundationEssentials/String/String+Path.swift
jmschonfeld: Sure - I updated it to use a specializing cast to determine if `self` is already `String`. I confirmed that when compiling with `-O` the optimizer is smart enough to elide the `_standardizingSlashes()` call entirely when called from an extension on `String`

jmschonfeld:
@swift-ci please test

PR #732, opened 2024-07-12
Disable testResolveSymlinksViaGetAttrList() on non-Darwin

jmschonfeld:
This test attempts to test an optimization that is only present on Darwin (using `getattrlist` to resolve symlinks when possible). I encountered a situation on Windows on my local device where calling this function would standardize other portions of the path resulting in test failures. Since this test isn't applicable on non-Darwin anyways, this disables the test to ensure we don't hit accidental failures depending on local device setups.

In my device's case, the expected destination path was

```
C:\\Users\\jmschonfeld\\AppData\\Local\\Temp\\FileManagerPlayground_F2BEE091-140C-478E-91B2-1EBB08B88E5C\\destination
```

However calling this function was standardizing this to

```
C:\\Users\\JMSCHO~1\\AppData\\Local\\Temp\\FileManagerPlayground_F2BEE091-140C-478E-91B2-1EBB08B88E5C\\destination
```

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -686,6 +686,9 @@ final class FileManagerTests : XCTestCase {
     }
     
     func testResolveSymlinksViaGetAttrList() throws {
+        #if !canImport(Darwin)
+        throw XCTSkip("This test is not applicable on this platform")
+        #else
         try FileManagerPlayground {
             "destination"
         }.test {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

compnerd (approved):
If this is a Darwin specific test, that makes sense to skip. However, the normalization seems to be a short path thing ... which is odd. I don't expect to see short paths in the results.

jmschonfeld:
> If this is a Darwin specific test, that makes sense to skip. However, the normalization seems to be a short path thing ... which is odd. I don't expect to see short paths in the results.

Yeah this test is really intended to test `String._resolvingSymlinksInPathUsingFullPathAttribute()` which is an optimization for Darwin but is a no-op on other platforms. If that normalization is unexpected we should definitely look into that, I think it must be coming from `GetFinalPathNameByHandleW`?

PR #734, opened 2024-07-15
Use correct calculation for Date's current time

parkera:
Correct the constant value used for Date's current time calculation, and put in a smoke test for the result. Also, use `Double` here so we get sub-1s precision.


parkera:
@swift-ci test

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

compnerd (approved):

PR #735, opened 2024-07-15
Introduce Decimal arithmetics

iCharlesHu:
This patch implements Decimal arithmetics (plus, minus, power, etc) which completes Decimal's implementation.

### DIFF
--- a/Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
+++ b/Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
@@ -22,7 +22,7 @@ extension Decimal : CustomStringConvertible {
         if let decimalSeparator = locale?.decimalSeparator {
             decimalString = decimalString.replacing(decimalSeparator, with: ".")
         }
-        guard let value = Decimal.decimal(from: decimalString.utf8, matchEntireString: false) else {
+        guard let value = Decimal.decimal(from: decimalString.utf8, matchEntireString: false).result else {
             return nil
         }
         self = value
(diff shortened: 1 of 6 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
```diff
@@ -310,6 +307,56 @@ extension Decimal {
 
     @available(*, unavailable, message: "Decimal does not yet fully adopt FloatingPoint.")
     public mutating func formTruncatingRemainder(dividingBy other: Decimal) { fatalError("Decimal does not yet fully adopt FloatingPoint") }
+
+    public var nextUp: Decimal {
```
itingliu, line 311: Double checking: these new `public var` aren't new API, right? Did you move them from somewhere else?

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
```diff
@@ -445,10 +538,8 @@ extension Decimal : Codable {
 }
 
 // MARK: - SignedNumeric
-// SwiftFoundation's `Decimal` does not fully conform to
-// SignedNumeric yet because no arithmetics have been implemented
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
-extension Decimal /* : SignedNumeric */ {
+extension Decimal : SignedNumeric {
```
itingliu, line 542: I presume this addition is for in `swift-foundation` only and not FOUNDATION_FRAMEWORK?

itingliu (approved):

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
iCharlesHu: They are technically new API for SwiftFoundation but not for Foundation in general (it was only this way because initially we were trying to only bring the minimal of what we needed).

iCharlesHu:
@swift-ci test macOS

iCharlesHu:
Added `Decimal+Math.swift` to CMakeLists.txt

iCharlesHu:
@swift-ci please test

PR #736, opened 2024-07-15
Limit Synchronization usage to FOUNDATION_FRAMEWORK

jmschonfeld:
There are some issues with using the `Synchronization` module from within the toolchain on Linux so for now we'll limit its usage to `FOUNDATION_FRAMEWORK` only


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #739, opened 2024-07-17
Enable reading files of indeterminate length on Linux

jmschonfeld:
Some files on Linux (such as those in `/proc`) report a size of `0` when stat-ing the files. However, these files do have contents, but the length is not able to be predetermined. On Linux, we now always attempt a read in chunks even if the file reports as empty and grow our buffer as necessary to accommodate the resulting size.


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

parkera (approved):

PR #740, opened 2024-07-17
Re-add sendable annotations

jmschonfeld:
We previously reverted some sendable annotations in https://github.com/apple/swift-foundation/pull/679 due to a compiler bug around the IRGen for `KeyPath & Sendable` parameters. While the compiler bug is not yet resolved, this adds these annotations back in place using a workaround to prevent the compiler bug from causing issues on ABI stable platforms. This ensures that our public API is the most correct, and in the future when the bug is resolved we can remove the additional AEIC functions that serve as a workaround currently.


jmschonfeld:
@swift-ci please test

itingliu (approved):
@swift-ci please test

PR #742, opened 2024-07-18
Re-Add Synchronization Import on non-Darwin

jmschonfeld:
Previously, we had issues with import `Synchronization` in the toolchain - this was due to the rpaths of the produced libraries. We previously applied `-no-toolchain-stdlib-rpath` so that the rpath of the libraries in the build folder had no rpaths set. When the libraries are installed into the toolchain, `$ORIGIN` was appended which is correct for the installed library. However, XCTest's tests link against these libraries directly in the build folder before they are installed. Loading the stdlib worked because XCTest's tests also link the stdlib and so the library was found in the cache, however since XCTest's tests do not link `Synchronization` it wasn't found in the cache and since the rpaths list was empty the library couldn't be found.

Instead, we now remove the `-no-toolchain-stdlib-rpath` argument so that the libraries in the build directory have rpaths pointing to the swift build folder. However, we also now set `INSTALL_REMOVE_ENVIRONMENT_RPATH` to `ON` so that when copying to the installation folder, this path is _replaced_ by `$ORIGIN` instead of just appending the `$ORIGIN` path to the list so that the produced toolchain doesn't contain local rpaths.

This should also resolve a related warning in the macro due to the lack of rpaths while loading the macro from the build folder to build FoundationEssentials itself:

```
/home/build-user/swift-foundation/Sources/FoundationEssentials/Predicate/Expression.swift:32:14: warning: external macro implementation type 'FoundationMacros.ExpressionMacro' could not be found for macro 'Expression'; failed to load library plugin '/home/build-user/build/buildbot_linux/foundation-linux-x86_64/lib/libFoundationMacros.so' in plugin server '/home/build-user/build/buildbot_linux/swift-linux-x86_64/lib/swift/host/libSwiftInProcPluginServer.so'; loader error: libswift_StringProcessing.so: cannot open shared object file: No such file or directory
```

### DIFF
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -79,11 +79,9 @@ if(NOT BUILD_SHARED_LIBS)
         "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend _FoundationCollections>")
 endif()
 
-target_link_options(FoundationEssentials PRIVATE
-    "SHELL:-no-toolchain-stdlib-rpath")
-
 set_target_properties(FoundationEssentials PROPERTIES
-    INSTALL_RPATH "$ORIGIN")
+    INSTALL_RPATH "$ORIGIN"
+    INSTALL_REMOVE_ENVIRONMENT_RPATH ON)
 
 set_property(GLOBAL APPEND PROPERTY SWIFT_FOUNDATION_EXPORTS FoundationEssentials)
 _swift_foundation_install_target(FoundationEssentials)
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
Full toolchain tests and a full toolchain build is running at https://github.com/swiftlang/swift/pull/75321

parkera (approved):
Awesome, thanks so much for tracking this down.

jmschonfeld:
@swift-ci please test

Azoy:
@swift-ci please test

jmschonfeld:
@swift-ci please test

parkera:
For now I had to remove the file that used Synchronization, so this PR actually becomes somewhat of a no-op in the short term. However, I am sure we will want Synchronization back soon.

jmschonfeld:
> For now I had to remove the file that used Synchronization, so this PR actually becomes somewhat of a no-op in the short term. However, I am sure we will want Synchronization back soon.

Sounds good - I've rebased on `main` so that we can still land this so that `PredicateExpressions.Variable` can still use `Synchronization` (and wherever else we need it in the future)

jmschonfeld:
@swift-ci please test

PR #743, opened 2024-07-18
Implement isoLatin1 and macOSRoman encoding

jmschonfeld:
As we're re-coring some of the String <--> data functions in swift-corelibs-foundation, we've found places in swift-corelibs-foundation that rely on being able to encode to/from the `isoLatin1` and `macOSRoman` text encodings. This implements conversion for those encodings in swift-foundation


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
Looks like on linux `fgetxattr` is failing with `EMULTIHOP` 🤔 - I'll take another look at this to see where we might be doing something wrong

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
I resolved the issue by excluding these encodings on Linux (for now) and Windows. @parkera / @itingliu / @iCharlesHu this is now ready for review to unblock re-enabling this functionality in swift-corelibs-foundation's tests

Sources/FoundationEssentials/String/String+IO.swift
```diff
@@ -14,8 +14,38 @@
 internal import _ForSwiftFoundation
 #endif
 
+internal import _FoundationCShims
+
 fileprivate let stringEncodingAttributeName = "com.apple.TextEncoding"
 
+private struct ExtendingToUTF16Sequence<Base: Sequence<UInt8>> : Sequence {
```
parkera, line 21: I found this Sequence adaptor pattern to have less than stellar performance in the past. It's probably good to get us started but we should note that it may be possible to optimize.

parkera (approved):

Sources/FoundationEssentials/String/String+IO.swift
jmschonfeld: Ah good to know, yeah this is probably worth a performance benchmark as a followup here

jmschonfeld:
@swift-ci please test

PR #719, opened 2024-07-05
Improve performance of FormatStyle formatting when used in multithreaded context

parkera:
Improves the performance of formatting dates and times, especially when used in multithreaded contexts. 

This patch reworks some of the caching in the Locale, TimeZone, and Calendar types to more efficiently store the singleton values.

## Before

```
parallel-and-serialized-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │   6884 │
╘═══════════════════════════╧════════╛

parallel-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │   8330 │
╘═══════════════════════════╧════════╛

serialized-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │   1226 │
╘═══════════════════════════╧════════╛
```

## After

```
parallel-and-serialized-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │   3607 │
╘═══════════════════════════╧════════╛

parallel-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │   4651 │
╘═══════════════════════════╧════════╛

serialized-number-formatting
╒═══════════════════════════╤════════╕
│ Time (total CPU) (μs) *   │    505 │
╘═══════════════════════════╧════════╛
```

PR #744, opened 2024-07-18
Revert formatting performance improvements

parkera:
There appears to be some kind of race or memory smash in ICU after these, and we need more time to investigate the full root cause. (reverts #719)


parkera:
@swift-ci test

itingliu (approved):

PR #747, opened 2024-07-19
Decimal.init(sign:exponent:significand:) should clamp exponent

iCharlesHu:
While I was here: imported more Decimal tests from swift-corelibs-foundation


iCharlesHu:
@swift-ci please test

parkera (approved):

PR #750, opened 2024-07-19
Decimal.toString should accept custom separator

iCharlesHu:
`NSDecimalString` accepts a dictionary in addition to to locale. Updating `Decimal.toString` to take the decimal separator directly instead of a locale.


iCharlesHu:
@swift-ci please test

PR #752, opened 2024-07-19
Fix URL percent-decoding of %00

jrflat:
`URLParser`'s percent-decoding function used `String(validatingUTF8:)`, which would truncate the string if we had just decoded `%00` to NULL. Instead of using a NULL-terminated `OutputBuffer`, I switched to using `String(_validating:as:)`, which handles/accepts NULL characters in the buffer.


jrflat:
@swift-ci please test

parkera (approved):

PR #753, opened 2024-07-22
build: install linker files for installed targets

compnerd:
Ensure that we install any linker files for the modules which are required on certain platforms.

### DIFF
--- a/Sources/FoundationMacros/CMakeLists.txt
+++ b/Sources/FoundationMacros/CMakeLists.txt
@@ -59,3 +59,5 @@ install(TARGETS FoundationMacros
     ARCHIVE DESTINATION lib/swift/host/plugins
     LIBRARY DESTINATION lib/swift/host/plugins
     RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(FILES $<TARGET_LINKER_FILE:FoundationMacros>
+  DESTINATION lib)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

Sources/FoundationInternationalization/CMakeLists.txt
```diff
@@ -54,3 +54,5 @@ set_target_properties(FoundationInternationalization PROPERTIES
 
 set_property(GLOBAL APPEND PROPERTY SWIFT_FOUNDATION_EXPORTS FoundationInternationalization)
 _swift_foundation_install_target(FoundationInternationalization)
+install(FILES $<TARGET_LINKER_FILE:FoundationInternationalization>
```
jmschonfeld, line 57: nit: should we just add this to `_swift_foundation_install_target` so that this is automatically applied for any targets we might install in the future?

compnerd: Sure, let me sink it into that.

compnerd:
@swift-ci please test

jmschonfeld:
Thanks, this seems fine to me! Before merging could we kick off a toolchain build or full swift test with this change just to get some good coverage over the CMake changes? We don't have that setup on this repo yet so it'll need to be kicked off via a PR to the swiftlang/swift repo (or the swift-corelibs-foundation repo) to double check this doesn't cause any issues in the toolchain build on linux+windows

compnerd:
Sure! Kicked that off at https://github.com/swiftlang/swift/pull/65907. I'll follow up with the installer changes later, I want to get this in to unblock CI

jmschonfeld (approved):
Awesome thanks! Approved and good to merge after that build finishes successfully

compnerd:
https://ci-external.swift.org/job/swift-PR-build-toolchain-windows/5374/ was the run.

PR #754, opened 2024-07-22
rdar://131793235 (Invalid JSON string causes precondition failure)

kperryua:
Very large number values, like 100e200 are technically valid JSON, and so are not flagged by the parsing layer as invalid. When decoded as a Decimal, which cannot hold this value, we need a signal that differentiates that the value is too large compared to the input being unparseable. This allows JSONDecoder to throw the correct kind of error instead of hitting a fatal error.

This also generalizes the parsing implementation on Collection<UTF8.CodeUnit> so that JSONDecoder doesn't have to instantiate a throw-away String.


kperryua:
@swift-ci test

parkera (approved):

PR #756, opened 2024-07-22
Use dynamic replacement instead of _typeByName for internationalization upcalls

jmschonfeld:
We currently use `_typeByName` to "up call" from types in FoundationEssentials to types in FoundationInternationalization (mainly for finding ICU-based versions of types defined in the essentials module). Now that swift-foundation exists in the toolchain, `_typeByName` is conflating the two versions of FoundationInternationalization when building a copy locally resulting in test crashes and failures. Instead, this transitions to the use of `@_dynamicReplacement` in order to swap out the up-call function at runtime for an implementation that returns the resulting type. Since the replacement implementation is statically type checked at compile time, it statically references the correct version of the type avoiding the failures.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLParser.swift
+++ b/Sources/FoundationEssentials/URL/URLParser.swift
@@ -158,13 +158,15 @@ package protocol UIDNAHook {
     static func decode(_ host: some StringProtocol) -> String?
 }
 
-dynamic package func _uidnaHook() -> UIDNAHook.Type? {
-    #if FOUNDATION_FRAMEWORK && canImport(_FoundationICU)
+#if FOUNDATION_FRAMEWORK && canImport(_FoundationICU)
+internal func _uidnaHook() -> UIDNAHook.Type? {
     UIDNAHookICU.self
-    #else
-    nil
-    #endif
 }
+#else
+dynamic package func _uidnaHook() -> UIDNAHook.Type? {
+    nil
+}
+#endif
 
 internal struct RFC3986Parser: URLParserProtocol {
     static let kind: URLParserKind = .RFC3986
--- a/Sources/FoundationEssentials/URL/URLParser.swift
+++ b/Sources/FoundationEssentials/URL/URLParser.swift
@@ -158,17 +158,16 @@ package protocol UIDNAHook {
     static func decode(_ host: some StringProtocol) -> String?
 }
 
+dynamic package func _uidnaHook() -> UIDNAHook.Type? {
+    #if FOUNDATION_FRAMEWORK && canImport(_FoundationICU)
+    UIDNAHookICU.self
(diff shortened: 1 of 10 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

Sources/FoundationInternationalization/URLParser+ICU.swift
```diff
@@ -15,7 +15,14 @@ import FoundationEssentials
...
 
-internal final class UIDNAHookICU: UIDNAHook {
+#if !FOUNDATION_FRAMEWORK
+@_dynamicReplacement(for: _uidnaHook())
+private func _uidnaHook_localized() -> UIDNAHook.Type? {
+    return UIDNAHookICU.self
+}
+#endif
+
+private struct UIDNAHookICU: UIDNAHook {
```
itingliu, line 25: Did I misread it or was `UIDNAHookICU` changed from `final class` to a `struct` 👀?  Is this intentional?

Sources/FoundationEssentials/URL/URLParser.swift
```diff
@@ -158,17 +158,16 @@ package protocol UIDNAHook {
     static func decode(_ host: some StringProtocol) -> String?
 }
 
+dynamic package func _uidnaHook() -> UIDNAHook.Type? {
```
itingliu, line 161: Nit:
From the [implementation PR](https://github.com/swiftlang/swift/pull/20333/files#diff-d425fbf14611e9cd6298e6ef29c1c915d9c757e739b9095e7b097858e6ee7740R244) it looks like dynamic replacement would work with properties too. So if we want to keep this as a read-only `var` I suppose we can keep doing that. I'm fine with this though.

Sources/FoundationInternationalization/URLParser+ICU.swift
jmschonfeld: Yeah, the only reason it was a `final class` was so that `_typeByName` would work (it can only lookup classes). I changed it to a `struct` to be more in-line with how we typically write swift types now that a class isn't required

Sources/FoundationEssentials/URL/URLParser.swift
jmschonfeld: Ah yeah I could do that too, I'm fine with either - happy to change it if you have a preference for the variable spelling

Sources/FoundationInternationalization/URLParser+ICU.swift
itingliu: Thanks that makes total sense.

itingliu (approved):

jmschonfeld:
@swift-ci please test

PR #757, opened 2024-07-23
[android] fix the android build

hyp:
Macros can't yet be built as there's no host build support yet

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
@@ -166,8 +166,8 @@ struct _FTSSequence: Sequence {
                 return
             }
 
-            state = [UnsafeMutablePointer(mutating: path), nil].withUnsafeBufferPointer { dirList in
-                guard let stream = fts_open(dirList.baseAddress!, opts, nil) else {
+            state = [Optional(UnsafeMutablePointer(mutating: path)), nil].withUnsafeBufferPointer { dirList in
+                guard let stream = fts_open(unsafeBitCast(dirList.baseAddress!, to: UnsafePointer<UnsafeMutablePointer<CChar>>.self), opts, nil) else {
                     return .error(errno, String(cString: path))
                 }
                 return .stream(stream)
--- a/Sources/CMakeLists.txt
+++ b/Sources/CMakeLists.txt
@@ -14,8 +14,8 @@
 
 add_subdirectory(_FoundationCShims)
 
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+# Disable the macro build on Windows and Android until we can correctly build it for the host architecture
+if(NOT (CMAKE_SYSTEM_NAME STREQUAL "Windows" OR CMAKE_SYSTEM_NAME STREQUAL "Android"))
 	add_subdirectory(FoundationMacros)
 endif()
 
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -48,7 +48,7 @@ add_subdirectory(String)
 add_subdirectory(TimeZone)
 add_subdirectory(URL)
 
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+if(NOT (CMAKE_SYSTEM_NAME STREQUAL "Windows" OR CMAKE_SYSTEM_NAME STREQUAL "Android"))
     # Depend on FoundationMacros
     add_dependencies(FoundationEssentials FoundationMacros)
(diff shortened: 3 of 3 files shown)


hyp:
@swift-ci please test

hyp:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
```diff
@@ -166,8 +166,13 @@ struct _FTSSequence: Sequence {
                 return
             }
 
-            state = [UnsafeMutablePointer(mutating: path), nil].withUnsafeBufferPointer { dirList in
-                guard let stream = fts_open(dirList.baseAddress!, opts, nil) else {
+            state = [Optional(UnsafeMutablePointer(mutating: path)), nil].withUnsafeBufferPointer { dirList in
+#if canImport(Android)
+                let baseAddress = unsafeBitCast(dirList.baseAddress!, to: UnsafePointer<UnsafeMutablePointer<CChar>>.self)
```
jmschonfeld, line 171: Could you elaborate a bit more on what error this change resolves? I would have assumed that the above optional wrapping wouldn't be necessary since it'd be inferred by the compiler, and it seems that this `unsafeBitCast` of an optional pointer to a non-optional pointer would be a trap waiting for a caller that tries to read that final `nil` element, right?

hyp: This is needed because the Android NDK has a nullability annotation on `fts_open`:

```
FTS* _Nullable fts_open(char* _Nonnull const* _Nonnull __path, int __options, int (* _Nullable __comparator)(const FTSENT* _Nonnull * _Nonnull  __lhs, const FTSENT* _Nonnull * _Nonnull __rhs)) __INTRODUCED_IN(21);
```

This makes the path pointer nonnullable . Therefore the type inference is then unable to infer the type for the array, and thus I need to explicitly make the first element optional, so that the second one can be null. Then the pointer need to be cast to match the type, as we still want to have a nil element in the array, but have non null pointer types.

hyp: In general this stems from the fact that this specific nullability annotation in the NDK is probably incorrect, as `fts_open` is documented to take an array that has NULL as the last element.

hyp: I added a comment in source.

hyp:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
finagolfin: This change should no longer be needed after swiftlang/swift#74829, as [noted on your earlier commit weeks ago](https://github.com/apple/swift-corelibs-foundation/pull/4889#discussion_r1664581775).

hyp: Good point, I need to retry with that change.

compnerd:
I think that we should consider #714 as an alternative here as that is a broader fix.

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
finagolfin: Note [the additional linux change that I mentioned there](https://github.com/swiftlang/swift/pull/74829#discussion_r1681611802), we'll need to do something like that for that change to register on the linux CI.

finagolfin: Oh wait, that pull isn't needed at all on the linux CI, as this function signature only varies in Bionic.

In that case, this pull can just be closed, if #714 gets macros working and another small apinotes pull avoids the need for this change.

hyp:
This now drops the fts_open change - as posix_filesystem.apinotes are now picked up correctly : https://github.com/swiftlang/swift/pull/75494

hyp:
> I think that we should consider https://github.com/apple/swift-foundation/pull/714 as an alternative here as that is a broader fix.
@compnerd Good point, let me test that change for Android instead.

hyp:
@swift-ci please test

hyp:
@swift-ci please test

hyp:
Ok, macros work for android with:

https://github.com/compnerd/swift-foundation/pull/1
https://github.com/swiftlang/swift/pull/75549

If we land host macros first, we can drop this PR. wdyt @compnerd, or should this land first .

hyp:
I talked to @compnerd and I think we can hold off from this one, as the host macros changes should land soon.

compnerd:
@hyp I think that this can be closed now?

jmschonfeld:
I think the changes here are either no longer necessary or have been applied via other PRs so I'm going to go ahead and close this. If any of these changes are still relevant, feel free to update the PR and re-open

PR #759, opened 2024-07-23
Only provide .busy file attribute when true

jmschonfeld:
The previous ObjC implementation only set the `.busy` attribute value when true and omitted the key when false. This restores the historic behavior for compatibility with existing clients.


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #760, opened 2024-07-23
Update availability release for FoundationPredicateRegex

jmschonfeld:
The new stdlib APIs that FoundationPredicateRegex relies upon are present in macOS 15 aligned releases, so we can update the availability accordingly. Additionally, while I was here I added the same "generation" logic to the CMake files as that in the Package manifest so adding/updating availability in the future is easier.


jmschonfeld:
@swift-ci please test

jmschonfeld:
Windows and Linux toolchain builds are successful with this change so the CMake changes appear OK

iCharlesHu (approved):

PR #761, opened 2024-07-23
Skip testSearchPaths_XDGEnvironmentVariables when XDG environment is set

jmschonfeld:
When an XDG environment variable is set, we should skip this test because it may impact the result of this test (where we'd prefer the XDG environment variable over the `HOME` environment variable)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #762, opened 2024-07-24
Avoid checking filesystem when constructing path errors

jmschonfeld:
When we throw a path-based error from `FileManager`, we attempt to construct a URL from this path to include in the user info dictionary which ends up checking the file system. This file system check is mostly unnecessary since it's highly likely that the file either doesn't exist, is inaccessible, or the path is malformed. Instead, let's infer whether the file is a directory via the path itself as a best effort attempt to improve performance while creating these errors


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #765, opened 2024-07-25
`Calendar.date(from: <DateComponents>)` returns incorrect result when…

itingliu:
… `year`, `weekOfYear`, `yearForWeekOfYear` are set.

Gregorian calendar's implementation for `date(from: <DateComponents>)` is incorrect for a `DateComponents` when `year`, `weekOfYear`, `yearForWeekOfYear` are set. Restore the behavior of `CalendarICU`, where `yearForWeekOfYear` is preferred to `year` when both are set. This would allow the calculation to use the `weekOfYear` field, which is set in the date components, instead an unset `day` field.

Drive-by fix: give `weekdayOrdinal` a higher priority over other week fields when multiple of them are set. This isn't really an issue as this configuration is ambiguous by nature, but doing so makes it consistent with `CalendarICU`'s behavior.

Fixes 130203724

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -134,19 +134,19 @@ enum ResolvedDateComponents {
                 adjustedYear = year
             }
             self = .day(year: adjustedYear, month: month, day: d, weekOfYear: components.weekOfYear)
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

parkera (approved):

itingliu:
@swift-ci please test

PR #766, opened 2024-07-25
URL: Don't percent-encode an appended component as a first path segment

jrflat:
`URL.appending(path:directoryHint:)` currently percent-encodes `path` like any other full path, where `:` is percent-encoded in the first path segment. However, `path` is generally a relative component, and in that case, `:` is percent-encoded when it shouldn’t be.


jrflat:
@swift-ci please test

parkera (approved):

jrflat:
Resolves https://github.com/apple/swift-corelibs-foundation/issues/5028

PR #767, opened 2024-07-25
Bare minimum [U]Int128 support for JSON encode/decode

stephentyrone:
This doesn't yet tackle _slowpath_unwrapFixedWidthInteger, but at least allows us to round-trip [U]Int128 with our own encoder and decoder.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONDecoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONDecoder.swift
@@ -1128,6 +1128,11 @@ extension JSONDecoderImpl : SingleValueDecodingContainer {
     func decode(_: Int64.Type) throws -> Int64 {
         try decodeFixedWidthInteger()
     }
+  
+    @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
+    func decode(_: Int128.Type) throws -> Int128 {
+      try decodeFixedWidthInteger()
+    }
 
     func decode(_: UInt.Type) throws -> UInt {
         try decodeFixedWidthInteger()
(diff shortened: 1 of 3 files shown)


kperryua:
This piece is fine. BPlist might be a separable piece, considering its complication, if you prefer to split them up.
However, XMLPlistDecoder should be pretty straightforward to include here since it ultimately uses the same `_parseHexIntegerDigits()` and `_parseIntegerDigits()` that `JSONDecoder` uses.

kperryua (approved):

Sources/FoundationEssentials/JSON/JSONDecoder.swift
```diff
@@ -1128,6 +1128,11 @@ extension JSONDecoderImpl : SingleValueDecodingContainer {
     func decode(_: Int64.Type) throws -> Int64 {
         try decodeFixedWidthInteger()
     }
+  
+    @available(macOS 15.0, iOS 18.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
```
jmschonfeld, line 1132: In other places in the project, we use availability like `FoundationPreview 0.4` (which "back deploys" to macOS 13.3) or `FoundationPredicateRegex 0.4` (which has macOS 15 aligned availability). Is it worth using/creating a FoundationPreview-based availability specifier for this, or should we just explicitly write out the availability here like it is now?

parkera: Another benefit of using the FoundationPreview macro is that it is defined to be the min deployment target when building the package, making it possible to test this implementation on older OSes instead of requiring a host OS of macOS 15.

jmschonfeld: I think in this case we likely do need a host OS of macOS 15 since I don't think `UInt128` will be present on macOS <15, right?

parkera: I see, that's why you mentioned the Predicate-specific availability here. I was hoping those wouldn't proliferate too much, but maybe we don't have a choice...

stephentyrone: Yeah, we need to match the availability of Int128, so I think it's fine to write these out explicitly.

parkera:
@swift-ci test

stephentyrone:
macOS failure is because the macOS tests run on a 13.x host, where Int128 doesn't exist.

stephentyrone:
@swift-ci test

stephentyrone:
@swift-ci test

PR #769, opened 2024-07-25
Bin compat: NSDecimalString should accept dictionary in addition to locale

iCharlesHu:
NSDecimalString historically accepted Any as the second parameter, which could be either a Locale or a Dictionary. We should keep that behavior.


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #773, opened 2024-07-25
Correctly set .fileTypeSymlink on windows

jmschonfeld:
When calculating file attributes for an item, we did not previously correctly detect symlinks on Windows. This updates the implementation to correctly detect symlinks and set the file type when populating the attributes dictionary.


jmschonfeld:
@swift-ci please test

compnerd (approved):
Ah, nice, you don't need to actually get the destination of the symlink, which makes this code easier.

PR #775, opened 2024-07-26
Add explicit void type parameter to C functions without parameters

kateinoigakukun:
C functions with `()` as parameter list can take any number of parameters. But WebAssembly requires static signature information for every function call, so we need to explicitly specify `(void)` to indicate that the function takes no parameters.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

jmschonfeld (approved):

PR #777, opened 2024-07-26
Enable wasi-libc emulation features

kateinoigakukun:
Those features require explicit macro definitions to be enabled, so add them to the package definition. Only affects WASI builds.


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

CMakeLists.txt
```diff
@@ -104,6 +104,14 @@ foreach(version ${_SwiftFoundation_versions})
     endforeach()
 endforeach()
 
+# wasi-libc emulation feature flags
+set(_SwiftFoundation_wasi_libc_flags)
```
jmschonfeld, line 108: nit: I think @etcwilde has mentioned to me before that you don't need to set a variable like this before appending to it, but perhaps I'm wrong on that

Package.swift
```diff
@@ -88,18 +93,27 @@ let package = Package(
...
+        ),
 
         // TestSupport (Internal)
-        .target(name: "TestSupport", dependencies: [
-            "FoundationEssentials",
-            "FoundationInternationalization",
-        ], swiftSettings: availabilityMacros + concurrencyChecking),
+        .target(
+          name: "TestSupport",
+            dependencies: [
```
jmschonfeld, line 110:
```suggestion
          name: "TestSupport",
          dependencies: [
```

nit: is the indentation off here?

CMakeLists.txt
kateinoigakukun: While we can skip the explicit declaration, I usually prefer doing that for conditionally appended list variables to avoid reference to uninitialized CMake variables. A reference to uninitialized CMake variable is a little bit tricky and CMake provides [a way to warn them](https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-warn-uninitialized), so I expect it's not a recommended way.

kateinoigakukun:
@swift-ci test

Sources/_FoundationCShims/include/_CStdlib.h
```diff
@@ -60,7 +60,17 @@
 #endif
 
 #if __has_include(<signal.h>)
-#include <signal.h>
+/* wasi-libc's signal.h is available only if _WASI_EMULATED_SIGNAL is defined */
+# if !defined(__wasi__) || defined(_WASI_EMULATED_SIGNAL)
```
jmschonfeld, line 64: One more small comment, could we use `TARGET_OS_WASI` here?

Also, if `_WASI_EMULATED_SIGNAL` isn't defined, what happens to `signal.h` on WASI? Does the file exist, it's just effectively empty and that's why the `__has_include` check passes?

kateinoigakukun: In wasi-libc, the `signal.h` header file itself is always available, but ff the macro isn't defined, it raises `#error` at compile-time: https://github.com/WebAssembly/wasi-libc/blob/b9ef79d7dbd47c6c5bafdae760823467c2f60b70/libc-top-half/musl/include/signal.h#L1-L3

The reason we need to conservative `_WASI_EMULATED_SIGNAL` guard here: We can assume all Foundation modules importing the CShim module defines `_WASI_EMULATED_SIGNAL`, so `defined(_WASI_EMULATED_SIGNAL)` shouldn't be false while building foundation. However, modules importing Foundation also transitively import the CShim and we can't assume `_WASI_EMULATED_SIGNAL` is defined while building those modules. So we need to be resilient to allow the both senario.

jmschonfeld: Ah makes sense, thanks for the explanation!

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):
This LGTM, but before we merge could we do a quick cross-repo test building the toolchain? That way we can test the CMake changes since CI on this repo doesn't (yet) build via CMake

kateinoigakukun:
@swift-ci test

kateinoigakukun:
Ok, confirmed CMake change does not break anything by cross repo CI

PR #779, opened 2024-07-26
Throw `.featureUnsupported` when attempting to create temp files on WASI

kateinoigakukun:
WASI does not have temp directory concept, and does not provide mktemp family of functions, so attempting to create a temporary file should be considered a feature unsupported.


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #781, opened 2024-07-26
Use platform shims for clock_gettime to support wasi-libc

kateinoigakukun:
This change adds platform shims for `clock_gettime` so that we can use it in Swift code even when the clock id macros can't be imported through ClangImporter like wasi-libc's definitions.

Also wasi-libc's `timespec.tv_sec` and `timespec.tv_nsec` are not imported as `Int` in Swift, so we need to cast them to `UInt64` before doing arithmetic operations on them.

### DIFF
--- a/Sources/_FoundationCShims/include/platform_shims.h
+++ b/Sources/_FoundationCShims/include/platform_shims.h
@@ -68,4 +68,17 @@ typedef enum {
 INTERNAL const char * _Nonnull _platform_shims_kOSThermalNotificationPressureLevelName();
 #endif
 
+#if TARGET_OS_WASI
+// Define clock id getter shims so that we can use them in Swift
+// even if clock id macros can't be imported through ClangImporter.
+
+#include <time.h>
+static inline _Nonnull clockid_t _platform_shims_clock_monotonic(void) {
+    return CLOCK_MONOTONIC;
+}
+static inline _Nonnull clockid_t _platform_shims_clock_realtime(void) {
+    return CLOCK_REALTIME;
+}
+#endif
+
 #endif /* CSHIMS_PLATFORM_SHIMS */
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -30,6 +30,7 @@ add_library(FoundationEssentials
     SortComparator.swift
     UUID_Wrappers.swift
     UUID.swift
+    WASILibc+Extensions.swift
     WinSDK+Extensions.swift)
 
 add_subdirectory(AttributedString)
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -129,7 +129,7 @@ final class _ProcessInfo: Sendable {
 #else
         var ts: timespec = timespec()
         clock_gettime(CLOCK_MONOTONIC_RAW, &ts)
-        let time: UInt64 = UInt64(ts.tv_sec * 1000000000 + ts.tv_nsec)
+        let time: UInt64 = UInt64(ts.tv_sec) * 1000000000 + UInt64(ts.tv_nsec)
 #endif
         let timeString = String(time, radix: 16, uppercase: true)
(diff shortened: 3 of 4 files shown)


compnerd (commented):
Does this really require the C shim? Why can we not just add an OS conditional?

kateinoigakukun:
`CLOCK_MONOTONIC` and `CLOCK_REALTIME` are defined in wasi-libc as follows, and we can't use them directly from Swift. Thus we need C shim at least for WASI, and I think it makes sense to use the shims for all unixy platforms consistently.

```c
extern const struct __clockid _CLOCK_MONOTONIC;
#define CLOCK_MONOTONIC (&_CLOCK_MONOTONIC)
extern const struct __clockid _CLOCK_REALTIME;
#define CLOCK_REALTIME (&_CLOCK_REALTIME)
```

https://github.com/WebAssembly/wasi-libc/blob/b9ef79d7dbd47c6c5bafdae760823467c2f60b70/libc-bottom-half/headers/public/__header_time.h#L17-L20

jmschonfeld:
@swift-ci please test

Sources/_FoundationCShims/platform_shims.c
```diff
@@ -67,3 +67,18 @@ mach_port_t _platform_mach_task_self() {
 }
 #endif
 
+int _platform_shims_clock_monotonic_raw_gettime(struct timespec * _Nonnull ts) {
```
jmschonfeld, line 70: Looks like the Windows failures are related:

```
C:\source\swift-foundation\Sources\_FoundationCShims\platform_shims.c:74:12: error: call to undeclared function 'clock_gettime'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   74 |     return clock_gettime(CLOCK_MONOTONIC_RAW, ts);
      |            ^
```

Do the implementation (and header) need to be guarded by a `__has_include` check?

kateinoigakukun: Since windows provides `time.h` header itself, we need to exclude them by `#if !defined(_WIN32)`

kateinoigakukun:
@swift-ci test

Sources/_FoundationCShims/include/platform_shims.h
```diff
@@ -68,4 +68,14 @@ typedef enum {
 INTERNAL const char * _Nonnull _platform_shims_kOSThermalNotificationPressureLevelName();
 #endif
 
+// Define clock_gettime shims for each clock type so that we can use them in Swift
+// even if clock id macros can't be imported through ClangImporter.
+
+#if !defined(_WIN32)
```
jmschonfeld, line 74: Could we use `TARGET_OS_WINDOWS` here to try to be consistent with our platform checks (I know some places might not, I plan to come and fix up a few but it'd be great to standardize on the target conditionals with new additions)

Sources/_FoundationCShims/platform_shims.c
```diff
@@ -67,3 +67,20 @@ mach_port_t _platform_mach_task_self() {
 }
 #endif
 
+#if !defined(_WIN32)
+int _platform_shims_clock_monotonic_raw_gettime(struct timespec * _Nonnull ts) {
+#if __wasi__ // WASI does not support CLOCK_MONOTONIC_RAW. Fall back to CLOCK_MONOTONIC.
```
jmschonfeld, line 72: Same comment here as above, could we use `TARGET_OS_WASI`?

kateinoigakukun:
@swift-ci test

jmschonfeld:
Just confirming - is this change still necessary for the WASI build? One thought that occurred to me - if the problem is `CLOCK_MONOTONIC`, I wonder if we could just shim that and limit the shim to WASI instead of shimming the whole function, does that seem plausible / do you think that'd be any better? Not sure if it's inherently better, just a thought

kateinoigakukun:
@jmschonfeld That totally makes sense to me. Now shims are isolated to only WASI.

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #782, opened 2024-07-26
Fix `operatingSystemVersion` on WASI

kateinoigakukun:
The `operatingSystemVersion` property type is a tuple but the it was returning an `OperatingSystemVersion` instance on unknown platforms.


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #783, opened 2024-07-26
Guard out user/group related code on WASI

kateinoigakukun:
This change guards out the user/group related code on WASI, as WASI does not have the concept of users or groups.


Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -907,6 +907,7 @@ extension _FileManagerImpl {
             let group = attributes[.groupOwnerAccountName] as? String
             let groupID = _readFileAttributePrimitive(attributes[.groupOwnerAccountID], as: UInt.self)
             
+            #if !os(WASI) // WASI does not have the concept of users or groups
```
jmschonfeld, line 910: Should we throw a `.featureUnsupported` error if the caller tries to provide these attribute values?

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -273,15 +273,23 @@ extension _FileManagerImpl {
 
 #if !os(Windows)
     static func _userAccountNameToNumber(_ name: String) -> uid_t? {
+        #if os(WASI) // WASI doesn't have getpwnam
```
jmschonfeld, line 276: We don't compile these at all for Windows since they don't make sense on that platform, should we just do the same here (i.e. `#if !os(Windows) && !os(WASI)` above) so that we don't accidentally try to call these on WASI?

Sources/FoundationEssentials/Platform.swift
```diff
@@ -174,7 +174,7 @@ extension Platform {
         // FIXME: bionic implements this as `return 0;` and does not expose the
         // function via headers. We should be able to shim this and use the call
         // if it is available.
-#if !os(Android)
+#if !os(Android) && !os(WASI)
         guard issetugid() == 0 else { return nil }
```
jmschonfeld, line 178: Just confirming: since WASI doesn't have the concept of users there's no equivalent that we should be calling instead, right?

kateinoigakukun: Right, WASI doesn't expose "taintedness" concept nor secure env var

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
kateinoigakukun: Ah, yes, we should throw in that case.

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
kateinoigakukun: Oh, we don't use these methods other than in `setAttributes`, so we can exclude them at all, thanks!

kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #784, opened 2024-07-26
Guard out extended or fs attributes related code on WASI

kateinoigakukun:
This commit guards out the extended attributes and file system attributes related code on WASI as WASI does not support these features. Just return nothing or ignore the set request.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -482,7 +482,7 @@ extension _FileManagerImpl {
 #endif
     }
 
-#if !os(Windows)
+#if !os(Windows) && !os(WASI)
     private func _extendedAttribute(_ key: UnsafePointer<CChar>, at path: UnsafePointer<CChar>, followSymlinks: Bool) throws -> Data? {
         #if canImport(Darwin)
         var size = getxattr(path, key, nil, 0, 0, followSymlinks ? 0 : XATTR_NOFOLLOW)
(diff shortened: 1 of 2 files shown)


kateinoigakukun:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -484,7 +484,13 @@ extension _FileManagerImpl {
 #endif
     }
 
-#if !os(Windows)
+#if os(Windows)
+#elseif os(WASI)
```
jmschonfeld, line 488: Should we do the same thing that we do on Windows for these functions and prevent calling it at all rather than returning `nil`?

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -173,7 +173,12 @@ extension _FileManagerImpl {
         #endif
     }
 
-#if !os(Windows)
+#if os(Windows)
+#elseif os(WASI)
```
jmschonfeld, line 177: Same question as above - would it be better to ensure we don't ever try to call this (like on Windows) rather than making it a no-op? That way we could throw a `.featureUnsupported` error if a developer tries to set it

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
kateinoigakukun: Ok, we can do that.

kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #785, opened 2024-07-26
Skip sticky-bit check in `isDeletableFile` on WASI

kateinoigakukun:
WASI does not surface the sticky bit and getuid, so we cannot check whether the file is actually deletable before attempting to delete it.


kateinoigakukun:
@swift-ci test

PR #786, opened 2024-07-26
Add explicit include of `wasi/libc-environ.h`

kateinoigakukun:
This is necessary to get the `__wasilibc_get_environ` function declaration.


compnerd:
@swift-ci please test

MaxDesiatov (approved):

PR #787, opened 2024-07-26
Implement `_copyRegularFile` for WASI without `sendfile`

kateinoigakukun:
WASI doesn't have `sendfile`, so we need to implement the copy in user space with `read` and `write`. It's not as efficient as `sendfile`, but it's the best we can do.


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #792, opened 2024-07-26
URL: Don't standardize relative file paths

jrflat:
Resolves https://github.com/apple/swift-foundation/issues/790 (thanks @nicklockwood for filing!)

We should not standardize the relative path outside of the context of its base URL, such as when calculating the absolute `.path()`. This allows us to properly handle `".."` segments at the beginning of the relative path, such as in

```
let baseURL = URL(filePath: "/docs/src/")
let result = URL(filePath: "../images/foo.png", relativeTo: baseURL)
XCTAssertEqual(result.path(), "/docs/images/foo.png")
```

Note that if the inputted path is absolute, e.g. `"/../images/foo.png"`, we should not resolve against the base URL. I've added a small unit test to cover both cases.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -2155,14 +2155,6 @@ extension URL {
             isDirectory = filePath.utf8.last == slash
         }
 
-        if !isAbsolute {
-            #if !NO_FILESYSTEM
-            filePath = filePath.standardizingPath
-            #else
-            filePath = filePath.removingDotSegments
-            #endif
-        }
-
         #if os(Windows)
         // Convert any "\" back to "/" before storing the URL parse info
         filePath = filePath.replacing(UInt8(ascii: "\\"), with: UInt8(ascii: "/"))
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

DmT021:
@jrflat Could it be the source of this error:
```
/home/build-user/swift-integration-tests/test-static-stdlib/test-static-stdlib.test:7:8: error: CHECK: expected string not found in input
CHECK: foo bar baz: file:/foo
       ^
<stdin>:1:1: note: scanning from here
foo bar baz: file:///foo
^

Input file: <stdin>
Check file: /home/build-user/swift-integration-tests/test-static-stdlib/test-static-stdlib.test

-dump-input=help explains the following input dump.

Input was:
<<<<<<
         1: foo bar baz: file:///foo 
check:7     X~~~~~~~~~~~~~~~~~~~~~~~~ error: no match found
>>>>>>

error: command failed with exit status: 1
```
I see it on CI in this build: https://ci.swift.org/job/swift-PR-Linux/16654/console

PR #793, opened 2024-07-26
[CMake] Fix macro install rpath

jmschonfeld:
This updates the macro library's rpath to reference the correct locations for the host and stldib libraries that it depends upon. This matches the rpath values for the existing macro libraries:

```
# readelf -d /usr/lib/swift/host/plugins/libSwiftMacros.so | grep RUNPATH
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/../../../swift/linux:$ORIGIN/..]
```

I confirmed that using a locally built `libFoundationMacros.so` with this change successfully finds the required libraries rather than erroring out with `compiler plugin '/usr/lib/swift/host/plugins/libFoundationMacros.so' could not be loaded;  libswift_StringProcessing.so: cannot open shared object file: No such file or directory` like before

### DIFF
--- a/Sources/FoundationMacros/CMakeLists.txt
+++ b/Sources/FoundationMacros/CMakeLists.txt
@@ -43,8 +43,9 @@ target_link_libraries(FoundationMacros PUBLIC
     SwiftSyntax::SwiftSyntaxBuilder
 )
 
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
Full swift toolchain build passed with this change (and passed the rpath packaging tests too)

iCharlesHu (approved):

PR #794, opened 2024-07-26
Benchmark Tweaks

glessard:
This fixes a number of issues for the swift-foundation benchmarks.

- Use URL-based file APIs for portability
- Tolerate `autoreleasepool` blocks when the Obj-C runtime isn't present
- Work around declaration ambiguity of `String.Encoding` when another import of `Foundation` inevitably exists.


glessard:
@swift-ci please test

Package.swift
```diff
@@ -78,6 +78,7 @@ let package = Package(
         .library(name: "FoundationPreview", targets: ["FoundationPreview"]),
         .library(name: "FoundationEssentials", targets: ["FoundationEssentials"]),
         .library(name: "FoundationInternationalization", targets: ["FoundationInternationalization"]),
+        .library(name: "FoundationBenchmarkSupport", targets: ["FoundationBenchmarkSupport"]),
```
iCharlesHu, line 81: We shouldn't export this since it's for testing only.

Sources/FoundationBenchmarkSupport/FoundationBenchmarkSupport.swift
```diff
@@ -0,0 +1,15 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import FoundationEssentials
+
+public typealias FoundationEssentialsStringEncoding = String.Encoding
```
iCharlesHu, line 15: I think this target should go under `Benchmarks` so you won't have to export it as a library.

glessard: That does work too! (done)

Package.swift
glessard: Moved this to the benchmark package as you suggested below

glessard:
@swift-ci please test

glessard:
Closing in favor of https://github.com/apple/swift-foundation/pull/810

PR #798, opened 2024-07-27
[6.0] URL: Don't standardize relative file paths

### LINKED ISSUE #790: `URL(fileURLWithPath:relativeTo:)` regression on Linux
The following code works as expected on macOS and earlier Linux versions but is broken on the latest nightly:

```swift
let path = "../images/foo.png"
let baseURL = URL(fileURLWithPath: "/docs/src/")
let result = URL(fileURLWithPath: path, relativeTo: baseURL)
print(result.path)
```

Expected:

```
/docs/images/foo.png
```

Actual:

```
/docs/src/images/foo.png
```

(i.e. the leading ../ in the path is ignored)

parkera:
release/6 cherry pick of https://github.com/apple/swift-foundation/pull/792

Resolves #790

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -2155,14 +2155,6 @@ extension URL {
             isDirectory = filePath.utf8.last == slash
         }
 
-        if !isAbsolute {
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci test

PR #801, opened 2024-07-29
Remove unneccessary _FileManagerImpl.delegate

jmschonfeld:
A file manager's delegate is stored as a property of the `FileManager` itself, rather than as a property of the underlying `_FileManagerImpl` type. However, when creating this type I accidentally left a stray `delegate` property here that was unused. To Avoid confusion, remove the extra, unused property and the function that referenced it so that it's clear that only `FileManager.delegate` is used.


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #802, opened 2024-07-29
Upgrade SwiftFoundationICU to version 74

iCharlesHu:
This patch upgrades `SwiftFoundationICU`'s version to 74.

WIP: I'll update the final URL once the new version of ICU is tagged.


Package.swift
```diff
@@ -62,7 +62,7 @@ var dependencies: [Package.Dependency] {
                 from: "1.1.0"),
             .package(
                 url: "https://github.com/apple/swift-foundation-icu",
-                exact: "0.0.9"),
+                branch: "release/6.0"),
```
jmschonfeld, line 65: Since this is being merged to the `main` branch, this should point at the `main` branch of `swift-foundation-icu` right?

iCharlesHu: Ahhh good point! Thanks!

iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #803, opened 2024-07-29
[6.0] [CMake] Fix macro install rpath

jmschonfeld:
Explanation: Fixes the rpath value for the Foundation macro library in the toolchain
Scope: Should only modify the rpath value for the macro library which is currently only built on linux
Original PR: https://github.com/apple/swift-foundation/pull/793
Risk: Minimal - minor change to fix an already-broken rpath value for a macro library which no other toolchain components are using yet
Testing: Testing done via local build and validation of the rpath as well as swift-ci toolchain builds and testing
Reviewer: @iCharlesHu


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #804, opened 2024-07-30
Fix to correct order of the alphabet

coffmark:
### Summary

I fixed to correct order of the alphabet in `BenchmarkString.swift` because it's missing the letter 's'.

### DIFF
--- a/Benchmarks/Benchmarks/String/BenchmarkString.swift
+++ b/Benchmarks/Benchmarks/String/BenchmarkString.swift
@@ -26,7 +26,7 @@ let benchmarks = {
     Benchmark.defaultConfiguration.metrics = [.cpuTotal, .wallClock, .throughput]
     
     // MARK: Encoding strings
-    let asciiSmallStr = "abcdefghijklmnopqrtuvwxyz"
+    let asciiSmallStr = "abcdefghijklmnopqrstuvwxyz"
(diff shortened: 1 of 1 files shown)


Benchmarks/Benchmarks/String/BenchmarkString.swift
```diff
@@ -26,7 +26,7 @@ let benchmarks = {
     Benchmark.defaultConfiguration.metrics = [.cpuTotal, .wallClock, .throughput]
     
     // MARK: Encoding strings
-    let asciiSmallStr = "abcdefghijklmnopqrtuvwxyz"
+    let asciiSmallStr = "abcdefghijklmnopqrstuvwxyz"
```
jmschonfeld, line 29: Fixing this seems fine with me, but just want to double check with @parkera / @Catfish-Man - by "Small String" here do we just mean "a small string" or "Swift's `SmallString`"? And if it's the latter, we're not pushing this over the small string boundary by going from 25 to 26 characters, right?

itingliu: @coffmark The lack of "`s`" here doesn't mean the benchmark test is incorrect. Can you clarify what this change aims to fix?

coffmark: @itingliu 
Thank you.I closed this pr.

PR #808, opened 2024-07-31
Prevent null character in Windows home directory

### LINKED ISSUE #807: On Windows, null byte added to user from `FileManager.default.homeDirectoryForCurrentUser`, resulting in infinite looping in `DirectoryEnumerator`
```swift
import Foundation

// Prints file:///C:/Users/alex%00/
// Notice the null byte after the user
print(FileManager.default.homeDirectoryForCurrentUser) 

// Same here
print(URL(fileURLWithPath: ("~" as NSString).expandingTildeInPath))

// Prints 
// file:///C:/Users/alex%00/alex/
// file:///C:/Users/alex%00/alex/alex/
// file:///C:/Users/alex%00/alex/alex/alex/
// file:///C:/Users/alex%00/alex/alex/alex/alex/
// file:///C:/Users/alex%00/alex/alex/alex/alex/alex/
// .. continues forever
let enumerator = FileManager.default.enumerator(at: FileManager.default.homeDirectoryForCurrentUser, includingPropertiesForKeys: nil)
while let url = enumerator?.nextObject() as? URL {
  print("\(url)\r\n")
}
```

jrflat:
Resolves #807

`String(decoding:as:)` was including the null-terminator from the Windows API, and when this is passed to `URL`, it's encoded in the form `file:///C:/Users/alex%00/`. When enumerating the directory, we get the file system representation of the `URL`, which decodes `%00` to back to `\0`. This was causing us to append a file name `alex` to the directory, getting `file:///C:/Users/alex%00/alex`, which "exists" because it was only checking for the existence of `file:///C:/Users/alex`. This cycle repeats, causing an infinite loop in the enumerator.

This PR 1) replaces `String(decoding:as:)` with `String(decodingCString:as:)` in the home directory function and 2) prevents `URL.fileSystemPath` from decoding `%00` to `\0`.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -374,7 +374,7 @@ extension String {
                     guard GetEnvironmentVariableW(pwszVariable, $0.baseAddress, dwLength) == dwLength - 1 else {
                         return nil
                     }
-                    return String(decoding: $0, as: UTF16.self)
+                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
                 }
             }
         }
@@ -436,7 +436,7 @@ extension String {
             guard GetUserProfileDirectoryW(hToken, $0.baseAddress, &dwcchSize) else {
                 fatalError("unable to query user profile directory")
             }
-            return String(decoding: $0, as: UTF16.self)
+            return String(decodingCString: $0.baseAddress!, as: UTF16.self)
         }
 #else
         #if targetEnvironment(simulator)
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

compnerd (approved):
Ah, the `W` terminator is `\0\0`. Nice catch!

jmschonfeld (approved):
Thanks!

compnerd:
@jrflat anything left to do here, or good to merge?

jrflat:
Will merge soon after some testing on Darwin

PR #811, opened 2024-08-02
Use `POSIXErrorCode` instead of direct errno E constants from wasi-libc

kateinoigakukun:
wasi-libc's E constants cannot be imported directly by ClangImporter, so we should use `POSIXErrorCode` from Swift instead.

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -94,7 +94,15 @@ extension CocoaError {
         var userInfo = userInfo
         
         // (130280235) POSIXError.Code does not have a case for EOPNOTSUPP
-        if errno != EOPNOTSUPP {
+        let _EOPNOTSUPP: Int32
+        #if os(WASI)
+        // wasi-libc's errno.h constants cannot be directly imported into Swift
+        // so we use Swift stdlib's POSIXErrorCode instead
+        _EOPNOTSUPP = POSIXErrorCode.EOPNOTSUPP.rawValue
+        #else
+        _EOPNOTSUPP = EOPNOTSUPP
+        #endif
+        if errno != _EOPNOTSUPP {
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

kateinoigakukun:
@swift-ci test

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -94,7 +94,15 @@ extension CocoaError {
         var userInfo = userInfo
         
         // (130280235) POSIXError.Code does not have a case for EOPNOTSUPP
-        if errno != EOPNOTSUPP {
+        let _EOPNOTSUPP: Int32
+        #if os(WASI)
+        // wasi-libc's errno.h constants cannot be directly imported into Swift
+        // so we use Swift stdlib's POSIXErrorCode instead
+        _EOPNOTSUPP = POSIXErrorCode.EOPNOTSUPP.rawValue
```
jmschonfeld, line 101: This workaround here was put in place because the `POSIXErrorCode.EOPNOTSUPP` does not exist on platforms like Darwin (likely an oversight / a missing `enum` case). Does that case actually exist on WASI? If so, we might as well avoid this check entirely and just always evaluate the true branch of the `if` statement rather than worrying about storing this variable on a per-platform basis.

Also, is this something that can/should be fixed in the WASI overlay? There are a few other places where we use these error code constants that I expect would hit the same issue

kateinoigakukun: > Also, is this something that can/should be fixed in the WASI overlay? There are a few other places where we use these error code constants that I expect would hit the same issue

Good pointing! This reminds me that we already provide errno constant shims in the overlay, and I just realized it missed some of them including `EOPNOTSUPP`. 

So this change should be unnecessary after https://github.com/swiftlang/swift/pull/75671

PR #813, opened 2024-08-02
[6.0] Prevent null character in Windows home directory

jrflat:
**Explanation:** Prevents a null-terminator from the Windows home directory API from ending up in the resulting `String` or `URL`. Prevents `%00` from being decoded to `\0` when getting a `URL`'s `fileSystemPath`, which is often converted to a C string for file system calls.
**Scope:** Windows-only change from `String(decoding:)` to `String(decodingCString:)` for `homeDirectoryPath(forUser:)`. Not decoding `%00` occurs for file paths of `URL`s on all platforms.
**Original PR:** https://github.com/apple/swift-foundation/pull/808
**Risk:** Low. Minor changes to prevent unexpected behavior when null-characters are included in file paths. The respective Windows APIs null-terminate their output, so `String(decodingCString:)` is safe.
**Testing:** swift-ci testing and local testing on Windows
**Reviewers:** @compnerd @jmschonfeld

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -374,7 +374,7 @@ extension String {
                     guard GetEnvironmentVariableW(pwszVariable, $0.baseAddress, dwLength) == dwLength - 1 else {
                         return nil
                     }
-                    return String(decoding: $0, as: UTF16.self)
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

compnerd (approved):

itingliu (approved):

PR #815, opened 2024-08-02
Add compatibility path for invalid UTF8 data

parkera:
Add a compatibility path for apps which parse invalid UTF8 data and expect it to be automatically repaired instead of returning `nil`.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #817, opened 2024-08-02
Remove _typeByName lookup of _FoundationNSNumberInitializer

jmschonfeld:
With https://github.com/apple/swift-corelibs-foundation/pull/5045, Foundation now dynamically replaces the implementation of this function when it is present so there is no need to perform a `_typeByName` lookup. We can safely remove this now and just return `nil` because we know that if this implementation is used, (the correct) Foundation is not loaded.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #820, opened 2024-08-02
Add FoundationNetworking hook for reading contents of remote URL

jmschonfeld:
This adds a hook for `FoundationNetworking` to override in order to support an up-call for cases where `Data` needs to read from a non-file URL


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #822, opened 2024-08-05
Concurrency-Safe Notifications

cthielen:
This PR introduces a proposal for adopting Swift Concurrency in `NotificationCenter`.

### DIFF
--- a/Proposals/NNNN-concurrency-safe-notifications.md
+++ b/Proposals/NNNN-concurrency-safe-notifications.md
@@ -222,7 +222,7 @@ Posting only requires passing an instance of the `NotificationCenter.Message`-co
 NotificationCenter.default.post(WillLaunchApplication(...))
 ```
 
-Like the existing `post()` methods, all observers will be called synchronously and in serial to ensure existing notification usage patterns like `will` / `did` are executed in order. This is possible because the isolation used to call `post()` will match the isolation defined by `NotificationCenter.Message`, and all observers will then be called from that same isolation.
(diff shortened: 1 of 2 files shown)


iCharlesHu (approved):

Proposals/NNNN-concurrency-safe-notifications.md
```diff
@@ -0,0 +1,386 @@
...
+extension NotificationCenter {
+    public protocol Message {
+        associatedtype Subject: AnyObject
+        static var name: Notification.Name { get }
+        
+        static func makeMessage(_ notification: Notification) -> Self?
+        static func makeNotification(_ message: Self) -> Notification
+    }
+    
+    public protocol MainActorMessage: Message {}
```
FranzBusch, line 186: I am wondering if this protocol should have a `@MainActor` annotation. Those types will only be sent and received from the MainActor anyways.

Proposals/NNNN-concurrency-safe-notifications.md
```diff
@@ -0,0 +1,386 @@
...
+        -> ObservationToken
+}
+```
+
+And for `AsyncMessage`:
+
+```swift
+@available(FoundationPreview 0.5, *)
+extension NotificationCenter {
+    public func addObserver<I: MessageIdentifier, M: AsyncMessage>(of subject: M.Subject,
```
FranzBusch, line 231: How can those methods take an `async` closure while not being async themselves? Where is the task spawned that calls the observer closures?

If the `addObserver` method takes an async closure I would have expected that method to become `async` itself. Something more like this

```
await center.observe(of: subject, for: identifier) { message
  // do something async with the message
  await process(message)
}
```

This way you also don't need `ObservationToken`s but can just rely on task cancellation which IMO fits better into the Concurrency paradigms.

itingliu:
@swift-ci please test

PR #824, opened 2024-08-05
[6.0] Remove _typeByName lookup of _FoundationNSNumberInitializer

jmschonfeld:
Explanation: Removes a _typeByName lookup in a FoundationEssentials up-call that can result in confusion between the toolchain and local builds
Scope: Should only impact up-calls from FileManager in FoundationEssentials to initialize NSNumbers
Original PR: https://github.com/apple/swift-foundation/pull/817
Risk: Minimal - the body of this function is already overridden by SCL-F, this just removes the `_typeByName` lookup for cases where Foundation isn't loaded (and therefore the lookup will only ever fail, or find the wrong Foundation)
Testing: Testing done via swift-ci testing
Reviewer: @itingliu


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #826, opened 2024-08-05
[6.0] Add FoundationNetworking hook for reading contents of remote URL

jmschonfeld:
Explanation: Adds a hook point for FoundationNetworking to allow `Data(contentsOf:)` to perform a remote URL request to fetch data
Scope: Should only impact calling `Data(contentsOf:)` with a non-file URL, turning an error into non-regressed behavior (but is inert without the matching SCL-F change)
Original PR: https://github.com/apple/swift-foundation/pull/820
Risk: Minimal - fixes a regression where the current behavior is just to `throw`, and fairly limited scope
Testing: Testing done via swift-ci testing and local testing
Reviewer: @iCharlesHu


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #827, opened 2024-08-06
rdar://132940984 (Regression: Swift Decoding of Double.greatestFiniteMagnitude as Int causes a crash)

kperryua:
Fixup from previous change here. Like the previous version of this code, we should immediately throw an error if we fail to decode any Double from the buffer. We'll evaluate if it's expected to be the "wrong" integer only if it succeeds.

The previous fix caused a preconditionFailure crash since we'd fall through to the `validateNumber` function for a valid syntax JSON number.


kperryua:
@swift-ci please test

PR #829, opened 2024-08-06
[6.0] rdar://132940984 (Regression: Swift Decoding of Double.greatestFiniteMagnitude as Int causes a crash) (#827)

kperryua:
Explanation: Restores previous behavior of throwing an error when a JSON floating point number (during slow-path integer parsing) fails to decode as `Double`.
Scope: Limited to JSONDecoder usages that attempt to parse very large floating point numbers as integers.
Original PR: https://github.com/apple/swift-foundation/pull/827
Risk: Minimal - fixes a regression that introduced an unexpected preconditionFailure
Reviewer: @itingliu


kperryua:
@swift-ci please test

itingliu (approved):

PR #830, opened 2024-08-06
do not import stdatomic as it doesn't work with c++ interop on windows

hyp:
…c++ interop enabled

Works around https://github.com/swiftlang/swift/issues/75720

### DIFF
--- a/Sources/_FoundationCShims/include/_CStdlib.h
+++ b/Sources/_FoundationCShims/include/_CStdlib.h
@@ -122,8 +122,14 @@
 #include <complex.h>
 #endif
 
+#if !defined(_WIN32)
+// On windows, atomic_signal_fence from clang headers is incompatible
+// with atomic_signal_fence definition from MSVC when C++ interop is enabled,
+// so drop the stdatomic include.
+// Works around: https://github.com/swiftlang/swift/issues/75720
 #if __has_include(<stdatomic.h>)
 #include <stdatomic.h>
+#endif
 #endif
 
 #if __has_include(<threads.h>)
--- a/Sources/_FoundationCShims/include/_CStdlib.h
+++ b/Sources/_FoundationCShims/include/_CStdlib.h
@@ -122,16 +122,6 @@
 #include <complex.h>
 #endif
 
-#if !defined(_WIN32)
-// On windows, atomic_signal_fence from clang headers is incompatible
-// with atomic_signal_fence definition from MSVC when C++ interop is enabled,
-// so drop the stdatomic include.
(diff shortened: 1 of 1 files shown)


hyp:
Tested together witH:
https://github.com/swiftlang/swift/pull/75721

hyp:
CC @compnerd

compnerd (commented):
Unfortunate, but lets get this fixed up. We should pull this into 6.0 as well.

compnerd (approved):

Sources/_FoundationCShims/include/_CStdlib.h
```diff
@@ -122,9 +122,15 @@
 #include <complex.h>
 #endif
 
+#if !defined(_WIN32)
```
jmschonfeld, line 125: If we need to special case windows here, can we use `TARGET_OS_WINDOWS` instead for consistency with the other headers?

Sources/_FoundationCShims/include/_CStdlib.h
```diff
@@ -122,9 +122,15 @@
...
+// On windows, atomic_signal_fence from clang headers is incompatible
+// with atomic_signal_fence definition from MSVC when C++ interop is enabled,
+// so drop the stdatomic include.
+// Works around: https://github.com/swiftlang/swift/issues/75720
 #if __has_include(<stdatomic.h>)
 #include <stdatomic.h>
```
jmschonfeld, line 131: Presumably this is in here because something in swift-foundation requires it - do we know what that is / what will be broken by not having this include? Just want to confirm which scenario we have:

1. This import isn't used at all - in which case we should just remove it instead of conditionalizing it
2. This import isn't used on Windows but is used on other platforms - in which case this is fine, but it'd be good to know what needs it
3. This import is used on all platforms - in which case CI should hopefully show us where Windows needs it

hyp: Good questions! Let me investigate.

hyp: It looks like we can drop `stdatomic` include alltogether, and nothing in swift-foundation uses, and swift-corelibs-foundation includes it too in CFInternal.h . Let me just do some more testing and I'll update this PR removing it alltogether.

hyp:
@swift-ci please test

Sources/_FoundationCShims/include/_CStdlib.h
hyp: Removed the include!

jmschonfeld (approved):

PR #831, opened 2024-08-06
Fix NSURLComponents bridging with percent-encoding

jrflat:
Resolves https://github.com/apple/swift-foundation/issues/828 (thanks @giginet for filing!)

Because the bridging code assigns non-percentEncoded variants of the components (e.g. `.path` instead of `.percentEncodedPath`), an `NSURLComponents` that is bridged to Swift `URLComponents` loses its percent-encoding. For example:

```
let nsURLComponents = NSURLComponents(
    string: "https://example.com?url=https%3A%2F%2Fapple.com"
)!
let urlComponents = nsURLComponents as URLComponents
print(nsURLComponents.string) // "https://example.com?url=https%3A%2F%2Fapple.com"
print(urlComponents.string) // "https://example.com?url=https://apple.com"
```

Note we only lose percent-encoding if the percent-encoding is not absolutely required, e.g. `"://"` does not require encoding in the query component, but this is something we should fix regardless.

This PR first checks if we can cast the `NSURLComponents` to a `_NSSwiftURLComponents` and take the components right out of it. If the cast fails, we fall back to assigning the `percentEncoded` properties.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLComponents.swift
+++ b/Sources/FoundationEssentials/URL/URLComponents.swift
@@ -1141,15 +1141,19 @@ extension URLComponents: ReferenceConvertible, _ObjectiveCBridgeable {
     }
 
     public static func _conditionallyBridgeFromObjectiveC(_ x: NSURLComponents, result: inout URLComponents?) -> Bool {
+        if let swiftComp = x as? _NSSwiftURLComponents {
+            result = swiftComp.components
+            return true
+        }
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

giginet (approved):
Thanks!

I couldn't address this because it requires `FOUNDATION_FRAMEWORK`

PR #834, opened 2024-08-08
Fix unexpected Decimal sign behavior from `Decimal(sign:exponent:significand:)`

### LINKED ISSUE #833: `Decimal(sign:exponent:significand:)` causes unexpected sign if `significand` is negative
With the release of iOS 18 Developer Beta 5 and Public Beta 3, my team received reports from customers that all negative values in our app were showing as positives, which led to major confusion.

We narrowed down the issue to a change to the initializer for Decimal: `Decimal(sign:exponent:significand)`. Prior to Beta 5, the sign passed into the initializer would be the sign of the Decimal. Passing `.plus` would result in a positive decimal, and passing `.minus` would result in a negative Decimal. This is regardless of the sign of the significant. In Beta 5, it seems that the sign passed into the init, and the sign of the significand are now negated. This means that passing `.minus` for sign and a negative Decimal for significand results in an unexpectedly positive Decimal value in Beta 5 alone. This behavior does not seem to be explicitly documented.

I created a quick playground to illustrate the issue. Here's the code:

```swift
// Expected Value: 20
// Actual Value: 20
let positiveDecimal = Decimal(sign: .plus, exponent: 1, significand: 2)

// Expected Value: 20
// Actual Value (15.4.0, 16.0 Beta 4): 20
// Actual Value (16.0 Beta 5): -20
let positiveDecimalWithNegativeSignificand = Decimal(sign: .plus, exponent: 1, significand: -2)

// Expected Value: -20
// Actual Value: -20
let negativeDecimal = Decimal(sign: .minus, exponent: 1, significand: 2)

// Expected Value: -20
// Actual Value (15.4.0, 16.0 Beta 4): -20
// Actual Value (16.0 Beta 5): 20
let negativeDecimalWithNegativeSignificand = Decimal(sign: .minus, exponent: 1, significand: -2)
```

I've tracked down the issue to [a PR in the swift-foundation repo](https://github.com/apple/swift-foundation/pull/741/commits/6ccff7b99331095da56fb3188bdd2ecb94096598) from 3 weeks ago, which seems to pull a commit from [the swift-corelibs-foundation repo](https://github.com/apple/swift-corelibs-foundation/pull/3068).

EthanHumphrey:
Resolves #833. Fixes an issue that caused the initialized Decimal's sign not to match the passed `sign` when `significand` is negative. More information in the linked issue.

### DIFF
--- a/Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
+++ b/Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
@@ -251,8 +251,11 @@ extension Decimal /* : FloatingPoint */ {
                 self = .nan
             }
         }
-        if sign == .minus {
-            negate()
+
+        guard self._length != 0 else { return }
+        _isNegative = switch sign {
+        case .plus: 0
+        case .minus: 1
         }
     }
 
--- a/Tests/FoundationEssentialsTests/DecimalTests.swift
+++ b/Tests/FoundationEssentialsTests/DecimalTests.swift
@@ -1130,9 +1130,9 @@ final class DecimalTests : XCTestCase {
         var x = -42 as Decimal
         XCTAssertEqual(x.significand.sign, .plus)
         var y = Decimal(sign: .plus, exponent: 0, significand: x)
-        XCTAssertEqual(y, -42)
-        y = Decimal(sign: .minus, exponent: 0, significand: x)
         XCTAssertEqual(y, 42)
+        y = Decimal(sign: .minus, exponent: 0, significand: x)
+        XCTAssertEqual(y, -42)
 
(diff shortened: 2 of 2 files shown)


EthanTheInnovator:
@swift-ci please test

vitorll (approved):

xwu:
This would break clients who expect the _documented_ behavior rather than the buggy one. Importantly, there are synchronized changes that would also cause some clients expecting the _buggy_ behavior to break.

pappalar:
Agreed the fix should not revert to the buggy behavior.
The actual fix is to make the RC follow ABI stability to not break existing clients with old toolchains

PR #836, opened 2024-08-08
Port directory enumeration related code to WASI

kateinoigakukun:
For now wasi-libc does not include fts(3) implementation, so mark features depending on it as unsupported on WASI. Once wasi-libc includes fts or we decide to implement and maintain our own fts-like API, we can remove these `#if os(WASI)` guards. 

I'm working on porting FTS to wasi-libc (https://github.com/WebAssembly/wasi-libc/pull/522), and we can share most of the enumerator code with other posix platforms after that.

Also, wasi-libc defines some constants in a way that ClangImporter can't understand, so we need to grab them manually through _FoundationCShims in function call form.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -512,7 +512,14 @@ enum _FileOperations {
             // We failed for a reason other than the directory not being empty, so throw
             throw CocoaError.removeFileError(errno, resolve(path: pathStr))
         }
-        
+
+        #if os(WASI)
+
+        // wasi-libc does not support FTS, so we don't support removing non-empty directories on WASI for now.
+        throw CocoaError.errorWithFilePath(.featureUnsupported, pathStr)
+
+        #else
+
         let seq = _FTSSequence(path, FTS_PHYSICAL | FTS_XDEV | FTS_NOCHDIR | FTS_NOSTAT)
         let iterator = seq.makeIterator()
         var isFirst = true
@@ -561,6 +568,7 @@ enum _FileOperations {
                 }
             }
         }
+        #endif
         
     }
     #endif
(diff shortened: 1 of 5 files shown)


kateinoigakukun:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -903,6 +911,43 @@ enum _FileOperations {
     }
     #endif
 
+    #if os(WASI)
+    private static func _linkOrCopyFile(_ srcPtr: UnsafePointer<CChar>, _ dstPtr: UnsafePointer<CChar>, with fileManager: FileManager, delegate: some LinkOrCopyDelegate) throws {
```
MaxDesiatov, line 915: line length readability nit
```suggestion
    private static func _linkOrCopyFile(
        _ srcPtr: UnsafePointer<CChar>, 
        _ dstPtr: UnsafePointer<CChar>, 
        with fileManager: FileManager, delegate: some LinkOrCopyDelegate
    ) throws {
```

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -903,6 +911,43 @@ enum _FileOperations {
...
+        var stat = stat()
+        guard lstat(srcPtr, &stat) == 0, !stat.isDirectory else {
+            // wasi-libc does not support FTS for now, so we don't support copying/linking
+            // directories on WASI for now.
+            throw CocoaError(.featureUnsupported)
```
jmschonfeld, line 920: Similar comment as above - should we include the path here via `String(cString: srcPtr)`?

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -903,6 +911,43 @@ enum _FileOperations {
     }
     #endif
 
+    #if os(WASI)
+    private static func _linkOrCopyFile(_ srcPtr: UnsafePointer<CChar>, _ dstPtr: UnsafePointer<CChar>, with fileManager: FileManager, delegate: some LinkOrCopyDelegate) throws {
+        var stat = stat()
+        guard lstat(srcPtr, &stat) == 0, !stat.isDirectory else {
```
jmschonfeld, line 917: We might want to consult the delegate here - calling `shouldPerformOnItemAtPath` first, and if failing try `throwIfNecessary` and just return otherwise if the delegate wants to continue past this directory

Sources/_FoundationCShims/include/platform_shims.h
```diff
@@ -79,6 +79,29 @@ static inline _Nonnull clockid_t _platform_shims_clock_monotonic(void) {
...
+// `d_name` as "flexible array member" which is not supported by ClangImporter yet.
+
+#include <dirent.h>
+
+static inline char * _Nonnull _platform_shims_dirent_d_name(struct dirent * _Nonnull entry) {
+    return entry->d_name;
+}
+
+// Define getter shims for constants because wasi-libc defines them as function-like macros
+// which are not supported by ClangImporter yet.
```
jmschonfeld, line 93: For things like this, is there any hope for us resolving this somehow beyond updating the ClangImporter (some sort of API notes or something at the toolchain layer? Not sure if that's possible today, but I'm curious what our options are since it seems like WASI requires a large amount of shims

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -512,7 +512,14 @@ enum _FileOperations {
             // We failed for a reason other than the directory not being empty, so throw
             throw CocoaError.removeFileError(errno, resolve(path: pathStr))
         }
-        
+
+        #if os(WASI)
+
+        // wasi-libc does not support FTS, so we don't support removing non-empty directories on WASI for now.
+        throw CocoaError(.featureUnsupported)
```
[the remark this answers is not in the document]
MaxDesiatov: Very good catch, +1 here

Sources/_FoundationCShims/include/platform_shims.h
kateinoigakukun: Yes, we have three directions, 1. provide those definitions in WASILibc overlay module, 2. change the upstream wasi-libc headers to be compatible with ClangImporter, 3. "fix" ClangImporter to import those constants from wasi-libc headers directly. 

1 is what we are doing for some frequently used constants now. Still, it's quite ad-hoc and not a scalable solution. 2 is an option worth considering to align with other platform's headers, but not sure if it's acceptable for the upstream and if they can maintain the compatibility with ClangImporter.
I think 3 is the way we should go, and there are several open tickets for this kind of ClangImporter enhancement around macros (https://github.com/swiftlang/swift/issues/46679, https://github.com/swiftlang/swift/issues/45008, https://github.com/swiftlang/swift/issues/43102, https://github.com/swiftlang/swift/issues/45009). We need to carefully do it since it can break source compatibilities, but I have a plan to work on this.

kateinoigakukun:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileOperations.swift
jmschonfeld: The first piece looks good, but I think we still need to check with the delegate before throwing below via `delegate.throwIfNecessary` - that way if the delegate wants to continue, we correctly continue on past this item

kateinoigakukun: Oops, I missed the second part.

kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #838, opened 2024-08-08
Get Swift-Foundation building for Musl/Static SDK

etcwilde:
This adds the missing imports to get Swift-Foundation building against Musl so we can get the static SDK going again.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -16,6 +16,8 @@ internal import os
 import Android
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(CRT)
 import CRT
 #endif
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -16,6 +16,8 @@ internal import os
 import Android
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif canImport(CRT)
 import CRT
 #endif
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -22,6 +22,8 @@ import Darwin
 import Android
 #elseif canImport(Glibc)
 import Glibc
+#elseif canImport(Musl)
+import Musl
 #elseif os(Windows)
 import CRT
 import WinSDK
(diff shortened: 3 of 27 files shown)


compnerd (commented):
I'd really rather that we just merge #714 rather than do the workaround that you are doing for the macros.

etcwilde:
We can do that, but since they aren't used in the build, we don't need to build them at all.

jmschonfeld:
> We can do that, but since they aren't used in the build, we don't need to build them at all.

*yet - eventually the macro will be needed to build Foundation itself, so this isn't an option that we can expect to be long-lived if we did add it

etcwilde:
The install/deployment story for macros is still very hazy and will likely need to change, but this is for 6.0.
Without changing SPM and the compiler, the only way to get them to work for 6.0 is for the compiler to pull the macro from the toolchain side of things rather than in the SDK itself. This is safe enough because the static SDK only works with an aligned toolchain, so the macro content in the toolchain and the SDK content in the static SDK can't be out-of-sync and still be a viable configuration.

Unless you're planning on making them required to build Swift-Foundation in the 6.0 release? In which case, yes, we should take the bits of https://github.com/apple/swift-foundation/pull/714 for ensuring that the macros are compiled for the machine doing the build rather than the machine that we're building for. At the same time, in that case, we should build the macro independently as a separate project and expose it through some `SwiftFoundationMacroConfig.cmake` file so that we can `find_package` it and re-use the same one across builds instead of rebuilding it for each architecture.

jmschonfeld:
> but this is for 6.0.

Ah I missed that this is for 6.0 - normally for 6.0 work we merge to `main` and then cherry-pick back to 6.0 filling out the nomination template, could we follow that pattern here to be consistent? That will help us on the Foundation team out a lot with the tooling that we use to manage the repo. But yes we don't plan to require the macro to build Foundation itself in the swift 6 timeframe, so if we need to exclude the macro build from the static SDK due the the incompatibilities at the moment, that seems reasonable. Is that what the stdlib macros do today?

etcwilde:
https://github.com/apple/swift-corelibs-foundation/pull/5060
@swift-ci please test

Sources/CMakeLists.txt
```diff
@@ -14,9 +14,11 @@
 
 add_subdirectory(_FoundationCShims)
 
-# Disable the macro build on Windows until we can correctly build it for the host architecture
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
-	add_subdirectory(FoundationMacros)
+# Macros are only useful for development, but not at runtime.
+option(SwiftFoundation_BUILD_MACROS "Include SwiftFoundation Macro Targets" ON)
```
iCharlesHu, line 18: Is this flag set to false by the build script on Windows?

iCharlesHu (approved):

iCharlesHu:
This change will need to be cherry-picked to `release/6.0`. Please tag @itingliu for review since she's the release manager for 6.0

al45tair (approved):

Sources/CMakeLists.txt
etcwilde: Good catch, is now: https://github.com/swiftlang/swift/pull/75814

itingliu (approved):
6.0 change looks fine to me

etcwilde:
@swift-ci please test

PR #839, opened 2024-08-09
Fix ProcessInfo.processName for Windows

iCharlesHu:
Instead of relying on hard coding PATH_SEPARATOR, use existing .lastPathComponent that already works on Windows


iCharlesHu:
@swift-ci please test

PR #842, opened 2024-08-10
[Android] Use the Bionic module in more places

finagolfin:
Also, use `canImport()` wherever importing APIs, reserving `os(Android)` for platform differences.

I've started porting this repo to Android: this is a first pull that keeps everything building while more narrowly scoping imports.

@compnerd and @hyp, try it out and let me know what you think.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -18,7 +18,7 @@ internal import os
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import unistd
 #elseif canImport(Glibc)
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -17,7 +17,7 @@ internal import DarwinPrivate.sys.content_protection
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import posix_filesystem
 #elseif canImport(Glibc)
--- a/Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
@@ -12,7 +12,7 @@
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import unistd
 #elseif canImport(Glibc)
--- a/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
@@ -109,7 +109,7 @@ struct _Win32DirectoryContentsSequence: Sequence {
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import posix_filesystem.dirent
 #elseif canImport(Glibc)
@@ -328,7 +328,7 @@ extension Sequence<_FTSSequence.Element> {
 struct _POSIXDirectoryContentsSequence: Sequence {
     #if canImport(Darwin)
     typealias DirectoryEntryPtr = UnsafeMutablePointer<DIR>
-    #elseif os(Android) || canImport(Glibc) || canImport(Musl) || os(WASI)
+    #elseif canImport(Android) || canImport(Glibc) || canImport(Musl) || os(WASI)
     typealias DirectoryEntryPtr = OpaquePointer
     #endif
     
--- a/Sources/FoundationEssentials/Platform.swift
+++ b/Sources/FoundationEssentials/Platform.swift
@@ -28,7 +28,7 @@ fileprivate let _pageSize: Int = {
 #elseif os(WASI)
 // WebAssembly defines a fixed page size
 fileprivate let _pageSize: Int = 65_536
-#elseif os(Android)
+#elseif canImport(Android)
 import Bionic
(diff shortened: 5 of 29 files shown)


Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -18,7 +18,7 @@ internal import os
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import unistd
```
finagolfin, line 23: Isn't this redundant since the prior Android overlay includes it?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -17,7 +17,7 @@ internal import DarwinPrivate.sys.content_protection
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import posix_filesystem
```
finagolfin, line 22: Same here

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -12,7 +12,7 @@
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import unistd
```
finagolfin, line 17: Same here

Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
```diff
@@ -109,7 +109,7 @@ struct _Win32DirectoryContentsSequence: Sequence {
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
 import Android
 import posix_filesystem.dirent
```
finagolfin, line 114: Same here

Sources/FoundationEssentials/Platform.swift
```diff
@@ -28,7 +28,7 @@ fileprivate let _pageSize: Int = {
 #elseif os(WASI)
 // WebAssembly defines a fixed page size
 fileprivate let _pageSize: Int = 65_536
-#elseif os(Android)
+#elseif canImport(Android)
```
iCharlesHu, line 31: Is it intentional that we check `canImport(Android)` but actually importing `Bionic` below? Do `canImport(Android)` and `canImport(Bionic)` always return the same value on `Android`?

I agree with you that using `canImport()` is better than `os()`, but it seems we are using `canImport(Android)` and `canImport(Bionic)` interchangeably in place of `os(Android)`. Can we standardize that to one?

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -14,7 +14,7 @@ internal import _FoundationCShims
 
 #if canImport(Darwin)
 import Darwin
-#elseif os(Android)
+#elseif canImport(Android)
```
iCharlesHu, line 17: Same here

Sources/FoundationEssentials/_ThreadLocal.swift
```diff
@@ -24,7 +24,7 @@ internal import threads
 #endif
 
 struct _ThreadLocal {
-#if canImport(Darwin) || os(Android) || canImport(Glibc)
+#if canImport(Darwin) || canImport(Bionic) || canImport(Glibc)
```
iCharlesHu, line 27: Here we are using `canImport(Bionic)` in place of `os(Android)` but we are using `canImport(Android)` above.

compnerd:
`os(Android)` does seem reasonable as the check to me as the libc (bionic) is part of the OS and we should be able to assume that the Android overlay is available.

Sources/FoundationEssentials/Platform.swift
finagolfin: > Is it intentional that we check canImport(Android) but actually importing Bionic below?

Yes, as `unistd` comes from the `Android` overlay. I did not add those imports, so I merely used the most appropriate `canImport()` here, the one that includes them both, since `Android` is a superset of `Bionic`.

> Do canImport(Android) and canImport(Bionic) always return the same value on Android?

Yes, they are interchangeable, as they are both defined in the same module map. However, `import Bionic` and `import Android` are not interchangeable, so I simply choose the `canImport()` to match the one we are actually importing.

> we are using canImport(Android) and canImport(Bionic) interchangeably... Can we standardize that to one?

No need to, as they _are_ interchangeable and one is simply chosen to match the non-interchangeable `import` used.

finagolfin:
> os(Android) does seem reasonable as the check to me as the libc (bionic) is part of the OS and we should be able to assume that the Android overlay is available.

It will work, but it is not correct as an indicator of what is happening. While both may be functionally interchangeable, we should be consistent with other platforms and only use `canImport()` when selecting platform C APIs and `os(Android)` for platform differences, as the few remaining uses of the latter I left do.

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
finagolfin: Same explanation as above

finagolfin:
Rebased and fixed merge conflict.

finagolfin:
Rebased again and fixed merge conflicts.

iCharlesHu:
@swift-ci please test

iCharlesHu (approved):

finagolfin:
Thanks @iCharlesHu, I was just able to run the tests for this repo natively on Android for the first time. 😄

Only five tests failed- two `FileManager` tests and three `ProcessInfo` tests- I will look into those and submit a pull for those next, before trying to build and run the full re-cored `swift-corelibs-foundation` tests natively on Android.

PR #843, opened 2024-08-10
Autolink wasi-libc emulation libraries

kateinoigakukun:
`_WASI_EMULATED_XXX` macros require linking against the corresponding emulation libraries. This patch adds `-Xfrontend -public-autolink-library` flags to the Swift compiler invocation to automatically link against the emulation libraries. Also enable getpid emulation explicitly, as it is available by default but emiting a warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,16 +114,25 @@ foreach(version ${_SwiftFoundation_versions})
     endforeach()
 endforeach()
 
-# wasi-libc emulation feature flags
+# wasi-libc emulation feature flags passed to the Swift compiler
 set(_SwiftFoundation_wasi_libc_flags)
+# wasi-libc emulation libraries to link against when building a shared library
+set(_SwiftFoundation_wasi_libc_libraries)
 if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
     list(APPEND _SwiftFoundation_wasi_libc_flags
         "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xcc -D_WASI_EMULATED_GETPID>"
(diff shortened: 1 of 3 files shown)


kateinoigakukun:
@swift-ci test

CMakeLists.txt
```diff
@@ -118,8 +118,12 @@ endforeach()
 set(_SwiftFoundation_wasi_libc_flags)
 if(CMAKE_SYSTEM_NAME STREQUAL "WASI")
     list(APPEND _SwiftFoundation_wasi_libc_flags
+        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xcc -D_WASI_EMULATED_GETPID>"
+        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend wasi-emulated-getpid>"
```
jmschonfeld, line 122: Previously we've used `-public-autolink-library` to have clients of the module link in the listed library which was needed for the static build. In this case, is this to support clients linking as part of the static SDK (if so should we put this in an `if(BUILD_SHARED_LIBS)` path?) or is this just for FoundationEssentials to link these libraries (if so, do we need a public auto link or is it enough to just add the equivalent of `-lwasi-emulated-getpid`?)

kateinoigakukun: Good catch! We don't support shared library on WASI for now, but for shared libraries, we should link emulation libraries to shared libraries instead of asking clients to link them.

kateinoigakukun:
@swift-ci test

jmschonfeld (approved):
This looks ok to me - since this makes some nontrivial changes to the CMake files and we don't have CI for that on this repo yet, could you kick off a cross-repo toolchain test run on a swift PR linked to this change so that we can validate this doesn't somehow inadvertently break the existing CMake build? After that passes we can go ahead and merge this

jmschonfeld:
Looks like the linked toolchain build passed. Thanks!

PR #844, opened 2024-08-12
Automatically build FoundationMacros for local CMake builds

jmschonfeld:
When building this project via CMake as a standalone project (not as part of a toolchain build) we automatically check out and build some dependencies such as swift-collections and swift-foundation-icu. Now that we've separated the macro build into a separate build, we need to do the same for FoundationMacros. This updates the CMake files so that when `SwiftFoundation_MACROS` is not specified, we add the macro target to the build graph and build a copy of the macro for the host to use while building the rest of the project. This builds the macro as an executable just like SwiftPM rather than as a library so that the macro can be built with a different copy of swift-syntax than what the compiler has.

While I was here, I also added some logging messages when checking out the dependencies to make it clear what build settings you can specify for local copies, as well as indicate why a cmake configuration was taking a while during git cloning

### DIFF
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -50,8 +50,26 @@ add_subdirectory(TimeZone)
 add_subdirectory(URL)
 
 if(SwiftFoundation_MACRO)
-  target_compile_options(FoundationEssentials PRIVATE
-    "SHELL:-plugin-path ${SwiftFoundation_MACRO}")
+    message(STATUS "SwiftFoundation_MACRO provided, using macros in ${SwiftFoundation_MACRO}")
+    # A path to Foundation macros was provided, so we use that path
+    target_compile_options(FoundationEssentials PRIVATE
+        "SHELL:-plugin-path ${SwiftFoundation_MACRO}")
+else()
+    message(STATUS "SwiftFoundation_MACRO not provided, building Foundation macros locally for host")
+    # No path to Foundation macros was provided, so we must build it ourselves
+    set(FoundationMacros_BuildLocalExecutable YES)
+    FetchContent_Declare(FoundationMacros
+        SOURCE_DIR ${CMAKE_SOURCE_DIR}/Sources/FoundationMacros)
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
Full toolchain tests which cover the CMake changes passed in the linked PR

iCharlesHu (approved):

Sources/FoundationEssentials/CMakeLists.txt
```diff
@@ -50,8 +50,26 @@ add_subdirectory(TimeZone)
...
-    "SHELL:-plugin-path ${SwiftFoundation_MACRO}")
+    message(STATUS "SwiftFoundation_MACRO provided, using macros in ${SwiftFoundation_MACRO}")
+    # A path to Foundation macros was provided, so we use that path
+    target_compile_options(FoundationEssentials PRIVATE
+        "SHELL:-plugin-path ${SwiftFoundation_MACRO}")
+else()
+    message(STATUS "SwiftFoundation_MACRO not provided, building Foundation macros locally for host")
+    # No path to Foundation macros was provided, so we must build it ourselves
+    set(FoundationMacros_BuildLocalExecutable YES)
+    FetchContent_Declare(FoundationMacros
```
etcwilde, line 61: The macros should build through ExternalProject. If we are cross-compiling Foundation, these macros will get built for the machine that we are cross-compiling to instead of the machine running the build, and therefore will not be usable.

Sources/FoundationEssentials/CMakeLists.txt
```diff
@@ -50,8 +50,26 @@ add_subdirectory(TimeZone)
...
+        SOURCE_DIR ${CMAKE_SOURCE_DIR}/Sources/FoundationMacros)
```
drodriguez, line 62: Wouldn't `CMAKE_SOURCE_DIR` be the `swift-corelibs-foundation` when this project is added as an external project from there? Would `${CMAKE_CURRENT_SOURCE_DIR}/../FoundationMacros` be more correct here?

PR #846, opened 2024-08-12
[android] fix the LP32 armv7/i686 android build

hyp:
This fixes the errors related to the st_mode structure being operated on with mode_t values, and the use of `pw_gecos` error.

### DIFF
--- a/Sources/FoundationEssentials/Android+Extensions.swift
+++ b/Sources/FoundationEssentials/Android+Extensions.swift
@@ -1,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+//
+//===----------------------------------------------------------------------===//
+
+#if os(Android) && (arch(i386) || arch(arm))
+import Android
+
+// struct stat.st_mode is UInt32
+internal func &(left: UInt32, right: mode_t) -> mode_t {
+    return mode_t(left) & right
+}
+#endif
--- a/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
+++ b/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
@@ -52,7 +52,7 @@ internal struct BuiltInUnicodeScalarSet {
 
     // CFUniCharIsMemberOf
     func contains(_ scalar: Unicode.Scalar) -> Bool {
-        let planeNo = Int((scalar.value >> 16) & 0xFF)
+        let planeNo = Int((scalar.value >> 16) & UInt32(0xFF))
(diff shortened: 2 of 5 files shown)


hyp:
cc @compnerd

hyp:
@swift-ci please test

compnerd (approved):

hyp:
CC @jmschonfeld

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
```diff
@@ -52,7 +52,7 @@ internal struct BuiltInUnicodeScalarSet {
 
     // CFUniCharIsMemberOf
     func contains(_ scalar: Unicode.Scalar) -> Bool {
-        let planeNo = Int((scalar.value >> 16) & 0xFF)
+        let planeNo = Int((scalar.value >> 16) & UInt32(0xFF))
```
jmschonfeld, line 55: Does `Unicode.Scalar.value` have a different type on Android than on other platforms?

Sources/FoundationEssentials/Android+Extensions.swift
```diff
@@ -0,0 +1,19 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
```
jmschonfeld, line 5:
```suggestion
// Copyright (c) 2024 Apple Inc. and the Swift project authors
```

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
hyp: No, it's just that the new operator conflicts during overload resolution, with this specific instance of using `&` on a constant that is not typed.

hyp:
@swift-ci please test

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
jmschonfeld: Hmm, based on that it seems like the new operator has the possibility of a widespread affect across the project, and I worry that changes like this which appear arbitrary to someone not knowledgeable of the android build are likely to cause the android build to be broken quite frequently. Is there anything we can do to target that operator to just the uses we want? How unruly would it be to perform that at each bitwise-and with a `mode_t` rather than using a global operator?

hyp: that should work too, let me see how invasive such change would be.

hyp:
@swift-ci please test

hyp:
@jmschonfeld I dropped the operator & overload, how does it look now?

compnerd (approved):

hyp:
@jmschonfeld could you take another look please 🙏 , we would like to get this in to make further build progress

jmschonfeld (approved):

hyp:
Thanks!

PR #849, opened 2024-08-14
FoundationMacros: use cross-compilation to build for host (#714)

compnerd:
Use `ExternalProject` to switch FoundationMacros to cross-compilation. This allows us to build the macros for the right OS/architecture when cross-compiling Foundation for other environments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,8 @@ set(BUILD_TESTING NO)
 set(COLLECTIONS_SINGLE_MODULE YES)
 set(COLLECTIONS_FOUNDATION_TOOLCHAIN_MODULE YES)
 
+set(SwiftFoundation_MACRO "" CACHE STRING "Path to Foundation macro plugin")
+
 # Make sure our dependencies exists
 include(FetchContent)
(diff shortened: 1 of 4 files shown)


compnerd:
**Explanation**: Build changes necessary to support building macros for Windows.
**Scope**: Enables building and distributing the Foundation macros on all platforms.
**Issue**:
**Original PR**: #714
**Risk**: We would need additional changes to build and distribute the macros.
**Testing**: The original changes were tested as part of building the toolchain and the installer for Windows as well as the built image for Linux.
**Reviewer**: @jmschonfeld

compnerd:
@swift-ci please test

jmschonfeld:
Backporting this sounds like a good idea to me, let's just make sure that's ok with @etcwilde since his Musl PR also tries to address this, and let's also make sure we back port the swift repo changes (for build-script and build.ps1) and the installer script changes to install the macro on Windows

etcwilde (approved):
If we're waiting on me for approval, I'm fine with this. I just need a decision landed so that I can get the static SDK tipped back upright.

PR #851, opened 2024-08-14
[android] don't enable WMO as the old driver fails to build with WMO …

hyp:
…on windows

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,7 +36,8 @@ if(NOT SWIFT_SYSTEM_NAME)
 endif()
 
 # Don't enable WMO on Windows due to linker failures
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+# Don't enable WMO on Android as swift's old driver has a bug with using it.
+if((NOT CMAKE_SYSTEM_NAME STREQUAL Windows) AND (NOT CMAKE_SYSTEM_NAME STREQUAL Android))
     # Enable whole module optimization for release builds & incremental for debug builds
     if(POLICY CMP0157)
         set(CMAKE_Swift_COMPILATION_MODE "$<IF:$<CONFIG:Release>,wholemodule,incremental>")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,9 +35,9 @@ if(NOT SWIFT_SYSTEM_NAME)
   endif()
 endif()
 
-# Don't enable WMO on Windows due to linker failures
-# Don't enable WMO on Android as swift's old driver has a bug with using it.
-if((NOT CMAKE_SYSTEM_NAME STREQUAL Windows) AND (NOT CMAKE_SYSTEM_NAME STREQUAL Android))
+# Don't enable WMO on Windows hosts due to linker failures, and the use of swift's
+# old driver from build.ps1 when building for windows/android on a windows host.
+if(NOT CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
     # Enable whole module optimization for release builds & incremental for debug builds
     if(POLICY CMP0157)
         set(CMAKE_Swift_COMPILATION_MODE "$<IF:$<CONFIG:Release>,wholemodule,incremental>")


hyp:
build.ps1 uses the old driver which has an issue with the WMO mode for android here.

hyp:
cc @compnerd  @jmschonfeld

hyp:
@swift-ci please test

jmschonfeld (approved):
> build.ps1 uses the old driver which has an issue with the WMO mode for android here.

Is this a bug we're tracking? It'd be nice to know so that we can hopefully fix it. Enabling WMO has a >4x improvement in performance for `JSONDecoder` in some cases so it'd be great if platforms like Windows and Android can eventually get that too 🙂

hyp:
@jmschonfeld interesting, thanks for mentioning the numbers. That's definitely something that we want to fix. I filed https://github.com/swiftlang/swift/issues/75890 to track this

CMakeLists.txt
```diff
@@ -36,7 +36,8 @@ if(NOT SWIFT_SYSTEM_NAME)
 endif()
 
 # Don't enable WMO on Windows due to linker failures
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+# Don't enable WMO on Android as swift's old driver has a bug with using it.
+if((NOT CMAKE_SYSTEM_NAME STREQUAL Windows) AND (NOT CMAKE_SYSTEM_NAME STREQUAL Android))
```
compnerd, line 40: I don't think that this is a good "test". If I am building on Darwin, this should still be enabled yes? Perhaps we should be doing:

```cmake
if(NOT CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
```

hyp: I don't know what the windows linker error is so I can't comment on whether there's a host/target windows system dependency or not. The android issue is host specific though so we can use the HOST check for sure

hyp: although it doesn't really make sense if windows is host/target since we're always going to be building for windows target on windows host. So yeah, actually your suggestion makes sense!

hyp:
@swift-ci please test

hyp:
@jmschonfeld how does it look now?

jmschonfeld (approved):
@hyp makes sense to me, could we run a quick test on the swift repo via a cross repo test? (sorry, we don't have any CI on the cmake build in this repo yet so a quick validation on that before merging would be great - we're hoping to enable that on this repo in the future)

CMakeLists.txt
```diff
@@ -35,8 +35,9 @@ if(NOT SWIFT_SYSTEM_NAME)
   endif()
 endif()
 
-# Don't enable WMO on Windows due to linker failures
-if(NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
+# Don't enable WMO on Windows hosts due to linker failures, and the use of swift's
+# old driver from build.ps1 when building for windows/android on a windows host.
+if(NOT CMAKE_HOST_SYSTEM_NAME STREQUAL Windows)
```
compnerd, line 40: Can we just drop this or conditionalise it on the CMake version? CMake has had `-wmo` as the default since 3.26. The compilation mode defaults were also part of the release when CMP0157 was introduced. Effectively, this is just to enable `-wmo` with CMake 3.24..<3.26.

compnerd: Well, the problem is that WMO doesn't work very well with the old driver :(. When we build the toolchain, we don't yet have swift-driver as we need Foundation to build swift-driver. We can elide the WMO on Windows using `-D CMAKE_Swift_COMPILER_USE_OLD_DRIVER=YES` to workaround the issue for now. We will want to consider a multi-stage build so we can build with the new driver.

hyp:
@compnerd I tested `CMAKE_Swift_COMPILER_USE_OLD_DRIVER=YES`, but it doesn't seem to help here for android specifically, so it looks like we still need this until we migrate over to the new driver.

@jmschonfeld do you mind merging this in ? I ran the cross-PR testing in https://github.com/swiftlang/swift/pull/75927

jmschonfeld (approved):

PR #852, opened 2024-08-14
URL.fileSystemPath should drop all trailing slashes

jrflat:
Currently `URL.fileSystemPath` only drops the last trailing slash. It should drop all of them so that we get:

```
let path = “/path///“
let url = URL(filePath: path)
print(url.fileSystemPath) // "/path" (previously "/path//" having one less trailing slash)
```


jrflat:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu (approved):

jrflat:
@swift-ci please test

PR #853, opened 2024-08-14
Hide Windows magic environment values from ProcessInfo.environment

iCharlesHu:
Windows GetEnvironmentStringsW API can return magic environment variables set by the cmd shell that starts with "=". We should hide these values to avoid surprising behavior.

resolves: https://github.com/apple/swift-foundation/issues/847

### DIFF
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -75,6 +75,16 @@ final class _ProcessInfo: Sendable {
             for env in environments {
                 let environmentString = String(cString: env)
 
+#if os(Windows)
+                // Windows GetEnvironmentStringsW API can return
+                // magic environment variables set by the cmd shell
+                // that starts with `=`
+                // We should exclude these values
+                if environmentString.utf8.first == ._equal {
+                    continue
+                }
+#endif // os(Windows)
+
                 guard let delimiter = environmentString.firstIndex(of: "=") else {
                     continue
                 }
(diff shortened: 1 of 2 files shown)


Tests/FoundationEssentialsTests/ProcessInfoTests.swift
```diff
@@ -189,4 +189,16 @@ final class ProcessInfoTests : XCTestCase {
         processInfo.processName = originalProcessName
         XCTAssertEqual(processInfo.processName, originalProcessName)
     }
+
+#if os(Windows)
```
jmschonfeld, line 193: Might as well run this test on all platforms - this behavior should be the same on macOS and Linux too

Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -75,6 +75,16 @@ final class _ProcessInfo: Sendable {
             for env in environments {
                 let environmentString = String(cString: env)
 
+#if os(Windows)
+                // Windows GetEnvironmentStringsW API can return
+                // magic environment variables set by the cmd shell
+                // that starts with `=`
+                // We should exclude these values
+                if environmentString.hasPrefix("=") {
```
jmschonfeld, line 83: Somewhat of a nit: while it's unlikely, it's possible that one of these magic variables could start with a grapheme extend character and this check would fail because this is checking based on the character view. Should we consider using `environmentString.utf8.first == UInt8(ascii: "=")` here to ensure that regardless of what follows the `=`, we catch anything that begins with the `=` scalar?

iCharlesHu: No that's a good point! I love defensive programming. Thanks!!

iCharlesHu:
Addressed Jeremy's comments.

iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #854, opened 2024-08-15
Use fallback home directory on Windows

jmschonfeld:
On macOS/Linux, when asking for the home directory of a user that doesn't exist, we fallback to `/var/empty` - we also have a handful of unit tests in swift-corelibs-foundation that verify this behavior. However, on Windows we currently crash with a `fatalError` when provided a non-existent user name. swift-corelibs-foundation previously defaulted to `/var/empty` on Windows too, but that directory doesn't make sense on Windows. Instead, we standardize on a fallback of `%SystemDrive%\Users\Public` for users that don't exist which is still a valid directory and is close enough to the macOS/Linux semantics despite not having a direct equivalent to `/var/empty`

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -369,33 +369,27 @@ extension String {
     #if !NO_FILESYSTEM
     internal static func homeDirectoryPath(forUser user: String? = nil) -> String {
 #if os(Windows)
-        func GetUserProfile() -> String? {
-            return "USERPROFILE".withCString(encodedAs: UTF16.self) { pwszVariable in
-                let dwLength: DWORD = GetEnvironmentVariableW(pwszVariable, nil, 0)
-                // Ensure that `USERPROFILE` is defined.
-                if dwLength == 0 { return nil }
-                return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/String/String+Path.swift
```diff
@@ -369,33 +369,27 @@ extension String {
...
-                // Ensure that `USERPROFILE` is defined.
-                if dwLength == 0 { return nil }
-                return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-                    guard GetEnvironmentVariableW(pwszVariable, $0.baseAddress, dwLength) == dwLength - 1 else {
-                        return nil
-                    }
-                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        if let user {
+            func fallbackUserDirectory() -> String {
+                guard let systemDrive = ProcessInfo.processInfo.environment["SystemDrive"] else {
```
compnerd, line 374: I think we should use `ALLUSERSPROFILE` instead (https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables).

jmschonfeld:
@swift-ci please test

compnerd (approved):
Thanks!

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #855, opened 2024-08-15
Fix hard-coded path to `FoundationEssentialsTests` resources

xedin:
SwiftPM used to incorrectly build tests for both host and target. That has been fixed by https://github.com/swiftlang/swift-package-manager/pull/7879.

The fix makes sure that if there are any direct macro dependencies in test targets xctest bundle is only built for "host".

This means that the hard-coded resources path needs to be updated to include "-tool" suffix.

### DIFF
--- a/Tests/FoundationEssentialsTests/ResourceUtilities.swift
+++ b/Tests/FoundationEssentialsTests/ResourceUtilities.swift
@@ -56,10 +56,23 @@ func testData(forResource resource: String, withExtension ext: String, subdirect
     // swiftpm drops the resources next to the executable, at:
     // ./swift-foundation_FoundationEssentialsTests.resources/Resources/
     // Hard-coding the path is unfortunate, but a temporary need until we have a better way to handle this
-    var path = URL(filePath: ProcessInfo.processInfo.arguments[0])
+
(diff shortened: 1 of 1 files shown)


xedin:
https://github.com/swiftlang/swift-package-manager/pull/7883
@swift-ci please test

jmschonfeld:
Looks like Windows failed because this change isn't compatible with older versions of SwiftPM (note: this repo does not support cross-repo testing, tests are run using the SwiftPM from the most recent nightly toolchain)

```
FoundationEssentialsTests/JSONEncoderTests.swift:2411: Fatal error: Unexpectedly found nil while unwrapping an Optional value
```

jmschonfeld:
> @jmschonfeld I triggered Windows CI in my swift PR, if that passes we should be good to go.

@xedin that won't be sufficient here I'm afraid. This package needs to build with the latest nightly toolchain because that's what our CI currently uses and what we use to develop at desk. We need a solution here that will work both with your change and without - even if your swift PR passes, CI in this repo will continue to fail since it's not using a just-built toolchain, and we will no longer be able to build just Foundation at desk

xedin:
@swift-ci please test

jmschonfeld (approved):

PR #858, opened 2024-08-15
Fix Windows symlink handling in FileManager APIs

jmschonfeld:
While getting SCL-F tests up and running on Windows, I noticed that those tests expose a handful of behavioral changes that swift-foundation made compared to swift-corelibs-foundation specifically around handling symlinks on Windows. In particular:

- `contentsEqual(atPath:andPath:)` would compare contents at the destination of symlinks rather than at the symlinks themselves
- `copyItem(atPath:toPath:)` would fail when copying symlinks
- `copyItem(atPath:toPath:)` would follow symlinks to directories and copy the directory contents rather than copying the symlinks themselves

This PR addresses these issues and adds a handful of unit tests to validate all of these behaviors.

Resolves https://github.com/apple/swift-foundation/issues/845

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
@@ -78,14 +78,14 @@ internal struct _FileManagerImpl {
     ) -> Bool {
 #if os(Windows)
         return (try? path.withNTPathRepresentation {
-            let hLHS = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, nil)
+            let hLHS = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS, nil)
             if hLHS == INVALID_HANDLE_VALUE {
                 return false
             }
             defer { CloseHandle(hLHS) }
 
             return (try? other.withNTPathRepresentation {
-                let hRHS = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS, nil)
+                let hRHS = CreateFileW($0, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_OPEN_REPARSE_POINT | FILE_FLAG_BACKUP_SEMANTICS, nil)
                 if hRHS == INVALID_HANDLE_VALUE {
                     return false
                 }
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -826,10 +835,10 @@ enum _FileOperations {
...
                         try linkOrCopyFile(src.appendingPathComponent(item.fileName), dst: dst.appendingPathComponent(item.fileName), with: fileManager, delegate: delegate)
                     }
-                } else if bCopyFile || faAttributes.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT {
+                } else if bCopyFile {
                     var ExtendedParameters: COPYFILE2_EXTENDED_PARAMETERS = .init()
                     ExtendedParameters.dwSize = DWORD(MemoryLayout<COPYFILE2_EXTENDED_PARAMETERS>.size)
-                    ExtendedParameters.dwCopyFlags = COPY_FILE_FAIL_IF_EXISTS | COPY_FILE_COPY_SYMLINK | COPY_FILE_NO_BUFFERING | COPY_FILE_OPEN_AND_COPY_REPARSE_POINT
+                    ExtendedParameters.dwCopyFlags = COPY_FILE_FAIL_IF_EXISTS | COPY_FILE_NO_BUFFERING
 
                     if FAILED(CopyFile2(pwszSource, pwszDestination, &ExtendedParameters)) {
```
jmschonfeld, line 843: As @jakepetroules mentioned in the GitHub issue - it seems there may be an issue with `CopyFile2` itself here. I was unable to figure out the issue, so instead we now just handle symlink logic separately by reading the symlink's destination and creating a new symlink at the copy destination (which is what swift-corelibs-foundation roughly did)

compnerd (approved):

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -122,11 +122,19 @@ internal struct _FileManagerImpl {
...
-                if fbiLHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT,
-                   fbiRHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT {
+                let lhsIsReparsePoint = fbiLHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT
+                let rhsIsReparsePoint = fbiRHS.FileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT
+                let lhsIsDirectory = fbiLHS.FileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY
+                let rhsIsDirectory = fbiRHS.FileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY
+                if lhsIsReparsePoint || rhsIsReparsePoint {
+                    guard lhsIsReparsePoint == rhsIsReparsePoint else {
+                        return false
+                    }
```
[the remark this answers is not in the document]
jmschonfeld: Ah yeah it could, let me see if it can make it a little cleaner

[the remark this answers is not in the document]
jmschonfeld: Cleaned it up a bit and simplified the conditions

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -817,7 +817,16 @@ enum _FileOperations {
             guard delegate.shouldPerformOnItemAtPath(src, to: dst) else { return }
 
             try dst.withNTPathRepresentation { pwszDestination in
-                if faAttributes.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY == FILE_ATTRIBUTE_DIRECTORY {
+                // Check for reparse points first because symlinks to directories are reported as both reparse points and directories, and we should copy the symlink not the contents of the linked directory
+                if faAttributes.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT == FILE_ATTRIBUTE_REPARSE_POINT {
```
itingliu, line 821: Would a non-symlinked directory happen to satisfy this constraint as well, and is that the desired behavior? By reading at this code, I'm not sure if `fileManager.createSymbolicLink` would succeed since it's not a symlink. And if it fails, we'd `throw` instead of proceeding to copy the content of the directory. Is that ok?

itingliu (approved):
I left a question, but I'm not sure if that's really a supported scenario. I don't mind addressing it later if that turns out to be an issue.

Sources/FoundationEssentials/FileManager/FileOperations.swift
jmschonfeld: Yeah plain directories won't have this flag set, but other types of reparse points that aren't supported by `destinationOfSymbolicLink` would likely fail here - I think we can address that later since most of the `FileManager` code in general likely doesn't do well with other types of reparse points

PR #861, opened 2024-08-16
Fix Windows home directory for specific user

jmschonfeld:
When looking up a home directory for a specific user, we currently have a bug that results in always using the fallback path. This PR updates the implementation to:

- Expect `LookupAccountNameW` and `RegGetValueW` to fail at the first call with an empty buffer (since there is no space in the buffer and it needs to return a valid buffer size
- Fix a minor typo in the registry key that ensures we interpolate the user name rather than putting the literal `\(String(decodingC...` text into the key string

The added unit test ensures we exercise this code path

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -382,7 +382,8 @@ extension String {
                 var cbSID: DWORD = 0
                 var cchReferencedDomainName: DWORD = 0
                 var eUse: SID_NAME_USE = SidTypeUnknown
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

compnerd (approved):
Whoops, thanks for fixing this!

jmschonfeld:
Hmm, it looks like `ProcessInfo.processInfo.userName` is returning an empty string causing this test to fail, I wonder if the CI environment is set up in such a way that we can't reliably get the current user's username?

jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #862, opened 2024-08-16
[6.0] Use fallback home directory on Windows

jmschonfeld:
Explanation: Restores Swift 5.10 behavior for getting a home directory for a non-existent user on Windows (using a reasonable default rather than asserting)
Scope: Only impacts looking up home directories for non-existent users on Windows and turns a crash into a successful return value
Original PR: https://github.com/apple/swift-foundation/pull/854
Risk: Minimal - scope is very small and it restores pre-recore behavior
Testing: Testing done via swift-ci testing and local testing
Reviewer: @compnerd @iCharlesHu


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #867, opened 2024-08-16
URL(filePath: path, directoryHint: .notDirectory) should strip trailing slashes

jrflat:
If a user supplies a `path` with trailing slash(es) and hints that it's `.notDirectory`, we should take their word for it and strip all trailing slashes. While that user input is slightly contradictory, this was the behavior before the Swift URL implementation, and at least it maintains the following consistency for any `path`:

```
let url = URL(filePath: path, directoryHint: .notDirectory)
XCTAssertEqual(url.hasDirectoryPath, false)
```

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -2022,24 +2022,28 @@ extension URL {
 
 #if !NO_FILESYSTEM
     private static func isDirectory(_ path: String) -> Bool {
-#if !FOUNDATION_FRAMEWORK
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jrflat:
Windows failure is real, the trailing slashes are not being stripped because they're backslashes at that point in the code, will push a fix

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #869, opened 2024-08-16
subpathsOfDirectory(atPath:) should throw on empty paths

jmschonfeld:
This should hopefully be the last of the Windows FileManager fixups for now 🙂

A swift-corelibs-foundation test revealed that `subpathsOfDirectory(atPath: "")` throws on Darwin/Linux but returns an empty array on Windows. We should make it behave the same by throwing an error (however Windows currently throws `.fileReadInvalidFileName` for empty paths unlike Darwin/Linux which reports that the file doesn't exist, so we maintain that consistency with other APIs here)

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
@@ -33,6 +33,11 @@ struct _Win32DirectoryContentsSequence: Sequence {
         init(path: String, appendSlashForDirectory: Bool, prefix: [String]) {
             self.slash = appendSlashForDirectory
 
+            guard !path.isEmpty else {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

compnerd (approved):
Any reason to not change the error to match the other platforms?

jmschonfeld:
> Any reason to not change the error to match the other platforms?

Not in particular, although I think that may be best done as a followup since there are a lot of FileManager APIs that return `fileReadInvalidFileName` on Windows but a variant of `fileReadNoSuchFile` on Linux/Darwin. I'm fine with making them consistent, but it might be easiest to do that in one chunk unless you think I should go ahead and update this one now while I'm here

parkera (approved):

PR #871, opened 2024-08-17
[Android] Enable more code and tests

finagolfin:
while disabling setting extended file attributes and a test creating a hard link, features not normally allowed on Android.

This got all the tests passing natively on Android 13 AArch64.

@hyp and @compnerd, try it out yourself and please review.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -956,6 +956,8 @@ extension _FileManagerImpl {
                 #if os(WASI)
                 // WASI does not support extended attributes
                 throw CocoaError.errorWithFilePath(.featureUnsupported, path)
+                #elseif canImport(Android)
+                // Android doesn't allow setting this for normal apps, so just skip it.
                 #else
                 try Self._setAttributes(extendedAttrs, at: fileSystemRepresentation, followSymLinks: false)
                 #endif
--- a/Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
+++ b/Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
@@ -158,7 +158,7 @@ struct TimeZoneCache : Sendable, ~Copyable {
                 }
             }
             
-#if os(Linux) && !os(WASI)
+#if os(Linux)
             // Try localtime
             tzset()
             var t = time(nil)
--- a/Tests/FoundationEssentialsTests/DataIOTests.swift
+++ b/Tests/FoundationEssentialsTests/DataIOTests.swift
@@ -238,7 +238,7 @@ class DataIOTests : XCTestCase {
     }
     
     func test_zeroSizeFile() throws {
-        #if !os(Linux)
+        #if !os(Linux) && !os(Android)
         throw XCTSkip("This test is only applicable on Linux")
         #else
         // Some files in /proc report a file size of 0 bytes via a stat call
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -23,6 +23,10 @@ import TestSupport
 @testable import Foundation
 #endif
 
+#if canImport(Android)
+import Android
+#endif
+
 extension FileManager {
     fileprivate var delegateCaptures: DelegateCaptures {
         (self.delegate as! CapturingFileManagerDelegate).captures
@@ -337,8 +341,13 @@ final class FileManagerTests : XCTestCase {
             XCTAssertTrue($0.delegateCaptures.isEmpty)
             try $0.linkItem(atPath: "foo", toPath: "bar")
             XCTAssertEqual($0.delegateCaptures.shouldLink, [.init("foo", "bar")])
+            #if os(Android) // Hard links are not normally allowed on Android.
+            XCTAssertEqual($0.delegateCaptures.shouldProceedAfterLinkError, [.init("foo", "bar", code: .fileWriteNoPermission)])
+            XCTAssertFalse($0.fileExists(atPath: "bar"))
+            #else
             XCTAssertEqual($0.delegateCaptures.shouldProceedAfterLinkError, [])
             XCTAssertTrue($0.fileExists(atPath: "bar"))
+            #endif
         }
         
         try FileManagerPlayground {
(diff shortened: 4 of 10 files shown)


Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
```diff
@@ -173,7 +173,7 @@ struct TimeZoneCache : Sendable {
                 }
             }
             
-#if os(Linux) && !os(WASI)
+#if os(Linux)
```
finagolfin, line 176: @kateinoigakukun, I assume the `os` can't be both `Linux` and `WASI`?

kateinoigakukun: Right, just a `os(Linux)` should be enough.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -956,6 +956,8 @@ extension _FileManagerImpl {
                 #if os(WASI)
                 // WASI does not support extended attributes
                 throw CocoaError.errorWithFilePath(.featureUnsupported, path)
+                #elseif canImport(Android)
+                // Android doesn't allow setting this for normal apps, so just skip it.
```
jmschonfeld, line 960: If Android doesn't allow setting this, should we throw an error similarly to WASI, either `.featureUnsupported` or perhaps an error indicating that the operation isn't allowed?

finagolfin: I considered simply enabling the WASI error block above on Android too, and would have if this method were just setting these extended attributes alone, but this `setAttributes()` method is also setting a bunch of other attributes and works fine for that on Android.

I don't know the situation on WASI but pretty much all files on Android appear to have some preset extended attributes, but normal users can't change them. Given that, it makes sense to just skip setting them on Android, whereas maybe WASI normally doesn't have them or something, so the error above is rare on that platform.

finagolfin:
Updated to allow reading from zero-sized system files on Android too and imported the Android overlay in a test file that now needs it and rebased, tested with the Sep. 4 trunk source snapshot natively on Android without a problem. I was hoping to get confirmation from TBC guys, who are preparing their own Android SDK, but no word from them for the last month on this pull, so might as well move ahead.

@parkera, just need someone to review and get this in, then I'd like to get this into release/6.0 next. Should be perfectly safe as this only affects Android, touches no code on any other platform.

parkera:
@swift-ci test

Tests/FoundationEssentialsTests/DataIOTests.swift
```diff
@@ -238,7 +238,7 @@ class DataIOTests : XCTestCase {
     }
     
     func test_zeroSizeFile() throws {
-        #if !os(Linux)
+        #if !os(Linux) && !os(Android)
```
jmschonfeld, line 241: We might have discussed this before, but is there any semantic difference between the `os(Android)` used here in the test files, and the `canImport(Android)` used in the implementation?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -331,6 +335,7 @@ final class FileManagerTests : XCTestCase {
     }
     
     func testLinkFileAtPathToPath() throws {
+        #if !os(Android) // Hard links are not normally allowed on Android.
```
jmschonfeld, line 338: I don't think I see anything in the implementation that's specific to android with this function (unless that was in a previous PR). Are we missing an Android check in the implementation to throw? I think rather than avoiding the test if that's the case maybe it makes sense to `XCTAssertThrowsError` here?

jmschonfeld (commented):
Overall LGTM, but a few small comments

Tests/FoundationEssentialsTests/DataIOTests.swift
finagolfin: None, I simply followed the convention I've seen elsewhere in the toolchain source of using `canImport(randomOS)` when invoking platform-specific APIs and `os(randomOS)` when separating actual OS differences. Since this `/proc` behavior is in the OS, I used `os(Android)` here. I did not go changing conditional compilation checks for other non-Android OS's that may be doing this wrong to use the same convention though.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -23,6 +23,10 @@ import TestSupport
 @testable import Foundation
 #endif
 
+#if canImport(Android)
+import Android
+#endif
```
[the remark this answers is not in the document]
finagolfin: This is the compilation error I get without it:
```
Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift:622:15: error: cannot find 'getuid' in scope
620 |     func testFileAccessAtPath() throws {
621 |         #if !os(Windows)
622 |         guard getuid() != 0 else {
    |               `- error: cannot find 'getuid' in scope
```
There are some strange issues with these platform overlays, [which I investigated a bit last month](https://github.com/apple/swift-certificates/pull/183#discussion_r1730366251). Since nobody I asked seems to have a good idea of what's going on, I chalked it up to some Swift module leak that can be fixed later and am just adding these imports now to shut the compiler up.

Tests/FoundationEssentialsTests/DataIOTests.swift
jmschonfeld: Ah ok, sounds good - as long as we have a reasoning for it and they're semantically equivalent, this seems fine to me

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jmschonfeld: If calling `link()` doesn't work on Android, I think we should replace that call in the implementation to something that will work on Android, or if Android does not have that concept then we should probably throw to indicate that it did not succeed. That way, either the test will pass as-is or we can update the test to check for the thrown error on Android

finagolfin: [Android doesn't allow hard linking on most partitions for regular users for the last decade](https://stackoverflow.com/questions/32365690/is-android-m-not-allowing-hard-links). Since it may work in those few remaining scenarios, I didn't really want to think about it and just disabled the test.

Since it will usually fail, I could replace it with this if you'd like instead:
```
#if os(Android)
XCTAssertEqual($0.delegateCaptures.shouldProceedAfterLinkError, [.init("foo", "bar", code: .fileWriteNoPermission)])
XCTAssertFalse($0.fileExists(atPath: "bar"))
#else
... current passing asserts ...
```

finagolfin: As discussed, added those Android-specific asserts to this pull

finagolfin:
Addressed all review feedback and updated, one last CI run and this can go in.

jmschonfeld:
@swift-ci please test

finagolfin:
Passed CI, ready for merge. I will submit for 6.0 next.

jmschonfeld (approved):

PR #873, opened 2024-08-18
Remove the `#include <setjmp.h>` from `_CStdlib.h`

kateinoigakukun:
It seems like we don't use neither setjmp nor longjmp in the Foundation codebase, so we can safely remove this include.
The main motivation for this change is to fix the build with the latest wasi-libc version. Recent wasi-libc versions started to provide `setjmp.h` header but it raises compilation errors unless Exception Handling feature (still under standardization process) is enabled.

See https://github.com/WebAssembly/wasi-libc/blob/wasi-sdk-22/libc-top-half/musl/include/setjmp.h

### DIFF
--- a/Sources/_FoundationCShims/include/_CStdlib.h
+++ b/Sources/_FoundationCShims/include/_CStdlib.h
@@ -55,10 +55,6 @@
 #include <math.h>
 #endif
 
-#if __has_include(<setjmp.h>)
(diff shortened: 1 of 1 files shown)


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #875, opened 2024-08-19
URL.host should not return percent-encoded host

### LINKED ISSUE #863: URL Parsing behaviour changed on Linux between 5.10 and 6.0
When attempting to parse the string `"ftp://user:password@*.xn--poema-9qae5a.com.br:4343/cat.txt"` as a URL, the result is different between 5.10 and 6.0 on Linux, and on Apple platforms as well. The difference is at the very least in the host component.

This can be shown by using the following test program:

```
import Foundation

func main() {
    let x = URL(string: "ftp://user:password@*.xn--poema-9qae5a.com.br:4343/cat.txt")!
    print(x.host!)
}

main()
```

- On Linux, Swift 5.10 (`Swift version 5.10.1 (swift-5.10.1-RELEASE)`), the result of this program is `*.xn--poema-9qae5a.com.br`.
- On Linux, Swift 6.0 (`Swift version 6.0-dev (LLVM 3751470251df3e4, Swift 026ffddec66fd08)`), the result of the program is `*.p%C3%A3oema%C3%A7%C3%A3.com.br`.
- On the latest macOS Sequoia 15.1 beta, the result is `*.xn--poema-9qae5a.com.br`.

Note that on Linux, 6.0, the URL type seems to have decided to thoroughly mangle the hostname. Specifically, it appears to have decoded the IDNA-encoded hostname, and then _percent-encoded_ it when we got it back out. That can't possibly be the right thing to do.

jrflat:
Before this change, `URL.host` would just call `.host()`, which defaults to `.host(percentEncoded: true)`. This meant that if the host was IDNA-encoded, `.host` would decode the IDNA-encoding and then percent-encode, returning the percent-encoded string.

This PR fixes `.host` to match the behavior before the Swift URL implementation:
- If the host was percent-encoded, return the percent-decoded host
- If the host was IDNA-encoded (or not encoded at all), return that host string

This is exactly the behavior of `.host(percentEncoded: false)`, which is now used instead of `.host()`.

Resolves #863 (thanks @Lukasa!)

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1188,7 +1188,7 @@ public struct URL: Equatable, Sendable, Hashable {
             return _url.host
         }
         #endif
-        return host()
+        return host(percentEncoded: false)
     }
 
     /// Returns the host component of the URL if present, otherwise returns `nil`.
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -622,6 +622,11 @@ final class URLTests : XCTestCase {
         XCTAssertEqual(url.path(), "/my/non/existent/path")
     }
 
+    func testURLHostRetainsIDNAEncoding() throws {
+        let url = URL(string: "ftp://user:password@*.xn--poema-9qae5a.com.br:4343/cat.txt")!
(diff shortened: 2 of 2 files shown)


jrflat:
@swift-ci please test

Lukasa:
No, thank _you_ @jrflat! 🎉

parkera (approved):

jmschonfeld (approved):

PR #878, opened 2024-08-20
CMake: Fixing cross-compiling Swift-Foundation

etcwilde:
The macros must build for the local machine running the build, not the machine that Swift-Foundation will run on, in order for the build to be able to use the macros.

To do this, the macro build must use ExternalProject, which treats the child project as an entirely independent project. One cannot introspect into the external project since it is configured at build time, rather than at configure time. This is what allows the external project to build for another platform.

The expectation is that the calling project will pull the built products of the ExternalProject from the install location. EPs have an internal implicit install prefix where they can install stuff to without dirtying the building machine or installed products, so we can make use of that. In order for that to work, the products must actually get installed though, so we have to install the FoundationMacros, even when built as an executable. To support that, I've exposed an option to tell the macro build to build the macros as an executable.

On the library side, I've exposed the Foundation macros as an interface library that only exposes the `-load-plugin-path` option needed for picking up the macro. Linking against this interface library will load the plugin as desired.

This results in a build that
   - can use macros, even when cross-compiling.
   - does not install the macros into the installed library, only to the build directory.


etcwilde:
@swift-ci please test

Sources/CMakeLists.txt
```diff
@@ -12,6 +12,31 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+if(NOT SwiftFoudation_MACRO)
+  include(ExternalProject)
```
jmschonfeld, line 16: The rest of our CMake files use 4-space indents, could we update to do the same here?

Sources/CMakeLists.txt
```diff
@@ -12,6 +12,31 @@
...
+  # The macros are required for building Swift-Foundation. Build them for the
+  # build machine.
+  ExternalProject_Add(FoundationMacros
+    SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/FoundationMacros"
+    INSTALL_DIR "${CMAKE_CURRENT_BINARY_DIR}/FoundationMacros"
+    CMAKE_ARGS
+      -DCMAKE_BUILD_TYPE=Release
+      -DCMAKE_Swift_COMPILER=${CMAKE_Swift_COMPILER}
+      -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
```
jmschonfeld, line 25: I haven't seen the `<INSTALL_DIR>` syntax before, out of curiosity, what does it mean?

Sources/FoundationEssentials/CMakeLists.txt
```diff
@@ -55,21 +55,7 @@ if(SwiftFoundation_MACRO)
     target_compile_options(FoundationEssentials PRIVATE
```
jmschonfeld, line 55: To keep the macro logic together, what do you think about moving this case up to `Sources/CMakeLists.txt` alongside the `ExternalProject` directives, and just creating an interface library that has this path? That way clients don't need to conditionalize on `SwiftFoundation_MACRO` and here we can just unconditionally link `FoundationBuildMacros`?

Sources/FoundationEssentials/CMakeLists.txt
```diff
@@ -55,21 +55,7 @@ if(SwiftFoundation_MACRO)
...
-    add_dependencies(FoundationEssentials FoundationMacros)
-    get_target_property(MacroDIR FoundationMacros RUNTIME_OUTPUT_DIRECTORY)
-    if(CMAKE_SYSTEM_NAME STREQUAL Windows)
-        set(MacroExecutable "${MacroDIR}/FoundationMacros.exe#FoundationMacros")
-    else()
-        set(MacroExecutable "${MacroDIR}/FoundationMacros#FoundationMacros")
-    endif()
-    target_compile_options(FoundationEssentials PUBLIC
-        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-load-plugin-executable ${MacroExecutable}>")
+  target_link_libraries(FoundationEssentials PRIVATE FoundationBuildMacros)
```
jmschonfeld, line 58: Does this need to be `PUBLIC` so that the `FoundationInternationalization` module (which imports `FoundationEssentials`) also gets the build path applied?

Sources/CMakeLists.txt
compnerd: This is a special placeholder that will be filled in with the install path. This is documented in the `ExternalProject` documentation IIRC.

Sources/FoundationEssentials/CMakeLists.txt
etcwilde: Hmm, not sure. We definitely don't want the builder-macros bleeding out of the build of the project since the file path is specific to this build, as is the build of the macro executable itself.

I've mostly been thinking of the macros as independent "libraries" from the library that exposes them, which is maybe wrong? Not sure. If they are part of that library's interface, that would make sense. Is the macro exposed as part of the interface of FoundationEssentials, while also being used to build FoundationEssentials, or is it only one or the other?

jmschonfeld: > Is the macro exposed as part of the interface of FoundationEssentials, while also being used to build FoundationEssentials

Yes, the

```
@freestanding(expression)
@available(FoundationPredicate 0.1, *)
public macro Predicate<each Input>(_ body: (repeat each Input) -> Bool) -> Predicate<repeat each Input> = #externalMacro(module: "FoundationMacros", type: "PredicateMacro")
```

Declaration is a public declaration exposed to all clients of `FoundationEssentials`. We can use `#Predicate` in `FoundationEssentials` itself which requires using it to build `FoundationEssentials`, but also clients need to use the macro. So if I'm understanding your question correctly yes the macro is both part of what's required to build FoundationEssentials itself as well as part of its interface

etcwilde: We'll have to include the macros in the exports, but yeah, that sounds like `PUBLIC` is the right visibility then.

Sources/CMakeLists.txt
etcwilde: Fixed

Sources/FoundationEssentials/CMakeLists.txt
etcwilde: Fixed

etcwilde:
@swift-ci please test

Sources/CMakeLists.txt
```diff
@@ -12,6 +12,36 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+add_library(FoundationBuildMacros INTERFACE)
+export(TARGETS FoundationBuildMacros FILE FoundationBuildMacros.cmake)
+if(SwiftFoudation_MACRO)
+    message(STATUS "SwiftFoundation_MACRO provided, using macros in ${SwiftFoundation_MACRO}")
+    target_compile_options(FoundationBuildMacros INTERFACE
+        "SHELL:$<BUILD_INTERFACE:-plugin-path ${SwiftFoudation_MACRO}>")
```
jmschonfeld, line 20: Does this need to indicate `$<COMPILE_LANGUAGE:Swift>`?

jmschonfeld (commented):
One minor comment, otherwise this looks good to me. Before merging, lets run a cross-repo CI test to get a full toolchain test going with this since this repo's CI doesn't test the cmake build yet

etcwilde:
@swift-ci please test

jmschonfeld (approved):
Approved assuming cross-repo test comes back clean, thanks for updating this!

jmschonfeld:
> -- SwiftSyntax_DIR not provided, checking out local copy of swift-syntax

Looks like we need to make sure `SwiftSyntax_DIR` gets passed along to the separate macro build

etcwilde:
> Looks like we need to make sure SwiftSyntax_DIR gets passed along to the separate macro build

Except that was built for the machine we're building the toolchain for, not the machine we're building the toolchain on.

jmschonfeld:
Actually I got my wires crossed - in the toolchain build, swift-foundation should already receive `SwiftFoundation_MACROS` from the build script so it shouldn't ever hit the `ExternalProject` path, right?

Sources/CMakeLists.txt
```diff
@@ -12,6 +12,36 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+add_library(FoundationBuildMacros INTERFACE)
+export(TARGETS FoundationBuildMacros FILE FoundationBuildMacros.cmake)
+if(SwiftFoudation_MACRO)
```
jmschonfeld, line 17: Aha, I think this is the problem

```suggestion
if(SwiftFoundation_MACRO)
```

Sources/CMakeLists.txt
```diff
@@ -12,6 +12,36 @@
...
+    message(STATUS "SwiftFoundation_MACRO provided, using macros in ${SwiftFoundation_MACRO}")
+    target_compile_options(FoundationBuildMacros INTERFACE
+        "SHELL:$<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:Swift>:-plugin-path ${SwiftFoudation_MACRO}>>")
```
finagolfin, line 20: Here too

etcwilde:
@swift-ci please test

Sources/CMakeLists.txt
etcwilde: Doh! Good catch. Thanks.

etcwilde:
@swift-ci please test

finagolfin:
Is this really necessary? When cross-compiling this repo on the CI, you use a current freshly-built host toolchain from the same source, which already has these macros compiled and installed for the host. I haven't had any problem cross-compiling these new Foundation repos that way.

What I'd like instead is an option to disable cross-compiling these macros too, which is only useful for cross-compiling the toolchain, but not for standalone SDKs. For example, I'm cross-compiling an Android SDK using `build-script` right now and had to [manually patch out cross-compiling these macros](https://github.com/finagolfin/swift-android-sdk/pull/171/files#diff-cbcc6ddabac6443cf0d596be79c7df906e2ae24ad569fda713876e00bf64a3a7R27) on there.

Btw @etcwilde, there appears to be a problem with building for Swift with the current CMake 3.30. You [added the `CMAKE_SHARED_MODULE_CREATE_Swift_FLAGS` when building shared libraries](https://gitlab.kitware.com/cmake/cmake/-/commit/56e5cea6008397f83479952d4c59964c32bc35fd#6f3fc16cba3bde22680edfd232f2e1ac93f00aff), then set it for `Apple-Swift` alone later in that commit, which likely worked fine because that variable was uninitialized on other platforms.

However, a little later [Brad default-initialized that variable for all languages to the C flags](https://gitlab.kitware.com/cmake/cmake/-/commit/ae1d5ef47fd4175a31f8bbfc8957cd6cb26d5e75#1d2a5ac04fbd802cdf0e72758f69624ff8b1b3c4_24_107), which appears to have broken linking Swift libraries whenever the C flags are set on non-Apple platforms, by passing in [C linker flags like `-Wl,--build-id=sha1` to the Swift compiler](https://github.com/finagolfin/swift-android-sdk/actions/runs/10978697104/job/30482150707#step:8:3870).

compnerd (approved):

etcwilde:
@swift-ci please test

PR #879, opened 2024-08-20
Remove RegexBuilder conditional in Package manifest

jmschonfeld:
Originally, we added this conditional because `RegexBuilder` was not available on Windows. However CI has been successfully running these tests for some time now, so we no longer need the conditional. This simplifies the manifest a bit to remove the unnecessary logic.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #880, opened 2024-08-21
Decimal bin compat: remove the usage of @_cdecl from 'public' methods defined in NSDecimal.h

iCharlesHu:
Rationale: @_cdecl emits a C symbol that calls back to the Swift symbol. These new swift symbols does not exist on older OSs, causing a crash due to sym bol not found. Instead of using @_cdecl to implement public methods directly, use it on an internal type so it will not introduce unexpected public symbols.

resolves: rdar://133371820


iCharlesHu:
@swift-ci please test

itingliu (approved):

PR #883, opened 2024-08-21
URLComponents: support http(s)+unix schemes

jrflat:
The common (but not standardized) way to represent an "HTTP over Unix socket" URL is to use the scheme `http+unix` and set the URL's host to the percent-encoded path of the socket, e.g.

`http+unix://%2Fvar%2Frun%2Fdocker.sock/info`

Before the swift-corelibs-foundation re-core, `URLComponents` would always percent-encode the `.host`, so this worked as expected. After the re-core, `URLComponents` now supports and favors IDNA-encoding of the host, breaking this use case.

This PR special-cases the `http+unix` and `https+unix` schemes to percent-encode the host like before.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLParser.swift
+++ b/Sources/FoundationEssentials/URL/URLParser.swift
@@ -222,6 +222,8 @@ internal struct RFC3986Parser: URLParserProtocol {
         "addressbook",
         "contact",
         "phasset",
+        "http+unix",
+        "https+unix",
     ])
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jrflat:
Fixed conflicts

jrflat:
@swift-ci please test

parkera (approved):

PR #884, opened 2024-08-21
Formatting performance improvements

parkera:
Re-apply the diff from #719 (reverted in #744). We discovered the root cause was use of ICU's ability to set the malloc functions used, which can race. We will be disabling those in swift-foundation-icu soon. In the meantime, we know that they are not used on Linux so this patch is safe.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #890, opened 2024-08-29
Re-enable disabled tests due to ICU-74

iCharlesHu:
We disabled some tests due to changes to ICU74. Now we upgraded ICU we should reenable these tests.

resolves: rdar://121399858


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #892, opened 2024-08-30
Change @_nonSendable to unavailable extensions

jmschonfeld:
`@_nonSendable` was designed as a detail for the compiler to use and was not intended to be written directly in source. As such, the compiler treats it slightly differently and has allowed us to commit some incorrect annotations, such as:

```swift
@_nonSendable
open class FileManager : @unchecked Sendable {
```
(a both non-sendable and unchecked-sendable class)

and

```swift
@_nonSendable
public struct KeyPathComparator<Compared>: SortComparator {
```
(a non-sendable struct that conforms to a `Sendable` protocol)

This PR replaces usages of `@_nonSendable` with the officially supported design (unavailable extensions). This has the drawbacks of needing to duplicate the type's availability and moving the non-sendable annotation away from the type itself, but it means that both of the errors above would be diagnosed as actual compilation errors rather than silently being accepted and potentially causing issues for clients.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeScope.swift
@@ -22,10 +22,16 @@ public protocol AttributeScope : DecodingConfigurationProviding, EncodingConfigu
     static var encodingConfiguration: AttributeScopeCodableConfiguration { get }
 }
 
-@_nonSendable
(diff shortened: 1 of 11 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #894, opened 2024-09-03
ProcessInfoTests.testOperatingSystemIsAtLeastVersion should not run on visionOS

iCharlesHu:
visionOS's version number is indeed lower than 5.

resolves: rdar://134731230


iCharlesHu:
@swift-ci please test

itingliu (approved):

PR #895, opened 2024-09-03
Implement negative power support to pow(_ x: Decimal, _ y: Int). 

iCharlesHu:
Historically Decimal has never supported negative power and will reproduce unexpected results

resolves: rdar://133875543


iCharlesHu:
@swift-ci please test

PR #896, opened 2024-09-03
Set `SWIFT_SYSTEM_NAME` for macro project in CMake

jmschonfeld:
When `FoundationMacros` was moved into its own CMake project, we did not copy over the `SWIFT_SYSTEM_NAME` setting. This results in the rpath for `libFoundationMacros.so` being `$ORIGIN/../../../swift/:$ORIGIN/..` instead of the correct value (`$ORIGIN/../../../swift/linux:$ORIGIN/..` for linux, for example). This fixes the macro build which allows `#Predicate` and `#Expression` to work as expected.

(note for context that this set value is used on line 90 of this file: `INSTALL_RPATH "$ORIGIN/../../../swift/${SWIFT_SYSTEM_NAME}:$ORIGIN/.."`)

### DIFF
--- a/Sources/FoundationMacros/CMakeLists.txt
+++ b/Sources/FoundationMacros/CMakeLists.txt
@@ -26,6 +26,14 @@ endif()
 project(FoundationMacros
   LANGUAGES Swift)
 
+if(NOT SWIFT_SYSTEM_NAME)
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #898, opened 2024-09-03
[6.x] Set SWIFT_SYSTEM_NAME for macro project in CMake

jmschonfeld:
Explanation: A configuration change that updates the rpath value of `libFoundationMacros.so` to be correct
Scope: Only impacts the rpath value of `libFoundationMacros.so`
Original PR: https://github.com/apple/swift-foundation/pull/896
Risk: Low - scope is narrow, and fixes a regression that causes the macro to not be loadable so it can't really make anything worse :)
Testing: Testing done via local testing
Reviewer: @parkera

### DIFF
--- a/Sources/FoundationMacros/CMakeLists.txt
+++ b/Sources/FoundationMacros/CMakeLists.txt
@@ -26,6 +26,14 @@ endif()
 project(FoundationMacros
   LANGUAGES Swift)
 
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #900, opened 2024-09-04
Properly `throw` in case of numeric overflow in Calendar mathematics

itingliu:
We have applied multiple different strategies with regard to overflow during calendrical calculation, including capping input `Date` and clipping Julian day calculation with artificial bounds. Those are still insufficient as there are too many paths where numeric overflow may happen.

This patch patches those places by `throw`ing whenever applicable. Also adds fuzzing tests to cover many more Calendar API.

Resolves rdar://133558250

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
@@ -3223,7 +3223,7 @@ final class GregorianCalendarTests : XCTestCase {
     }
 
     // MARK: - Day Of Year
-    func test_dayOfYear() {
+    func test_dayOfYear() throws {
         // An arbitrary date, for which we know the answers
         let date = Date(timeIntervalSinceReferenceDate: 682898558) // 2022-08-22 22:02:38 UTC, day 234
         let leapYearDate = Date(timeIntervalSinceReferenceDate: 745891200) // 2024-08-21 00:00:00 UTC, day 234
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -3257,10 +3257,10 @@ final class GregorianCalendarTests : XCTestCase {
         XCTAssertEqual(cal.range(of: .dayOfYear, in: .year, for: leapYearDate), 1..<367)
 
         // Addition
-        let d1 = cal.add(.dayOfYear, to: date, amount: 1, inTimeZone: .gmt)
+        let d1 = try? cal.add(.dayOfYear, to: date, amount: 1, inTimeZone: .gmt)
```
jmschonfeld, line 3260: If we don't expect this to throw, could we mark the test as `throws` and just do `try` here? That'd mean that if an error is thrown, the test would fail and the failure would indicate the error information rather than just a `nil` value for `d1` which would probably help with debugging

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -1220,7 +1220,34 @@ final class CalendarTests : XCTestCase {
...
+            let components = DateComponents(year: 9223372036854775556)
+            let added = calendar.date(byAdding: components, to: date)
+            XCTAssertNil(added)
+        }
+    }
+
+    func test_dateComponentsFromDateOverflow() {
+        let calendar = Calendar(identifier: .gregorian)
+        let dc = calendar.dateComponents([.year], from: Date(timeIntervalSinceReferenceDate: Double(Int64.max)))
+
```
jmschonfeld, line 1246: Just confirming - was there anything else that needed to be here or is this just testing that this doesn't crash?

iCharlesHu (approved):

Tests/FoundationInternationalizationTests/CalendarTests.swift
itingliu: Thanks. Updated

jmschonfeld (approved):

PR #902, opened 2024-09-04
[6.0] URL.host should not return percent-encoded host

jrflat:
**Explanation:** Fixes `URL.host` to match the behavior before the Swift URL implementation:
- If the host was percent-encoded, return the percent-decoded host
- If the host was IDNA-encoded (or not encoded at all), return that host string

**Scope:** Only impacts deprecated (but still used) `URL.host` for URLs with an encoded host, restoring previous behavior
**Original PR:** https://github.com/apple/swift-foundation/pull/875
**Risk:** Low - minimal scope, one line change to restore behavior
**Testing:** Local, swift-ci, stable in `main` for 2 weeks
**Reviewer:** @parkera @jmschonfeld

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1188,7 +1188,7 @@ public struct URL: Equatable, Sendable, Hashable {
             return _url.host
         }
         #endif
-        return host()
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

PR #907, opened 2024-09-09
Benchmarks for URL and URLComponents

jrflat:
Initial performance benchmarks for `URL` and `URLComponents`. Results look great, particularly between Swift 5.10 and 6.0 toolchains (old `swift-corelibs-foundation` vs new `swift-foundation` implementations). I tested on the latest Ubuntu 22.04 Swift 5.10 and nightly-6.0 images. (`swift-6.0-DEVELOPMENT-SNAPSHOT-2024-08-22-a`)

From 5.10 to 6.0, we see a **5-10x** speedup for almost all benchmarks, with even better outliers for `URL-ParseValidASCII` (**29x!**) and `URL-ParseInvalid` (**94x!**), as well as a whopping **254x** speedup for `URLComponents-GetComponentRanges` (which is probably due to an implementation difference since we now store the Swift `Range<String.Index>` directly). These results were consistent across many runs.

I think the 29x speedup of `URL(string:)` for valid URL strings is the most significant improvement because that's the most common use case.

URLBenchmarks Ubuntu 22 04 Swift 5 10 vs 6 0

### DIFF
--- a/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
+++ b/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
@@ -1,0 +1,234 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

PR #908, opened 2024-09-09
Migrate swift-foundation to swift-testing

jmschonfeld:
This PR contains the changes to migrate swift-foundation to swift-testing. All swift-foundation tests have been fully migrated to swift-testing with this change. A few notes about the changes here:

- The existing TestSupport module was removed and replaced with a re-purposed module to just contain a few utility helpers. It no longer contains `expectEqual`-like wrappers around XCTest and it no longer contains a list of type aliases
- `FoundationEssentialsTests` no longer links `FoundationInternationalization`. As a result, some tests needed to be moved from `FoundationEssentialsTests` to `FoundationInternationalizationTests` because they required having the i18n module loaded
- swift-testing runs tests in parallel by default, so some tests were updated to no longer rely on the current `Locale`/`Calendar`/`TimeZone`, and tests that did need to use these or other global state (like the CWD) were moved into nested serial suites (see `FilePlaygroundTests` as an example)
- We use the toolchain-provided version of swift-testing: this was done to avoid regressions to our test build time, and swift-foundation tests already require a pre-built toolchain with SwiftPM present so we are guaranteed to have a toolchain with swift-testing at that time


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
All tests now pass in our CI, just waiting for confirmation that tests pass when run in the full toolchain suite before merging

itingliu (approved):

jmschonfeld:
Closing as this has been completed by a series of other PRs and is now just about complete

PR #912, opened 2024-09-10
Adapt some of the original swift-corelibs-foundation 5.10 implementation to regain performance in JSONEncoder

kperryua:


Sources/FoundationEssentials/JSON/JSONWriter.swift
```diff
@@ -131,134 +69,165 @@ internal struct JSONWriter {
     private let sortedKeys: Bool
     private let withoutEscapingSlashes: Bool
 
-    var data = Data()
+    var bytes = [UInt8]()
```
fabianfett, line 72: reserve some space here. otherwise the first `{` is very expensive. I think we should always assume at least 256bytes of data? After that the reallocs will be very, very rare.

PR #918, opened 2024-09-17
URLComponents: support ws(s)+unix schemes

jrflat:
Similar to https://github.com/apple/swift-foundation/pull/883, the common way to represent a "WebSocket over Unix socket" URL is to use the scheme `ws+unix` and set the URL's host to the percent-encoded path of the socket.

This PR adds `ws+unix` and `wss+unix` to the set of schemes where we percent-encode the host instead of IDNA-encode.


jrflat:
@swift-ci please test

parkera (approved):

PR #919, opened 2024-09-17
PropertyListEncoder: Reference the property on the class itself instead of the options struct

parkera:
Use the `var` on `PropertyListEncoder` instead of the options struct directly, to support overrides in subclasses.


parkera:
@swift-ci test

kperryua:
Appears a type mismatch is causing a compiler error.

parkera:
Yah, that should be `PropertyListDecoder.PropertyListFormat` actually (which is a type alias for each platform).

parkera:
@swift-ci test

kperryua (approved):

parkera:
@swift-ci test

PR #927, opened 2024-09-19
String+Path parsing performance improvements

jrflat:
Improved some String+Path functions' performance with the following changes:

- Used `.utf8` where applicable (such as when searching for ascii `UInt8._slash`, which happens a lot)
- Refactored some path components functions, fixed an issue where `.deletingPathComponent` didn't skip over consecutive trailing slashes
- Refactored `.removingDotSegments` to directly modify the String buffer
- Replaced automount prefix matching with `UInt8` arrays

Non-implementation related:

- Replaced my previous ambiguous use of a `var input` with `utf8` to avoid confusion
- Used `UInt8._dot` and `._slash` instead of re-defining them in-function
- Made state enum names more specific to avoid potential naming conflicts
- Renamed helper functions for clarity
- Added unit tests

Results: We generally see **2-6x** speedup in the respective functions.

Old vs New `String`

String+Path Performance Old vs New

`NSString` vs `String`

String+Path Performance NSString vs String

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -27,7 +27,7 @@ import WASILibc
 internal import _FoundationCShims
 
 extension StringProtocol {
-    fileprivate func _standardizingSlashes() -> String {
+    fileprivate func _convertingSlashesIfNeeded() -> String {
         #if os(Windows)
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #931, opened 2024-09-19
Fix windows creation of relative symlinks to directories

jmschonfeld:
On windows, we need to specify an extra flag if we are creating a symlink to a directory. To do this, we check if the provided destination is a directory using `FileManager.fileExists(atPath:isDirectory:)`. However, the destination path is relative to the _symlink_ and not relative to the current working directory like `fileExists(atPath:)` expects so we must first convert the relative path into an absolute path using the symlink path as a base.

Resolves https://github.com/apple/swift-foundation/issues/928

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
@@ -59,7 +59,8 @@ extension _FileManagerImpl {
     ) throws {
 #if os(Windows)
         var bIsDirectory = false
-        _ = fileManager.fileExists(atPath: destPath, isDirectory: &bIsDirectory)
+        let absoluteDestPath = URL(filePath: destPath, relativeTo: URL(filePath: path, directoryHint: .notDirectory)).path
+        _ = fileManager.fileExists(atPath: absoluteDestPath, isDirectory: &bIsDirectory)
 
         try path.withNTPathRepresentation { lpSymlinkFileName in
             try destPath.withFileSystemRepresentation {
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
```diff
@@ -59,7 +59,8 @@ extension _FileManagerImpl {
     ) throws {
 #if os(Windows)
         var bIsDirectory = false
-        _ = fileManager.fileExists(atPath: destPath, isDirectory: &bIsDirectory)
+        let absoluteDestPath = URL(filePath: destPath, relativeTo: URL(filePath: path, directoryHint: .notDirectory)).path
```
compnerd, line 62: What happens if `path` is absolute?

jmschonfeld: If `path` (the symlink) is absolute, then the behavior doesn't change - the destination is interpreted as relative to that path (so if `path` is `/foo/bar` and `destPath` is `file`, then `absoluteDestPath` is `/foo/file`). If `destPath` (the destination) is absolute, then just the destination path is used (so if `destPath` is `/some/file` then regardless of what `path` is, `absoluteDestPath` is `/some/file`)

compnerd (approved):
LGTM, but the test case would be nice.

Sources/FoundationEssentials/FileManager/FileManager+SymbolicLinks.swift
jmschonfeld: Yeah I can see if I can form an absolute path based on the CWD for a test case, that's a fair point

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld:

iCharlesHu (approved):

PR #949, opened 2024-09-27
Remove unused Swift URL code

jrflat:
Removes old, unused Swift URL code so that only the new Swift implementation is left.


jrflat:
@swift-ci please test

PR #950, opened 2024-09-30
Create benchmarks for base64

fabianfett:
### Motivation

We love benchmarks. In order to ensure that our base64 performance does not regress we should have a bunch of benchmarks that test different aspects. Since we will have a bunch of those in the future lets get them into a separate target.

### Changes

- Create `Base64Benchmarks` target
- Add Base64 benchmarks
    - Use scaling factor one if the example data is sufficiently large (1MB)
    - Use scaling factor kilo if the example data is so small that measuring perfs overhead is to large (mostly in the jwt head examples)

### Result

We have baselines to test against.

### DIFF
--- a/Benchmarks/Benchmarks/Base64/Base64.swift
+++ b/Benchmarks/Benchmarks/Base64/Base64.swift
@@ -1,0 +1,200 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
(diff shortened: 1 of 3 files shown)


kperryua:
@swift-ci test

parkera:
@swift-ci test

parkera:
Oops - @kperryua already did that 😛

parkera (approved):

PR #945, opened 2024-09-25
 Follow up for dcd7a97: Trim an empty line

itingliu:

PR #937, opened 2024-09-20
Short-circuit logic to get locale that requires special case handling.

itingliu:
This fix targets at `__CFLocaleGetDoesNotRequireSpecialCaseHandling`. Currently, this function calls into `class _NSSwiftLocale`, then `struct Locale`, then `any LocaleProtocol` (`LocaleICU` here), which calls into a `static func` of `struct Locale`. All the hoop jumping contributes to a lot of retain and release calls.

It turns out that this function is only used by this one call site (`_CFStrGetSpecialCaseHandlingLanguageIdentifierForLocale`). This change simplifies the calling chain so that we call the static function directly from `_NSSwiftLocale`, and the result is cached inside this class directly.

I've verified that this change brings down the time spent in `CFStringCompareWithOptionsAndLocale` of the reproducable case in the radar down from 904ms to 397ms.

Also added a BenchmarkLocale target.

`CFStringCompareWithOptionsAndLocale` benchmark result **before** the change:

```
╒═══════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│ Metric                    │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞═══════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│ Malloc (total) *          │       0 │       0 │       0 │       0 │       0 │       0 │       0 │     448 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Throughput (# / s) (M)    │     163 │     153 │     150 │     146 │     144 │     129 │     123 │     448 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (total CPU) (ns) *   │       6 │       7 │       7 │       7 │       7 │       7 │       8 │     448 │
├───────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│ Time (wall clock) (ns) *  │       6 │       7 │       7 │       7 │       7 │       8 │       8 │     448 │
╘═══════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

And **after** the change:

```
[...]

PR #956, opened 2024-10-02
Trim the empty lines

lamtrinhdev:
Refer #945 of @itingliu that trim the empty lines at `Locale_ICU.swift`.


itingliu:
Are you seeing any issues with these empty new lines that you felt they have to be removed? 

#945 was a follow up for #937, which I had as a commit locally that I forgot to include in the same PR, and it introduced a merge conflict with another WIP branch. 

While we don't have clear guidelines that discourages formatting-only changes, I personally find it unworthy to overwrite git history unless there's a good reason for doing so.

lamtrinhdev:
Dear @itingliu ,

Thank you for your feedback. I am clear now. Let me close this ticket.

Thanks,
Lam

PR #959, opened 2024-10-02
Update to swift-syntax 600.0.0

jmschonfeld:
Now that swift-syntax 600.0.0 is released we can update our manifests to point at this tag. Additionally, this change removes a deprecation warning that was introduced with 600.0.0 and removes an availability check that is no longer needed.


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #960, opened 2024-10-02
Disable GregorianCalendarCompatibilityTests

jmschonfeld:
These tests are compatibility tests that are extensive and long running. Their goal is not to check each change but rather provide exhaustive compatibility validations and they're intended to be run at desk rather than routinely in CI. This disables the tests - they can be enabled locally to validate changes on-demand.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #961, opened 2024-10-03
URL(filePath:) should not treat "~" as absolute

jrflat:
Paths beginning with `~` should not be treated as absolute for `URL(filePath:)`. This is because when constructing a `URLComponents` for a file path we consider absolute, we do:

```
components.file = "file"
components.host = ""
components.path = path
```

This would create the string like `file://~/`. However, `URLComponents` rejects this string because the `~` is ambiguous--it looks like a host. This can lead `components.url!` to crash when force-unwrapping.

This PR changes the `isAbsolute` check to only consider `/` paths absolute. If the path starts with `~`, we will try to expand the tilde, then ensure that the result is absolute.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1688,41 +1688,47 @@ extension URL {
     /// Checks if a file path is absolute and standardizes the inputted file path on Windows
     /// Assumes the path only contains `/` as the path separator
     internal static func isAbsolute(standardizing filePath: inout String) -> Bool {
+        if filePath.utf8.first == ._slash {
+            return true
+        }
         #if os(Windows)
-        var isAbsolute = false
         let utf8 = filePath.utf8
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1688,41 +1688,47 @@ extension URL {
     /// Checks if a file path is absolute and standardizes the inputted file path on Windows
     /// Assumes the path only contains `/` as the path separator
     internal static func isAbsolute(standardizing filePath: inout String) -> Bool {
+        if filePath.utf8.first == ._slash {
+            return true
+        }
         #if os(Windows)
```
jrflat, line 1694: Refactored the Windows part and moved up the `.first == ._slash` check, which is valid for Windows and non-Windows platforms.

jmschonfeld (approved):

PR #965, opened 2024-10-07
URL path extension APIs should strip trailing slashes

jrflat:
When supplying a directory to `URL` like `URL(filePath: "/my/directory", directoryHint: .checkFileSystem)`, the initializer will automatically append a trailing slash, so the `.path()` is `/my/directory/`.

The current logic for `.appendingPathExtension("zip")` then directly appends the path extension, creating the unexpected path: `/my/directory/.zip`. A similar issue occurs where `.deletingPathExtension()` does not search before the trailing slash.

This PR restores the behavior from before the Swift `URL` implementation, which strips the trailing slash before appending or deleting the path extension, then appends a slash if the path was a directory. E.g. after appending the path extension, the resulting `.path()` is now `/my/directory.zip/` as expected.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -186,7 +186,11 @@ extension String {
         guard let lastDot = utf8.lastIndex(of: ._dot) else {
             return self
         }
-        return String(self[..<lastDot])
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #971, opened 2024-10-09
Guard old behavior in URL tests

jrflat:
Add guards to make these tests compatible with both the old and new `URL` implementation.


jrflat:
@swift-ci please test

parkera (approved):

jrflat:
The `.path` check wasn't correct on Windows since we now strip the leading slash from drive letter paths. Should be fixed now.

jrflat:
@swift-ci please test

Issue #966, opened 2024-10-07
[Windows] TimeZone.current returns GMT despite system settings


jmschonfeld:
Running `TimeZone.current` on Windows always returns GMT despite any present system settings.

This appears to be because on Windows we check `TZDEFAULT`: https://github.com/swiftlang/swift-foundation/blob/5141cf69577f26cdbabd65f6b12ff643b93f4685/Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift#L116-L131

However, `TZDEFAULT` does not appear to be set in the `WinSDK`/`ucrt` modules so we instead pickup the default value from our CShims: https://github.com/swiftlang/swift-foundation/blob/5141cf69577f26cdbabd65f6b12ff643b93f4685/Sources/_FoundationCShims/include/_CStdlib.h#L151-L157

These default paths seem unsuitable for Windows and result in us failing to detect the current timezone and falling back to GMT

jmschonfeld:
@compnerd do you happen to know if we'd expect `TZDEFAULT` to be defined for Windows somewhere, or do we need to instead do something like what CoreFoundation has historically done: https://github.com/swiftlang/swift-corelibs-foundation/blob/e55e1d88001997be830fbc01086564431d405dad/Sources/CoreFoundation/CFTimeZone.c#L723-L734

compnerd:
I don't expect `TZDEFAULT` to be set on Windows, we should be doing what CF did historically here.

PR #975, opened 2024-10-10
[Windows] Fix incorrect `TimeZone.current` lookup logic

jmschonfeld:
The current implementation of `TimeZone.current` attempts to read from `/etc/localtime` which is not a path that exists on Windows. This leads to `TimeZone.current` always being GMT on Windows. Instead, this updates the implementation to call `GetTimeZoneInformation`, convert the Windows name to an ICU system timezone name, and initialize a timezone with that name (this matches the behavior of `TimeZone` on Swift 5.10 when it was based on `CFTimeZone`). This also restricts the default values for `TZDEFAULT`/`TZDIR` to Darwin + Linux where the paths are reasonable unlike Windows where we know these constants should never exist to avoid accidental usage in the future.

There's likely more work to be done here for Windows for a full solution (to allow in some way for developers to interface with Windows time zone identifiers so that `TimeZone` can interop with Windows system APIs) but this change at least fixes the critical issue and restores the Swift 5.10 behavior, we can look into adding additional API for working with Windows time zone identifiers in the future.

Resolves https://github.com/swiftlang/swift-foundation/issues/966, rdar://137451317

### DIFF
--- a/Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
+++ b/Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
@@ -32,6 +32,13 @@ private func _timeZoneICUClass_localized() -> _TimeZoneProtocol.Type? {
 }
 #endif
 
+#if os(Windows)
+@_dynamicReplacement(for: _timeZoneIdentifier(forWindowsIdentifier:))
+private func _timeZoneIdentifier_ICU(forWindowsIdentifier windowsIdentifier: String) -> String? {
+    _TimeZoneICU.getSystemTimeZoneID(forWindowsIdentifier: windowsIdentifier)
+}
+#endif
+
 internal final class _TimeZoneICU: _TimeZoneProtocol, Sendable {
     init?(secondsFromGMT: Int) {
         fatalError("Unexpected init")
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
```diff
@@ -309,6 +316,23 @@ internal final class _TimeZoneICU: _TimeZoneProtocol, Sendable {
         return result
     }
 
+    #if os(Windows)
+    internal static func getSystemTimeZoneID(forWindowsIdentifier identifier: String) -> String? {
+        let timeZoneIdentifier = Array(identifier.utf16)
+        let result: String? = timeZoneIdentifier.withUnsafeBufferPointer { identifier in
+            return _withResizingUCharBuffer { buffer, size, status in
+                let len = ucal_getTimeZoneIDForWindowsID(identifier.baseAddress, Int32(identifier.count), nil, buffer, size, &status)
+                if status.isSuccess {
```
compnerd, line 325: return status.isSuccess ? len : nil feels like a better way to write this.

compnerd (approved):

Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
jmschonfeld: Yeah that seems fine too, I was just copying the format of the above `getCanonicalTimeZoneID` function for consistency

PR #984, opened 2024-10-16
[wasm] Fix WASI build around TimeZone

kateinoigakukun:
https://github.com/swiftlang/swift-foundation/pull/975 started to restrict the fallback value for `TZDIR` and it revealed that WASI platform implicitly depends on TZDIR even though it won't have such directory. This patch explicitly handles the case for WASI platform for timezone operations.


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

PR #985, opened 2024-10-17
Fix 'var currentDirectoryPath: String' crash when the directory is removed

ypli-chn:
Forced unpacking caused a crash.

```
let dest = "/tmp/a/b/c"
let fm = FileManager.default
try fm.createDirectory(atPath: dest, withIntermediateDirectories: true, attributes: nil)
let _ = fm.changeCurrentDirectoryPath(dest)
let _ = fm.currentDirectoryPath // it's ok
try fm.removeItem(atPath: dest)
let _ = fm.currentDirectoryPath // will crash
```
The same code in the old version(<5.10) of swift will be empty string.

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -729,9 +729,6 @@ final class FileManagerTests : XCTestCase {
             XCTAssertEqual(try $0.subpathsOfDirectory(atPath: ".").sorted(), ["bar", "dir", "dir/foo"])
             XCTAssertFalse($0.changeCurrentDirectoryPath("does_not_exist"))
             
-            XCTAssertThrowsError(try $0.removeItem(atPath: $0.currentDirectoryPath)) {
-                XCTAssertEqual(($0 as? CocoaError)?.code, .fileReadNoSuchFile)
-            }
             // Test get current directory path when it's parent directory was removed.
             XCTAssertTrue($0.changeCurrentDirectoryPath("dir"))
 #if os(Windows)
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -729,10 +729,14 @@ final class FileManagerTests : XCTestCase {
             XCTAssertEqual(try $0.subpathsOfDirectory(atPath: ".").sorted(), ["bar", "dir", "dir/foo"])
             XCTAssertFalse($0.changeCurrentDirectoryPath("does_not_exist"))
             
+            #if !os(Windows)
             // Test get current directory path when it's parent directory was removed.
+            // It's will be remove failed on Windows when directory is using.
             XCTAssertTrue($0.changeCurrentDirectoryPath("dir"))
             try $0.removeItem(atPath: $0.currentDirectoryPath)
             XCTAssertEqual($0.currentDirectoryPath, "")
+            #endif
+            
         }
     }
     
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -732,7 +732,7 @@ final class FileManagerTests : XCTestCase {
             // Test get current directory path when it's parent directory was removed.
             XCTAssertTrue($0.changeCurrentDirectoryPath("dir"))
 #if os(Windows)
-            // It's will be remove failed on Windows when directory is using.
+            // Removing the current working directory fails on Windows because the directory is in use.
             XCTAssertThrowsError(try $0.removeItem(atPath: $0.currentDirectoryPath)) {
                 XCTAssertEqual(($0 as? CocoaError)?.code, .fileWriteNoPermission)
             }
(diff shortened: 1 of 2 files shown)


ypli-chn:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -729,10 +729,14 @@ final class FileManagerTests : XCTestCase {
             XCTAssertEqual(try $0.subpathsOfDirectory(atPath: ".").sorted(), ["bar", "dir", "dir/foo"])
             XCTAssertFalse($0.changeCurrentDirectoryPath("does_not_exist"))
             
+            #if !os(Windows)
             // Test get current directory path when it's parent directory was removed.
+            // It's will be remove failed on Windows when directory is using.
```
jmschonfeld, line 734: I'm not sure I'm entirely following this comment, could you elaborate a bit more about why this behavior doesn't apply to Windows?

ypli-chn: image
After debugging, I always get an error here under Windows.
```
The process cannot access the file because it is being used by another process.
```

jmschonfeld: Ah I see, so Windows doesn't actually allow you to remove the current working directory it sounds like? If so, it'd be great if the unit test accounted for that behavior and verified that it throws an error with a particular code rather than limiting the test to only non-Windows. Would you be able to update the unit test to check for the thrown error on Windows and mention that behavior in the comment? Otherwise this change looks like it's on the right track, thanks!

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -728,6 +728,21 @@ final class FileManagerTests : XCTestCase {
             XCTAssertTrue($0.changeCurrentDirectoryPath(".."))
             XCTAssertEqual(try $0.subpathsOfDirectory(atPath: ".").sorted(), ["bar", "dir", "dir/foo"])
             XCTAssertFalse($0.changeCurrentDirectoryPath("does_not_exist"))
+            
+            XCTAssertThrowsError(try $0.removeItem(atPath: $0.currentDirectoryPath)) {
```
jmschonfeld, line 732: Thanks for updating, could you explain a bit about what this test is testing? Since the previous `changeCurrentDirectoryPath` failed I'd expect that `$0.currentDirectoryPath` is currently `..` here - is this removal attempting to test something in specific about `..` here?

ypli-chn: $0.currentDirectoryPath will only fail after removeItem is executed. In this case, $0.currentDirectoryPath called before removeItem, so there is no issue. Here, deleting $0.currentDirectoryPath removes the current folder, while `..` moves up to the parent directory. Both can actually achieve the same test purpose.

jmschonfeld: > In this case, $0.currentDirectoryPath called before removeItem, so there is no issue. Here, deleting $0.currentDirectoryPath removes the current folder

I think that's the source of my confusion with this particular line - the rest of the test seems ok, but it seems like this line asserts that deleting the parent directory fails, but I wouldn't expect that to be the case on Linux/Darwin where you can remove the current directory. Are you able to explain a little bit more about why this specific `removeItem(atPath:)` fails / what it is testing and how it is different from the `removeItem(atPath:)` which succeeds on non-Windows below?

ypli-chn: Ohhhhh...Sorry, my mistake. There was an issue when I was merging the code, but the problematic code has already been removed.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -728,6 +728,18 @@ final class FileManagerTests : XCTestCase {
             XCTAssertTrue($0.changeCurrentDirectoryPath(".."))
             XCTAssertEqual(try $0.subpathsOfDirectory(atPath: ".").sorted(), ["bar", "dir", "dir/foo"])
             XCTAssertFalse($0.changeCurrentDirectoryPath("does_not_exist"))
+            
+            // Test get current directory path when it's parent directory was removed.
+            XCTAssertTrue($0.changeCurrentDirectoryPath("dir"))
+#if os(Windows)
+            // It's will be remove failed on Windows when directory is using.
```
jmschonfeld, line 735:
```suggestion
            // Removing the current working directory fails on Windows because the directory is in use.
```

jmschonfeld (commented):
One minor wording suggestion about the comment, otherwise I think this change looks good to me! Adding @jrflat as a reviewer just in case you see any issue with the behavior of using the empty path for cases where the CWD doesn't exist on-disk

jmschonfeld:
@swift-ci please test

jrflat:
> One minor wording suggestion about the comment, otherwise I think this change looks good to me! Adding @jrflat as a reviewer just in case you see any issue with the behavior of using the empty path for cases where the CWD doesn't exist on-disk

Should be okay from `URL` side. The base URL generated for a relative path in `URL(filePath:)` will be `nil` whether `currentDirectoryPath` returns `nil` or `""`. This also potentially fixes a crash when calling `URL.currentDirectory()`.

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

Issue #921, opened 2024-09-18
Appending to an empty file path results in an absolute path


omochi:
**Problematic Code:**

```swift
import Foundation

let dir = URL(fileURLWithPath: "")
print(dir.path) // => /tmp/ee78646dddfd6304
let file = dir.appendingPathComponent("main.swift")
print(file.path) // => /main.swift
```

As shown in the example, even though `dir` holds the path to the current directory, appending a file to it results in an absolute path starting with a slash. This behavior is unexpected because appending to a current directory URL should ideally maintain a relative path, not convert it to an absolute one.

Issue #922, opened 2024-09-18
Removing the last path component of a standalone file path results in an empty path


omochi:
As described in the title, the following occurs:

```swift
import Foundation

let file = URL(fileURLWithPath: "main.swift")
let dir = file.deletingLastPathComponent()
print(dir.relativePath) // => 
print(dir.appendingPathComponent("main.swift").path) // => /main.swift
```

Since `main.swift` represents a relative path from the current directory, removing the last component should result in `.` to indicate the current directory. However, it incorrectly results in an empty path ``.

When combined with the bug #921, this behavior causes the path to change unexpectedly when re-adding the removed component, resulting in a different path from the original file.

On macOS, the result is `.` as expected, so re-adding the removed component restores the path to the original file.

jrflat:
Resolved by #989, thanks!

PR #988, opened 2024-10-18
URL: Appending to an empty file path results in an absolute path

jrflat:
When appending `path` to a URL with an empty relative path, the current code does: `"" + "/" + path`. This is problematic because it makes the relative path absolute, so it no longer resolves against the base URL.

E.g. for the following code,

```
let currentDir = URL(filePath: "")
print(currentDir.path) // "/my/cwd"
let file = currentDir.appending(path: "main.swift")
print(file.path) // "/main.swift"
```

we should get an output of `"/my/cwd/main.swift"`, not `"/main.swift"`.

This PR updates `URL.appending(path:)` so if the current path is empty (and thus relative), we don't append a slash which would make the path absolute. That is unless we have an authority. If we have an authority, we must add a slash to separate the path from the authority, e.g. `URL("http://example.com").appending(path: "path") == http://example.com/path`.

Resolves https://github.com/swiftlang/swift-foundation/issues/921 (thanks @omochi for filing!)

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -2214,9 +2214,8 @@ extension URL {
         var path = String(path)
         #endif
 
-        var newPath = relativePath()
         var insertedSlash = false
-        if !newPath.isEmpty && path.utf8.first != ._slash {
+        if !relativePath().isEmpty && path.utf8.first != ._slash {
             // Don't treat as first path segment when encoding
             path = "/" + path
             insertedSlash = true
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jrflat:
Thanks again for bringing this to our attention! I believe this also resolves #922 (the relative path will still be empty, but we will handle it correctly afterwards). I've opened #989 to fix some other special cases of `.deletingLastPathComponent()` as well

jrflat:
Updated the behavior slightly to be more compatible for clients expecting the relative path to start with `"./"`.

jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #993, opened 2024-10-21
Prevent crash when bridging an invalid CFURL to Swift URL

jrflat:
Bridging an invalid `CFURL` like `https://2001:db8::1:433` to `URL` in Swift causes a crash since the Swift parser returns `nil`. For compatibility, we can allow bridging to proceed by storing the string in a `URLParseInfo` with all `nil` ranges. This prevents the crash and allows `URL` to be a dummy container for the string, but otherwise will not return info that may be ambiguous.


jrflat:
@swift-ci please test

parkera (approved):
Let's cherry pick to release/6.0 as well

Issue #982, opened 2024-10-13
Crash When Streaming Files with Vapor - Related to swift-foundation


omrd:
After much discussion with @gwynne, it seems there's an issue with thread safety. Specifically, the problem is that `getgrgid(3)` is used in a manner that is not thread-safe. Possibley `getpwuid(3)` as well, for that matter.

The following backtrace shows an issue when accessing file properties
```
Debugging will be degraded due to missing types. Rebuilding the project will regenerate the needed module files.
* thread #2, name = 'omrd', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x13)
  * frame #0: 0x0000000004dff208 omrd`__libc_realloc + 168
    frame #1: 0x0000000004e131b8 omrd`getdelim + 244
    frame #2: 0x0000000004e0ef80 omrd`__getgrent_a + 76
    frame #3: 0x0000000004e0eaec omrd`__getgr_a + 132
    frame #4: 0x0000000004e0ea18 omrd`getgrgid + 60
    frame #5: 0x0000000003fd3afc omrd`partial apply for closure #1 in _FileManagerImpl.attributesOfItem(atPath:) [inlined] FoundationEssentials._nameFor(gid=$ss6UInt32VD @ scalar) -> Swift.Optional<Swift.String> at FileManager+Files.swift:50:21 [opt]
    frame #6: 0x0000000003fd3af4 omrd`partial apply for closure #1 in _FileManagerImpl.attributesOfItem(atPath:) at FileManager+Files.swift:195:28 [opt]
    frame #7: 0x0000000003fd37fc omrd`partial apply for closure #1 in _FileManagerImpl.attributesOfItem(atPath:) at FileManager+Files.swift:649:41 [opt]
    frame #8: 0x0000000003fd36ec omrd`partial apply for closure #1 in _FileManagerImpl.attributesOfItem(atPath:) at <compiler-generated>:0 [opt]
    frame #9: 0x000000000410f0a8 omrd`partial apply for closure #1 in String.withFileSystemRepresentation<A>(_:) [inlined] closure #1 ($0=<unavailable>, block=<unavailable>) throws -> τ_0_0 in Swift.String.withFileSystemRepresentation<τ_0_0>((Swift.Optional<Swift.UnsafePointer<Swift.Int8>>) throws -> τ_0_0) throws -> τ_0_0 at String+Internals.swift:152:17 [opt]
[...]

PR #994, opened 2024-10-22
Fetching user/group info causes race conditions

jmschonfeld:
We call a variety of functions like `getpwnam`/`getpwuid`/`getgrgid` to fetch information about users and groups. These functions are not thread safe (causing crashes during execution of the functions) and also return inner pointers which can be overwritten out from under us. To resolve these race conditions, this switches to the `_r` variants of the functions which use a client-supplied buffer for populating the result. I consolidated these calls into functions on the `Platform` namespace and updated call sites throughout swift-foundation to avoid the functions that can race.

Resolves https://github.com/swiftlang/swift-foundation/issues/982

### DIFF
--- a/Sources/FoundationEssentials/Platform.swift
+++ b/Sources/FoundationEssentials/Platform.swift
@@ -135,6 +135,70 @@ extension Platform {
         }
         return result
     }
+    
+    #if canImport(Darwin)
+    typealias Operation<Input, Output> = (Input, UnsafeMutablePointer<Output>?, UnsafeMutablePointer<CChar>?, Int, UnsafeMutablePointer<UnsafeMutablePointer<Output>?>?) -> Int32
+    #else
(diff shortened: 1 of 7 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/Platform.swift
```diff
@@ -135,6 +135,70 @@ extension Platform {
...
+            var ptr: UnsafeMutablePointer<Output>?
+            let error = operation(input, &result, $0.baseAddress!, bufferLen, &ptr)
+            guard error == 0 && ptr != nil else {
+                return nil
+            }
+            return try block(result)
+        }
+    }
+    
+    static func uidFor(name: String) -> uid_t? {
```
parkera, line 161: We don't have to block this PR on it, but I think the more canonically named version with be `uid(forName name: String)` (etc.)

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #996, opened 2024-10-23
Ensure empty AttributedSubstrings produce no runs

jmschonfeld:
Empty `AttributedSubstring`s should contain no runs, however there was a small index calculation bug in which empty substrings in the middle of the string (neither at the start or end indices) would incorrectly expand their ranges to the next run. This resolves the bug by ensuring that the `Runs.startIndex == Runs.endIndex` for all empty substrings regardless of their location in the string.

Resolves rdar://138432081


jmschonfeld:
@swift-ci please test

hristost (approved):

PR #1000, opened 2024-10-23
Don't skip anchors with strict matching in Calendar.RecurrenceRule. Resolve #881

### LINKED ISSUE #881: Unexpected results with Calendar.RecurrenceRule
I tried to use the new `Calendar.Recurrence` objects and functionality, and I've bumped into some surprising (wrong?) output.

Given the following date range:  `2024-01-01...2024-12-31`,  I used the `.monthly` recurrence rule to find every 5th Friday in the given range, using the `strict` matching rule.

The result was the (correct) following:

```
2024-03-29
2024-05-31
2024-08-30
2024-11-29
```

Now if I change the lower bound (start date) of the range to `2024-01-31` the result is unexpectedly the following:

```
2024-03-29
2024-05-31
// Missing date here
2024-08-30
```

The `2024-11-29` is missing from the output.

Output becomes even stranger if the calendar's `firstWeekday` is changed. If `firstWeekday == 2`, and the lower bound (start day) is `2024-01-01`  then the output is the following:

```
2024-03-29
2024-05-31
2024-08-30
2024-09-27 // Month with only 4 Fridays
2024-11-29
2024-12-27 // Month with only 4 Fridays
```

If the lower bound (start date) is  `2024-01-31` then the result is the following.

```
2024-03-29
2024-05-31
2024-08-30
// Missing date here
2024-12-27 // Month with only 4 Fridays
```

I believe that this is a bug, unless I am missing something in the way the algorithm works.

Some code if someone wants to quickly run some tests:

```
var calendar = Calendar(identifier: .gregorian)
// calendar.firstWeekday = 2

// let startComponents = DateComponents(year: 2024, month: 1, day: 1)
let startComponents = DateComponents(year: 2024, month: 1, day: 31)
let startDate = calendar.date(from: startComponents)!

let endComponents = DateComponents(year: 2024, month: 12, day: 31)
let endDate = calendar.date(from: endComponents)!

let rule = Calendar.RecurrenceRule.monthly(
    calendar: calendar,
    interval: 1,
    matchingPolicy: .strict,
    weekdays: [.nth(5, .friday)]
)

let output = rule.recurrences(of: startDate, in: startDate..<endDate)
```

hristost:
In recurrence rules that expand days or weekdays in a month, we first use a base recurrence to calculate "anchors" in the month, and then change the day of month or weekday to find results. Because the base recurrence used to match the day of month of the start date, we could miss anchors if matching was set to `.strict`. This change makes sure that if we know that the day of month is known to change, we reset it to 1 in the base recurrence.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -230,12 +230,42 @@ extension Calendar {
                     case .monthly:  [.second, .minute, .hour, .day]
                     case .yearly:   [.second, .minute, .hour, .day, .month]
                 }
-                let componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
+                var componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
                 
-                let rangeForBaseRecurrence: Range<Date>? = nil
+                let expansionChangesDay = dayOfYearAction == .expand || dayOfMonthAction == .expand || weekAction == .expand || weekdayAction == .expand
+                let expansionChangesMonth = dayOfYearAction == .expand || monthAction == .expand || weekAction == .expand
+
+                if expansionChangesDay, componentsForEnumerating.day != nil {
+                    // If we expand either the day of the month or weekday, then
+                    // the day of month is likely to not match that of the start
+                    // date. Reset it to 1 in the base recurrence as to not skip
+                    // "invalid" anchors, such as February 30
+                    componentsForEnumerating.day = 1
+                }
+                if expansionChangesMonth, componentsForEnumerating.month != nil {
+                    // Likewise, if we will be changing the month, reset it to 1 
+                    // in case the start date falls on a leap month
+                    componentsForEnumerating.month = 1
+                    componentsForEnumerating.isLeapMonth = nil
+                }
+                if expansionChangesDay || expansionChangesMonth, weekAction == .expand, weekdayAction != .expand {
+                    // If we are expanding weeks, all expansions in a given year
+                    // will have the same weekday. Above we have reset the month
+                    // or the day of the month, so we also changed that weekday.
+
+                    // To specify a yearly recurrence which starts from the same
+                    // weekday, and which doesn't start from a leap day / month,
+                    // simply use `dayOfYear` of the start date
+                    componentsForEnumerating.day = nil
+                    componentsForEnumerating.month = nil
+                    componentsForEnumerating.isLeapMonth = nil
+                    let daysInWeek = recurrence.calendar.maximumRange(of: .weekday)!.count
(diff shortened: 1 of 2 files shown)


hristost:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -230,9 +230,23 @@ extension Calendar {
...
+                    // the day of month is likely to not match that of the start
+                    // date. Reset it to 1 in the base recurrence as to not skip
+                    // "invalid" anchors, such as February 30
+                    componentsForEnumerating.day = 1
+                }
+                if componentsForEnumerating.month != nil, monthAction == .expand {
+                    // Likewise, if we will be changing the month, reset it to 1 
+                    // in case the start date falls on a leap month
+                    componentsForEnumerating.month = 1
+                    componentsForEnumerating.isLeapMonth = nil
```
parkera, line 248: Should we add a test case for a calendar with leap months?

hristost: I was actually working on that in a separate change - #1005

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -230,9 +230,23 @@ extension Calendar {
...
+                }
```
itingliu, line 249: Would we also need to handle nth-weekday and yearly combination?

itingliu (approved):

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -346,6 +379,11 @@ extension Calendar {
                 if weekAction == .expand {
                     // In a yearly recurrence with weeks specified, results do not land on any specific month
                     componentCombinations.weeksOfYear = recurrence.weeks
+                    if componentCombinations.weekdays == nil {
+                        if let weekdayIdx = components.weekday, let weekday = Locale.Weekday(weekdayIdx) {
+                            componentCombinations.weekdays = [.every(weekday)]
```
hristost, line 384: In adding tests, I found another regression from #981: if we are expanding weeks, we should always have a weekday set. The test case we had (`testYearlyRecurrenceWithWeekNumberExpansion`) just happened to work without the fix since the anchor was the very start of the search range.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
hristost: Yes, good catch! Added more exhaustive checks and cleaned up the logic a bit to be easier to read

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
hristost: Actually no, I was misled -- we already handle this above. I just went down the wrong rabbit hole while writing test cases.

hristost:
@swift-ci please test

hristost:
@swift-ci please test

PR #1001, opened 2024-10-23
JSONDecoder should parse base64 via a buffer instead of creating a string

iCharlesHu:
resolves: rdar://132749546


iCharlesHu:
@swift-ci please test

kperryua (approved):

PR #1006, opened 2024-10-25
Fix JSONEncoder performance regression in 6.0 toolchain

kperryua:
Major refactoring of Darwin Foundation's (and hence swift-foundation's) `JSONEncoder` took place a couple of years ago, with the rewrite being entirely in Swift, no longer relying on `NSJSONSerialization`. This effort resulted in a significant performance boost on Darwin. However, when swift-corelibs-foundation was re-cored on top of swift-foundation in 6.0, a performance regression was discovered there compared to the 5.10 toolchain.

After examining the code from the 5.10 toolchain's `JSONEncoder`, I've adapted the swift-foundation implementation to use its general architecture, with some additional important correctness fixes and other adjustments. I also included some other optimizations on top of it. All of this resulted in slightly *improved* performance in usage within the toolchain (over 5.10), as well via the swift-foundation package and Darwin Foundation.framework over previous.

To summarize the various changes:

- Serialize to a `[UInt8]` instead of a `Data`. It's unfortunately noticeably more expensive to serialize to `Data` right now.
- Avoid unnecessary `String` allocations when serializing escaped characters in `String`s.
- Replace `JSONReference` (mutable reference / leaf value combination) with `JSONEncoderValue` (immutable value type) + `JSONFuture` (container of either a value or one of the reference types) + `JSONFuture.RefArray` + `JSONFuture.RefObject` (pure reference types). This approach resulted in less overall retain/release traffic, at the cost of some rare "unwrapping" of `JSONEncoderValue` values back into the reference types.
- Make each `__JSONEncoder` responsible for one layer of decoding, instead of reusing the same one multiple times. This enabled some performance optimizations over what the 5.10 SCL-F implementation had, including `CodingKey` and `CodingPath` optimizations (each encoder holds a `CodingKey`, and each nested container has a `CodingPathNode` linked list descending from the original encoder; the path, when needed (rarely) is constructed by walking up this chain of encoders), and `__JSONEncoder` instance reuse to avoid allocation churn.
- Reimplemented the `_JSONDirectArrayEncodable` optimization path to use `[UInt8]` buffers directly (with substring lengths where needed) instead of holding on to `String`s. This will enable us to have the stdlib encode straight to our buffer instead of jumping through an unnecessary transient `String` allocation if such an API is ever introduced.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONEncoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONEncoder.swift
@@ -347,14 +347,14 @@ open class JSONEncoder {
     /// - throws: An error if any value throws an error during encoding.
     open func encode<T : Encodable>(_ value: T) throws -> Data {
         try _encode({
-            try $0.wrapGeneric(value, for: .root)
+            try $0.wrapGeneric(value)
         }, value: value)
     }
     
     @available(FoundationPreview 0.1, *)
     open func encode<T : EncodableWithConfiguration>(_ value: T, configuration: T.EncodingConfiguration) throws -> Data {
         try _encode({
-            try $0.wrapGeneric(value, configuration: configuration, for: .root)
+            try $0.wrapGeneric(value, configuration: configuration)
         }, value: value)
     }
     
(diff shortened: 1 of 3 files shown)


kperryua:
@swift-ci Please test

itingliu (approved):
Is there any performance statistics you can share?

kperryua:
Sure!

@jmschonfeld ran our JSON encoding benchmarks for me on Ubuntu, comparing the 5.10, 6.0, and 6.0+this PR versions

**Canada-encodeToJSON:**

- 5.0: 15 iters/s
- 6.0: 13 iters/s
- This PR: 20 iters/s

**Twitter-encodeToJSON:**

- 5.0: 549 iters/s
- 6.0: 353 iters/s
- This PR: 658 iters/s

PR #1007, opened 2024-10-25
Fix typo in String.replacingTildeWithRealHomeDirectory

jmschonfeld:
Fixes a typo where we created a path with just the user's name rather than the user's home directory due to a recent change in this function.


jmschonfeld:
@swift-ci please test

jrflat (approved):

Issue #957, opened 2024-10-02
URL.init(string:) handling of IPv6 addresses has changed in Swift 6


rnro:
Initializing a URL from an IPv6 address in Swift 6 on Linux now does not strip enclosing square-brackets. In earlier Swift versions and on macOS the square-brackets are removed.

As well as being a behavior change this can cause issues when passing the host component on e.g. to `inet_pton` which doesn't tolerate the square braces.

On Linux Swift 5.10:
```
  1> import Foundation
  2> let u = Foundation.URL(string: "http://[::1]")
  [snip]
  3> print(u!.host)
Optional("::1")
```

On Linux Swift 6.0.1:
```
  1> import Foundation
  2> let u = Foundation.URL(string: "http://[::1]")
  [snip]
  3> print(u!.host)
Optional("[::1]")
```

jrflat:
This is expected from the update of `URL` to RFC 3986, which defines:

```
host          = IP-literal / IPv4address / reg-name
IP-literal    = "[" ( IPv6address / IPvFuture  ) "]"
```

with `host` including the `[` and `]`. We'd like to maintain this behavior for `.host`, as it unifies the behaviors of `URL.host`, `URL.host()`, and `URLComponents.host`, but we'll keep an eye out if there's more bincompat issues from this change.

YOCKOW:
Something about this that may impact existing code is discrepancy between macOS and Linux:
`print(URL(string: "http://[::1]")!.host!)` prints "::1" (without square brackets) on macOS even in Swift 6.

I think we'd better to resolve such discord at least.

PR #1008, opened 2024-10-25
Restore `URL.host` bracket stripping for compatibility

jrflat:
As part of the new Swift `URL` implementation, we unified `URL.host` to behave like `URLComponents.host`, which will return an IP literal with the square brackets included. While RFC 3986 does say that `host` includes these square brackets, this change caused bincompat issues for clients who take IPv6 addresses from `URL.host` and pass them directly to APIs like `inet_pton(3)`.

Because `URL.host` has long behaved by stripping the brackets, and no longer stripping them can cause immediate client breakage that requires workaround, this PR restores the previous behavior. Thank you to all who worked around it for the time being.

Resolves https://github.com/swiftlang/swift-foundation/issues/957

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1213,21 +1213,38 @@ public struct URL: Equatable, Sendable, Hashable {
             return nil
         }
         #endif
-        guard let encodedHost else { return nil }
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

parkera:
Should cherry pick for release/6.0

PR #1018, opened 2024-10-28
Fix `IS_DECOMPOSABLE` flag when bridging to `CFURL`

jrflat:
`CFURL` considers a URL decomposable if it can be decomposed into its component parts according to RFC 1808. This varies slightly from an analogous definition of decomposable for RFC 3986, specifically when it comes to relative paths as found in `data:` URLs.

This PR adjusts Swift URL's `CFURL` creation logic to add the `IS_DECOMPOSABLE` flag only if it does not have a scheme or if there is a slash directly following the scheme.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1019, opened 2024-10-28
[6.0.x] URL: Appending to an empty file path results in an absolute path (#988)

jrflat:
**Explanation:** Corrects `URL.appending(path:)` behavior to prevent an empty relative path from becoming absolute, which would no longer resolve against the base URL.
**Scope:** Only impacts `URL.appending(...)` APIs when the relative path is empty.
**Original PR:** #988 
**Risk:** Low - Small change to fix breaking behavior in this edge case.
**Testing:** Added unit tests, swift-ci, stable in `main`
**Reviewer:** @jmschonfeld 

Resolves https://github.com/swiftlang/swift-foundation/issues/921 for `release/6.0`

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -2203,9 +2203,8 @@ extension URL {
         var path = String(path)
         #endif
 
-        var newPath = relativePath()
         var insertedSlash = false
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

PR #1020, opened 2024-10-28
Calendar.RecurrenceRule: Only match weekday components when filtering by weekday

hristost:
When filtering by weekdays, we match the dates against a set of date components. These components happened to contain the hour, minute and second components from the anchor date, so filtering would undo previous expansions by hour, minute and second.


hristost:
@swift-ci please test

hristost:
@swift-ci please test

itingliu (approved):

PR #1026, opened 2024-10-30
Add flag for CFURLs originating from Swift

jrflat:
This flag is used to accurately compare `CFURL`s bridged from Swift `URL` to `CFURL`s originating from Obj-C.


jrflat:
@swift-ci please test

jmschonfeld (approved):

Issue #1034, opened 2024-11-05
Race condition in `_FileManagerImpl.currentDirectoryPath`


ahoppen:
Reading the following code, I noticed a race condition that causes us to return `nil` for the `currentDirectoryPath` even though there’s always a current directory path on Windows.

https://github.com/swiftlang/swift-foundation/blob/a0147acdc4e51255bbda829572f57ce110c0e663/Sources/FoundationEssentials/FileManager/FileManager%2BDirectories.swift#L467-L476

Race: 
- We execute `let dwLength: DWORD = GetCurrentDirectoryW(0, nil)` and allocate a buffer of the length `dwLength`
- The current directory path gets changed to be longer
- The temporary allocation is not sufficient to fit the new (longer) directory path
- The second `GetCurrentDirectoryW` call returns a non-zero value
- `_Filemanager.currentDirectoryPath` returns `nil` even though a current directory existed.

PR #1035, opened 2024-11-05
FileManager: avoid a TOCTOU issue in computing CWD

compnerd:
On Windows, we could potentially return a `nil` for the current working directory in the rare case that the current working directory was changed during the computation:

```swift
let dwLength: DWORD = GetCurrentDirectoryW(0, nil)                                // 1
return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
    if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {           // 2
        return String(decodingCString: $0.baseAddress!, as: UTF16.self)
    }
    return nil                                                                    // 3
}
```

Consider the case where at step 1, we receive $n$. We then are interrupted, the CWD changed. We then perform step 2, where we receive $m$ (st $m != n$). We would then proceed to point 3, where we return `nil`. Avoid this TOCTOU issue by repeating this operation to a fixed point.

Because we are guaranteed a current directory on Windows (unless the initial query for the buffer size fails), we will eventually succeed. In order to avoid a DoS attack vector, limit the attempt to quiescence to a fixed number.

Fixes: #1034

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -466,13 +466,22 @@ extension _FileManagerImpl {
     
     var currentDirectoryPath: String? {
 #if os(Windows)
-        let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
-        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-            if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        var dwLength: DWORD = GetCurrentDirectoryW(0, nil)
+        guard dwLength > 0 else { return nil }
+
+        for _ in 0 ... 8 {
+            if let szCurrentDirectory = withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength), {
+                let dwResult: DWORD = GetCurrentDirectoryW(dwLength, $0.baseAddress)
+                if dwResult == dwLength - 1 {
+                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+                }
+                dwLength = dwResult
+                return nil
+            }) {
+                return szCurrentDirectory
             }
-            return nil
         }
+        return nil
 #else
         withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
             guard getcwd(buffer.baseAddress!, FileManager.MAX_PATH_SIZE) != nil else {


compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -466,13 +466,20 @@ extension _FileManagerImpl {
...
 #if os(Windows)
-        let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
-        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-            if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        for _ in 0 ... 8 {
+            let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
+            guard dwLength > 0 else { return nil }
+
+             if let szCurrentDirectory = withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength), {
```
ahoppen, line 473: Looks like this has one additional space before `if`

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -466,13 +466,20 @@ extension _FileManagerImpl {
...
+                if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {
```
ahoppen, line 474: Would it be possible to re-use the return value of this call to `GetCurrentDirectoryW` for the next iteration instead of doing another `GetDirectoryW(0, nil)` call?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -466,13 +466,20 @@ extension _FileManagerImpl {
     
     var currentDirectoryPath: String? {
 #if os(Windows)
-        let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
-        return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-            if GetCurrentDirectoryW(dwLength, $0.baseAddress) == dwLength - 1 {
-                return String(decodingCString: $0.baseAddress!, as: UTF16.self)
+        for _ in 0 ... 8 {
+            let dwLength: DWORD = GetCurrentDirectoryW(0, nil)
+            guard dwLength > 0 else { return nil }
```
ahoppen, line 471: Do you know when `dwLength` can be 0 here? Just because I’m assuming that `currentDirectoryPath` is not `nil` in https://github.com/swiftlang/swift-foundation/pull/1033.

compnerd: No, I don't know _when_ that can be 0, but the docs indicate that it is possible for the call to fail and in such a case `0` will be returned.

https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getcurrentdirectory

> If the function fails, the return value is zero. To get extended error information, call [GetLastError](https://learn.microsoft.com/en-us/windows/desktop/api/errhandlingapi/nf-errhandlingapi-getlasterror).

At the same time, it does document that each process has a working directory, so it is certainly unclear when and how this function could fail, especially when simply querying the required buffer size.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
ahoppen: Hmm, OK. I guess returning `nil` is the best option.

ahoppen (approved):
LGTM but @jmschonfeld should also look at the change.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -466,13 +466,22 @@ extension _FileManagerImpl {
...
+                return nil
+            }) {
+                return szCurrentDirectory
             }
-            return nil
         }
+        return nil
 #else
         withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
             guard getcwd(buffer.baseAddress!, FileManager.MAX_PATH_SIZE) != nil else {
```
jmschonfeld, line 487: One quick question - on non-Windows we just allocate a buffer of `MAX_PATH_SIZE` (1024 bytes IIRC) and assume that the path will fit (returning `nil` if for some reason it doesn't). Just confirming - I assume there's no equivalent for Windows that we should use either instead of requesting the buffer size or using in the case of the size check failing?

compnerd: Yeah, there is no real equivalent for Windows. The majority of the API surface on Windows is designed to assume that you will query the required sizes. In the case of the file system, the only place there is a small buffer size guarantee is the Win32 API surface and in Foundation we will bypass that layer entirely when working with paths to ensure that long paths work.

kperryua:
Is this an established pattern of calling this API? Or: how do other libraries handle this race?

compnerd:
@kperryua the established pattern is to call it in a loop until it succeeds.

However, this does open up a possible DOS attack vector where the attacker has access to your address space and is able to inject a detour for `GetCurrentWorkingDirectoryW` to return the value and change the directory. That is the reason for the bounded retry. It is arguable that if the attacker has the ability to detour the function the DoS attack vector is meaningless as they can execute arbitrary code.

The actual race condition that exists with the loop variant is only possible in a multithreaded scenario where a thread is constantly changing the working directory to a monotonically increasing size (which will obviously be bound as you cannot grow ad infinitum) and you are forced to yield control in between the calls. I'm not sure how likely that is to be hit in practice though.

compnerd:
@jmschonfeld oaky to merge?

jmschonfeld (approved):

PR #1036, opened 2024-11-06
Fix incorrect unicode escape sequence encodings

kperryua:
We had a testing gap in the encoding of strings containing low unicode escape sequences. The recent JSONWriter re-write broke these, and the missing test means we didn't catch it.

Single digit hex values were getting / 16 when % 16 was more correct (though actually unnecessary).
Double digit hex values were in reverse order.


kperryua:
@swift-ci Please test

jmschonfeld (approved):

Issue #976, opened 2024-10-10
`URL.withUnsafeFileSystemRepresentation` drops trailing backslash even if it is significant on Windows


ahoppen:
```swift
let url = URL(fileURLWithPath: #"C:\"#)
url.withUnsafeFileSystemRepresentation { filePath in
    print(String(cString: filePath! , encoding: .utf8))
}
```

prints `C:`. Dropping the trailing backslash is problematic here because `C:\` is considered a root directory in Windows but `C:` is not (see table of examples in https://learn.microsoft.com/en-us/windows/win32/api/pathcch/nf-pathcch-pathcchisroot).

Issue #977, opened 2024-10-10
`URL.path` includes leading Slash if the URL points to the root directory of a drive on Windows


ahoppen:
```swift
print(URL(fileURLWithPath: #"C:\"#).path)
```

prints `/C:`. It should print `C:\`. 

Also see https://github.com/swiftlang/swift-foundation/issues/976, which is related.

jmschonfeld:
I think this was likely resolved by https://github.com/swiftlang/swift-foundation/pull/964

ahoppen:
It specific about `C:\`. I don't see the issue with non-root URLs

PR #1038, opened 2024-11-08
`URL.path` should not strip trailing slash for root paths on Windows

jrflat:
We should not strip the trailing slash when getting the `fileSystemPath` for `C:\` on Windows, since `C:\` represents the root while `C:` does not. Similarly, we should not strip the trailing slash for any UNC path, since the trailing slash might be root.

Resolves https://github.com/swiftlang/swift-foundation/issues/976 and https://github.com/swiftlang/swift-foundation/issues/977

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1349,31 +1349,36 @@ public struct URL: Equatable, Sendable, Hashable {
         }
     }
 
-    private static func windowsPath(for posixPath: String) -> String {
-        let utf8 = posixPath.utf8
-        guard utf8.count >= 4 else {
-            return posixPath
+    private static func windowsPath(for urlPath: String) -> String {
+        let utf8 = urlPath.utf8
+        guard !utf8.starts(with: [._slash, ._slash]) else {
+            // UNC path, don't strip any trailing slash, which might be root
+            return decodeFilePath(urlPath)
         }
         // "C:\" is standardized to "/C:/" on initialization
-        let array = Array(utf8)
-        if array[0] == ._slash,
-           array[1].isAlpha,
-           array[2] == ._colon,
-           array[3] == ._slash {
-            return String(Substring(utf8.dropFirst()))
+        var iter = utf8.makeIterator()
+        guard iter.next() == ._slash,
+              let driveLetter = iter.next(), driveLetter.isAlpha,
+              iter.next() == ._colon,
+              iter.next() == ._slash else {
+            return decodeFilePath(urlPath._droppingTrailingSlashes)
         }
-        return posixPath
+        // Strip trailing slashes from the path, which preserves a root "/"
+        let path = String(Substring(utf8.dropFirst(3)))._droppingTrailingSlashes
+        // Don't include a leading slash before the drive letter
+        return "\(Unicode.Scalar(driveLetter)):\(decodeFilePath(path))"
+    }
+
+    private static func decodeFilePath(_ path: some StringProtocol) -> String {
+        let charsToLeaveEncoded: Set<UInt8> = [._slash, 0]
+        return Parser.percentDecode(path, excluding: charsToLeaveEncoded) ?? ""
     }
 
     private static func fileSystemPath(for urlPath: String) -> String {
-        let charsToLeaveEncoded: Set<UInt8> = [._slash, 0]
-        guard let posixPath = Parser.percentDecode(urlPath._droppingTrailingSlashes, excluding: charsToLeaveEncoded) else {
-            return ""
-        }
         #if os(Windows)
-        return windowsPath(for: posixPath)
+        return windowsPath(for: urlPath)
         #else
-        return posixPath
+        return decodeFilePath(urlPath._droppingTrailingSlashes)
         #endif
     }
 
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -13,6 +13,10 @@
 public struct URLResourceKey {}
 #endif
 
+#if os(Windows)
+import WinSDK
+#endif
+
 #if FOUNDATION_FRAMEWORK
 internal import _ForSwiftFoundation
 internal import CoreFoundation_Private.CFURL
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1349,31 +1349,36 @@ public struct URL: Equatable, Sendable, Hashable {
         }
     }
 
-    private static func windowsPath(for posixPath: String) -> String {
-        let utf8 = posixPath.utf8
-        guard utf8.count >= 4 else {
-            return posixPath
+    private static func windowsPath(for urlPath: String) -> String {
+        let utf8 = urlPath.utf8
+        guard !utf8.starts(with: [._slash, ._slash]) else {
```
compnerd, line 1354: What about `\\server\share\directory\`? Should that not be `\\server\share\directory`?

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1349,31 +1349,36 @@ public struct URL: Equatable, Sendable, Hashable {
...
+            return decodeFilePath(urlPath)
         }
         // "C:\" is standardized to "/C:/" on initialization
-        let array = Array(utf8)
-        if array[0] == ._slash,
-           array[1].isAlpha,
-           array[2] == ._colon,
-           array[3] == ._slash {
-            return String(Substring(utf8.dropFirst()))
+        var iter = utf8.makeIterator()
```
compnerd, line 1359: Can we assert that the path starts with `[._slash, alpha, ._colon, ._slash]`?

Tests/FoundationEssentialsTests/URLTests.swift
```diff
@@ -340,13 +340,39 @@ final class URLTests : XCTestCase {
...
+        XCTAssertEqual(url.path, "C:/")
+        XCTAssertEqual(url.fileSystemPath, "C:/")
+
+        url = URL(filePath: #"\C:\"#, directoryHint: .isDirectory)
+        XCTAssertEqual(url.absoluteString, "file:///C:/")
+        XCTAssertEqual(url.path(), "/C:/")
+        XCTAssertEqual(url.path, "C:/")
+        XCTAssertEqual(url.fileSystemPath, "C:/")
+
+        let base = URL(filePath: #"\d:\path\"#, directoryHint: .isDirectory)
```
compnerd, line 369: Is Foundation supposed to take this path?

Sources/FoundationEssentials/URL/URL.swift
jrflat: Ideally, yeah. I think it gets a bit complicated to make sure we do the right thing for all paths, though. I'm not super familiar with UNC paths, so please correct me if I'm wrong, but I think we could have:

```
\\.\c:\
\\.\BootPartition\
\\server\share\
\\.\server\share\
\\.\UNC\server\share\
```

so we couldn't rely on just stripping after X segments. And it still might be complex to check for `.` or `?` in the first segment since we could have either `driveName` or `server\share` following it.

Do you think we should call `PathIsRootW` to determine whether we should strip the slash? I'm a bit hesitant to do filesystem I/O in `URL.path`, but I think that would get us the correct answer.

Tests/FoundationEssentialsTests/URLTests.swift
jrflat: I think we should handle filePath inputs of `\d:\` and `d:\` equivalently on Windows since both will be standardized to the same RFC URL path.

Sources/FoundationEssentials/URL/URL.swift
compnerd: Yeah, `PathIsRootW` is the right thing to use IMO. `\\.\` paths are device paths, there are also kernel paths `\\?\`. Additionally, you might also encounter a root local path with a `\??\` prefix. For the UNC path, you can also label it via link, e.g. `\\?\UNC\server\share\directory\`.

jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
jrflat: Switched to checking `PathIsRootW` for non-drive letter paths in [9d325ee](https://github.com/swiftlang/swift-foundation/pull/1038/commits/9d325ee65d411131f89a1ce34349c53e007efd52)

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -1349,31 +1353,43 @@ public struct URL: Equatable, Sendable, Hashable {
...
-           array[1].isAlpha,
-           array[2] == ._colon,
-           array[3] == ._slash {
-            return String(Substring(utf8.dropFirst()))
+        // There are many flavors of UNC paths, so use PathIsRootW to ensure
+        // we don't strip a trailing slash that represents a root.
+        let path = decodeFilePath(urlPath)
+        return path.replacing(._slash, with: ._backslash).withCString(encodedAs: UTF16.self) { pwszPath in
+            guard !PathIsRootW(pwszPath) else {
+                return path
```
compnerd, line 1376: I think that we still get one path incorrect now: `/S:`. This a non-UNC, drive-relative path. We could either resolve that to an absolute path or return it as `S:`. You can also have a more complex path such as `S:utils\build.ps1` (which I believe would be encoded as `/s:utils/build.ps1`.

jrflat: Yeah, more work definitely needs to be done to support drive-relative paths in `URL`. Currently, `S:utils\build.ps1` is actually treated as a relative path, relative to the current working directory (but not necessarily the cwd of `S:`), so `.path` won't resolve it correctly.

I think we should get the cwd of `S:` during `URL(filePath:)` initialization and use that cwd path as the base URL, then strip `S:` from the relative path. If that sounds good I can open up a follow-on PR, since it'll be slightly more involved than this change to fix `C:\` and other absolute paths' behaviors.

compnerd: I think that we should do some checking if the call needs to be done, but otherwise, this sounds like a good way forward. Doing that in a follow up is totally fine.

compnerd (approved):

Issue #1040, opened 2024-11-08
Calendar.RecurrenceRule.recurrences(of:in) can find dates excluded by the range


Salynn:
If a recurrence occurs on the upper bound of the limiting range, it is still included in the returned sequence. This violates the definition of Range<T> which states that the range does not include the upper bound. Either the type of the `of` parameter should be a ClosedRange for the current behavior, or the upper bound should be excluded from returned sequence.

```swift
let start = try Date.ISO8601FormatStyle().year().month().day().parse("2024-01-01")
let end = try Date.ISO8601FormatStyle().year().month().day().parse("2024-03-01")
let rule = Calendar.RecurrenceRule(calendar: .current, frequency: .monthly, matchingPolicy: .previousTimePreservingSmallerComponents)
for date in rule.recurrences(of: start, in: start..<end) {
    assert((start..<end).contains(date))
}
```

Expected behavior:
Code executes successfully

Actual behavior:
Code asserts on the last item in the sequence.

Also filed as [FB15728756](https://feedbackassistant.apple.com/feedback/15699570).

Salynn:
Looks like the bug is on [Calendar_Recurrence.swift line 468](https://github.com/swiftlang/swift-foundation/blob/main/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift#L468). The conditional check should be `>=` instead of `>`.
I'll see if I can get setup to test the fix and submit a PR.

itingliu:
Fixed in https://github.com/swiftlang/swift-foundation/pull/1041

PR #1041, opened 2024-11-09
Fix upper bound in Calendar.RecurrenceRule.recurrences(of:in:)

Salynn:
This commit fixes the use of the `in:` bound in
Calendar.RecurrenceRule.recurrences(of:in:) to properly exclude the upper bound as revealed by #1040.

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
@@ -766,4 +766,19 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
         XCTAssertEqual(results.next(), Date(timeIntervalSince1970: 1695630600.0)) // 2023-09-25T08:30:00-0000
         XCTAssertEqual(results.next(), Date(timeIntervalSince1970: 1695631500.0)) // 2023-09-25T08:45:00-0000
     }
+
+    func testRangeExcludesUpperBounds() {
+        let start = Date(timeIntervalSince1970: 1695304800.0) // 2023-09-21T14:00:00-0000
+        let end   = Date(timeIntervalSince1970: 1697896800.0) // 2023-10-21T14:00:00-0000
+
+        let rule = Calendar.RecurrenceRule(calendar: gregorian, frequency: .monthly)
+
+        let eventStart = Date(timeIntervalSince1970: 1285077600.0) // 2010-09-21T14:00:00-0000
+        let results = Array(rule.recurrences(of: eventStart, in: start..<end))
+
+        let expectedResults: [Date] = [
+            Date(timeIntervalSince1970: 1695304800.0), // 2023-09-21T14:00:00-0000
+        ]
+        XCTAssertEqual(results, expectedResults)
+     }
 }
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -465,7 +465,7 @@ extension Calendar {
                             return nil
                         }
                         if let range = self.range {
-                            if date > range.upperBound {
+                            if date >= range.upperBound {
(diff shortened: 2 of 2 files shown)


Salynn:
@swift-ci please test

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -739,4 +739,16 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
...
+    func testRangeExcludesUpperBounds() {
+         let start = Date(timeIntervalSince1970: 1695304800.0) // 2023-09-21T14:00:00-0000
+         let end   = Date(timeIntervalSince1970: 1729519200.0) // 2024-10-21T14:00:00-0000
+
+         let rule = Calendar.RecurrenceRule(calendar: gregorian, frequency: .monthly)
+
+         let eventStart = Date(timeIntervalSince1970: 1285077600.0) // 2010-09-21T14:00:00-0000
+         let results = Array(rule.recurrences(of: eventStart, in: start..<end))
+
+        XCTAssertNotEqual(results.last!, end)
```
itingliu, line 752: Instead of doing XCTAssertNotEqual, can you XCTAssertEqual what the last result should be?

Salynn: I can, but that changes the intent of the test, which is to ensure that `end` is not part of the recurrence. What the last value is doesn't actually matter for the purpose of this test.

Salynn: How about change the assertion to `XCTAssertLessThan`?

itingliu: I understand what you want to test, but this line, `XCTAssertNotEqual(results.last!, end)`, alone can mean anything if I hadn't looked at the code change. For example, this assert would pass if `results` is just an arbitrary value. Same goes with `XCTAssertLessThan`.

Also, instead of force unwrapping `results.last!`, it's useful to write it as `XCTUnwrap` instead of just force unwrapping it so we get a nice test report if it fails

Salynn: I was unaware of XCTUnwrap! That's neat! I'll make the changes later today!

Salynn: I decided to go with a pattern established elsewhere and just use a smaller list of recurrences and validate the entire array. Let me know if that doesn't work for some reason.

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -766,4 +766,19 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
         XCTAssertEqual(results.next(), Date(timeIntervalSince1970: 1695630600.0)) // 2023-09-25T08:30:00-0000
         XCTAssertEqual(results.next(), Date(timeIntervalSince1970: 1695631500.0)) // 2023-09-25T08:45:00-0000
     }
+
+    func testRangeExcludesUpperBounds() {
+         let start = Date(timeIntervalSince1970: 1695304800.0) // 2023-09-21T14:00:00-0000
+        let end   = Date(timeIntervalSince1970: 1697896800.0) // 2023-10-21T14:00:00-0000
```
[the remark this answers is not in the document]
Salynn: Nope, it's weird. Not sure how the autoformatter let that through. I'll fix it and get that pushed later today.

itingliu (commented):
@swift-ci please test

itingliu:
@swift-ci please test

parkera:
Thank you @Salynn !

itingliu (approved):
LGTM. Thanks

PR #1043, opened 2024-11-12
rdar://133742708 (JSONDecoder crashes whole app on invalid input instead of throwing Error)

kperryua:
These `try!` assertions were incorrect. They were focused on the fact that the type of the value in the input JSON was guaranteed to be a string, but with the on-demand parsing behavior, there is no longer any initial UTF-8 validation step, which means decoding the keys into String values could fail. This needs to actually throw an error instead of trapping. Fortunately, this is an (embarrassingly) easy change to make.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONDecoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONDecoder.swift
@@ -786,8 +786,8 @@ extension JSONDecoderImpl: Decoder {
 
         var iter = jsonMap.makeObjectIterator(from: region.startOffset)
         while let (keyValue, value) = iter.next() {
-            // Failing to unwrap a string here is impossible, as scanning already guarantees that dictionary keys are strings.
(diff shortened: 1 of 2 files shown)


kperryua:
@swift-ci Please test

parkera (approved):

jhoughjr:
I'll share my xp is it might be relevant. I just spent 12 hours yesterday chasing a bug tangent to this issue.
Long long story short, somehow a non printable character got into a folder name in Xcode's UI I assume.
This unprintable in the filename is how I crashed due to those force unwraps.
What would have been nice, was if at the site of the error, where it lists the line, column and index of the error, if it could also list what file threw.

I have to assume if it has a line number and char index, it also has the name at the same scope.
In the meantime I'm looking at finding or making a utility to scan a repo for any unprintable.

Because JSONDecoder gave incomplete context it took far longer to identify why it crashed.
I think a catch block would allow the increased context I desire to manifest.

PR #1046, opened 2024-11-13
Refine format of Concurrency-Safe Notifications

li-bei:
- Rename all `makeMessage(:Notification)` and `makeNotification(:Self)` to `makeMessage(_:)` and `makeNotification(_:)`, which using the correct Swift method signature style.
- Use `NSWorkspace.applicationUserInfoKey` instead of `"applicationUserInfoKey"`.
- Make all multi-line method definitions and calls using same style.


cthielen:
@li-bei Thanks, these are good improvements.

I'll incorporate these changes in our next revision to avoid churn during the review period.

PsychoH13:
There's a typo in the file you could fix as part of this PR:
```swift
    public struct WillLaunchApplication: NotificationCenter.MainActorMessage {
```

the W in WillLaunchApplication shouldn't be capitalized.

PR #1050, opened 2024-11-18
NSRange to Range<AttributedString.Index> conversion should be relative to sliced boundaries

jmschonfeld:
When converting an `NSRange` to a `Range<AttributedString.Index>` via our API, the conversion currently does not take the provided `AttributedStringProtocol`'s `startIndex`/`endIndex` into account. This means that the `NSRange`'s location is considered relative to the base's start for an `AttributedSubstring` rather than the substring's start, leading to conversion failing because the `NSRange` is often considered beyond the bounds of the slice.

This updates the conversion to operate on a slice of the underlying `BigString` so that the range calculation is performed from the slice's `startIndex` rather than the base's `startIndex`.


jmschonfeld:
@swift-ci please test

PR #1051, opened 2024-11-18
Improve performance for calling `Locale.identifier` from `NS/CFLocale`

itingliu:
- use @_effects(releasenone) to solemnly swear to the compiler that we won't release anything inside these getters, eliminating refcounting to call them

- micro-optimize identifierDoesNotRequireSpecialCaseHandling, allowing us to delete the cache of it without regressing perf

On the test app this goes from "so slow I didn't bother to see how long it takes to finish" to "finishes in around 1 second"

Resolves 137885111

Co-authored-by: @Catfish-Man


itingliu:
According to my measurement with the said test app: Before: ~13 seconds. After: ~0.4 seconds. So 32x of improvement

jmschonfeld (approved):

PR #1056, opened 2024-11-21
[android] fix the tm_zone access, which is nullable on Android

hyp:
Regression after https://github.com/swiftlang/swift-foundation/pull/1053


hyp:
The nullability annotation on Android appears to be correct, as there as some cases where `tm_zone` is set to NULL in Android's bionic

compnerd (approved):

jmschonfeld (approved):

PR #1057, opened 2024-11-21
Add FoundationPreview 6.2 as an allowed availability annotation

jmschonfeld:
Now that we've branched for `release/6.1`, changes targeting main will need to be annotated via a new 6.2-aligned availability. This PR adds `FoundationPreview 6.2` as an allowed availability annotation so that future API work can use this to properly mark availability on platforms where Foundation ships as part of an SDK.


jmschonfeld:
@swift-ci please test

PR #1059, opened 2024-11-22
Properly write non-ASCII file names on Windows for file creation

jmschonfeld:
Previously, using `Data.write(to:)` (or `FileManager.createFile(atPath:)` which calls to that function) with a path that contained non-ASCII characters would crash. This crash was caused by an incorrect buffer length calculation that used `String.count` instead of the actual length of the buffer (which will differ for non-ASCII paths). Fixing the crash however revealed incorrect behavior where file creation would fail due to providing a UTF-8 file name to `_sopen_s`. Instead, we need to transcode the file name to UTF-16 and open the file with `_wsopen_s` to ensure that non-ASCII file names can be opened. The added unit test previously crashed (and then failed when the crash was resolved) but now successfully creates the file.

Resolves https://github.com/swiftlang/swift-foundation/issues/1058

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -37,18 +37,22 @@ import WASILibc
 
 // MARK: - Helpers
 
+#if os(Windows)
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

compnerd (approved):
Thank you!

compnerd:
Might be good to pull this into the 6.0 branch.

Issue #1052, opened 2024-11-19
Consider exposing common `Locale.NumberingSystem` as static variables


itingliu:
`Locale.LanguageCode` and `Locale.ScriptCode` defines common language and script codes as static vars for easy access, [such as](https://github.com/swiftlang/swift-foundation/blob/94adc6a958e10d759304dc37bcdb3b59f011e996/Sources/FoundationEssentials/Locale/Locale%2BComponents.swift#L899)

```swift
@_alwaysEmitIntoClient
public static var danish: Locale.LanguageCode { Locale.LanguageCode("da") }
```

We can consider exposing common `Locale.NumberingSystem` as well.

Discussion: https://forums.swift.org/t/fou-locale-components-language-and-language-components/54084/17

PR #1061, opened 2024-11-24
[Proposal] Add static properties for Locale.NumberingSystem

glebfann:
This proposal addresses [#1052](https://github.com/swiftlang/swift-foundation/issues/1052) by adding static properties to `Locale.NumberingSystem` for standard numbering systems defined in [Unicode CLDR](https://github.com/unicode-org/cldr/blob/latest/common/bcp47/number.xml).


glebfann:
@itingliu Hi! Do you have any thoughts/feedback on this proposal?

itingliu:
> @itingliu Hi! Do you have any thoughts/feedback on this proposal?

Sorry for not replying here. I've been following the discussion on the forum; Thank you for the discussion there! 

One comment about the proposal: Can you please include all the proposed variables in this proposal? I understand that's a long list, but having everything listed in the proposal allows consolidating the discussion around, say, naming, in the same place.

glebfann:
> One comment about the proposal: Can you please include all the proposed variables in this proposal? I understand that's a long list, but having everything listed in the proposal allows consolidating the discussion around, say, naming, in the same place.

Done, added all the numbering systems with their documentation.
Thank you for discussing this on the forum as well!

glebfann:
@itingliu Hi! Any updates here? Let me know if anything needs to be adjusted.

PR #1064, opened 2024-12-02
Correct buffer size logic when calling `confstr`

jmschonfeld:
Temporary directory lookup has faulty logic on Darwin:

```swift
if length > 0 && length < buffer.count {
    return String(decoding: buffer, as: UTF8.self)
}
```

1. We already checked 1length > 0` above, so we can eliminate that check
2. `length < buffer.count` will never be true, since we pass `capacity: length` to the `withUnsafeTemporaryAllocation` call which means that `buffer.count` will always be `length`
3. Since `buffer.count == length == finalLength`, `buffer.count` includes the null terminator but `String(decoding:)` does not stop at null terminators (in other words, the resulting string would contain a null byte at the end

This means we never actually read the value from `confstr` and always fell back to `TMPDIR`. To fix it, we just remove the unnecessary checks and switch to `String(cString:)` (and initialize the buffer to `0`s) since `confstr` is guaranteed to always copy a null-terminated string.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1066, opened 2024-12-03
Implement AttributedString UTF8 and UTF16 views

jmschonfeld:
This adds a new UTF-8 and UTF-16 view to `AttributedString`, vending APIs similar to the underlying `BigString`'s UTF-8 and UTF-16 view

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
@@ -33,10 +33,6 @@ extension AttributedString {
             _guts = guts
             _range = range
         }
-
-        public init() {
-            self.init(Guts())
-        }
     }
 
     public var utf16: UTF16View {
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
@@ -1,0 +1,139 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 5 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
```diff
@@ -0,0 +1,139 @@
...
+        internal init(_ guts: AttributedString.Guts) {
+            self.init(guts, in: guts.stringBounds)
+        }
+
+        internal init(_ guts: Guts, in range: Range<BigString.Index>) {
+            _guts = guts
+            _range = range
+        }
+
+        public init() {
```
itingliu, line 37: Is this intentionally `public`? It wasn't included in the proposal. I don't think you can `init` a `String.UTF16View` either.

Sources/FoundationEssentials/AttributedString/AttributedString+UTF8View.swift
```diff
@@ -0,0 +1,139 @@
...
+        .init(_range.upperBound)
+    }
+
+    public var count: Int {
+        _utf8.count
+    }
+
+    public func index(before i: AttributedString.Index) -> AttributedString.Index {
+        precondition(i >= startIndex && i <= endIndex, "AttributedString index out of bounds")
+        let j = Index(_guts.string.utf8.index(before: i._value))
```
itingliu, line 81: Does this throw if `i == startIndex`? I believe calling `string.utf8.index(before: string.utf8.startIndex)` throws the index out of bound error. If so, should we just enforce this at the precondition above?

jmschonfeld: We still do hit a precondition from swift-collections:

```
_RopeModule/BigString+Contents.swift:292: Precondition failed: Can't advance below start index
```

But we might as well account for this here, good call - I'll update the above `>=` to be `>`

Sources/FoundationEssentials/AttributedString/AttributedString+UTF16View.swift
jmschonfeld: Good call, this was copied from the `CharacterView` and `UnicodeScalarView` code, but I didn't include it in the proposal since there's no need to ever create this yourself especially since it's immutable - I'll remove this from the implementation

jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1069, opened 2024-12-04
Darwin SearchPaths without expanding tilde produce non-tilde-based paths

jmschonfeld:
The `String`, ObjC-based search path enumeration calls the Swift implementation which produces a `URL` and then converts each URL back to a `String` path. This process is actually making paths absolute which is fine for Swift use cases, but the ObjC API has an option to prevent expansion of the tilde directory. By passing those unexpanded paths through `URL`, we end up with paths like `/Users/USER/Library` which are incorrect because the tilde is expanded.

To avoid these behaviors and maintain support for the "legacy" ObjC behavior for string paths, I shuffled the functions around just a bit so that the implementation creates `String`s instead of `URL`s and only converts to `URL` for the Swift API. This also allows us to test the `String` results directly in our unit tests to add a test that validates this behavior even in non-`FOUNDATION_FRAMEWORK` builds.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/SearchPaths/FileManager+SearchPaths.swift
+++ b/Sources/FoundationEssentials/FileManager/SearchPaths/FileManager+SearchPaths.swift
@@ -57,41 +57,11 @@ extension FileManager.SearchPathDomainMask {
     }
 }
 
-#if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1070, opened 2024-12-04
Avoid attaching removefile callbacks when delegate does not implement functions

jmschonfeld:
The previous ObjC implementation of `-[NSFileManager removeItemAtPath:]` only attached callbacks to the `removefile_state_t` when the delegate was present and it implemented the equivalent functions. However, the new Swift implementation unconditionally attaches callbacks that simply return "default" values when no delegate responds. This causes a behavioral difference in `removefile` when removing items that are not materialized (ex. remote iCloud documents that do not currently exist on disk). Calling the confirm callback requires iterating the children of the removed directory and performing this iteration causes `removefile` to materialize the file on-disk. We should avoid this materialization when no callback exists - this change does this by only attaching callbacks to the `removefile_state_t` when we have a delegate that actually cares.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -152,6 +152,24 @@ extension FileManager {
         return delegateResponse ?? true
     }
     
+    fileprivate func _delegateImplementsRemoveItemConfirmation() -> Bool {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1071, opened 2024-12-05
Guard macro declarations behind `hasFeature(Macros)`

DougGregor:
To enable building this package with a "minimal" Swift toolchain whose compiler does not have support for macros, guard the macro declarations behind `#if hasFeature(Macros)`. This has no effect on anything but the toolchain bootstrap process.

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Expression.swift
+++ b/Sources/FoundationEssentials/Predicate/Expression.swift
@@ -27,6 +27,8 @@ public struct Expression<each Input, Output> : Sendable {
     }
 }
 
+#if hasFeature(Macros)
 @freestanding(expression)
 @available(FoundationPredicate 0.4, *)
 public macro Expression<each Input, Output>(_ body: (repeat each Input) -> Output) -> Expression<repeat each Input, Output> = #externalMacro(module: "FoundationMacros", type: "ExpressionMacro")
+#endif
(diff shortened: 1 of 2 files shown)


DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Predicate/Expression.swift
```diff
@@ -27,6 +27,8 @@ public struct Expression<each Input, Output> : Sendable {
     }
 }
 
+#if hasFeature(Macros)
```
jmschonfeld, line 30: Just confirming, do we need to worry about this in our unit tests which use the macro or are we not expected to ever try to build tests with a minimal compiler?

Additionally, do we need to handle avoiding building the macro itself? IIRC if the cmake build is not provided with a path to the built macro, it will attempt to build the macro automatically and SwiftPM builds will always build the macro. I suspect that we don't expect to be able to build via SwiftPM with a minimal toolchain, but is there anything that needs to be done on the CMake side to avoid invoking a build of the macro library when the compiler doesn't have this feature?

DougGregor: No, we don't have to worry about tests. We don't care whether that first stage passes tests, it just needs to build the second stage.

I don't think we need to avoid building the macro library, either. It's just unused.

jmschonfeld (approved):

PR #1073, opened 2024-12-06
[6.1] Home directory for non-existent user should not fall back to /var/empty or %ALLUSERSPROFILE%

jmschonfeld:
- **Explanation**: Requesting the home directory for a non-existent user now returns `nil` as it used to instead of `/var/empty` or `%ALLUSERSPROFILE%`. The current user's home directory still falls back to `/var/empty`/`%ALLUSERSPROFILE%` as it has previously.
    
  - **Scope**: Impacts requesting the home directory for a username that does not exist across all platforms
    
  - **Original PRs**: https://github.com/swiftlang/swift-foundation/pull/1072
    
  - **Risk**: Low-Medium - this does change behavior, however it restores swift 5.10 behavior, this is heavily tested by unit tests, and I don't expect this behavior to be fairly load bearing
    
  - **Testing**: local testing, swift-ci testing
    
  - **Reviewers**: @jrflat @kperryua

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -48,7 +48,13 @@ extension _FileManagerImpl {
     }
     
     func homeDirectory(forUser userName: String?) -> URL? {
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1074, opened 2024-12-07
Fix parsing foreign currency strings

itingliu:
Currently parsing a currency string would fail if the currency code does not match `FormatStyle`'s locale region. 

For example,
```swift
let style = Decimal.FormatStyle.Currency(code: "GBP", locale: .init(identifier: "en_US")).presentation(.isoCode)
```
    
This formats 3.14 into `"GBP\u{0xa0}3.14"`, but parsing such string fails. 

Fix this by always set the ICU formatter's currency code.
    
Resolves rdar://138179737

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
@@ -123,7 +123,7 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
     enum NumberFormatType : Hashable, Codable {
         case number(NumberFormatStyleConfiguration.Collection)
         case percent(NumberFormatStyleConfiguration.Collection)
-        case currency(CurrencyFormatStyleConfiguration.Collection, /*currency code*/ String)
+        case currency(CurrencyFormatStyleConfiguration.Collection, currencyCode: String)
         case descriptive(DescriptiveNumberFormatConfiguration.Collection)
     }
 
@@ -143,7 +143,7 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
                 }
             case .percent(_):
                 icuType = .percent
-            case .currency(let config, let _):
+            case .currency(let config, _):
                 icuType = config.icuNumberFormatStyle
             case .descriptive(let config):
                 icuType = config.icuNumberFormatStyle
--- a/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
@@ -123,7 +123,7 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
     enum NumberFormatType : Hashable, Codable {
         case number(NumberFormatStyleConfiguration.Collection)
         case percent(NumberFormatStyleConfiguration.Collection)
-        case currency(CurrencyFormatStyleConfiguration.Collection)
+        case currency(CurrencyFormatStyleConfiguration.Collection, /*currency code*/ String)
         case descriptive(DescriptiveNumberFormatConfiguration.Collection)
     }
 
@@ -143,7 +143,7 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
                 }
             case .percent(_):
                 icuType = .percent
-            case .currency(let config):
+            case .currency(let config, let _):
                 icuType = config.icuNumberFormatStyle
             case .descriptive(let config):
                 icuType = config.icuNumberFormatStyle
@@ -178,12 +178,13 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
                     }
                 }
 
-            case .currency(let config):
+            case .currency(let config, let currencyCode):
                 setMultiplier(config.scale, formatter: formatter)
                 setPrecision(config.precision, formatter: formatter)
                 setGrouping(config.group, formatter: formatter)
                 setDecimalSeparator(config.decimalSeparatorStrategy, formatter: formatter)
                 setRoundingIncrement(config.roundingIncrement, formatter: formatter)
+                try setTextAttribute(.currencyCode, formatter: formatter, value: currencyCode)
 
                 // Currency specific attributes
                 if let sign = config.signDisplayStrategy {
--- a/Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
+++ b/Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
@@ -595,6 +595,39 @@ final class NumberFormatStyleTests: XCTestCase {
         XCTAssertEqual((-922337203685477 as Decimal).formatted(baseStyle.rounded(rule: .awayFromZero, increment: 100)), "-$1000T")
     }
 
+    func testCurrency_Codable() throws {
+        let gbpInUS = Decimal.FormatStyle.Currency(code: "GBP", locale: enUSLocale)
+        let encoded = try JSONEncoder().encode(gbpInUS)
+        let json_gbp = String(data: encoded, encoding: String.Encoding.utf8)!
+        print(json_gbp)
+
+        let previouslyEncoded = """
+{
+    "collection":
+    {
+        "presentation":
+        {
+            "option": 1
+        }
+    },
+    "currencyCode": "GBP",
+    "locale":
+    {
+        "current": 0,
+        "identifier": "en_US"
+    }
+}
+""".data(using: .utf8)
+
+        guard let previouslyEncoded else {
+            XCTFail()
+            return
+        }
+
+        let decoded = try JSONDecoder().decode(Decimal.FormatStyle.Currency, from: previouslyEncoded)
+        XCTAssertEqual(decoded, gbpInUS)
+    }
+
     func testCurrency_scientific() throws {
         let baseStyle = Decimal.FormatStyle.Currency(code: "USD", locale: Locale(identifier: "en_US")).notation(.scientific)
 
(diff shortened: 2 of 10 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
```diff
@@ -1905,23 +1905,33 @@ final class FormatStylePatternMatchingTests : XCTestCase {
         _verifyMatching("€52,249", formatStyle: floatStyle, expectedUpperBound: nil, expectedValue: nil)
         _verifyMatching("€52,249", formatStyle: decimalStyle, expectedUpperBound: nil, expectedValue: nil)
 
-        // Different locale
         let frenchStyle: IntegerFormatStyle<Int>.Currency = .init(code: "EUR", locale: frFR)
-
-        let frenchPrice = "57 379 €"
-        _verifyMatching(frenchPrice, formatStyle: frenchStyle, expectedUpperBound: frenchPrice.endIndex, expectedValue: 57379)
-        _verifyMatching(frenchPrice, formatStyle: floatStyle.locale(frFR), expectedUpperBound: frenchPrice.endIndex, expectedValue: 57379)
```
itingliu, line 1913: The tests incorrectly assumed that a currency style with "USD" code but "fr_FR" locale can correctly match a currency string with EUR symbol (`"57 379,00 €"`). This isn't correct, and is exactly what this bug is about: we only respect the locale regional currency, but ignore the currency code.

Tests/FoundationInternationalizationTests/Formatting/NumberParseStrategyTests.swift
```diff
@@ -157,6 +157,67 @@ final class NumberParseStrategyTests : XCTestCase {
...
+            -87650000, -8765000, -876500, -87650, -8765, -876, -87, -8
+        ]
+
+        func _verifyRoundtripCurrency(_ testData: [Int], _ style: IntegerFormatStyle<Int>.Currency, _ testName: String = "", file: StaticString = #filePath, line: UInt = #line) {
+            for value in testData {
+                let str = style.format(value)
+                let parsed = try! Int(str, strategy: style.parseStrategy)
+                XCTAssertEqual(value, parsed, "\(testName): formatted string: \(str) parsed: \(parsed)", file: file, line: line)
+
+                let nonLenientParsed = try! Int(str, format: style, lenient: false)
```
theMomax, line 198: Using `try?` would probably be good in case this ever regresses in the future.

Sources/FoundationInternationalization/Formatting/Number/NumberFormatStyleConfiguration.swift
```diff
@@ -332,6 +332,7 @@ public enum CurrencyFormatStyleConfiguration {
         var roundingIncrement: RoundingIncrement?
         var presentation: Presentation
         var notation: Notation?
+        var currencyCode: String
```
theMomax, line 335: This would break the backwards compatibility of the `Decodable` conformance, i.e. decoding a currency format style that was encoded by a prior version of foundation would fail. We should probably make this optional and fall back to the locale's currency code if the value is not present here.

theMomax (commented):
Generally looks good! I think we just need to make sure the serialization format is backwards compatible, then this should be good to go!

Sources/FoundationInternationalization/Formatting/Number/NumberFormatStyleConfiguration.swift
itingliu: You're right, thanks. I'll fix that up and mark this ready for review once that's done.

Slightly related, we started requiring explicit `Codable` implementation a while ago, but these types predated that practice. I'm really tempted to do that together here although not sure if that's a good idea 😞

itingliu (commented):
I like this approach better than the previous one. We already have the currency code stored in the `FormatStyle` as a top-level property. There's no need to store it again in the nested configuration.

Also this should address @theMomax your concern about changing `Codable` representation. This should not change the serialized output anymore since we're not touching any `Codable` type in public types.

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
```diff
@@ -595,6 +595,39 @@ final class NumberFormatStyleTests: XCTestCase {
         XCTAssertEqual((-922337203685477 as Decimal).formatted(baseStyle.rounded(rule: .awayFromZero, increment: 100)), "-$1000T")
     }
 
+    func testCurrency_Codable() throws {
+        let gbpInUS = Decimal.FormatStyle.Currency(code: "GBP", locale: enUSLocale)
+        let encoded = try JSONEncoder().encode(gbpInUS)
+        let json_gbp = String(data: encoded, encoding: String.Encoding.utf8)!
+        print(json_gbp)
```
theMomax, line 602: Left over debug print?

Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
```diff
@@ -595,6 +595,39 @@ final class NumberFormatStyleTests: XCTestCase {
...
+        }
+    },
+    "currencyCode": "GBP",
+    "locale":
+    {
+        "current": 0,
+        "identifier": "en_US"
+    }
+}
+""".data(using: .utf8)
```
theMomax, line 620: You should be able to maintain proper indentation here. Any indentation that is present on the closing `"""` won't appear in the actual string.

Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
```diff
@@ -123,7 +123,7 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
     enum NumberFormatType : Hashable, Codable {
         case number(NumberFormatStyleConfiguration.Collection)
         case percent(NumberFormatStyleConfiguration.Collection)
-        case currency(CurrencyFormatStyleConfiguration.Collection)
+        case currency(CurrencyFormatStyleConfiguration.Collection, /*currency code*/ String)
```
theMomax, line 126:
```suggestion
        case currency(CurrencyFormatStyleConfiguration.Collection, currencyCode: String)
```
Swift actually allows this syntax, but not sure about the Foundation style guidelines 😄

theMomax, line 126: This type is part of the `Codable` format of a number of public types conforming to `ParseStrategy`, e.g. `IntegerParseStrategy`. So while this solution means the Codable conformance for `Decimal.FormatStyle.Currency` is backwards compatible, your decoding test would still fail for `IntegerParseStrategy`.

theMomax (commented):
This new approach is definitely better because I think people serializing `FormatStyle`s is much more common than serializing `ParseStrategy`s. But I think we should still make it backwards compatible in any scenario. Could we make `currencyCode` an optional property on `CurrencyFormatStyleConfiguration.Collection`? That way you could still avoid writing out a custom codable conformance.

Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
itingliu: > This type is part of the Codable format of a number of public types conforming to ParseStrategy, e.g. IntegerParseStrategy

I'm not sure what you meant by this. This is part of `ICULegacyNumberFormatter`, and I don't think we're encoding this type anywhere, unless I missed something?

What's a test that would fail because of this change?

itingliu: Ah, you meant this is a var in `FloatingPointParseStrategy`... OK we need test for that too.

itingliu:
> Could we make currencyCode an optional property on CurrencyFormatStyleConfiguration.Collection? That way you could still avoid writing out a custom codable conformance.

I'm flipping back and forth. A custom `Codable` conformance feels brittle as well because anyone could easily change the struct without realizing the `Codable` implementation also needs to be updated. But on the other hand with a `Codable` implementation it at least makes it clear what properties are expected in the serialization

Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
itingliu: - [ ] Add codable test for `ParseStrategy`

itingliu: Another attempt: In [9a1e617](https://github.com/swiftlang/swift-foundation/pull/1074/commits/9a1e617c6affea242c994cdbf45fddb686f6115b) I removed the stored property `numberFormatType` from `IntegerParseStrategy` and `FloatingPointParseStrategy`. That means the we no longer preserve `numberFormatType` in serialization, and `numberFormatType` in previous `Codable` representation will be ignored from now on. 

This is ok because the information is already provided by the other variable, `formatStyle`. In other words, `numberFormatType` has always been redundant. I also liked this because the previous serialization output feels like we're leaking internal information. 

There is indeed a slight chance of regression -- if you manually craft an archive where `numberFormatType` represents different configurations from that of `formatStyle`, previously we'd favor what's represented by `numberFormatType`, but now we'd favor `formatStyle`. I think this risk is low enough though. Besides, I don't think we ever guarantee compatibility for manually crafted archives.

itingliu:
@swift-ci please test

theMomax (approved):
That sounds like a good solution, thank you!

itingliu:
@swift-ci please test

PR #1077, opened 2024-12-09
Properly adjust invalid dates in RecurrenceRule enumeration

hristost:
In recurrence rules without a strict matching policy, we'd sometimes come across dates which don't exist (such as February 29, 2009). When adjusting these dates, we were passing unadjusted date components. That resulted in adjusted dates with the highest unspecified component different than the original date. For example, adjusting February 29, 2009 to match components {month: 2, day: 29} would result February 29, 2012, instead of Match 1 or February 28 of the same year.

Adjusting the date components (that is, setting the year to 2009 for the example above) fixes this. This is what we already do in Calendar.dates(byMatching: ...)

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
@@ -426,6 +426,29 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
         ]
         XCTAssertEqual(results, expectedResults)
     }
+
+    func testYearlyRecurrenceMovingToFeb29() {
+        /// Rule for an event that repeats on February 29th of each year, or closest date after
+        var rule = Calendar.RecurrenceRule(calendar: gregorian, frequency: .yearly, matchingPolicy: .nextTimePreservingSmallerComponents)
+        rule.months = [2]
+        rule.daysOfTheMonth = [29]
+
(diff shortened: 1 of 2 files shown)


hristost:
@swift-ci please test

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -426,6 +426,29 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
...
+    func testYearlyRecurrenceMovingToFeb29() {
+        /// Rule for an event that repeats on February 29th of each year, or closest date after
+        var rule = Calendar.RecurrenceRule(calendar: gregorian, frequency: .yearly, matchingPolicy: .nextTimePreservingSmallerComponents)
+        rule.months = [2]
+        rule.daysOfTheMonth = [29]
+
+        let rangeStart = Date(timeIntervalSince1970: 946684800.0) // 2000-01-01T00:00:00-0000
+        let rangeEnd = Date(timeIntervalSince1970: 1262304000.0) // 2010-01-01T00:00:00-0000
+        var birthdays = rule.recurrences(of: rangeStart, in: rangeStart..<rangeEnd).makeIterator()
+        //                               ^ The initial date here does not matter, since the rule will change the month and day
```
itingliu, line 439: I thought the initial date does matter in a sense -- it does affect hour, minute and second of the results. Perhaps rephrase this comment to clarify that the initial date does not affect specifically the month and day of the results?

itingliu (approved):

hristost:
@swift-ci please test

PR #1079, opened 2024-12-10
Updated SF-0007 to version 5

iCharlesHu:
This PR updates the proposal for `SF-0007 Swift Subprocess` to version 5 and added header docs. Here are the changes since version 3:

- **v4**: Minor updates (primarily name changes):
    - Dropped the `executing` label from `.run()`.
    - Removed references to `Subprocess.AsyncBytes`:
        - Instead, we use an opaque type: `some AsyncSequence<UInt8, Error>`.
        - When `typed throws` is ready, we can then update to the actual error type.
    - Updated `.standardOutput` and `.standardError` properties on `Subprocess` and `CollectedResult` to be non-optional (now they use `fatalError` instead).
        - Rationale: These properties can only be `nil` in two cases:
            1. The corresponding `.output`/`.error` was not set to `.redirectToSequence` when `run()` was called.
            2. These properties are accessed multiple times. This is because these `AsyncSequence`s are reading pipes under the hood, and pipes can only be read once.
            - Both cases can’t be resolved until the source code is updated; they are therefore considered programming errors.
    - Updated `StandardInputWriter` to be non-sendable.
    - Renamed `PlatformOptions.additionalAttributeConfigurator` to `Platform.preSpawnAttributeConfigurator`; renamed `PlatformOptions.additionalFileAttributeConfigurator` to `Platform.preSpawnFileAttributeConfigurator`.
    - Updated all `closeWhenDone` parameter labels to `closeAfterSpawningProcess`.
    - Renamed `Subprocess.Result` to `Subprocess.ExecutionResult`.
    - Added `Codable` support to `TerminationStatus` and `ExecutionResult`.
    - Renamed `TerminationStatus.exit`:
        - From `.exit` to `.exited`.
        - From `.wasUnhandledException` to `.isUnhandledException`.
    - Added two sections under `Future Direction`: `Support Launching Long Running Processes` and `Process Piping`.
    - Added Linux-specific `PlatformOptions`: `.closeAllUnknownFileDescriptors`.
        - This option attempts to emulate Darwin’s `POSIX_SPAWN_CLOEXEC_DEFAULT` behavior. It is the default value on Darwin.
        - Unfortunately, `posix_spawn` does not support this flag natively, hence on Linux this behavior is opt-in, and we will fall back to a custom implementation of `fork/exec`.
- **v5**: Platform-specific changes, `Subprocess.runDetached`, and others:
    - Added `Hashable`, `CustomStringConvertable` and `CustomDebugStringConvertible` conformance to `Subprocess.Configuration` and friends
    - `Subprocess.Arguments`:
        - Add an array initializer to `Subprocess.Arguments`:
            - `public init(_ array: [String])`
            - `public init(_ array: [Data])`.
    - `Subprocess.CollectedOutputMethod`:
        - Combined `.collect` and `.collect(upTo:)`
    - `Subprocess.PlatformOptions` (all platforms):
        - Changed from `.default` to using empty initializer `.init()`.
        - Changed to prefer platform native types such as `gid_t` over `Int`.
    - Darwin Changes:
        - Updated `PlatformOptions.createProcessGroup` to `PlatformOptions.processGroupID`.
            - Also changed the public init.
        - Combined `PlatformOptions.preSpawnAttributeConfigurator` and `PlatformOptions.preSpawnFileAttributeConfigurator` into `PlatformOptions.preSpawnProcessConfigurator` to be consistant with other platforms.
    - Linux Changes:
        - Updated `PlatformOptions` for Linux.
        - Introduced `PlatformOptions.preSpawnProcessConfigurator`.
    - Windows Changes:
        - Removed `Arguments` first argument override and non-string support from Windows because Windows does not support either.
        - Introduced `PlatformOptions` for Windows.
        - Replaced `Subprocess.send()` with
            - `Subprocess.terminate(withExitCode:)`
            - `Subprocess.suspend()`
            - `Subprocess.resume()`
    - `Subprocess.runDetached`:
        - Introduced `Subprocess.runDetached` as a top level API and sibling to all `Subprocess.run` methods. This method allows you to spawn a subprocess **WITHOUT** needing to wait for it to finish.
    - Updated `.standardOutput` and `.standardError` properties on `Subprocess` to be `AsyncSequence<Data, any Error>` instead of `AsyncSequence<UInt8, any Error>`.
        - The previous design, while more "traditional", leads to performance problems when the subprocess outputs large amount of data
    - Teardown Sequence support (for Darwin and Linux):
        - Introduced `Subprocess.teardown(using:)` to allow developers to gracefully shutdown a subprocess.
        - Introuuced `PlatformOptions.teardownSequence` that will be used to gracefully shutdown the subprocess if the parent task is cancelled.

### DIFF
--- a/Proposals/0007-swift-subprocess.md
+++ b/Proposals/0007-swift-subprocess.md
@@ -3,7 +3,7 @@
 * Proposal: [SF-0007](0007-swift-subprocess.md)
 * Authors: [Charles Hu](https://github.com/iCharlesHu)
 * Review Manager: [Tina Liu](https://github.com/itingliu)
-* Status: **Active Review: Feb 28, 2024...Mar 06, 2024**
+* Status: **2nd Review, Active: Dec 12, 2024...Dec 19, 2024**
 * Bugs: [rdar://118127512](rdar://118127512), [apple/swift-foundation#309](https://github.com/apple/swift-foundation/issues/309)
 
 
--- a/Proposals/0007-swift-subprocess.md
+++ b/Proposals/0007-swift-subprocess.md
@@ -14,17 +14,71 @@
     - Switched `AsyncBytes` to be backed by `DispatchIO`.
     - Introduced `resolveExecutablePath(withEnvironment:)` to enable explicit lookup of the executable path.
     - Added a new option, `closeWhenDone`, to automatically close the file descriptors passed in via `.readFrom` and friends.
-    - Introduced a new parameter, `shouldSendToProcessGroup`, in the `sendSignal` function to control whether the signal should be sent to the process or the process group.
+    - Introduced a new parameter, `shouldSendToProcessGroup`, in the `send()` function to control whether the signal should be sent to the process or the process group.
     - Introduced a section on "Future Directions."
 * **v3**: Minor updates:
     - Added a section describing `Task Cancellation`
     - Clarified for `readFrom()` and `writeTo()` Subprocess will close the passed in file descriptor right after spawning the process when `closeWhenDone` is set to true.
     - Adjusted argument orders in `Arguments`.
     - Added `Subprocess.run(withConfiguration:...)` in favor or `Configuration.run()`.
+- **v4**: Minor updates (primarily name changes):
+    - Dropped the `executing` label from `.run()`.
+    - Removed references to `Subprocess.AsyncBytes`:
+        - Instead, we use an opaque type: `some AsyncSequence<UInt8, Error>`.
+        - When `typed throws` is ready, we can then update to the actual error type.
+    - Updated `.standardOutput` and `.standardError` properties on `Subprocess` and `CollectedResult` to be non-optional (now they use `fatalError` instead).
+        - Rationale: These properties can only be `nil` in two cases:
+            1. The corresponding `.output`/`.error` was not set to `.redirectToSequence` when `run()` was called.
+            2. These properties are accessed multiple times. This is because these `AsyncSequence`s are reading pipes under the hood, and pipes can only be read once.
+            - Both cases can’t be resolved until the source code is updated; they are therefore considered programming errors.
+    - Updated `StandardInputWriter` to be non-sendable.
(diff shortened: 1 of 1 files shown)


Proposals/0007-swift-subprocess.md
```diff
@@ -286,92 +459,305 @@ Both styles of the `run` methods provide convenient overloads that allow develop
...
+    ///   - executable: The executable to run.
+    ///   - arguments: The arguments to pass to the executable.
+    ///   - environment: The environment to use for the process.
+    ///   - workingDirectory: The working directory for the process.
+    ///   - platformOptions: The platform specific options to use for the process.
+    ///   - input: A file descriptor to bind to the subprocess' standard input.
+    ///   - output: A file descriptor to bind to the subprocess' standard output.
+    ///   - error: A file descriptor to bind to the subprocess' standard error.
+    /// - Returns: the process identifier for the subprocess.
+    public static func runDetached(
```
FranzBusch, line 553: Is this in the end just spawning an unstructured Task and calling `run` inside that task? If that's the case I would really prefer we don't offer that API and just expect users to construct that unstructured task themselves. This will allow them to still have some ownership of the task and cancel it if needed.

itingliu (changes requested):
- [x] Please change the status (on line 6) to 

```
* Status: **2nd Review, Active: Dec 12, 2024...Dec 19, 2024**
```

itingliu:
@swift-ci please test

Proposals/0007-swift-subprocess.md
iCharlesHu: No there is no unstructured task. This API is not `async` at all. It simply spawns the process and returns the PID to you as if you were calling `posix_spawn` yourself.

iCharlesHu:
> * [x]  Please change the status (on line 6) to
> 
> ```
> * Status: **2nd Review, Active: Dec 12, 2024...Dec 19, 2024**
> ```

Updated! Thanks!

iCharlesHu:
@swift-ci please test

itingliu (approved):

PR #1081, opened 2024-12-11
Ensure that FileManager.copyItem cannot copy directory metadata to files

jmschonfeld:
`FileManager.copyItem(atPath:toPath:)` has a TOCTOU bug due to how it recursively copies directories. As we recursively copy directories, we copy the directory itself with default metadata/permissions, copy the contents of the directory, then copy the metadata over (which may change permissions on the directory, including potentially making it read-only). This TOCTOU bug allows another process with write access to the source directory to (with the right timing) cause a process that uses `FileManager.copyItem` to copy and apply metadata to a file rather than a directory.

To fix the TOCTOU issue, the post-order traversal which copies the metadata now:

- Opens a file descriptor to both the source and destination directories
- `fstats` the source and destination directories to validate that they are directories (and not files)
- Then copies metadata from one open file descriptor to the other
    - This uses `fcopyfile` on Darwin, and a custom function on Linux where `fcopyfile` does not exist

This ensures that we only copy metadata between directories and cannot accidentally apply metadata to a file.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -948,7 +948,7 @@ enum _FileOperations {
         #endif
     }
     #endif
-
+    
     #if !canImport(Darwin)
     private static func _copyDirectoryMetadata(srcFD: CInt, srcPath: @autoclosure () -> String, dstFD: CInt, dstPath: @autoclosure () -> String, delegate: some LinkOrCopyDelegate) throws {
         // Copy extended attributes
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:

jmschonfeld:
@swift-ci please test macOS Platform

Issue #1080, opened 2024-12-11
`URL.appendingPathExtension("")` appends a trailing dot


kateinoigakukun:
```swift
import Foundation
let url = URL(fileURLWithPath: "/tmp/x")
print(url.appendingPathExtension(""))
// Swift 5.10: file:///tmp/x
// Swift 6.0: file:///tmp/x.
```

This behavior change might be tricky as `url.deletingPathExtension().appendingPathExtension(url.pathExtension)` won't be equal to `url`.

Https://github.com/swiftlang/swift-driver/pull/1756 revealed this change. Could you check if this is an intentional change? @jrflat

jrflat:
Not intended, I'll push a fix, thanks for finding this!

kateinoigakukun:
@jrflat Thanks! Can we cherry-pick this fix to 6.1 so that we will be able to remove a workaround once we drop 6.0 support?

jrflat:
Yes sounds good I'll cherry-pick to 6.1

PR #1082, opened 2024-12-11
URL.appendingPathExtension("") should not append a trailing dot

jrflat:
`URL.appendingPathExtension("")` should return the URL unchanged instead of appending a trailing dot.

Resolves https://github.com/swiftlang/swift-foundation/issues/1080


jrflat:
@swift-ci please test

jmschonfeld (approved):

Issue #1084, opened 2024-12-12
Permission error on Windows isn't providing the user variant key


cmcgee1024:
I'm getting the following error in a complex application (swiftpm) on Windows and I don't know what kind of operation led to the error, such as remove/copy/move/link so that I can try and remedy it.

```
Error Domain=NSCocoaErrorDomain Code=513 "You don't have permission."UserInfo={NSFilePath=C:\Users\cmcgee\swiftpm\foo\.build\x86_64-unknown-windows-msvc\Intermediates.noindex\swift-argument-parser.build\Debug-windows\ArgumentParserToolInfo_Module.build\ArgumentParserToolInfo.modulemap, NSUnderlyingError=Win32Error(code: 5), NSURL=file:///C:/Users/cmcgee/swiftpm/foo/.build/x86_64-unknown-windows-msvc/Intermediates.noindex/swift-argument-parser.build/Debug-windows/ArgumentParserToolInfo_Module.build/ArgumentParserToolInfo.modulemap, NSSourceFilePathErrorKey=C:\Users\cmcgee\swiftpm\foo\.build\x86_64-unknown-windows-msvc\Intermediates.noindex\swift-argument-parser.build\Debug-windows\ArgumentParserToolInfo_Module.build\ArgumentParserToolInfo.modulemap, NSDestinationFilePath=C:\Users\cmcgee\swiftpm\foo\.build\x86_64-unknown-windows-msvc\Intermediates.noindex\GeneratedModuleMaps\windows}
```

PR #1085, opened 2024-12-12
[Windows] Include variant in file path errors

jmschonfeld:
A small typo causes file path errors on Windows to not include the variant of the API that produced the error. This patch ensures we pass the variant down to the error constructor

Resolves https://github.com/swiftlang/swift-foundation/issues/1084


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #1092, opened 2024-12-12
Add linked-on-or-after check for empty URL.host()

jrflat:
This PR adds a linked-on-or-after check for apps which expect `URL.host()` to return `nil` for URLs like `https:///`. The new `.host()` implementation returns an empty string because according to RFC 3986, a host always exists if there is an authority component, it just might be empty.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1093, opened 2024-12-13
AttributedString Scope Enumeration

jmschonfeld:
Adds a new `AttributeScope` property to provide a public sequence of attribute keys within the scope


jmschonfeld:
@swift-ci please test

PR #1094, opened 2024-12-13
Fix WASI build of `_copyDirectoryMetadata`

kateinoigakukun:
Extended attributes don't exist in WASI, so we need to exclude the use of xattr-related APIs including `flistxattr`.


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #1097, opened 2024-12-16
Add compatibility check for malformed URL(filePath:) input

jrflat:
Add linked-on-or-after check for apps passing a full `file:` URL string to `URL(fileURLWithPath:)` instead of just the file path. In the old implementation, this would rarely happen to work if the app immediately called `.appendingPathComponent(_:)` after, which previously would interpret a relative string starting with `"file:"` as an absolute file URL string when creating the new `URL`.


jrflat:
@swift-ci please test

parkera (approved):

Issue #1100, opened 2024-12-19
Static Linux SDK: `SIGSEGV: invalid address` when concurrently creating `Calendar` instances.


barnard-b:
Creating instances of `Calendar` concurrently on Linux (Static SDK) reliably results in a segmentation fault. It occurs in both debug and release configurations. The same code runs without issue on macOS.

**Host:** macOS 15.2 (24C101), Xcode 16.2 (16C5032a)
**Swift:** Apple Swift version 6.0.3 (swift-6.0.3-RELEASE), arm64-apple-macosx15.0
**Static Linux SDK:** swift-6.0.3-RELEASE_static-linux-0.0.1
**Target:** Alma Linux 8.10 (x86_64 Static SDK)

### FoundationTest.swift
```swift
import Foundation

@main
struct App {
    static func main() async throws {
        await withTaskGroup(of: Void.self) { group in
            for _ in 0...1_000_000 {
                group.addTask {
                    _ = Calendar(identifier: .gregorian)
                }
            }
        }
    }
}
```

### Package.swift
```swift
// swift-tools-version: 6.0

import PackageDescription

let package = Package(
    name: "FoundationTest",
    platforms: [
        .macOS(.v12)
    ],
    targets: [
        .executableTarget(name: "FoundationTest")
    ]
)
```

#### Build for Linux
```shell
xcrun --toolchain swift swift build --swift-sdk x86_64-swift-linux-musl
```
#### Build for macOS
```shell
xcrun --toolchain swift swift build
```

## Core Dump (Linux)
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000000007df69a in swift_isUniquelyReferenced_nonNull_native ()
[Current thread is 1 (LWP 11637)]
(gdb) bt
#0  0x00000000007df69a in swift_isUniquelyReferenced_nonNull_native ()
#1  0x0000000000a6ae14 in $sSD8_VariantV8setValue_6forKeyyq_n_xtFSi_20FoundationEssentials17_TimeZoneProtocol_pTg5Tf4enn_n ()
#2  0x0000000000a66908 in $s20FoundationEssentials13TimeZoneCacheV5StateV011findCurrentcD0AA0cD0VyF ()
#3  0x0000000000a68ead in $s20FoundationEssentials13TimeZoneCacheV7defaultAA0cD0VvgAfC5StateVzYbXEfU_ ()
[...]

PR #1101, opened 2024-12-20
Implement LockedState for Musl

jmschonfeld:
Unfortunately, we're missing a few conditionals in the `LockedState` implementation that left `LockedState` acting as a no-op for the static linux SDK. This results in the lack of synchronization of caches in swift-foundation for that platform. I went ahead and added the right conditionals for Musl (same as Glibc) and also added an explicit `#else` case with a `#error` so that we can more quickly and easily detect platforms for which we have not implemented synchronization primitives here.

I've confirmed that running the sample program in the linked issue produces the segmentation fault but with this change the fault no longer occurs.

Resolves https://github.com/swiftlang/swift-foundation/issues/1100

### DIFF
--- a/Sources/FoundationEssentials/LockedState.swift
+++ b/Sources/FoundationEssentials/LockedState.swift
@@ -46,17 +46,19 @@ package struct LockedState<State> {
         fileprivate static func initialize(_ platformLock: PlatformLock) {
 #if canImport(os)
             platformLock.initialize(to: os_unfair_lock())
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

itingliu (approved):

iCharlesHu (approved):

parkera (approved):

PR #1103, opened 2024-12-20
Restore behavior of URL(string: "") returning nil

### LINKED ISSUE #962: URL.init(string:) handling of empty strings has changed in Swift 6
Initializing a URL from an empty string in Swift 6 on Linux now returns a non-nil URL value. In earlier Swift versions and on macOS it returns `nil`.

On Linux Swift 5.10:
```
  1> import Foundation
  2> let u = Foundation.URL(string: "")
  [snip]
  3> print(u)
nil
```

On Linux Swift 6.0.1:
```
  1> import Foundation
  2> let u = Foundation.URL(string: "")
  [snip]
  3> print(u)
Optional()
```

jrflat:
After receiving more reports of bincompat issues, we're restoring the previous behavior of `URL(string: "")` returning `nil`. `URL(string: "", relativeTo: baseURL)` will still be valid input, allowing us to pass all the RFC 3986 examples.

Resolves #962

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -777,19 +777,12 @@ public struct URL: Equatable, Sendable, Hashable {
     ///
     /// Returns `nil` if a `URL` cannot be formed with the string (for example, if the string contains characters that are illegal in a URL, or is an empty string).
     public init?(string: __shared String) {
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1106, opened 2024-12-31
Don't try to set extended attributes on Android

finagolfin:
Normal users don't have permission to change these, even for their own files.

@jmschonfeld, as [done previously for Android](https://github.com/swiftlang/swift-foundation/pull/871#discussion_r1722073325).


finagolfin:
@swift-ci please test

jmschonfeld (approved):

PR #1108, opened 2025-01-02
(140288075) Use-after-free in Swift implementation of NSData

phausler:
This is a confluence of a number of other fixes that exposed a flaw when combined - a UAF bug could occur when causing a retain of an underlying buffer from a no-copy initializer. This forces all bridging of no-copy cases to actually copy by robbing a bit from the capacity.


jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #1109, opened 2025-01-02
AttributedString Index Tracking

jmschonfeld:
This implements tracking of `AttributedString` indices across mutations

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
@@ -1,0 +1,168 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if FOUNDATION_FRAMEWORK
+@_spi(Unstable) internal import CollectionsInternal
+#elseif canImport(_RopeModule)
+internal import _RopeModule
+#elseif canImport(_FoundationCollections)
+internal import _FoundationCollections
+#endif
+
+// MARK: - Internal Index Updating
+
+extension AttributedString.Guts {
+    func _prepareTrackedIndicesUpdate(mutationRange: Range<BigString.Index>) {
+        // Move any range endpoints inside of the mutation range to outside of the mutation range since a range should never end up splitting a mutation
+        for idx in 0 ..< trackedRanges.count {
+            let lowerBoundWithinMutation = trackedRanges[idx].lowerBound > mutationRange.lowerBound && trackedRanges[idx].lowerBound < mutationRange.upperBound
+            let upperBoundWithinMutation = trackedRanges[idx].upperBound > mutationRange.lowerBound && trackedRanges[idx].upperBound < mutationRange.upperBound
+            switch (lowerBoundWithinMutation, upperBoundWithinMutation) {
+            case (true, true):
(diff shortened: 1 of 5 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
```diff
@@ -0,0 +1,168 @@
...
+            var upperBound = trackedRanges[idx].upperBound
+            
+            // Shift the lower bound if either:
+            //      A) The lower bound is greater than the start of the mutation (meaning it must be after the mutation due to the prepare step)
+            //      B) The lower bound is equal to the start of the mutation, but the mutation is an insertion (meaning the text is inserted before the start offset)
+            if lowerBound.utf8Offset > mutationStartOffset || (lowerBound.utf8Offset == mutationStartOffset && isInsertion), utf8LengthDelta != 0 {
+                lowerBound = string.utf8.index(string.startIndex, offsetBy: lowerBound.utf8Offset + utf8LengthDelta)
+            } else {
+                // Form new indices even if the offsets don't change to ensure the indices are valid in the newly-mutated rope
+                string.utf8.formIndex(&lowerBound, offsetBy: 0)
```
itingliu, line 59: Is this effectively the same as `string.utf8.startIndex`? and if so is there a reason for choosing one over the other?

Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
```diff
@@ -0,0 +1,168 @@
...
+    func _finalizeTrackedIndicesUpdate(mutationStartOffset: Int, isInsertion: Bool, utf8LengthDelta: Int) {
+        // Update indices to point to the correct offsets based on the mutation deltas
+        for idx in 0 ..< trackedRanges.count {
+            var lowerBound = trackedRanges[idx].lowerBound
+            var upperBound = trackedRanges[idx].upperBound
+            
+            // Shift the lower bound if either:
+            //      A) The lower bound is greater than the start of the mutation (meaning it must be after the mutation due to the prepare step)
+            //      B) The lower bound is equal to the start of the mutation, but the mutation is an insertion (meaning the text is inserted before the start offset)
+            if lowerBound.utf8Offset > mutationStartOffset || (lowerBound.utf8Offset == mutationStartOffset && isInsertion), utf8LengthDelta != 0 {
```
itingliu, line 55: The comment suggests that you want to shift the lower bound if either of these condition satisfies, but it doesn't mention how `utf8LengthDelta` affects this.  Just checking that we're not missing this just because `utf8LengthDelta == 0`

itingliu (approved):

Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
jmschonfeld: Not quite - `lowerBound` may not be the start index of the `AttributedString`, we just know that the raw UTF-8 offset doesn't need to change based on the mutation. This is just a way to "regenerate" a new `BigString.Index` with a valid rope path based on the existing offset. That existing offset may be at the start of the string or somewhere in the middle, but it doesn't change - we just want to make sure that any additional information in the index about the tree structure gets updated too based on the new tree layout.

Sources/FoundationEssentials/AttributedString/AttributedString+IndexTracking.swift
jmschonfeld: Ah right the comment doesn't mention that - that check is just so that if the mutation doesn't actually change the length of the string (i.e. "ABC" was replaced with "DEF") then there's no need to re-calculate offsets, we just do the same thing we would have done if the index was before the mutation. I can add a note to the comment here to make that a bit clearer

jmschonfeld:
@swift-ci please test

PR #1110, opened 2025-01-03
(142076445) Allow URL.standardized to return an empty string URL

jrflat:
Calling `URL.standardized` on a URL like `../../../../` removes the dot segments and produces an empty string. `NSURL` allows the empty string, and in the old implementation, this would return a `URL` that wraps an empty string `NSURL`.

In the new implementation, we ultimately call `URL(string: "")` when initializing this standardized URL, which returns `nil` again after #1103. Since `URL.standardized` returns a non-nil `URL`, this would crash when unwrapped.

We can allow `URL.standardized` (and other methods that call `URLComponents.url`) to return an empty string `URL` as they did previously by using `URL(string:relativeTo:)` instead.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1847,7 +1847,7 @@ public struct URL: Equatable, Sendable, Hashable {
         var components = URLComponents(parseInfo: _parseInfo)
         let newPath = components.percentEncodedPath.removingDotSegments
         components.percentEncodedPath = newPath
-        return components.url(relativeTo: baseURL)!
+        return components.url(relativeTo: baseURL) ?? self
     }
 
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

itingliu:
Would it make sense to avoid force unwrapping before returning from `standardized` at all in case there are other fallout paths? `URL(string:relativeTo:)` returns an optional after all, so the chance of invalid force unwrap is still there.

jrflat:
Yeah, I was thinking it could limit our ability to find other cases, but that is probably the safer approach, and I could always re-add the force-unwrap when testing inputs at desk. I'll add a `?? self`.

jrflat:
@swift-ci please test

itingliu (approved):

jrflat:
@swift-ci please test

PR #1113, opened 2025-01-08
Compatibility behaviors for Swift URL

jrflat:
This PR addresses main behavioral differences of the new Swift `URL` implementation in order to preserve bincompat.

- Adds a compatibility path to the parser allowing the scheme to be empty for `URL`. This was the behavior in the old implementation, and can cause `URL` to return different components for URL strings starting with `:`.
- Returns `nil` host for URLs like `https:///` instead of returning empty string. RFC 3986 suggests this should be an empty string, but I don't think this conformance outweighs compatibility here.
- Linked-on-or-after-check where `URL(filePath:)` will check for full `file:` and `http:` URL strings that were incorrectly passed. It drops the `file:` scheme or passes the `http:` string to `URL(string:)` instead. It also logs an "API MISUSE" fault in these cases. This fixes apps that rely on old buggy behavior where passing a full URL string as `filePath` could happen to work itself out if a different API like `.appendingPathComponent(_:)` is called immediately after.
- Makes `.appending(component:)` behave like `.appending(path:)` for file URLs, so that it does not percent-encode `/` to `%2F`. The old implementation had a bug where this actually didn't happen for file URLs, so we should maintain that old behavior to prevent breakage.
- Restores `URL(string:relativeTo:)` returning `nil` for empty string and instead uses an internal `URL(stringOrEmpty:relativeTo:)` in cases where we do need to initialize with the empty string, such as when `.deletingLastPathComponent()` of a single-component path. We can still pass the empty string RFC tests using `URL(stringOrEmpty:relativeTo:)`.
- Restores file path initializers' behavior of replacing an empty input path with `./` or `.`

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -616,6 +616,10 @@ internal func foundation_swift_url_enabled() -> Bool {
 internal func foundation_swift_url_enabled() -> Bool { return true }
 #endif
 
+#if canImport(os)
+internal import os
+#endif
+
 /// A URL is a type that can potentially contain the location of a resource on a remote server, the path of a local file on disk, or even an arbitrary piece of encoded data.
 ///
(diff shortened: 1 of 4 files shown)


jrflat:
@swift-ci please test

parkera (approved):
Looks like a good compromise for compatibility.

PR #1114, opened 2025-01-08
Compatibility path for apps looping on URL.deletingLastPathComponent().standardized

jrflat:
Compatibility path for apps that recurse on `url = url.deletingPathComponent().standardized` until `url.path.isEmpty`.

Before Swift `URL`, this used to work due to a combination of two bugs where:
```
URL("/").deletingLastPathComponent == URL("/../")
URL("/../").standardized == URL("")
```
both of which are incorrect behaviors. This PR adds a compatibility path to prevent the infinite loop and crash.


jrflat:
@swift-ci please test

itingliu (approved):

PR #1118, opened 2025-01-10
Allow URLParser to be called directly from ObjC

jrflat:
Allow compatibility-mode Swift URL parsing to be called directly from ObjC, saving an `NSURLComponents` allocation and improving performance.


jrflat:
@swift-ci please test

itingliu (approved):

PR #1119, opened 2025-01-13
URL.absoluteString crashes if baseURL starts with colon

jrflat:
Recent regression from reverting to compatibility `URL` behaviors in #1113.

`URL.absoluteString` uses `URLComponents` to build the string from the correct parts of the relative and base URLs. After allowing `URL` to have an empty scheme for compatibility in #1113, `URL.absoluteString` would crash if we used an empty scheme from the base URL. The crash occurs when we assign `components.scheme = ""` since `URLComponents` does not consider an empty scheme valid.

Since we are only using `URLComponents` to produce an absolute string, we can bypass the scheme validation with an internal function, since the scheme used for the URL's `absoluteString` has already been validated by the base `URL`.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1123,7 +1123,9 @@ public struct URL: Equatable, Sendable, Hashable {
         }
 
         if let baseScheme = _baseParseInfo.scheme {
-            result.scheme = String(baseScheme)
+            // Scheme might be empty, which URL allows for compatibility,
+            // but URLComponents does not, so we force it internally.
+            result.forceScheme(String(baseScheme))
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test macOS Platform

jrflat:
@swift-ci please test macOS Platform

jrflat:
@swift-ci please test macOS Platform

jrflat:
@swift-ci please test macOS Platform

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1121, opened 2025-01-15
Fix `_copyDirectoryMetadata` compilation error on FreeBSD

michael-yuji:
`_copyDirectoryMetadata` does not compile on FreeBSD since `f{list,get}xattr` is not available. Use `extattr_*_fd` to copy extended attributes recursively instead.

Unlike the Linux variant, `extxattr_*(2)` does not return an entry list with null terminating attribute names. Instead, each list entry consists of a single byte containing the length of the attribute name follow by the non-null terminating name itself.

To iterate each entry:
- we prepare a null-terminating terminating temporary buffer of to store the entrylist, plus 1 null terminator.
- we first get the length of the first entry
- in each iteration, we store the length of next entry and replace the length byte of the next entry with NULL, this effectively make the current attribute name a NULL-terminating string, that we can feed to `extattr_{get, set}_fd`. Finally we set the `current` pointer to the next attribute name.

Apart from the difference in how the entry list is iterated, unlike Linux, the extended attributes namespaces are not determined by the prefix the attribute names. Instead, we need to iterator each namespace (`EXTATTR_NAMESPACE_SYSTEM`, `EXTATTR_NAMESPACE_USER`) separately.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -960,6 +960,45 @@ enum _FileOperations {
     private static func _copyDirectoryMetadata(srcFD: CInt, srcPath: @autoclosure () -> String, dstFD: CInt, dstPath: @autoclosure () -> String, delegate: some LinkOrCopyDelegate) throws {
         #if !os(WASI) && !os(Android)
         // Copy extended attributes
+        #if os(FreeBSD)
+        // FreeBSD uses the `extattr_*` calls for setting extended attributes. Unlike like, the namespace for the extattrs are not determined by prefix of the attribute
+        for namespace in [EXTATTR_NAMESPACE_SYSTEM, EXTATTR_NAMESPACE_USER] {
+            // if we don't have permission to list attributes in system namespace, this returns -1 and skips it
+            var size = extattr_list_fd(srcFD, namespace, nil, 0)
+            if size > 0 {
+                try withUnsafeTemporaryAllocation(of: CChar.self, capacity: size + 1) { keyList in
+                    // The list of entry returns by `extattr_list_*` contains the length(1 byte) of the attribute name, follow by the Non-NULL terminated attribute name. (See exattr(2))
+                    size = extattr_list_fd(srcFD, namespace, keyList.baseAddress!, size)
+                    var keyLength = Int(keyList.baseAddress!.pointee)
+                    var current = keyList.baseAddress!.advanced(by: 1)
+                    let end = keyList.baseAddress!.advanced(by: keyList.count)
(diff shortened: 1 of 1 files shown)


Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -960,6 +960,45 @@ enum _FileOperations {
     private static func _copyDirectoryMetadata(srcFD: CInt, srcPath: @autoclosure () -> String, dstFD: CInt, dstPath: @autoclosure () -> String, delegate: some LinkOrCopyDelegate) throws {
         #if !os(WASI) && !os(Android)
         // Copy extended attributes
+        #if os(FreeBSD)
+        // FreeBSD uses the `extattr_*` calls for setting extended attributes. Unlike like, the namespace for the extattrs are not determined by prefix of the attribute
+        for namespace in [EXTATTR_NAMESPACE_SYSTEM, EXTATTR_NAMESPACE_USER] {
+            // if we don't have permission to list attributes in system namespace, this returns -1 and skips it
+            var size = extattr_list_fd(srcFD, namespace, nil, 0)
+            if size > 0 {
+                try withUnsafeTemporaryAllocation(of: CChar.self, capacity: size + 1) { keyList in
```
jmschonfeld, line 969: It took me a moment to discern from reading the code, but it seems like the extra `+ 1` byte is to have room to add a null terminator at the end of the last name. Could you add a comment explaining that here since at first I expected this to just allocate `size` bytes?

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -960,6 +960,45 @@ enum _FileOperations {
...
+                    // The list of entry returns by `extattr_list_*` contains the length(1 byte) of the attribute name, follow by the Non-NULL terminated attribute name. (See exattr(2))
+                    size = extattr_list_fd(srcFD, namespace, keyList.baseAddress!, size)
+                    var keyLength = Int(keyList.baseAddress!.pointee)
+                    var current = keyList.baseAddress!.advanced(by: 1)
+                    let end = keyList.baseAddress!.advanced(by: keyList.count)
```
jmschonfeld, line 974: I think it'd be good to have a `size > 0` check before accessing `keyList.baseAddress` on the lines above (in case this call failed) and perhaps `end` should be offset by `size` instead of `keyList.count` since we don't want to read more bytes than was written (in case the number of attributes changes between when we allocate the buffer and when we populate it)

jmschonfeld (commented):
2 small comments, but overall this looks like a good change to make. Thanks for putting this together!

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #1124, opened 2025-01-16
Prevent potential buffer over-reads of file system representations

jmschonfeld:
The CoreFoundation-level file system representation APIs allow for filling a provided buffer and returning a boolean value indicating success or failure. In some cases, we've seen clients partially filling a buffer, failing, and then reading the buffer despite the `false` return value. We can't avoid writing partial results to the buffer (as some clients depend on the partial results) but we can ensure that the buffer ends with a null terminator even in cases of failure. While the buffer may contain incorrect values, it will at least prevent simple buffer over-reads caused by immediate calls to functions like `strlen` on the return value.


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1128, opened 2025-01-21
Clean up for #1074

itingliu:
- Remove an unused variable.
- Calling `JSONDecoder`'s `func decode<T: Decodable>(_ type: T.Type, from data: Data) throws -> T` with `Decimal.FormatStyle.Currency` as the type isn't correct. But for some reason, the compiler allows it incorrectly. Fix it.
- Clean up whitespaces while we're here.


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1129, opened 2025-01-22
Allow non-numeric port when bridging from CFURL

jrflat:
`CFURL` historically allowed URLs such as `scheme://host:port-string` as long as the port string only contained valid URL characters (characters that can exist in at least some part of a URL string). RFC 3986 only allows digits in the port, so this breaks that use case when a `CFURL` is bridged to `URL`. This PR allows `URL` to have non-numeric ports with valid URL characters, but only in the case of bridging.


jrflat:
@swift-ci please test

parkera (approved):

PR #1134, opened 2025-01-22
Do not cache NSLocale.current if we fail to fetch preferences

itingliu:
Currently, we do not cache the current Swift locale if we cannot fetch proper preferences. This allows us to update the cached value if subsequent access succeeds.

We should be doing that for `currentNSLocale` too, but we are not. Fix the logic so that the behaviors match.

Resolves rdar://142699797


itingliu:
@swift-ci please test

parkera (approved):

PR #1136, opened 2025-01-23
Include String+Bridging.swift in CMakeLists.txt

cthielen:
This PR includes the new `String+Bridging.swift` file in `CMakeLists.txt` to ensure `swift-foundation` builds correctly.


Sources/FoundationEssentials/String/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ target_sources(FoundationEssentials PRIVATE
     String+IO.swift
     String+Internals.swift
     String+Path.swift
+    String+Bridging.swift
```
jmschonfeld, line 26: Not sure if we care too much, but keeping this list in alphabetical order would put `String+Bridging.swift` above `String+Comparison.swift` rather than below `String+Path.swift`

jmschonfeld (approved):

PR #1139, opened 2025-01-24
Don't encode colon if URLComponents path starts with colon

jrflat:
Apps can take a URL string like

```
://host/path
```

which `URL` now allows (after the compatibility change to go back to allowing an empty scheme), and pass it to `URLComponents(string:)`. `URLComponents` follows RFC 3986, which says a scheme cannot be empty, so it interprets the whole string as a path. The app could then use `components.queryItems = ...` to assign query items and call `components.url`, which needs to generate the new `.string`.

During string generation, `URLComponents` percent encodes the colon at the beginning of the path, so the `URL` returned does not have the scheme, host, and path it started out with. We can fix this by not percent-encoding any colons in the first path segment if the colon is the first character in the string.

This still preserves the behaviors of both `URL` and `URLComponents`: `URL` can interpret the colon as an empty scheme like before, and `URLComponents`, knowing that a scheme cannot be empty according to RFC 3986, can still interpret the string as a relative path.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLComponents.swift
+++ b/Sources/FoundationEssentials/URL/URLComponents.swift
@@ -367,14 +367,23 @@ public struct URLComponents: Hashable, Equatable, Sendable {
         }
 
         private var percentEncodedPathNoColon: String {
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

PR #1140, opened 2025-01-27
URL inits should resolve file reference paths

jrflat:
Previously on Darwin, all `URL` initializers created an `NSURL` then ran it through a conversion function to resolve any file reference URLs (with `/.file/id=` paths). This was done to prevent `URL` from ever being a file reference URL, since that changes the nullability of many functions' return values.

With the Swift `URL` implementation, I kept the conversion when bridging `NSURL` to `URL`, but I forgot that resolving the file reference needs to occur in _every_ initializer, e.g. to prevent `URL(string: "file:///.file/id=xxxxxx")` from storing a file reference.

This PR changes all `URL` inits to check if we have a file reference path, and if so, convert it to an actual path.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -634,6 +634,8 @@ public struct URL: Equatable, Sendable, Hashable {
     }()
 #endif
 
+    internal static let fileIDPrefix = Array("/.file/id=".utf8)
+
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

jmschonfeld (approved):

PR #1142, opened 2025-01-27
Increase package deployment target to macOS 15

jmschonfeld:
As we add new APIs and features to Foundation, we plan to adopt new APIs from the Swift standard library that were introduced in the macOS 15-aligned releases. Using these features requires that we increase the minimum deployment target of the SwiftPM build (used for ease of at-desk to macOS 15). This increases the deployment target accordingly so that we can begin to adopt these new APIs without guarding for macOS 15 availability.

In order to bumping the deployment target version, this change also:
- Updates some code in a handful of unit tests to deal with new ambiguity discovered by the compiler (these ambiguity issues stem from new APIs added to Foundation.framework in macOS 15 that were previously disfavored due to lacking matching availability but are now equally favored due to the increased deployment target).
- Resolves new deprecation warnings caused by uses of APIs deprecated between the old deployment target and the new one
- Removes now unnecessarily availability checks/annotations in the unit tests for versions lower than the new deployment target

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,17 +8,15 @@ import CompilerPluginSupport
 
 let availabilityTags: [_Availability] = [
     _Availability("FoundationPreview"), // Default FoundationPreview availability,
-    _Availability("FoundationPredicate", availability: .macOS14_0), // Predicate relies on pack parameter runtime support
-    _Availability("FoundationPredicateRegex", availability: .macOS15_0) // Predicate regexes rely on new stdlib APIs
+    _Availability("FoundationPredicate"), // Predicate relies on pack parameter runtime support
+    _Availability("FoundationPredicateRegex") // Predicate regexes rely on new stdlib APIs
 ]
 let versionNumbers = ["0.1", "0.2", "0.3", "0.4", "6.0.2", "6.1", "6.2"]
 
 // Availability Macro Utilities
 
 enum _OSAvailability: String {
-    case alwaysAvailable = "macOS 13.3, iOS 16.4, tvOS 16.4, watchOS 9.4" // This should match the package's deployment target
-    case macOS14_0 = "macOS 14, iOS 17, tvOS 17, watchOS 10"
-    case macOS15_0 = "macOS 15, iOS 18, tvOS 18, watchOS 11"
+    case alwaysAvailable = "macOS 15, iOS 18, tvOS 18, watchOS 11" // This should match the package's deployment target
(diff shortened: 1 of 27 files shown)


Package.swift
```diff
@@ -8,17 +8,15 @@ import CompilerPluginSupport
 
 let availabilityTags: [_Availability] = [
     _Availability("FoundationPreview"), // Default FoundationPreview availability,
-    _Availability("FoundationPredicate", availability: .macOS14_0), // Predicate relies on pack parameter runtime support
-    _Availability("FoundationPredicateRegex", availability: .macOS15_0) // Predicate regexes rely on new stdlib APIs
+    _Availability("FoundationPredicate"), // Predicate relies on pack parameter runtime support
+    _Availability("FoundationPredicateRegex") // Predicate regexes rely on new stdlib APIs
```
iCharlesHu, line 12: Do we still want keep these since now they just resolve to base availability?

Tests/FoundationEssentialsTests/StringTests.swift
```diff
@@ -1009,7 +1009,7 @@ final class StringTests : XCTestCase {
     func test_dataUsingEncoding_preservingBOM() {
         func roundTrip(_ data: Data) -> Bool {
             let str = String(data: data, encoding: .utf8)!
-            let strAsUTF16BE = str.data(using: .utf16BigEndian)!
+            let strAsUTF16BE = str.data(using: String._Encoding.utf16BigEndian)!
```
iCharlesHu, line 1012: Hmm I wonder why are they suddenly an issue now (as in you have to spill out `String._Encoding`)...

iCharlesHu (approved):

Package.swift
jmschonfeld: I think for now (to avoid a mass refactor) we can keep them but note that we'll come back to this at a later time to remove them and replace them with the "real" availability

Tests/FoundationEssentialsTests/StringTests.swift
jmschonfeld: It's because `String.data(using:)` was added in macOS 15 (only `StringProtocol.data(using:)` existed before that version) so previously the `String.data(using:)` from the package was favored over the `StringProtocol.data(using:)` from the package and from the SDK, but now that `String.data(using:)` is both in the package and in the SDK it becomes ambiguous

PR #1145, opened 2025-01-28
AttributedString APIs for operations over noncontiguous ranges

jmschonfeld:
This PR implements the API changes proposed in [SF-0014](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0014-attributed-string-discontiguous-operations.md) by adding a new `DiscontiguousAttributedSubstring` type (the result of slicing an `AttributedString` with a `RangeSet`) which vends out slices of each underlying view, an upgraded `Runs` view which now handles discontiguous segments, and performs operations over the discontiguous subranges.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/DiscontiguousAttributedSubstring.swift
+++ b/Sources/FoundationEssentials/AttributedString/DiscontiguousAttributedSubstring.swift
@@ -1,0 +1,320 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if FOUNDATION_FRAMEWORK
+@_spi(Unstable) internal import CollectionsInternal
+#elseif canImport(_RopeModule)
+internal import _RopeModule
+#elseif canImport(_FoundationCollections)
+internal import _FoundationCollections
+#endif
+
+@dynamicMemberLookup
+@available(FoundationPreview 6.2, *)
+public struct DiscontiguousAttributedSubstring: Sendable {
+    /// The guts of the base attributed string.
+    internal var _guts: AttributedString.Guts
+    
+    internal var _indices: RangeSet<BigString.Index>
+    
+    internal var _identity: Int = 0
+    
+    internal init(_ guts: AttributedString.Guts, in indices: RangeSet<BigString.Index>) {
+        self._guts = guts
+        // Forcibly resolve bounds and round them down to nearest scalar boundary.
+        var ranges = Array(indices.ranges)
+        for i in ranges.indices {
+            let slice = _guts.string.unicodeScalars[ranges[i]]
+            ranges[i] = Range(uncheckedBounds: (slice.startIndex, slice.endIndex))
+        }
+        self._indices = RangeSet(ranges)
+    }
+}
+
+@available(FoundationPreview 6.2, *)
+extension DiscontiguousAttributedSubstring {
+    public var base: AttributedString {
+        return AttributedString(_guts)
+    }
+}
+
+@available(FoundationPreview 6.2, *)
+extension DiscontiguousAttributedSubstring : CustomStringConvertible {
+    public var description: String {
+        _guts.description(in: _indices)
+    }
+}
+
+@available(FoundationPreview 6.2, *)
+extension DiscontiguousAttributedSubstring : Equatable {
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        AttributedString.Guts._characterwiseIsEqual(lhs.runs, to: rhs.runs)
+    }
+}
+
+@available(FoundationPreview 6.2, *)
+extension DiscontiguousAttributedSubstring : Hashable {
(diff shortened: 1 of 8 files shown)


Sources/FoundationEssentials/AttributedString/DiscontiguousAttributedSubstring.swift
```diff
@@ -0,0 +1,320 @@
...
+extension DiscontiguousAttributedSubstring : CustomStringConvertible {
+    public var description: String {
+        _guts.description(in: _indices)
+    }
+}
+
+@available(FoundationPreview 6.2, *)
+extension DiscontiguousAttributedSubstring : Equatable {
+    public static func == (lhs: Self, rhs: Self) -> Bool {
+        AttributedString.Guts._characterwiseIsEqual(lhs.runs, to: rhs.runs)
```
itingliu, line 60: Is this equivalent to looping over the ranges, and compare their characters one by one, like this?
```
for range in  _indices.ranges {
    AttributedString.Guts.characterwiseIsEqual(lhs.guts, in: range, to: rhs.guts, in: range)
}
```

Asking because in the `hash` implementation below, you're doing it range by range. That looks like a different way to think about this.

Sources/FoundationEssentials/AttributedString/DiscontiguousAttributedSubstring.swift
```diff
@@ -0,0 +1,320 @@
...
+                    }
+                })
+        }
+        for range in fixupRanges {
+            // FIXME: Collect boundary constraints.
+            _guts.enforceAttributeConstraintsAfterMutation(in: range, type: .attributes)
+        }
+    }
+    
+    public subscript(bounds: some RangeExpression<AttributedString.Index>) -> DiscontiguousAttributedSubstring {
```
itingliu, line 127: Is this a new API or a protocol requirement? It doesn't look like a protocol requirement for `AttributedStringAttributeMutation`. Am I missing something?

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/AttributedString/AttributedStringDiscontiguousTests.swift
```diff
@@ -0,0 +1,347 @@
...
+            var copy = str
+            copy[ranges] = str2[ranges2]
+            XCTAssertEqual(String(copy.characters), "ZbYdXfg")
+        }
+    }
+    
+    func testSubCharacterSlicing() {
+        // This test validates that DiscontiguousAttributedSubstring.characters (i.e. DiscontiguousSlice<AttributedString.CharacterView>) behaves the same as DiscontiguousSlice<String> w.r.t. slicing on sub-character boundaries
+        // Note: Country flag emojis are excluded from the text below because AttributedString.CharacterView is not self-slicing and Slice<AttributedString.CharacterView> does not properly handle sub-character slicing within country flag emojis
+        let complexString = "ABCDáëĩoō№Ωאבג🎺🌈WXYZ"
```
lorentey, line 323: This string appears to have no multi-scalar Characters -- neither (1) in its input string, nor (2) in the slices it generates. There are no "sub-character boundaries", so the test is not validating them -- what is the point of this test, then? Ideally we should verify behavior in both of these cases.

Regarding point (1): I expect that the flag emoji problem applies to all multi-scalar characters in the input string: `a\u{301}`, `집`, `🤞🏾`, etc. Given that there will be no easy way to avoid slicing causing crashes, this greatly limits the usefulness of attributed string slices, although hopefully it doesn't entirely eliminate it. (Note that UI-level text editing operations do not use the same grapheme breaking algorithms that Swift implements, so users are sometimes able to position cursors and select text on what Swift considers to be sub-character positions.) Solving this is probably outside the scope of this change; it may indeed be best to treat it as a temporary limitation.

However, point (2) above is independent of this. The issue with `CharacterView` incorrectly using `Slice` for its `SubSequence` does not (immediately) affect slicing that produces multi-scalar characters within _the result_, though -- so this test can carry some weight by exercising that at least. For example, take this input string:

```
let str = "a\n\u{301}"
```

`str` has three single-scalar Characters, `a`, `\n`, and `\u{301}` (the newline does not combine with the combining accent following it). Cutting the `a` and `\u{301}` off into a discontiguous slice produces a result that technically should (according to Unicode) consist of a single two-scalar Character, `á`. We expect that both `String` and `AttributedString` would handle this slicing without traps/crashes, although we expect them to both count the ASCII `a` and the combining accent as distinct `Character`s. Can we use this test to verify that the two types produce consistent behavior in this second case?

jmschonfeld: Apologies, you're right on both counts. Indeed any slicing within a grapheme cluster causes the same crash as I experienced with the country flag emojis. For point (1) I agree that's definitely something we should look into soon as a followup, my hope is that a temporary limitation is ok as this limitation has existed without much uproar for noncontiguous slicing since `AttributedString`'s inception. However we should definitely put a plan in place to resolve it so it doesn't continue to linger.

For point (2) that's a great point, and I've added a unit test in place of this one that asserts the behavior of `String` is the same as `AttributedString.CharacterView` in this regard to make sure that the behavior is the same (and indeed you are correct that both count the ASCII `a` and the combining accent as distinct `Character`s.

jmschonfeld:
@swift-ci please test

lorentey (approved):
👍

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1146, opened 2025-01-29
Prevent crashing when calling quotactl with UIDs greater than Int32.max

jmschonfeld:
When calling `quotactl` on Darwin as part of `FileManager.attributesOfFileSystem(atPath:)`, we take the effective user ID (a `uid_t`) and attempt to pass it to `quotactl` (whose parameter is a C `int` imported as `Int32`). The cast we perform currently causes a crash when the current EUID is greater than `Int32.max`. This is an annotation issue with `quotactl` which should be taking an unsigned integer instead of a signed integer. For now, to prevent the crash and restore the pre-swift-rewrite behavior, we can change this to a bit-pattern cast so that when `quotactl`'s implementation reinterprets the number as an unsigned integer it works.

Unfortunately there's no way to add a unit test for this as it requires the current EUID to be an abnormally large value which is not possible in our test environment since the tests are not guaranteed to be launched as root and only a super user can change the EUID to a value like that.

rdar://143724213


parkera (approved):
Might be worth a comment explaining why we have an unusual casting pattern here.

PR #1149, opened 2025-01-29
Use _CalendarGregorian for ISO8601

parkera:
Use the `_CalendarGregorian` implementation for the ISO8601 calendar, instead of ICU. This lowers it into FoundationEssentials. Previously, the same behavior could be achieved by customizing the init of a Gregorian calendar, but this makes it easier to do the right thing.

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
@@ -3876,5 +3876,25 @@ final class GregorianCalendarTests : XCTestCase {
         test(.minute, expected: -120)
         test(.second, expected: -7200)
     }
+    
+    // MARK: ISO8601
+    
+    func test_iso8601Gregorian() {
+        var calendar1 = Calendar(identifier: .iso8601)
+        let calendar2 = Calendar(identifier: .iso8601)
+        XCTAssertEqual(calendar1, calendar2)
+        
+        XCTAssertEqual(calendar1.firstWeekday, 2)
+        XCTAssertEqual(calendar1.minimumDaysInFirstWeek, 4)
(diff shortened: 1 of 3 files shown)


parkera:
@swift-ci test

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -3876,5 +3876,25 @@ final class GregorianCalendarTests : XCTestCase {
...
+        let calendar2 = Calendar(identifier: .iso8601)
+        XCTAssertEqual(calendar1, calendar2)
+        
+        XCTAssertEqual(calendar1.firstWeekday, 2)
+        XCTAssertEqual(calendar1.minimumDaysInFirstWeek, 4)
+        XCTAssertEqual(calendar1.timeZone, .gmt)
+        XCTAssertEqual(calendar1.locale, .unlocalized)
+        
+        // Verify that the properties are still mutable
+        let tz = TimeZone(secondsFromGMT: -3600)!
```
jmschonfeld, line 3893: Minor suggestion: it can be helpful to use `try XCTUnwrap` rather than force-unwrapping these values since test failures are a lot easier to debug in CI than crashes (in the off chance we break this initializer and it returns `nil`)

parkera: Fair enough. We use this same pattern a lot in this file and other tests.

jmschonfeld: Ah ok if it's just using the same pattern then maybe it makes sense to keep it as is. Not strongly opinionated either way since we don't really expect this to fail. Will let you decide what you think is best 🙂

jmschonfeld (approved):

PR #1151, opened 2025-01-30
Disable a test that fails with ICU-76

itingliu:
The behavior regarding parsing a POSIX string will change in ICU-76. While we haven't updated ICU dependency to ICU-76 yet, test will fail when we do. Disable the test for now.

I'll follow up with ICU owners to figure out the expected behavior and update the test or implementation if needed.

resolves 143438257


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1153, opened 2025-01-30
AttributesString.Runs.subscript(nsAttributedStringKeys:) should accept array of keys rather than variadic

jmschonfeld:
Minor update to this SPI to make it more usable for its callers


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1162, opened 2025-02-05
Re-add existing SPI for enumerating NSAttributedString.Key Runs

jmschonfeld:
With further testing, it's not safe to remove the existing SPI here in favor of the new one so this change adds the existing SPI back alongside the new SPI


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1163, opened 2025-02-06
Fix appending empty query items changing the URL

z2oh:
We experienced some regressions in our URL handling code after upgrading to the latest Foundation. Some of them I think are real bug-fixes (for instance: appending a single component `path/subpath` now appends `path%2Fsubpath` which is what it always should have done), but this one I suspect is a new bug.

When appending an empty list of `URLQueryItem`, the current implementation will still append the `?` to the end of the URL. I think this is because the underlying `_query` is unintentionally set to `""` instead of `nil`, which causes [this check](https://github.com/swiftlang/swift-foundation/blob/74275a96e87170eab064bf720367ca8e63150cbb/Sources/FoundationEssentials/URL/URLComponents.swift#L483) on the Optional to pass, appending the string `"?\(percentEncodedQuery)"` which just expands to `"?"` (as `percentEncodedQuery` is the empty string).

### DIFF
--- a/Sources/FoundationEssentials/URL/URLComponents.swift
+++ b/Sources/FoundationEssentials/URL/URLComponents.swift
@@ -592,7 +592,7 @@ public struct URLComponents: Hashable, Equatable, Sendable {
                 return
             }
             guard !newValue.isEmpty else {
-                _query = ""
+                _query = nil
                 return
             }
 
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -450,6 +450,11 @@ final class URLTests : XCTestCase {
             testAppendQueryItems.absoluteString,
             "https://www.example.com?id=42&color=blue"
         )
+        let emptyQueryItems: [URLQueryItem] = []
+        XCTAssertEqual(
+            base.appending(queryItems: emptyQueryItems).absoluteString,
+            "https://www.example.com"
+        )
 
         // Appending components
         XCTAssertEqual(


jrflat:
@z2oh Thanks for digging into this! Would you mind expanding on the regression you're seeing?

I believe this has been the behavior since `NSURLComponents` was originally introduced ~10 years ago. I think the reasoning is to differentiate between setting the query items to `nil`, meaning remove the query entirely, and setting the query items to an empty array, meaning we have a query, but have no query items (empty string). `URL.appending(queryItems:)`, which uses `URLComponents`, also hasn't changed from when it was introduced as far as I'm aware.

Unfortunately, changing the behavior of `URL` or `URLComponents` very often leads to breakage of apps that were relying on the old behavior, so I don't think we'll be able to make this particular change unless we confirm it's a regression. (For this same reason, we unfortunately had to revert/provide compatibility paths for some of those real bug-fixes you mentioned, too.)

z2oh:
Ah you are right! We have this test in our project:
```
let emptyQueryItems: [URLQueryItem] = []
XCTAssertEqual(
    base.appending(queryItems: emptyQueryItems).absoluteString,
    "https://www.example.com"
)
```
which started failing after upgrading the Windows toolchain to one that included the new Foundation. I assumed this was a regression in Foundation and traced it to the line changed in this PR, but now after taking a second look I realize we have a custom URL+Append extension shim in our project which is only defined when the old Foundation is present (via `#if !canImport(FoundationEssentials)`). It is in our shim where the behavior differs, not in Foundation, and I will update this on our end to match Foundation semantics.

Thank you for the thoughtful review!

PR #1175, opened 2025-02-15
always @preconcurrency import Glibc/Musl/Android/Bionic/WASILibc

weissi:
Related issues:
- https://github.com/swiftlang/swift-format/issues/933
- https://github.com/swiftlang/swift/issues/79414
- https://github.com/swiftlang/swift-corelibs-foundation/pull/5170

I've got a working program

```
@preconcurrency import Glibc
import Foundation

await Task {
  _ = fputs("hello\n", stderr)
}.value
```

but alas, if `swift-format` reformats it, it makes it (note that the two imports are reordered)

```
import Foundation
@preconcurrency import Glibc

await Task {
  _ = fputs("hello\n", stderr)
}.value
```

which unfortunately doesn't compile

```
$ swift format --in-place test.swift && swiftc -swift-version 6 test.swift && echo OK
test.swift:5:24: error: reference to var 'stderr' is not concurrency-safe because it involves shared mutable state
3 | 
4 | await Task {
5 |   _ = fputs("hello\n", stderr)
  |                        `- error: reference to var 'stderr' is not concurrency-safe because it involves shared mutable state
6 | }.value
7 | 

SwiftGlibc.stderr:1:12: note: var declared here
1 | public var stderr: UnsafeMutablePointer<FILE>!
  |            `- note: var declared here
```

the problem is that `Foundation` imports `Glibc`/`Musl` but doesn't mark them `@preconcurrency`.

---

This has been used to rewrite the sources:

```bash
find Sources/ -name '*.swift' | while read -r line; do sed -r -i '' 's/^( *)(@_exported )?import (Musl|Android|Bionic|Glibc|WASILibc)/\1\2@preconcurrency import \3/g' "$line"; done
```

### DIFF
--- a/Benchmarks/Benchmarks/Base64/Base64.swift
+++ b/Benchmarks/Benchmarks/Base64/Base64.swift
@@ -24,7 +24,7 @@ private func autoreleasepool<T>(_ block: () -> T) -> T { block() }
 #endif
 
 #if canImport(Glibc)
-import Glibc
+@preconcurrency import Glibc
 #endif
 #if canImport(Darwin)
 import Darwin
--- a/Benchmarks/Benchmarks/DataIO/BenchmarkDataIO.swift
+++ b/Benchmarks/Benchmarks/DataIO/BenchmarkDataIO.swift
@@ -24,7 +24,7 @@ private func autoreleasepool<T>(_ block: () -> T) -> T { block() }
 #endif
 
 #if canImport(Glibc)
-import Glibc
+@preconcurrency import Glibc
 #endif
 #if canImport(Darwin)
 import Darwin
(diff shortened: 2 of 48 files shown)


parkera:
@swift-ci test

weissi:
> Interesting. I wonder whether this is really the right fix, or whether we should be using API notes to annotate `stderr` et al properly.

That seems like a pretty big project and will always miss something, no? But of course, I'd be fine with that too.

But crucially until that's done, I think this is what we should do. We can remove the `@preconcurrency import` once that has been completed.

weissi:
@swift-ci test macos

parkera (approved):

parkera:
@weissi if this one is also for 6.1, please cherry pick

weissi:
> @weissi if this one is also for 6.1, please cherry pick

That one's already done and merged: https://github.com/swiftlang/swift-foundation/pull/1194

PR #1177, opened 2025-02-17
AttributedString Index Validity APIs

jmschonfeld:
This PR implements the index validity checking APIs as proposed in [SF-0015](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0015-attributedstring-tracking-indices.md) by adding a new version number tracked within each `AttributedString` and `AttributedString.Index`. As neither type is frozen, we have the flexibility in the future to optimize this (for example, by taking advantage of the version number stored within the already-nested `BigString.Index` if the swift-collections package provided access to it) if we deem necessary.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
@@ -60,11 +60,10 @@ extension AttributedString.Index {
 @available(FoundationPreview 6.2, *)
 extension Range<AttributedString.Index> {
     public func isValid(within text: some AttributedStringProtocol) -> Bool {
+        // Note: By nature of Range's lowerBound <= upperBound requirement, this is also sufficient to determine that lowerBound <= endIndex && upperBound >= startIndex
         self.lowerBound._version == text.__guts.version &&
         self.lowerBound >= text.startIndex &&
-        self.lowerBound <= text.endIndex &&
         self.upperBound._version == text.__guts.version &&
-        self.upperBound >= text.startIndex &&
         self.upperBound <= text.endIndex
     }
     
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
@@ -1,0 +1,93 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Synchronization)
+internal import Synchronization
+#endif
+
+extension AttributedString.Guts {
+    typealias Version = UInt
+    
+    #if canImport(Synchronization)
+    private static let _nextVersion = Atomic<Version>(0)
+    
+    static func createNewVersion() -> Version {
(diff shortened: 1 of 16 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:

Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
```diff
@@ -0,0 +1,93 @@
...
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(Synchronization)
```
itingliu, line 13: In other places we have `#if canImport(Synchronization) && FOUNDATION_FRAMEWORK`. Is that wrong?

Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
```diff
@@ -0,0 +1,93 @@
...
+#endif
+
+extension AttributedString.Guts {
+    typealias Version = UInt
+    
+    #if canImport(Synchronization)
+    private static let _nextVersion = Atomic<Version>(0)
+    
+    static func createNewVersion() -> Version {
+        _nextVersion.wrappingAdd(1, ordering: .relaxed).oldValue
```
itingliu, line 24: Nit: I'd probably put the `#if canImport(Synchronization)` inside this function since that's implementation detail

itingliu (approved):

Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
```diff
@@ -0,0 +1,93 @@
...
+}
+
+@available(FoundationPreview 6.2, *)
+extension Range<AttributedString.Index> {
+    public func isValid(within text: some AttributedStringProtocol) -> Bool {
+        self.lowerBound._version == text.__guts.version &&
+        self.lowerBound >= text.startIndex &&
+        self.lowerBound <= text.endIndex &&
+        self.upperBound._version == text.__guts.version &&
+        self.upperBound >= text.startIndex &&
```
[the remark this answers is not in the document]
jmschonfeld: You're right `Range` has that precondition so both checks aren't required. I can update this to simplify the expressions down a bit

Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
jmschonfeld: Previously we needed the `&& FOUNDATION_FRAMEWORK` because non-`FOUNDATION_FRAMEWORK` builds ran on a very old version of macOS in CI where we back deployed back to a version that didn't have `Synchronization` (even though the `canImport(Synchronization)` check passes because it was in the SDK). Now that we've bumped the deployment target of the package, we can safely use the `Synchronization` module on the macOS version CI is back deployed to, so no need to guard with `FOUNDATION_FRAMEWORK` (but there are still some edge case situations where the module doesn't exist at all that we need to guard for with the `canImport` check)

Sources/FoundationEssentials/AttributedString/AttributedString+IndexValidity.swift
jmschonfeld: Sure I can do that if you prefer - I still would need a separate `canImport` check around the `private static let` variables since their types are different, but if you want I can split it into two `canImport` checks, one for around the variable and one within the body of the function

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1180, opened 2025-02-18
New XML PropertyListDecoder parser accepts malformed tags

kperryua:
A mistake was made during translation of PropertyListSerialization XML parsing to native PropertyListDecoder with regard to open tags.

`PropertyListSerialization` accepts open tags that look like any of the following:

```
`<` <VALIDTAG> `>`
`<` <VALIDTAG> `/` `>`
`<` <VALIDTAG> <WHITESPACE> <ANYTHING> `>`
```

The latter case is to handle tags with attributes, which are typically only used on the `plist` tag, e.g. `<plist version="1.0">`.

The adapted code was too permissive and accepted:

```
`<` <VALIDTAG> <ANYTHING> `>`
`<` <VALIDTAG> <ANYTHING> `/` `>`
```

Which means we accepted input like:

```
<stringGARBAGE>foo</string>
```

This PR fixes that by checking the next character after a valid tag name before accepting it.

### DIFF
--- a/Sources/FoundationEssentials/PropertyList/XMLPlistScanner.swift
+++ b/Sources/FoundationEssentials/PropertyList/XMLPlistScanner.swift
@@ -908,68 +908,80 @@ internal struct XMLPlistScanner {
         }
     }
     
-    func determineTag() throws -> XMLPlistTag? {
+    mutating func readTag() throws -> XMLPlistTag? {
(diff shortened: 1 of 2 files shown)


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #1181, opened 2025-02-20
[6.1] Update SwiftPM/CMake dependencies to match toolchain build

jmschonfeld:
- **Explanation**: Updates the dependencies declared in the SwiftPM package manifest and `CMakeLists.txt` files to match the versions of the dependencies actually used when building the toolchain. Having these be correct will be important when making lockstep changes (for example, between swift-foundation and swift-syntax) so that if a dependency like swift-syntax makes a breaking change on `main` it will not break the build of our package's `release/6.1` branch due to incorrectly depending on the wrong branch.
  - **Scope**: Only affects the version of used dependencies and does not change the project's code itself. Also only impacts "local" builds and not any toolchain builds
  - **Issues**: N/A
  - **Original PRs**: N/A
  - **Risk**: Low - does not impact toolchain build and change is intended to have no effect currently
  - **Testing**: SwiftPM build is tested in CI, and I've tested both the SwiftPM and standalone CMake build locally
  - **Reviewers**: N/A

### DIFF
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -71,11 +71,11 @@ switch usePackage {
 
     case .useGitHubPackage:
         #if os(macOS)
-            packageDependency.append(.package(url: "https://github.com/apple/swift-foundation", branch: "main"))
+            packageDependency.append(.package(url: "https://github.com/apple/swift-foundation", branch: "release/6.1"))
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

PR #1183, opened 2025-02-20
Fix ambiguity on macOS in PropertyListEncoderTests

jmschonfeld:
https://github.com/swiftlang/swift-foundation/pull/1180 accidentally introduces a build issue on macOS due to new ambiguity between `Foundation.String.Encoding` and `FoundationEssentials.String.Encoding` but we weren't able to detect that since macOS CI is currently nonfunctional. This fixes the ambiguity using the same trick we use elsewhere in our unit tests. I've validated locally that this fixes the build issues on macOS.


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1191, opened 2025-02-25
Set CMP0157 to OLD for swift-collections

Steelskin:
When building the toolchain, we may not have an early swift-driver build. This causes issues when building swift-collections with CMP0157 set to NEW. As a workaround, this sets CMP0157 to OLD while including the swift-collections project.

The minimum CMake version required is increased to 3.25 to use the `block()` command.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 ##
 ##===----------------------------------------------------------------------===##
 
-cmake_minimum_required(VERSION 3.24)
+cmake_minimum_required(VERSION 3.25)
 
 if(POLICY CMP0156)
     # Deduplicate linked libraries where appropriate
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -70,18 +70,27 @@ else()
         GIT_REPOSITORY https://github.com/apple/swift-foundation-icu.git
         GIT_TAG 0.0.9)
 endif()
+FetchContent_MakeAvailable(SwiftFoundationICU)
```
compnerd, line 73: `FetchContent_MakeAvailable` should really be invoked once at the end as per the documentation.

CMakeLists.txt
```diff
@@ -70,18 +70,27 @@ else()
...
-        GIT_REPOSITORY https://github.com/apple/swift-collections.git
-        GIT_TAG 1.1.2)
-endif()
-FetchContent_MakeAvailable(SwiftFoundationICU SwiftCollections)
+block(SCOPE_FOR POLICIES)
+    # swift-collections does not build properly without swift-driver, which
+    # may not be available during the toolchain build. Set CMP0157 to OLD to
+    # work around this issue.
+    if(POLICY CMP0157)
+        cmake_policy(SET CMP0157 OLD)
```
jmschonfeld, line 80: If this only applies to swift-collections, can we make this change to swift-collections instead of to swift-foundation? There are other things in the toolchain that use swift-collections and I think it'd be best to fix it at the source rather than fixing it in each project that uses swift-collections in the toolchain (even if swift-foundation is the only project this affects, it'd be best to fix this in swift-collections since that's where this policy is actually set). cc @lorentey

Steelskin:
My apologies, this does not work because the whole project is generated as a single ninja project at the end, which is when the policy takes effect. The only possible workaround is to set CMP0157 to OLD in the top-level project (in this case swift-corelibs-foundation). Closing.

PR #1196, opened 2025-03-03
[Base64] Add new decoding options

fabianfett:


Tests/FoundationEssentialsTests/DataTests.swift
```diff
@@ -2097,6 +2096,27 @@ extension DataTests {
         XCTAssertEqual(Data([1, 2, 3, 4, 5, 6]), Data(base64Encoded: "AQIDBAUG    ", options: .ignoreUnknownCharacters))
     }
 
+    func test_base64Decode_testOmittingPaddingCharacters() {
```
itingliu: Like I mentioned in #1195, can you rebase and update to swift-testing for these tests?

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -957,11 +985,26 @@ extension Base64 {
...
     }
 
-    static func withUnsafeDecodingTablesAsBufferPointers<R, E: Swift.Error>(options: Data.Base64DecodingOptions, _ body: (UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>) throws(E) -> R) throws(E) -> R {
-        let decoding0 = Self.decoding0
-        let decoding1 = Self.decoding1
-        let decoding2 = Self.decoding2
-        let decoding3 = Self.decoding3
+    static func withUnsafeDecodingTablesAsBufferPointers<R, E: Swift.Error>(
+        options: Data.Base64DecodingOptions,
+        _ body: (UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>, UnsafeBufferPointer<UInt32>) throws(E) -> R
```
itingliu, line 990: As a separate follow up, I think it'd be very nice if we can use `Span` instead of `UnsafeBufferPointer` here

itingliu (approved):

Sources/FoundationEssentials/Data/Data+Base64.swift
fabianfett: Once this has landed, we can rebase:
https://github.com/swiftlang/swift-foundation/pull/2171

PR #1171, opened 2025-02-14
New Calendar Identifiers

karan-misra:
Exposed identifiers for 12 calendars. API Proposal: [SF-017](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0017-expanded-calendar-support.md)

This is take two, as we needed to revert the [first attempt](https://github.com/swiftlang/swift-foundation/pull/1168) due to CI failures.

141795986

PR #1197, opened 2025-03-04
Remove Thai Calendar

karan-misra:
Due to lack of an authoritative source of calendrical calculations, we’re not able to add support for it in ICU and correspondingly in Swift Foundation. Removing the identifier that was previously exposed with #1171.

145890794


itingliu:
@swift-ci please test

itingliu:
This is API change. Please also update the proposal

karan-misra:
Moved to #1203

PR #1186, opened 2025-02-22
[Proposal] URI Templating

danieleggert:
This PR introduces a proposal for adding [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570) _URI templates_ to the `URL` type.

PR #1198, opened 2025-03-04
URI Templating

danieleggert:
This adds [RFC 6570](https://datatracker.ietf.org/doc/html/rfc6570) _URI templates_ to the Swift URL type.

Evolution proposal: #1186


jrflat:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationEssentials/URL/URITemplate.swift
```diff
@@ -0,0 +1,98 @@
+#if canImport(CollectionsInternal)
```
parkera, line 1: We'll need the standard copyright header on this and all other files, please.

Sources/FoundationEssentials/URL/URITemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+
+// MARK: - Parse
+
+extension URL.Template {
+    /// Creates a new template from its text form.
+    ///
+    /// The template string needs to be a valid RFC 6570 template.
+    ///
+    /// This will parse the template and throw an error if the template is invalid.
+    public init(_ template: String) throws {
```
parkera, line 53: Do we need to add availability annotations to this, too, even though the type it is an `extension` of is new?

Sources/FoundationEssentials/URL/URITemplate_Expression.swift
```diff
@@ -0,0 +1,272 @@
...
+        }
+    }
+}
+
+private struct InvalidTemplateExpression: Swift.Error {
+    var text: String
+}
+
+extension Substring {
+    mutating func popPrefixMatch<Output>(_ regex: Regex<Output>) throws -> Regex<Output>.Match? {
```
parkera, line 47: If possible I'd like to mark these kinds of helpers as `fileprivate`, to make it clear where they are used. This makes it far easier to refactor in the future.

Sources/FoundationEssentials/URL/URITemplate_Expression.swift
```diff
@@ -0,0 +1,272 @@
...
+            /// name=value pairs separated by "&"; and,
+            case queryComponent = "?"
+            /// `&`   Continuation of query-style &name=value pairs within
+            /// a literal query component.
+            case continuation = "&"
+        }
+    }
+}
+
+private struct InvalidTemplateExpression: Swift.Error {
```
parkera, line 42: Hm, I wonder how we can make it easier to ensure this error isn't leaked into any public functions.

Sources/FoundationEssentials/URL/URITemplate_Substitution.swift
```diff
@@ -0,0 +1,143 @@
...
+    /// This will fail if variable expansion does not produce a valid,
+    /// well-formed URL.
+    ///
+    /// All text will be converted to NFC (Unicode Normalization Form C) and UTF-8
+    /// before being percent-encoded if needed.
+    ///
+    /// - Parameters:
+    ///   - template: The RFC 6570 template to be expanded.
+    ///   - variables: Variables to expand in the template.
+    public init?(
```
parkera, line 21: Availability is missing here.

Sources/FoundationEssentials/URL/URITemplate_VariableName.swift
```diff
@@ -0,0 +1,36 @@
...
+        init(_ key: Substring) {
+            self.key = String(key)
+        }
+    }
+}
+
+// MARK: -
+
+extension String {
+    public init(_ key: URL.Template.VariableName) {
```
parkera, line 19: Was this part of the proposal? Adding a public, no-arg init to `String` seems very broad.

Tests/FoundationEssentialsTests/URITemplatingTests/URITemplate_ExpressionTests.swift
```diff
@@ -0,0 +1,265 @@
...
+
+#if canImport(FoundationEssentials)
+@testable import FoundationEssentials
+import struct FoundationEssentials.URL
+#endif
+#if FOUNDATION_FRAMEWORK
+@testable import Foundation
+import struct Foundation.URL
+#endif
+import XCTest
```
parkera, line 21: Why use XCTest for this, but Swift Testing for the others?

Tests/FoundationEssentialsTests/URITemplatingTests/URITemplate_ExpressionTests.swift
```diff
@@ -0,0 +1,265 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
```
parkera, line 5: Can you update the copyright years on this header, and others?

parkera (commented):
Could you also add a benchmark suite for this functionality? You can pretty easily get a start by copying one of the existing formatting ones.

Sources/FoundationEssentials/URL/URITemplate_VariableName.swift
parkera: Oh, I see it now. I'm not sure about it now that I see it in code. I know it is strongly typed, but it can cause confusion in documentation or at call sites. How critical is it to do this conversion in this way at a call site?

Sources/FoundationEssentials/URL/URITemplate_PercentEncoding.swift
```diff
@@ -0,0 +1,244 @@
...
+        if allowed.isAllowedCodeUnit(next) {
+            write(next)
+        } else {
+            write(UInt8(ascii: "%"))
+            write(hexToAscii(next >> 4))
+            write(hexToAscii(next & 0xf))
+        }
+    }
+    write(0)
+    return String(cString: outputBuffer.baseAddress!)
```
jrflat, line 75: I think it might be preferable to do `String(decoding: outputBuffer[..<outputIndex], as: UTF8.self)` to avoid the null-terminator and cString initializer.

Sources/FoundationEssentials/URL/URITemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+        self.init()
+
+        var remainder = template[...]
+
+        func copyLiteral(upTo end: String.Index) {
+            guard
+                remainder.startIndex < end
```
jrflat, line 60: Nit: The rest of the code base uses
```
guard condition else { return }
```
or
```
guard condition else {
  return
}
```
so I think we should probably match that style.

Sources/FoundationEssentials/URL/URITemplate.swift
```diff
@@ -0,0 +1,98 @@
...
+            let literal = remainder[remainder.startIndex..<end]
+            let escaped = String(literal).normalizedAddingPercentEncoding(
+                withAllowedCharacters: .unreservedReserved
+            )
+            elements.append(.literal(escaped))
+        }
+
+        while let match = remainder.firstMatch(of: URL.Template.Global.shared.uriTemplateRegex) {
+            defer {
+                remainder = remainder[match.range.upperBound..<remainder.endIndex]
```
jrflat, line 71: Is deferring here different than just doing `remainder = ...` in the last line of the while loop?

jrflat (commented):
Thanks for pitching and implementing this! I left some comments, some nits and some questions on whether we can merge percent-encoding functionality with implementations in `URLParser.swift`.

Another minor nit: would we want to rename the files to `URLTemplate...` given that the new type is based on `URL`? (I realize the RFC does name it "URI".)

I think we will also need to add the sources to `CMakeLists` in the URL folder.

Sources/FoundationEssentials/URL/URITemplate.swift
danieleggert: IIRC, having `@available` on the type is sufficient.

Sources/FoundationEssentials/URL/URITemplate_Expression.swift
danieleggert: None of the public functions are throwing. That should be enough, no?

Tests/FoundationEssentialsTests/URITemplatingTests/URITemplate_ExpressionTests.swift
danieleggert: Let me convert the others. I had originally written a lot of this a long time ago, before Testing was available.

Sources/FoundationEssentials/URL/URITemplate_PercentEncoding.swift
```diff
@@ -0,0 +1,244 @@
+internal import RegexBuilder
+
+extension String {
+    /// Convert to NFC and percent-escape.
+    func normalizedAddingPercentEncoding(
+        withAllowedCharacters allowed: URL.Template.Expression.Operator.AllowedCharacters
+    ) -> String {
+        switch allowed {
+        case .unreserved:
+            return addingPercentEncodingToNFC(
```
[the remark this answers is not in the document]
danieleggert: I tried to clean it up a bit. Let me know what you think.

danieleggert:
> Could you also add a benchmark suite for this functionality? You can pretty easily get a start by copying one of the existing formatting ones.

The info about building and running *Benchmarks* is broken / missing / misleading.
```
SWIFTCI_USE_LOCAL_DEPS=.. swift package benchmark
```
doesn’t work, it needs to be `../..` — and then (because the top-level `Package.swift` also uses `SWIFTCI_USE_LOCAL_DEPS`, the top-level project also wants to use local dependencies.

I worked around it with local modificatoins to the `Package.swift`, but this could use some TLC.

jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URLTemplate_VariableName.swift
```diff
@@ -0,0 +1,48 @@
...
+        init(_ key: Substring) {
+            self.key = String(key)
+        }
+    }
+}
+
+// MARK: -
+
+extension String {
+    public init(_ key: URL.Template.VariableName) {
```
itingliu, line 31: This is also missing

```suggestion
   @available(FoundationPreview 6.2, *)
    public init(_ key: URL.Template.VariableName) {
```

Sources/FoundationEssentials/URL/URLParser.swift
```diff
@@ -1225,4 +1225,11 @@ fileprivate extension UTF8.CodeUnit {
             return (allowed & (UInt64(1) << (self - 64))) != 0
         }
     }
+
+    /// Is the character in `unreserved + reserved` from RFC 3986.
+    internal var isUnreservedURLCharacter: Bool {
```
jrflat, line 1230: The comment should just be `unreserved`, right?

Sources/FoundationEssentials/URL/URITemplate_PercentEncoding.swift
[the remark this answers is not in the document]
jrflat: Looks good to me, thanks!

jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URITemplate_VariableName.swift
jrflat: @parkera would removing this from the implementation require another round of review/block this PR? Or could we maybe mark it internal for now?

danieleggert: I think the right way forward is to remove `ExpressibleByStringLiteral`. This is only truly useful in small demo code. In “real” code, you’d usually have a fixed set of variable names, and would probably do something like:
```swift
extension URL.Template.VariableName {
    static var query: URL.Template.VariableName { URL.Template.VariableName("query") }
    static var number: URL.Template.VariableName { URL.Template.VariableName("number") }
}
```
and then use that when constructing the dictionary.

parkera: Agreed, or even

```
static var query: URL.Template.VariableName { .init("query") }
```

parkera (approved):
Looks good to me - @itingliu can you also review?

itingliu:
@swift-ci please test

itingliu (approved):
LGTM in general. I left some comments, but feel free to address them in a follow up or dismiss them if not valid.

itingliu:
@swift-ci please test

Sources/FoundationEssentials/URL/URLTemplate_Expression.swift
```diff
@@ -0,0 +1,286 @@
...
+}
+
+extension URL.Template.Expression.Element: CustomStringConvertible {
+    var description: String {
+        "\(name)\(maximumLength.map { ":\($0)" } ?? "")\(explode ? "*" : "")"
+    }
+}
+
+extension URL.Template.Expression {
+    init(_ input: String) throws {
```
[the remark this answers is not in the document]
danieleggert: This may also throw errors from RegEx parsing. I took a stab at using typed throws, but it ended up making the code more complex and more difficult to read.

Sources/FoundationEssentials/URL/URLTemplate_PercentEncoding.swift
```diff
@@ -0,0 +1,224 @@
...
+        outputBuffer[outputIndex] = a
+        outputIndex += 1
+    }
+
+    while let next = remainingInput.popFirst() {
+        // Any (valid) existing escape sequences need to be copied to the output verbatim.
+        // But any `%` that are not part of a valid escape sequence, need to be encoded.
+        guard next != UInt8(ascii: "%") || remainingInput.count < 2 else {
+            // Is this a valid escape sequence?
+            if remainingInput[remainingInput.startIndex].isValidHexDigit && remainingInput[remainingInput.startIndex + 1].isValidHexDigit {
```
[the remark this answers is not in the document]
danieleggert: The `remainingInput.count < 2` in the `guard` protects against that. The `// Is this a valid escape sequence?` code will only run if `next == UInt8(ascii: "%") && remainingInput.count >= 2`.

danieleggert:
@itingliu Thanks for the comments / feedback. I’ve pushed 2 additional commits.

Sources/FoundationEssentials/URL/URLTemplate_PercentEncoding.swift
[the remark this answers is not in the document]
itingliu: My bad. I kept reading it as `remainingInput.count > 2` 🤦‍♀️

itingliu:
@swift-ci please test

vanvoorden:
I'm trying to run benchmarks locally from `main` and seeing a lot of errors like:

```text
only available in macOS 26.0 or newer
```

AFAIK the benchmarks themselves do not look gated around that requirement? Was this going to be updated soon?

PR #1199, opened 2025-03-04
FSR writes bytes to end of buffer even on success cases

jmschonfeld:
When clients ask Foundation to fill a provided buffer with a string's file system representation, with some incorrect uses of the API the buffer length they provide is longer than the buffer really is. Previously, these clients likely got away with this if filling the buffer succeeded with a small portion of it used, but recently we started unconditionally writing a null byte to the end of the buffer, even on success cases, which now corrupts memory in the caller frames.

Instead to mitigate this issue for existing clients while still preserving the security mitigation of not allowing non-null-terminated buffers to be returned - we can ensure that we only write one null byte even in success cases. In the success case, that null byte immediately follows the written contents leaving the remainder of the buffer untouched/uninitialized. The same behavior is used for failure cases that aren't due to buffer size (for example, malformed UTF-8/UTF-16 contents). For failure cases where the buffer has been completely filled, we still maintain the behavior of replacing the final byte with a null byte to ensure null termination.

There are no unit tests here because this only affects the internal `NSString` API that fills a provided buffer whereas the Swift entry points in swift-foundation ensure that the buffer is never returned at all in failure cases

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -233,7 +233,16 @@ extension UnsafeBufferPointer {
             if nullTerminated {
                 // Ensure buffer is always null-terminated even on failure to prevent buffer over-reads
                 // At this point, the buffer is known to be non-empty, so it must have space for at least a null terminating byte (even if it overwrites the final output byte in the buffer)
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1138, opened 2025-01-24
Expanded calendar support

DraganBesevic:
This proposal offers support for additional calendars.

PR #1202, opened 2025-03-06
[SF-0017] Expanded Calendar Support (Update)

karan-misra:
Due to lack of an authoritative source of calendrical calculations, we’re not able to add support for the Thai lunisolar calendar in ICU and correspondingly in Swift Foundation. This is to update the proposal to remove the identifier that was previously exposed with #1138.

API Update in #1203.


itingliu (approved):
@swift-ci please test

PR #1204, opened 2025-03-06
Use lstat() semantics for URL directory detection

jrflat:
Historically, `NS/URL` checks if a path is a directory using `getattrlist` with `FSOPT_NOFOLLOW`. This is equivalent to `lstat()` in the sense that it won't follow a symlink at the end of the path, and it will report information about the symlink itself.

In the Swift `URL` implementation, I called into `FileManager` to check if a path is a directory, but `FileManager` uses `stat()` for this purpose (and historically always has).

Since `URL` historically doesn't follow a symlink at the end of the path, this PR switches to using `lstat()` on UNIX-like platforms instead of calling into `FileManager`.

On Windows, we can create the file handle with `FILE_FLAG_OPEN_REPARSE_POINT` to achieve the same semantics.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -13,8 +13,18 @@
 public struct URLResourceKey {}
 #endif
 
-#if os(Windows)
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Android)
+@preconcurrency import Android
+#elseif canImport(Glibc)
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif os(Windows)
 import WinSDK
+#elseif os(WASI)
+@preconcurrency import WASILibc
 #endif
 
 #if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -2200,16 +2210,29 @@ extension URL {
 #if !NO_FILESYSTEM
     private static func isDirectory(_ path: String) -> Bool {
         #if os(Windows)
+        guard !path.isEmpty else { return false }
```
compnerd, line 2213: Why not hoist this to be universal?

Sources/FoundationEssentials/URL/URL.swift
```diff
@@ -2200,16 +2210,29 @@ extension URL {
...
+        return (try? path.withNTPathRepresentation { pwszPath in
+            // If path points to a symlink (reparse point), get a handle to
+            // the symlink itself using FILE_FLAG_OPEN_REPARSE_POINT.
+            let handle = CreateFileW(pwszPath, GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE | FILE_SHARE_DELETE, nil, OPEN_EXISTING, FILE_FLAG_BACKUP_SEMANTICS | FILE_FLAG_OPEN_REPARSE_POINT, nil)
+            guard handle != INVALID_HANDLE_VALUE else { return false }
+            defer { CloseHandle(handle) }
+            var info: BY_HANDLE_FILE_INFORMATION = BY_HANDLE_FILE_INFORMATION()
+            guard GetFileInformationByHandle(handle, &info) else { return false }
+            if (info.dwFileAttributes & FILE_ATTRIBUTE_REPARSE_POINT) == FILE_ATTRIBUTE_REPARSE_POINT { return false }
+            return (info.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY) == FILE_ATTRIBUTE_DIRECTORY
```
jrflat, line 2224: Hi @compnerd, is creating the file with `FILE_FLAG_OPEN_REPARSE_POINT` and then checking the attributes for `FILE_ATTRIBUTE_REPARSE_POINT` the correct way to achieve these semantics on Windows: "return false if the path ends in a symlink, otherwise return true if the path ends at a directory"?

compnerd (approved):

Sources/FoundationEssentials/URL/URL.swift
compnerd: Yeah, I think that this is the correct way. You need to check the `FILE_ATTRIBUTE_REPARSE_POINT` first as `FIEL_TYPE_DIRECTORY` can also be set to indicate a junction.

jmschonfeld (approved):

Sources/FoundationEssentials/URL/URL.swift
jrflat: Fixed!

jrflat:
@swift-ci please test

PR #1206, opened 2025-03-11
Introduce compatibility behavior of `Calendar.date(byAdding:value:to:wrappingComponents)`

itingliu:
Some app is passing in `value: -9223372036854775808` to the function below, and force unwrapping the result. It then crashes when the force unwrap fails.

```swift
/// - returns: A new date, or nil if a date could not be calculated with the given input.
public func date(byAdding component: Component, value: Int, to date: Date, wrappingComponents: Bool = false) -> Date?
```

The crash is caused by #1149. Prior to the change, we were using `_CalendarICU`'s implementation, where we truncate the input value to `Int32`, which becomes 0 in this case. That results in us returning the input `date` unchanged.

Now with #1149, we truthfully return `nil` because the calculation cannot be done.

We could restore the old behavior, but that implementation is incorrect. It's also clearly a wrong assumption on the client side. Add a compatibility check and restore the old behavior if needed.

Resolves 145862455

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -601,7 +601,14 @@ public struct Calendar : Hashable, Equatable, Sendable {
             return nil
         }
         
-        return self.date(byAdding: dc, to: date.capped, wrappingComponents: wrappingComponents)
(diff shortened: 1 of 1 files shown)


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1213, opened 2025-03-14
Move temporarily relocated FileManager extensions back to their original locations

cthielen:
This PR restores the original location of a handful of `FileManager` extensions that were temporarily relocated to work around a Swift compiler bug.


cthielen:
@swift-ci please test

PR #1215, opened 2025-03-17
Use the `package` access level instead of `@testable` imports

ahoppen:
`@testable` imports have a number of build-time drawbacks, including:
- Building the package in `release` configuration and then running tests requires the entire package to be rebuilt because SwiftPM needs to pass `--enable-testing` to the compiler invocations.
- The usage of `@testable` means that swift-foundation cannot be included in the unified (aka. multiroot) build of all the other packages in Swift CI, which means that it needs to re-build swift-syntax, unnecessarily increasing CI time.

To fix this, make all `@testable` imports normal imports and mark the declarations that were previously accessed through `@testable` as `package`.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+Guts.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+Guts.swift
@@ -11,15 +11,15 @@
 //===----------------------------------------------------------------------===//
 
 #if FOUNDATION_FRAMEWORK
-@_spi(Unstable) internal import CollectionsInternal
+@_spi(Unstable) package import CollectionsInternal
 #elseif canImport(_RopeModule)
-internal import _RopeModule
+package import _RopeModule
 #elseif canImport(_FoundationCollections)
-internal import _FoundationCollections
+package import _FoundationCollections
 #endif
 
 extension AttributedString {
-    internal final class Guts : @unchecked Sendable {
+    package final class Guts : @unchecked Sendable {
         typealias Index = AttributedString.Index
         typealias Runs = AttributedString.Runs
         typealias AttributeMergePolicy = AttributedString.AttributeMergePolicy
@@ -30,7 +30,7 @@ extension AttributedString {
         typealias _AttributeStorage = AttributedString._AttributeStorage
 
         var version: Version
-        var string: BigString
+        package var string: BigString
         var runs: _InternalRuns
         var trackedRanges: [Range<BigString.Index>]
 
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -38,7 +38,7 @@ extension Date {
         timeIntervalSinceReferenceDate / 86400 + Self.julianDayAtDateReference
     }
 
-    func julianDay() throws (GregorianCalendarError) -> Int {
+    package func julianDay() throws (GregorianCalendarError) -> Int {
         let jd = (julianDate + 0.5).rounded(.down)
         guard jd <= Double(Self.maxJulianDay), jd >= Double(Self.minJulianDay) else {
             throw .overflow(nil, self, nil)
@@ -51,7 +51,7 @@ extension Date {
         self.init(julianDate: Double(julianDay))
     }
 
-    init(julianDate: Double) {
+    package init(julianDate: Double) {
         let secondsSinceJulianReference = (julianDate - Self.julianDayAtDateReference) * 86400
         self.init(timeIntervalSinceReferenceDate: secondsSinceJulianReference)
     }
@@ -165,13 +165,13 @@ enum ResolvedDateComponents {
 
 
 /// Internal-use error for indicating unexpected situations when finding dates.
-enum GregorianCalendarError : Error {
+package enum GregorianCalendarError : Error {
     case overflow(Calendar.Component?, Date? /* failing start date */, Date? /* failing end date */)
     case notAdvancing(Date /* next */, Date /* previous */)
 }
 
 /// This class is a placeholder and work-in-progress to provide an implementation of the Gregorian calendar.
-internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
+package final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
 
 #if canImport(os)
     internal static let logger: Logger = {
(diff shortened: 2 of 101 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/AttributedString/AttributedString+Guts.swift
```diff
@@ -30,7 +30,7 @@ extension AttributedString {
         typealias _AttributeStorage = AttributedString._AttributeStorage
 
         var version: Version
-        var string: BigString
+        package var string: BigString
```
jmschonfeld, line 33: It looks like this is only used from one place: `NumberFormatStyleTests` which could easily do `String(self.characters)` instead of `String(self._guts.string)` which would save us from needing to mark this and the imports above as `package`.

For cases like this, would you rather we update them in a separate change and keep this as just a `@testable` --> `package` change, or would you want to incorporate that into this to reduce the number of things that we import as `package` like the collections modules above?

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -38,7 +38,7 @@ extension Date {
         timeIntervalSinceReferenceDate / 86400 + Self.julianDayAtDateReference
     }
 
-    func julianDay() throws (GregorianCalendarError) -> Int {
+    package func julianDay() throws (GregorianCalendarError) -> Int {
```
jmschonfeld, line 41: I don't see uses of this, the initializer below, or the error type from our tests, was this an accidental addition?

Sources/FoundationEssentials/FileManager/SwiftFileManagerDelegate.swift
```diff
@@ -35,7 +35,7 @@ public protocol FileManagerDelegate : AnyObject, Sendable {
 }
 
 // Default implementations for String-based requirements
-extension FileManagerDelegate {
+package extension FileManagerDelegate {
```
jmschonfeld, line 38: I think this should actually be _public_ instead of just package visibility, since we want clients to get these default implementations too (I guess this goes to show how uncommon it is for people to use this delegate on Linux/Windows)

Tests/FoundationEssentialsTests/AttributedString/AttributedStringCOWTests.swift
```diff
@@ -15,9 +15,9 @@ import TestSupport
 #endif
 
 #if FOUNDATION_FRAMEWORK
-@testable import Foundation
```
jmschonfeld, line 18: Since `TestSupport` has an `@_exported import Foundation`, we can just remove these few lines entirely. The only reason they're spelled out here is because `@_exported` doesn't "re-export" the `@testable`-ness of the import. Same goes for any files that have changes like this

Tests/FoundationInternationalizationTests/Formatting/ICUPatternGeneratorTests.swift
```diff
@@ -11,17 +11,17 @@ import TestSupport
...
 
 #if FOUNDATION_FRAMEWORK
-@testable import Foundation
+import Foundation
 #endif
 
 final class ICUPatternGeneratorTests: XCTestCase {
 
-    typealias DateFieldCollection = Date.FormatStyle.DateFieldCollection
+    package typealias DateFieldCollection = Date.FormatStyle.DateFieldCollection
```
jmschonfeld, line 24: Is this needed? The test module isn't a dependency of any other module, so it shouldn't need anything more permissive than `internal`

jmschonfeld (commented):
One overall comment: I just want to make sure that in all scenarios (building as a package, being built as part of the toolchain etc.) that making these `package` will not allow any client to start using them (unless they're doing some `@_silgen_name` shenanigans) since we don't actually install the package swift interface file anywhere, right?

Sources/FoundationEssentials/AttributedString/AttributedString+Guts.swift
ahoppen: I’d prefer if this is just a mechanical `@testable` -> `package` change. It’s fairly big already and I don’t want to hide any other functionality change in it.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
ahoppen: It’s used here: https://github.com/swiftlang/swift-foundation/blob/3179457ef5a0c1f6a07ad042c192012d35a0b81f/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift#L69

Sources/FoundationEssentials/FileManager/SwiftFileManagerDelegate.swift
ahoppen: Similar here, do you want to make a separate PR for that so we don’t hide functionality change in here?

Tests/FoundationEssentialsTests/AttributedString/AttributedStringCOWTests.swift
ahoppen: Removed all imports to `Foundation(Essentials|Internationalization|)` when `TestSupport` is imported here: a23e682df57dc0fc0f59573e8d1cf0f64d49ea98

Tests/FoundationInternationalizationTests/Formatting/ICUPatternGeneratorTests.swift
ahoppen: Yes, I changed this by mistake

ahoppen:
> One overall comment: I just want to make sure that in all scenarios (building as a package, being built as part of the toolchain etc.) that making these `package` will not allow any client to start using them (unless they're doing some `@_silgen_name` shenanigans) since we don't actually install the package swift interface file anywhere, right?

I am pretty sure that we don’t install `.package.swiftinterface` but to double-check I just kicked off toolchain builds in https://github.com/swiftlang/swift/pull/37710 and will check them.

Sources/FoundationEssentials/FileManager/SwiftFileManagerDelegate.swift
jmschonfeld: Yeah we can put this in another PR

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
jmschonfeld: Ah yeah that's probably why I missed it in my search since it didn't actually call the function - ok thanks for the explanation

ahoppen:
Looks like the package symbols are being successfully stripped.

I checked the Linux build:
```
ls -lah usr/lib/swift/linux/Foundation.swiftmodule 
total 9912
drwxr-xr-x@  4 alex  staff   128B Mar 21 17:23 ./
drwxr-xr-x@ 67 alex  staff   2.1K Mar 24 13:43 ../
-rw-r--r--@  1 alex  staff   278K Mar 21 17:18 x86_64-unknown-linux-gnu.swiftdoc
-rw-r--r--@  1 alex  staff   4.6M Mar 21 17:18 x86_64-unknown-linux-gnu.swiftmodule
```

Grepping both `.swiftmodule` and `.swiftdoc` for `numericStringRepresentationForBinaryInteger` did not return any results

Windows also only contains `.swiftmodule` files (no `.swiftinterface`). macOS toolchains don’t contain Foundation, so there’s nothing to check here.

parkera (changes requested):
I think it is extremely unfortunate to lose the meaning of what internal vs. package means, even for the benefit of improved CI times. This will also make all of these exported symbols on Darwin. Let's look for another solution.

Issue #1216, opened 2024-12-30
swift 6 cannot open correctly a windowsCP1252 file in linux (worked on swift 5 and macos)


Dracks:
### Description

Opening a file that is an export of transactions from an spanish bank (Caixa enginyers) which is an xls (html+latin1252) it gives a bad format using swift 6 compiler in linux. In mac-os or using swift 5.10 compiler works fine

### Reproduction

Download the sample file: 
https://github.com/Dracks/mr-scrooge/blob/mr-scrooge-3.0.0/src/swift-server-tests/test_files/MovimientosCuenta.xls

```swift
	guard let fileData = FileManager.default.contents(atPath: filePath)
				else {
					return
				}
				guard let fileString = String(data: fileData, encoding: .windowsCP1252)
				else {
					return
				}
```

### Expected behavior

get's the contents of the file in fileString variable. (Now the second ward execute the else case)

### Environment

Swift version 6.0.2 (swift-6.0.2-RELEASE)
Target: x86_64-unknown-linux-gnu

### Additional information

The link is a sample file exported from my bank, and anonymized, that I've got to run tests of my app. But today I expecuted it on my machine using the original file downloaded from the bank account and the same happened.

parkera:
The cause of the bug is that when `data(using:)` dropped from swift-corelibs-foundation into swift-foundation, we lost the ability to decode all of the encodings that only CoreFoundation has built-in converters for. We will need to add an upcall into swift-corelibs-foundation when it is present [here](https://github.com/swiftlang/swift-foundation/blob/fa43c96f5f91baaa0a86a1c44f9158ee36861c93/Sources/FoundationEssentials/String/StringProtocol%2BEssentials.swift#L251).

PR #1217, opened 2025-03-19
Add an upcall point to swift-corelibs-foundation for String encoding conversion

parkera:
Add an upcall to swift-corelibs-foundation for converting String encodings that swift-foundation does not implement.

Partially resolves https://github.com/swiftlang/swift-foundation/issues/1216.

### DIFF
--- a/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
+++ b/Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
@@ -91,6 +91,14 @@ extension UInt16 {
 
 // These provides concrete implementations for String and Substring, enhancing performance over generic StringProtocol.
 
+#if !FOUNDATION_FRAMEWORK
+@_spi(SwiftCorelibsFoundation)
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci test

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
```diff
@@ -91,6 +91,14 @@ extension UInt16 {
 
 // These provides concrete implementations for String and Substring, enhancing performance over generic StringProtocol.
 
+#if !FOUNDATION_FRAMEWORK
+@_spi(SwiftCorelibsFoundation)
+dynamic public func _cfStringEncodingConvert(string: String, using encoding: UInt, allowLossyConversion: Bool) -> Data? {
```
jmschonfeld, line 96: Should we add a similar up-call for the bytes --> `String` direction so that we support decoding as well?

parkera: Good idea, done.

jmschonfeld (approved):

parkera:
@swift-ci test

PR #1219, opened 2025-03-19
Fix ISO Latin 1 Encoding/Decoding issues

jmschonfeld:
As mentioned in https://github.com/swiftlang/swift-foundation/issues/1216, the ISO Latin 1 Encoding/Decoding written natively in swift-foundation does not behave correctly. I accidentally referenced an incorrect code page table when writing the original implementation which limited the supported characters to a subset of those actually supported. In reality, ISO Latin 1 is just the 0x0 through 0xFF unicode code points encoded as single bytes. This means that encoding should simply encode the code point value (and fail for any code point greater than 0xFF) and decoding can simply extend each byte to 16 bits and parse as UTF-16 since all bytes are valid by definition.

This updates the implementation and adds some extra characters to the unit test to validate this behavior.

### DIFF
--- a/Sources/FoundationEssentials/String/String+IO.swift
+++ b/Sources/FoundationEssentials/String/String+IO.swift
@@ -18,33 +18,6 @@ internal import _FoundationCShims
 
 fileprivate let stringEncodingAttributeName = "com.apple.TextEncoding"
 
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1224, opened 2025-03-21
Improve performance for comparing AttributedStrings with differing character counts

jmschonfeld:
Two `AttributedString`s with differing character counts will never compare equal. When we can cheaply determine that the character counts are inequal, we should return early from `AttributedString.==` rather than comparing each runs contents. This situation can be quite common (for example, just appending a single character to the end of the last run like when a user types into a text box) and it turns a full run-by-run comparison of the entire string into a single constant time evaluation. The benchmark results highlight that for this worst-case, but common, scenario the wins are significant:

```
╒══════════════════════════════════════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╤═════════╕
│          Throughput (# / s) (M)          │      p0 │     p25 │     p50 │     p75 │     p90 │     p99 │    p100 │ Samples │
╞══════════════════════════════════════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╪═════════╡
│                   main                   │      92 │      91 │      90 │      88 │      83 │      71 │      71 │      89 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│                  branch                  │  480077 │  461567 │  453119 │  452863 │  444671 │  353023 │   11964 │   10000 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│                    Δ                     │  479985 │  461476 │  453029 │  452775 │  444588 │  352952 │   11893 │    9911 │
├──────────────────────────────────────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┼─────────┤
│              Improvement %               │  521723 │  507116 │  503366 │  514517 │  535648 │  497115 │   16751 │    9911 │
╘══════════════════════════════════════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╧═════════╛
```

Note: there's an explanation left in the comments of the code but this optimization is currently only performed for `AttributedString.==` and not for `AttributedSubstring.==` or `DiscontiguousAttributedSubstring.==`.

### DIFF
--- a/Benchmarks/Benchmarks/AttributedString/BenchmarkAttributedString.swift
+++ b/Benchmarks/Benchmarks/AttributedString/BenchmarkAttributedString.swift
@@ -414,12 +414,21 @@ let benchmarks = {
 #endif
     
     let manyAttributesString2 = createManyAttributesString()
+    let manyAttributesString3 =  {
+        var str = createManyAttributesString()
+        str.characters.append("a")
+        return str
+    }()
     let manyAttributesStringRange = manyAttributesString.characters.index(manyAttributesString.startIndex, offsetBy: manyAttributesString.characters.count / 2)...
     let manyAttributesSubstring = manyAttributesString[manyAttributesStringRange]
     let manyAttributes2Substring = manyAttributesString2[manyAttributesStringRange]
 
     Benchmark("equality") { benchmark in
         blackHole(manyAttributesString == manyAttributesString2)
+    }
+    
+    Benchmark("equalityDifferingCharacters") { benchmark in
+        blackHole(manyAttributesString == manyAttributesString3)
     }
     
     Benchmark("substringEquality") { benchmark in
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

lorentey (approved):
Nit: Can we replace the `fullString:` argument with a simple `_isPartial` (or similar) predicate on `AttributedString.Runs`?

It's fast/easy to test if the instance is over a full string; we do not need to pass this bit down explicitly. Eliminating the argument allows the fast path to trigger on substrings that happen to cover the entire string. It also simplifies flow, keeping all comparison logic in the `_characterwiseIsEqual` function.

jmschonfeld:
@lorentey sure thing! Updated the implementation to add an `_isPartial` property rather than passing it from the caller and confirmed that the benchmark results are unchanged.

jmschonfeld:
@swift-ci please test

PR #1229, opened 2025-03-27
Use the default time zone instead of gmt

parkera:
Match the previous behavior of an `iso8601` calendar by setting the time zone to the `default` (which falls back to `current`) instead of using only `gmt`.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #1233, opened 2025-03-28
Resolve deprecation warning in Foundation Macros concerning FunctionCallExprSyntax initializer

jmschonfeld:
This resolves a deprecation warning that was introduced by a swift-syntax change. We resolve the deprecation by switching to the recommended initializer.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1235, opened 2025-03-31
[wasm] Add O_NONBLOCK shim

kateinoigakukun:
05975959e67f57063445baabb652262dd764cc72 added a use of `O_NONBLOCK` but the constant cannot be imported from wasi-libc through ClangImporter directly, so we need to add a shim for it as well as other constants.


kateinoigakukun:
@swift-ci test

parkera (approved):

parkera:
Thanks!

Issue #1029, opened 2024-11-01
Swift should support kCFStringEncodingDOSLatinUS on Linux


microtherion:
This might be another issue classified under the #1015 umbrella.

Summary:
The `String(data:encoding:)` constructor fails on Linux when encoding is kCFStringEncodingDOSLatinUS.

Steps To Reproduce:
[ZIPFoundation](https://github.com/weichsel/ZIPFoundation) is doing the following:

```
        let dosLatinUS = 0x400
        let dosLatinUSEncoding = CFStringEncoding(dosLatinUS)
        let dosLatinUSStringEncoding = CFStringConvertEncodingToNSStringEncoding(dosLatinUSEncoding)
        let codepage437 = String.Encoding(rawValue: dosLatinUSStringEncoding)
        let encoding = self.centralDirectoryStructure.usesUTF8PathEncoding ? .utf8 : codepage437
```

This code works on Apple platforms, but fails on Linux.

The reason CP437 is needed for ZIP file decoding is that the ZIP Specification specifies that the encoding is CP437, unless a UTF-8 flag is set. So for legacy ZIP files, CP437 must be assumed (although in practice, encodings were far more random).

YOCKOW:
Thank you. Added this to #1015.
However, I want to note that `kCFStringEncodingDOSLatinUS` is not one of [`CFStringBuiltInEncodings`](https://developer.apple.com/documentation/corefoundation/cfstringbuiltinencodings?language=objc) but one of [External String Encodings](https://developer.apple.com/documentation/corefoundation/cfstring/external_string_encodings?language=objc).

In the context of [the pitch to add string encoding names](https://forums.swift.org/t/pitch-foundation-string-encoding-names/74623), a suggestion to add new static properties to `String.Encoding` received some kind of negative feedbacks.

davedelong:
Any advice on working around this? I'm using [CoreXLSX](https://github.com/CoreOffice/CoreXLSX) (which relies on ZIPFoundation) to parse xlsx files on linux, and the file entries in the spreadsheet are all encoded in CP437, which means it can't locate the spreadsheet's data because all the entry names fail to decode.

YOCKOW:
First of all, we have to expose the encoding as a static property of `String.Encoding` like this[^cf-ns-encoding-value]:

```swift
extension String.Encoding {
  public static let dosCP437 = Encoding(rawValue: 0x80000400)
}
```

[...]

PR #1243, opened 2025-04-08
[Proposal] Add "String Encoding Names" proposal.

YOCKOW:
This proposal allows `String.Encoding` to be converted to and from various names.

For example:

```swift
print(String.Encoding.utf8.ianaName!) // Prints "UTF-8"
print(String.Encoding(ianaName: "ISO_646.irv:1991") == .ascii) // Prints "true"
```

-----

Implementation: https://github.com/swiftlang/swift-foundation/pull/1286

### DIFF
--- a/Proposals/NNNN-String-Encoding-Names.md
+++ b/Proposals/NNNN-String-Encoding-Names.md
@@ -1,0 +1,317 @@
+# String Encoding Names
+
+* Proposal: Not assigned yet <!-- [FOU-NNNN](NNNN-filename.md) -->
+* Author(s): [YOCKOW](https://GitHub.com/YOCKOW)
+* Review Manager: TBD
+* Status: **Awaiting review**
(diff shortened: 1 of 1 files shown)


YOCKOW:
@itingliu

I think #1029 implies legacy encodings are still necessary in some areas.
While this proposed feature doesn't support some legacy encodings, this feature will make it easy to implement more string conversions since we already have [the string converter](https://github.com/swiftlang/swift-foundation/blob/a8bee5bfc71210168fa1b973fb1a1deb8bde2047/Sources/FoundationInternationalization/ICU/ICU%2BStringConverter.swift) that wraps ICU APIs and that requires IANA Charset Names to create an instance of naive ICU converter.

I mean I believe this feature is useful not only as public properties but also for internal use.
I hope this proposal will come up to the official process.

itingliu:
@swift-ci please test

PR #1244, opened 2025-04-08
Markdown List Delimiters

jmschonfeld:
This proposal adds a new `AttributedString` attribute that will be populated when parsing markdown with the delimiter used for entries in a list.


jmschonfeld:
@swift-ci please test

PR #1234, opened 2025-03-29
[Proposal] Writing Direction Attribute

theMomax:
Adds an `AttributedStringKey` for the base writing direction of a paragraph.

PR #1245, opened 2025-04-10
Writing Direction Attribute Implementation

theMomax:
The implementation for https://github.com/swiftlang/swift-foundation/pull/1234

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/FoundationAttributes.swift
+++ b/Sources/FoundationEssentials/AttributedString/FoundationAttributes.swift
@@ -523,7 +523,7 @@ extension AttributeScopes.FoundationAttributes {
     @frozen
     public enum WritingDirectionAttribute: CodableAttributedStringKey {
         public typealias Value = AttributedString.WritingDirection
-        public static let name: String = "Foundation.WritingDirection"
+        public static let name: String = "Foundation.WritingDirectionAttribute"
 
         public static let runBoundaries: AttributedString.AttributeRunBoundaries? = .paragraph
         public static let inheritedByAddedText = false
--- a/Sources/FoundationEssentials/AttributedString/FoundationAttributes.swift
+++ b/Sources/FoundationEssentials/AttributedString/FoundationAttributes.swift
@@ -34,7 +34,14 @@ extension AttributeScopes {
         
         @available(macOS 13, iOS 16, tvOS 16, watchOS 9, *)
         public let durationField: DurationFieldAttribute
-        
+
+        /// The base writing direction of a paragraph.
+        #if FOUNDATION_FRAMEWORK
+        @_spi(AttributedStringWritingDirection)
+        #endif
+        @available(FoundationPreview 6.2, *)
+        public let writingDirection: WritingDirectionAttribute
+
 #if FOUNDATION_FRAMEWORK
         @available(FoundationPreview 0.1, *)
         public let agreementConcept: AgreementConceptAttribute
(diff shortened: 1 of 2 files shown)


theMomax:
@swift-ci please test

Sources/FoundationEssentials/AttributedString/FoundationAttributes.swift
```diff
@@ -507,7 +514,21 @@ extension AttributeScopes.FoundationAttributes {
...
+
+    /// The attribute key for the base writing direction of a paragraph.
+    #if FOUNDATION_FRAMEWORK
+    @_spi(BringUpWritingDirectionAPI)
+    #endif
+    @available(FoundationPreview 6.2, *)
+    @frozen
+    public enum WritingDirectionAttribute: CodableAttributedStringKey {
+        public typealias Value = AttributedString.WritingDirection
+        public static let name: String = "Foundation.WritingDirection"
```
jmschonfeld, line 526: I know that we're a bit inconsistent with the existing attributes today, but most attributes use the form `Foundation.WritingDirectionAttribute` as the name rather than `Foundation.WritingDirection`. Could we update this to follow that convention just to try to stick with what the majority of attributes use?

Tests/FoundationEssentialsTests/AttributedString/AttributedStringTests.swift
```diff
@@ -2599,4 +2599,26 @@ E {
         XCTAssertEqual(testConstrainedContainer.filter(inheritedByAddedText: true), AttributeContainer.testInt(2).testParagraphConstrained(3).testCharacterConstrained(4))
         XCTAssertEqual(testConstrainedContainer.filter(inheritedByAddedText: false), AttributeContainer.testNonExtended(5))
     }
+
+    func testWritingDirectionBehavior() {
+        // Indicate that this sentence is primarily right to left, because the English term "Swift" is embedded into an Arabic sentence.
+        var string = AttributedString("Swift مذهل!", attributes: .init().writingDirection(.rightToLeft))
+
+        // To remove the information about the writing direction, set it to `nil`:
+        string.writingDirection = nil
```
jmschonfeld, line 2608: Is this intentional / did you mean to have some form of assertion between when you set the attribute above and when you removed it here?

Tests/FoundationEssentialsTests/AttributedString/AttributedStringTests.swift
```diff
@@ -2599,4 +2599,26 @@ E {
...
+
+        let range = string.range(of: "Swift")!
```
jmschonfeld, line 2610: small nit to avoid test crashes if a bug is introduced
```suggestion
        let range = try XCTUnwrap(string.range(of: "Swift"))
```

Tests/FoundationEssentialsTests/AttributedString/AttributedStringTests.swift
```diff
@@ -2599,4 +2599,26 @@ E {
...
+
+        // When setting or removing the value from a certain range, the value will always be applied to the entire paragraph(s) that intersect with that range:
+        string[range].writingDirection = .leftToRight
+        assert(string.runs[\.writingDirection].count == 1)
```
jmschonfeld, line 2614:
```suggestion
        XCTAssertEqual(string.runs[\.writingDirection].count, 1)
```

In general we prefer `XCTest` assertions here rather than `assert` since `assert` is only evaluated in debug builds and will crash the test runner if there's a bug in the output rather than providing a helpful message / output.

Tests/FoundationEssentialsTests/AttributedString/AttributedStringTests.swift
theMomax: Forgot the assert 👍

theMomax:
@swift-ci please test

theMomax:
@swift-ci please test

jmschonfeld (approved):

PR #1246, opened 2025-04-10
Fix FileManager.isExecutableFile for emulated x86 processes on ARM systems

### LINKED ISSUE #860: FileManager.isExecutableFile fails for wow64 processes
The following code, if compiled as an x86_64 process and run on an arm64 system, will claim that the command shell binary is not executable. This is because GetBinaryTypeW returns ERROR_BAD_EXE_FORMAT for emulated processes.

This is not necessarily a bug in Foundation, but a quirk in the Win32 APIs. It seems worth documenting this caveat, at least.

```swift
import Foundation
import WinSDK

let fm = FileManager()
let path = "C:\\Windows\\system32\\cmd.exe"
print("\(path) isExecutable = \(fm.isExecutableFile(atPath: path))") // false

var dwBinaryType: DWORD = .max
let ret = path.withCString(encodedAs: UTF16.self) { ptr in
    GetBinaryTypeW(ptr, &dwBinaryType)
}
let err = GetLastError()
print("\(ret) \(dwBinaryType) \(err)")

// err == ERROR_BAD_EXE_FORMAT
```

jakepetroules:
GetBinaryType will return ERROR_BAD_EXE_FORMAT when querying an arm64 executable from an x86 process running on an ARM system. This change switches the implementation to use SHGetFileInfoW, which isn't subject to this quirk.

This also makes isExecutableFile behave more similarly to other platforms -- e.g. isExecutableFile already returns true for any file with the execute bit, even for an arm64 executable on an x86_64 macOS system (which it can't actually run). The spirit of the API is that the file is of an executable type, not necessarily that the running system is capable of executing it.

The practical consequence of fixing this bug is that queries like:

```swift
FileManager.default.isExecutableFile(atPath: "C:\\Windows\\system32\\cmd.exe")
```

will now correctly return true regardless of what architecture the binary is compiled for or what type of system it's running on.

Closes #860


jakepetroules:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -448,8 +448,8 @@ extension _FileManagerImpl {
     func isExecutableFile(atPath path: String) -> Bool {
 #if os(Windows)
         return (try? path.withNTPathRepresentation {
-            var dwBinaryType: DWORD = 0
-            return GetBinaryTypeW($0, &dwBinaryType)
+            // Use SHGetFileInfo instead of GetBinaryType because the latter returns the wrong answer for x86 binaries running under emulation on ARM systems.
+            return (SHGetFileInfoW($0, 0, nil, 0, numericCast(SHGFI_EXETYPE)) & 0xFFFF) != 0
```
compnerd, line 452: What do you think of adding `SHGFI_EXETYPE` to `WinSDK+Extensions.swift` to avoid the `numericCast`?

compnerd (approved):

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
jakepetroules: Done

jakepetroules:
@swift-ci test

PR #1248, opened 2025-04-14
[wasm] Stop requiring `TZDIR` and `TZDEFAULT` on WASI

kateinoigakukun:
Those constants are not used on WASI in Swift side, so there's no need to require them in the C headers. The `#error` was introduced by 03fe46f43b3a2d8d9159d46a1f56517acc525e69.


kateinoigakukun:
@swift-ci test

kateinoigakukun:
Yes, the platform doesn't have the time zone concept.

PR #1249, opened 2025-04-14
Add Hashable conformance to Measurement<UnitType>

beltradini:
### Summary
This PR introduces `Hashable` conformance to `Measurement<UnitType>` when `UnitType` conforms to `Hashable`. This enables `Measurement` to be used in sets, dictionaries, and other standard collections that rely on hashing.

### Motivation
Currently, `Measurement` in `swift-foundation` does not conform to `Hashable`, unlike its counterpart in Foundation. This restricts its usage in Swift collections and functional APIs. By introducing this conformance, we bring the type closer to parity with Foundation and improve its flexibility in Swift-native environments.

### Implementation
- Added `Hashable` conformance to `Measurement<UnitType>` using conditional constraints.
- Implemented `hash(into:)` based on `value` and `unit`.
- Implemented `==` comparison using both `value` and `unit`.

### Testing
Created a new test suite `MeasurementTests.swift` with full coverage, including:
- Set membership behavior
- Dictionary key usage
- Hash value consistency
- Equality comparison
- Edge cases (zero, negative, NaN)
All tests pass via `swift test`.

### Additional context
This is an additive enhancement with no impact on existing APIs. The implementation is consistent with Swift's design patterns and follows the behavior of `Measurement` in Foundation on Apple platforms.

Let me know if any changes or adjustments are needed — happy to iterate!

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Measurement+FormatStyle+Stub.swift
+++ b/Sources/FoundationInternationalization/Formatting/Measurement+FormatStyle+Stub.swift
@@ -13,7 +13,25 @@
 #if !FOUNDATION_FRAMEWORK
 
 // stub
-public struct Measurement<UnitType> {}
+public struct Measurement<UnitType: Hashable>: Hashable {
+    public var value: Double
+    public var unit: UnitType
+    
+    public init(value: Double, unit: UnitType) {
+        self.value = value
+        self.unit = unit
+    }
+    
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(value)
+        hasher.combine(unit)
+    }
+    
+    public static func == (lhs: Measurement<UnitType>, rhs: Measurement<UnitType>) -> Bool {
+        return lhs.value == rhs.value && lhs.unit == rhs.unit
+    }
+}
+
 public class Dimension {}
 public class UnitDuration: Dimension {}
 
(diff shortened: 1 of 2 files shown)


jmschonfeld:
Hey @beltradini - thanks for the contribution! Unfortunately I don't think this change actually accomplishes the goal you outlined here. This `Measurement` type is not actually exposed as API, it's just a stub implementation here in swift-foundation to allow declaring subtypes in `FoundationInternationalization` as `Measurement` has not yet been moved to swift-foundation. The real implementation exposed via the `Foundation` module on Linux and Windows exists at https://github.com/swiftlang/swift-corelibs-foundation/blob/fefb1bf4b76279a43f7d7543b9f8432706adf377/Sources/Foundation/Measurement.swift#L26. Additionally, that public version of `Measurement` is already declared as `Hashable`, so the type should already be `Hashable` on all platforms. Are you experiencing issues with using `Measurement` as a `Hashable` type today? I don't think we intend for this stub type to become a full replacement for `Measurement` yet, but if it's causing problems we can address those if you have a sample project that hits issues

beltradini:
Thanks for the detailed response, @jmschonfeld ! That makes total sense. I wasn't aware that the current `Measurement` in `swift-foundation` was a stub and not yet meant to act as the functional API. I appreciate the clarification and the link to the core implementation. My intention was to help bring this closer to parity with Foundation, but I know understand that the actual `Hashable` conformance already exists where it matters. Thanks again for taking the time to explain!

itingliu:
> Are there any areas within swift-foundation currently open to contribution where implementations are API-facing and not just stubs? @jmschonfeld 

You can find a list of issues in the issues tab.

itingliu:
Since this isn't a valid fix, please close the pull request.

PR #1251, opened 2025-04-15
[Proposal]: `#bundle` macro

Tantalum73:
This proposal adds a macro to find the bundle that contains resources useful for the current context: `#bundle`.

### DIFF
--- a/Proposals/0024-CurrentBundle.md
+++ b/Proposals/0024-CurrentBundle.md
@@ -1,10 +1,10 @@
 # Introduce `#bundle`
 
 
-* Proposal: [SF-NNNN](NNNN-filename.md)
+* Proposal: [SF-0024](0024-filename.md)
 * Authors:[Matt Seaman](https://github.com/matthewseaman), [Andreas Neusuess](https://github.com/Tantalum73)
-* Review Manager: TBD
-* Status: **Awaiting review**
+* Review Manager: [Tina L](https://github.com/itingliu)
+* Status: **Accepted**
 
 
 ## Revision history
--- a/Proposals/NNNN-CurrentBundle.md
+++ b/Proposals/NNNN-CurrentBundle.md
@@ -10,6 +10,7 @@
 ## Revision history
 
 * **v1** Initial version
+* **v1.1** Remove `#bundleDescription` and add 2 initializers to `LocalizedStringResource`
 
 ## Introduction
 
@@ -62,16 +63,6 @@ label.text = String(
     )
 ```
 
-We will also introduce an equivalent macro for usage with `LocalizedStringResource.BundleDescription`.
-
-```swift
-let string = LocalizedStringResource(
-    "She didn't clean the camera!",
-    bundle: #bundleDescription,
-    comment: "Comment of astonished bystander"
-    )
-```
-
 ## Detailed design
 
 We propose introducing a `#bundle` macro as follows:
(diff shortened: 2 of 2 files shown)


itingliu:
@swift-ci please test

Proposals/NNNN-CurrentBundle.md
```diff
@@ -0,0 +1,184 @@
+# Introduce `#bundle`
+
+
+* Proposal: [SF-NNNN](NNNN-filename.md)
```
itingliu, line 4:
```suggestion
* Proposal: [SF-0024](NNNN-filename.md)
```

Proposals/NNNN-CurrentBundle.md
```diff
@@ -0,0 +1,184 @@
...
+* Authors:[Matt Seaman](https://github.com/matthewseaman), [Andreas Neusuess](https://github.com/Tantalum73)
+* Review Manager: TBD
```
itingliu, line 6:
```suggestion
* Review Manager: [Tina L](https://github.com/itingliu)
```

itingliu (approved):

itingliu:
@swift-ci please test

Proposals/0024-CurrentBundle.md
```diff
@@ -0,0 +1,184 @@
...
+## Revision history
+
+* **v1** Initial version
+* **v1.1** Remove `#bundleDescription` and add 2 initializers to `LocalizedStringResource`
+
+## Introduction
+
+API which loads localized strings assumes `Bundle.main` by default. This works for apps, but code that runs in a framework, or was defined in a Swift package, needs to specify a different bundle. The ultimate goal is to remove this requirement in the future. One step towards that goal is to provide an easy accessor to the bundle that stores localized resources: `#bundle`.
+
+## Motivation
```
stmontgomery, line 19: The example of localized resources in a framework or library is a solid motivator for this feature. But if you're interested in citing other motivators, I have one to suggest:

Swift Testing allows and recommends using structs (value types) instead of classes for grouping related test functions. Often, test authors need to reference resources in their test bundle for things like fixture date files. Historically it has been straightforward to use `Bundle(forClass:)` and pass the XCTestCase subclass in the current file to accomplish this. But when using Swift Testing idiomatically, users may not have a class in the test bundle module which can conveniently serve this purpose. And of course, as its name implies, `Bundle(forClass:)` does require a class meta type and doesn't accept the type of a struct. So this feature will benefit Swift Testing users and allow them to use value types for suites even more easily.

stmontgomery: (I mean `Bundle.init(for:)` which accepts `AnyClass`, but you get the idea.)

Proposals/0024-CurrentBundle.md
```diff
@@ -0,0 +1,184 @@
...
+
+The type `LocalizedStringResource` (LSR) doesn't operate on instances of `Bundle`, but `LocalizedStringResource.BundleDescription`. They can easily be converted into each other.  
+To make the new macro work well with LSR, we suggest adding two new initializers. We mark them as `@_alwaysEmitIntoClient` and `@_disfavoredOverload`, to avoid ambiguity over the initializers accepting a `BundleDescription` parameter:
+
+```swift
+@available(FoundationPreview 6.2, *)
+extension LocalizedStringResource {
+    @_alwaysEmitIntoClient
+    @_disfavoredOverload
+    public init(_ keyAndValue: String.LocalizationValue, table: String? = nil, locale: Locale = .current, bundle: Bundle, comment: StaticString? = nil)
```
stmontgomery, line 125: Is there a reason these _couldn't_ declare a default value of `#bundle` for the `bundle:` parameter now, in this proposal, leveraging SE-0422 immediately? I see that mentioned in Alternatives Considered, but it's unclear why it couldn't or shouldn't be done now when this new extension is being introduced.

Tantalum73: Thanks for bringing this up!

The other existing, similar initializer to this one is this:

```swift
LocalizedStringResource(
    _ keyAndValue: String.LocalizationValue,     
    table: String? = nil,     
    locale: Locale = .current,     
    bundle: LocalizedStringResource.BundleDescription = .main,     
    comment: StaticString? = nil
)
```

If our new initializer only took parameters with default values, one had to be marked as disfavored, probably the new one:
```swift
@_disfavoredOverload
LocalizedStringResource(
    _ keyAndValue: String.LocalizationValue, 
    table: String? = nil, 
    locale: Locale = .current, 
    bundle: Bundle = #bundle, 
    comment: StaticString? = nil
)
```

With the new initializer being disfavored, its default value would never take effect: `LocalizedStringResource("Hello")` always uses the existing initializer, and its default values, but never the new initializer (with its default values).  
To use the new initializer, one has to write `LocalizedStringResource("Hello", bundle: <...>)`, in which case a default value for `bundle` doesn't matter.

As eluded to in Future Directions, we can't change the default value for all `bundle: ` parameters this year. For consistency's sake, and because a default value here wouldn't apply anyway, I think it's better to not default anything to `#bundle` for now.

Proposals/0024-CurrentBundle.md
Tantalum73: Great example, I haven't thought about Swift Testing!

PR #1252, opened 2025-04-16
Remove URL workaround for Linux crash once SCL-F is rebuilt

jrflat:
Tested with a recent nightly-main container, and removing the Linux-only workaround `unsafeBitCast == 0x1` check works now that swift-corelibs-foundation was built with the new swift-foundation `URL` changes in #1238.

Removing the other `private var _padding` workaround still causes a crash in libXCTest on Linux, so that will require further investigation.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -690,18 +690,6 @@ public struct URL: Equatable, Sendable, Hashable {
     ///
     /// Returns `nil` if a `URL` cannot be formed with the string (for example, if the string contains characters that are illegal in a URL, or is an empty string).
     public init?(string: __shared String, relativeTo url: __shared URL?) {
-        #if os(Linux)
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jrflat:
Works locally on my Ubuntu 22.04 aarch64 image, but fails in CI's Ubuntu 22.04 x86 image

jrflat:
@swift-ci please test Linux Platform

jrflat:
Tested on an x86 Ubuntu 22.04 image locally with `swift-DEVELOPMENT-SNAPSHOT-2025-04-12-a` and it worked. Looks like CI was using an older `swift-DEVELOPMENT-SNAPSHOT-2025-04-03-a` based on the commit it logged. Let's try again

jrflat:
@swift-ci please test Linux Platform

jrflat:
`swift-foundation` tests pass in CI now that the newer toolchain is being used. I also ran Swift smoke tests to sanity check, and those passed, too.

PR #1256, opened 2025-04-21
AttributedString character insertion doesn't invalidate text dependent attributes

jmschonfeld:
When inserting a character into an attribute run, if that run contains any text-dependent attributes we should invalidate those attributes. However if the text is inserted between two runs of that attribute value, neither surrounding run should be invalidated. This updates `enforceAttributeConstraintsBeforeMutation(to:)` to ensure that we handle an empty range mutation (i.e. an insertion) correctly rather than skipping constraint enforcement for empty ranges.

rdar://149623441

### DIFF
--- a/Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
+++ b/Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
@@ -509,6 +509,14 @@ class TestAttributedStringConstrainingBehavior: XCTestCase {
         verify(string: result, matches: [("Hello, world", 1, 2), ("ABC", 1, nil)], for: \.testInt, \.testCharacterDependent)
         
         result = str
+        result.characters.insert("A", at: result.startIndex)
+        verify(string: result, matches: [("A", 1, nil), ("Hello, world", 1, 2)], for: \.testInt, \.testCharacterDependent)
+        
+        result = str
+        result.characters.insert("A", at: result.index(afterCharacter: result.startIndex))
+        verify(string: result, matches: [("HAello, world", 1, nil)], for: \.testInt, \.testCharacterDependent)
+        
+        result = str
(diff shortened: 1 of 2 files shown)


Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
```diff
@@ -566,5 +574,13 @@ class TestAttributedStringConstrainingBehavior: XCTestCase {
...
     }
-
+    
+    func testInvalidationCharacterInsertionBetweenRuns() {
+        var str = AttributedString("Hello", attributes: .init().testInt(1).testCharacterDependent(2))
+        str += AttributedString("World", attributes: .init().testInt(1).testCharacterDependent(3))
+        
+        // Inserting text between two runs should not invalidate text dependent attributes in either of the surrounding runs
+        str.characters.insert("|", at: str.index(str.startIndex, offsetByCharacters: 5))
+        verify(string: str, matches: [("Hello", 1, 2), ("|", 1, nil), ("World", 1, 3)], for: \.testInt, \.testCharacterDependent)
```
itingliu, line 584: Was the previous behavior`[("Hello", 1, nil), ("|", 1, nil), ("World", 1, nil)]`?

itingliu (approved):

Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
jmschonfeld: The previous behavior here would have been `[("Hello", 1, 2), ("|", 1, nil), ("World", 1, 3)]` as well, because we previously ignored invalidation for all insertions. But now that we're handling insertions and invalidating them in other places, I just added this unit test to ensure this doesn't break the invariant that invalidations shouldn't happen when inserting between runs. The test that previously failed without this change is the one above where inserting "A" into "Hello" would result in the surrounding "Hello" still having the text dependent attribute value set.

PR #1257, opened 2025-04-21
Transparently add the \\?\ prefix to Win32 calls for extended length path handling

jakepetroules:
On Windows, there is a built-in maximum path limitation of 260 characters under most conditions. This can be extended to 32767 characters under either of the following two conditions:

- Adding the longPathAware attribute to the executable's manifest AND enabling the LongPathsEnabled system-wide registry key or group policy.
- Ensuring fully qualified paths passed to Win32 APIs are prefixed with \?\

Unfortunately, the former is not realistic for the Swift ecosystem, since it requires developers to have awareness of this specific Windows limitation, AND set longPathAware in their apps' manifest AND expect end users of those apps to change their system configuration.

Instead, this patch transparently prefixes all eligible paths in calls to Win32 APIs with the \?\ prefix to allow them to work with paths longer than 260 characters without requiring the caller of Foundation to manually prefix the paths.

See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation for more info.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -23,7 +23,12 @@ import Darwin
 import WinSDK
 
 extension String {
-    package func withNTPathRepresentation<Result>(_ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
+    /// Invokes `body` with a resolved and potentially `\\?\`-prefixed version of the pointee,
+    /// to ensure long paths greater than MAX_PATH (260) characters are handled correctly.
+    ///
+    /// - parameter relative: Returns the original path without transforming through GetFullPathNameW + PathCchCanonicalizeEx, if the path is relative.
+    /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+    package func withNTPathRepresentation<Result>(relative: Bool = false, _ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
         guard !isEmpty else {
             throw CocoaError.errorWithFilePath(.fileReadInvalidFileName, "")
         }
@@ -35,15 +40,42 @@ extension String {
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
         return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString(encodedAs: UTF16.self) { pwszPath in
+            if relative && PathIsRelativeW(pwszPath) {
+                return try body(pwszPath)
+            }
+
             // 1. Normalize the path first.
+            // Contrary to the documentation, this works on long paths independently
+            // of the registry or process setting to enable long paths (but it will also
+            // not add the \\?\ prefix required by other functions under these conditions).
             let dwLength: DWORD = GetFullPathNameW(pwszPath, 0, nil, nil)
-            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-                guard GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil) > 0 else {
+            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { pwszFullPath in
+                guard (1..<dwLength).contains(GetFullPathNameW(pwszPath, DWORD(pwszFullPath.count), pwszFullPath.baseAddress, nil)) else {
                     throw CocoaError.errorWithFilePath(self, win32: GetLastError(), reading: true)
                 }
 
-                // 2. Perform the operation on the normalized path.
-                return try body($0.baseAddress!)
+                // 1.5 Leave \\.\ prefixed paths alone since device paths are already an exact representation and PathCchCanonicalizeEx will mangle these.
+                if let base = pwszFullPath.baseAddress,
+                    base[0] == UInt16(UInt8._backslash),
+                    base[1] == UInt16(UInt8._backslash),
+                    base[2] == UInt16(UInt8._period),
+                    base[3] == UInt16(UInt8._backslash) {
+                    return try body(base)
+                }
+
+                // 2. Canonicalize the path.
+                // This will add the \\?\ prefix if needed based on the path's length.
+                var pwszCanonicalPath: LPWSTR?
+                let flags: ULONG = PATHCCH_ALLOW_LONG_PATHS
+                let result = PathAllocCanonicalize(pwszFullPath.baseAddress, flags, &pwszCanonicalPath)
+                if let pwszCanonicalPath {
+                    defer { LocalFree(pwszCanonicalPath) }
+                    if result == S_OK {
+                        // 3. Perform the operation on the normalized path.
+                        return try body(pwszCanonicalPath)
+                    }
+                }
+                throw CocoaError.errorWithFilePath(self, win32: WIN32_FROM_HRESULT(result), reading: true)
             }
         }
     }
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -1072,4 +1072,92 @@ final class FileManagerTests : XCTestCase {
             XCTAssertEqual($0.contents(atPath: "a\u{301}/test"), data)
         }
     }
+
+    /// Tests that Foundation can correctly handle "long paths" (paths of 260 to 32767 chacters long) on Windows.
+    func testWindowsLongPathSupport() throws {
+        #if !os(Windows)
+        throw XCTSkip("This test is not applicable for this platform")
+        #else
+        // Create a directory with the absolute maximum path _component_ length of 255;
+        // this will guarantee the full playground path is well over 260 characters.
+        // Throw some Unicode in there for good measure, since only wide-character APIs support it.
+        let dirName = String(repeating: UUID().uuidString, count: 7) + "你好！"
+        XCTAssertEqual(dirName.count, 255)
+        XCTAssertEqual(dirName.utf16.count, 255)
+
+        try FileManagerPlayground {
+            Directory(dirName) {
+            }
+        }.test {
+            // Call every function that can call into withNTPathRepresentation with an overlong path and ensure it succeeds.
+            let fileName = UUID().uuidString
+            let cwd = try XCTUnwrap($0.currentDirectoryPath)
+
+            XCTAssertTrue($0.createFile(atPath: dirName + "/" + fileName, contents: nil))
+
+            let dirURL = URL(filePath: dirName, directoryHint: .checkFileSystem)
+            XCTAssertTrue(dirURL.hasDirectoryPath)
+
+            let fileURL = URL(filePath: dirName + "/" + fileName, directoryHint: .checkFileSystem)
+            XCTAssertFalse(fileURL.hasDirectoryPath)
+
+            XCTAssertTrue($0.fileExists(atPath: dirName + "/" + fileName))
+            XCTAssertTrue($0.isReadableFile(atPath: dirName + "/" + fileName))
+            XCTAssertTrue($0.isWritableFile(atPath: dirName + "/" + fileName))
+
+            // SHGetFileInfoW is documented to be limited to MAX_PATH, but appears to support long paths anyways (or at least does for SHGFI_EXETYPE).
+            // https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shgetfileinfow
+            XCTAssertNoThrow(try Data().write(to: URL(fileURLWithPath: dirName + "/" + fileName + ".bat")))
+            XCTAssertTrue($0.isExecutableFile(atPath: dirName + "/" + fileName + ".bat"))
+            XCTAssertFalse($0.isExecutableFile(atPath: dirName + "/" + fileName))
+
+            XCTAssertNoThrow(try $0.attributesOfItem(atPath: dirName + "/" + fileName))
+            XCTAssertNoThrow(try $0.setAttributes([.modificationDate: Date()], ofItemAtPath: dirName + "/" + fileName))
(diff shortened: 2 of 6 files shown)


jakepetroules:
@swift-ci test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,40 @@ extension String {
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
         return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString(encodedAs: UTF16.self) { pwszPath in
+            // lt MAX_PATH not lte MAX_PATH because MAX_PATH includes the null terminator
```
compnerd, line 43: This comment is not very clear to me.

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -23,7 +23,12 @@ import Darwin
 import WinSDK
 
 extension String {
-    package func withNTPathRepresentation<Result>(_ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
+    /// Invokes `body` with a resolved and potentially `\\?\`-prefixed version of the pointee,
+    /// to ensure long paths greater than 260 characters are handled correctly.
+    ///
+    /// - parameter preserveRelativePaths: Returns the original path without transforming through GetFullPathNameW + PathCchCanonicalizeEx, if the path is shorter than MAX_PATH.
+    /// - seealso: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+    package func withNTPathRepresentation<Result>(preserveRelativePaths: Bool = false, _ body: (UnsafePointer<WCHAR>) throws -> Result) throws -> Result {
```
compnerd, line 31: I don't think that `preserveRelativePaths` is the right term here - this isn't just preserving a relative path, its allowing for un-extended paths. `permitWin32Paths` perhaps? (The 261 limitation is Win32, not NT, `\\?\` bypaasses the Win32 layer and passes the path right to the NT subsystem).

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,40 @@ extension String {
...
+            }
+
             // 1. Normalize the path first.
+            // Contrary to the documentation, this works on long paths independently
+            // of the registry or process setting to enable long paths (but it will also
+            // not add the \\?\ prefix required by other functions under these conditions).
             let dwLength: DWORD = GetFullPathNameW(pwszPath, 0, nil, nil)
-            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-                guard GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil) > 0 else {
+            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { fullPath in
```
compnerd, line 53: The windows code path tends to use the windows style, so `pwszFullPath` seems better

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,40 @@ extension String {
...
+                guard GetFullPathNameW(pwszPath, DWORD(fullPath.count), fullPath.baseAddress, nil) > 0 else {
                     throw CocoaError.errorWithFilePath(self, win32: GetLastError(), reading: true)
                 }
 
-                // 2. Perform the operation on the normalized path.
-                return try body($0.baseAddress!)
+                // 1.5 Leave \\.\ prefixed paths alone since device paths are already an exact representation and PathCchCanonicalizeEx will mangle these.
+                if let base = fullPath.baseAddress, base[0] == UInt8(ascii: "\\"), base[1] == UInt8(ascii: "\\"), base[2] == UInt8(ascii: "."), base[3] == UInt8(ascii: "\\") {
```
compnerd, line 59: Can we use `._backslash` to check the character?

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,40 @@ extension String {
...
+                    return try body(base)
+                }
+
+                // 2. Canonicalize the path.
+                // This will add the \\?\ prefix if needed based on the path's length.
+                let capacity = Int16.max
+                return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: numericCast(capacity)) { outBuffer in
+                    let flags: ULONG = PATHCCH_ALLOW_LONG_PATHS
+                    let result = PathCchCanonicalizeEx(outBuffer.baseAddress, numericCast(capacity), fullPath.baseAddress, flags)
```
compnerd, line 68: Rather than the `switch` on the result, why not `guard` since we only consider one case?

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Removed, since this was refactored (see https://github.com/swiftlang/swift-foundation/pull/1257#discussion_r2052997473)

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Good point... I actually do want to preserve relative paths, because the CreateSymbolicLinkW will behave differently if lpTargetFileName is absolute vs relative. There's simply no way to specify an extended-length absolute path, since the extended length prefix requires fully qualifying a path. I changed the implementation to match the expectations.

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Done

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Done

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Done, I also changed this to call PathAllocCanonicalize which could theoretically be more efficient by allocating a smaller buffer.

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

jakepetroules:
@swift-ci test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,42 @@ extension String {
...
+
             // 1. Normalize the path first.
+            // Contrary to the documentation, this works on long paths independently
+            // of the registry or process setting to enable long paths (but it will also
+            // not add the \\?\ prefix required by other functions under these conditions).
             let dwLength: DWORD = GetFullPathNameW(pwszPath, 0, nil, nil)
-            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) {
-                guard GetFullPathNameW(pwszPath, DWORD($0.count), $0.baseAddress, nil) > 0 else {
+            return try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength)) { pwszFullPath in
+                guard GetFullPathNameW(pwszPath, DWORD(pwszFullPath.count), pwszFullPath.baseAddress, nil) == dwLength - 1 else {
```
jakepetroules, line 53: We now use `== dwLength - 1` rather than `> 0` since as pointed out by @lorentey in https://github.com/apple/swift-system/pull/226#discussion_r2052936771, the working directory could have changed between `GetFullPathNameW` invocations.

jakepetroules:
@swift-ci test

Sources/FoundationEssentials/String/String+Internals.swift
compnerd: I think that `relative:` is a better label then.

compnerd (approved):
Other than the labelling, LGTM

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1072,4 +1072,90 @@ final class FileManagerTests : XCTestCase {
...
+    /// Tests that Foundation can correctly handle "long paths" (paths of 260 to 32767 chacters long) on Windows.
+    func testWindowsLongPathSupport() throws {
+        #if !os(Windows)
+        throw XCTSkip("This test is not applicable for this platform")
+        #endif
+        // Create a directory with the absolute maximum path _component_ length of 255;
+        // this will guarantee the full playground path is well over 260 characters.
+        // Throw some Unicode in there for good measure, since only wide-character APIs support it.
+        let dirName = String(repeating: UUID().description, count: 7) + "你好！"
+        XCTAssertEqual(dirName.count, 255)
```
jmschonfeld, line 1085:
```suggestion
        let dirName = String(repeating: UUID().uuidString, count: 7) + "你好！"
        XCTAssertEqual(dirName.utf16.count, 255)
```

Two nits here:
1. Could we use `uuidString` instead of `description`? `uuidString`'s value is defined by an RFC, whereas (although unlikely) `description` is not meant to be parsed and could be changed in the future (and I also just find it clearer to read since at first I was unsure if the description was something like "UUID(bytes: ...)"
2. Is grapheme cluster count the right thing we want to test here? I assume we really want UTF-16 length since that's the "view" that the path APIs in the system will use, right?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1072,4 +1072,90 @@ final class FileManagerTests : XCTestCase {
...
+
+        try FileManagerPlayground {
+            Directory(dirName) {
+            }
+        }.test {
+            // Call every function that can call into withNTPathRepresentation with an overlong path and ensure it succeeds.
+            let fileName = UUID().description
```
jmschonfeld, line 1092:
```suggestion
            let fileName = UUID().uuidString
```

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1072,4 +1072,90 @@ final class FileManagerTests : XCTestCase {
             XCTAssertEqual($0.contents(atPath: "a\u{301}/test"), data)
         }
     }
+
+    /// Tests that Foundation can correctly handle "long paths" (paths of 260 to 32767 chacters long) on Windows.
+    func testWindowsLongPathSupport() throws {
+        #if !os(Windows)
+        throw XCTSkip("This test is not applicable for this platform")
+        #endif
```
jmschonfeld, line 1080: Could we change this to `#else` and put the rest of this function in the else block? The current spelling produces this warning on non-Windows platforms which would be nice to avoid:

```
/build/swift-foundation/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift:1084:9: warning: code after 'throw' will never be executed
```

jmschonfeld (commented):
A few minor comments on the unit test, but overall this looks good - thanks for putting this together! Added @jrflat as a reviewer as well just for another perspective from the `URL` side of things in case there's any considerations/impact there.

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Done

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jakepetroules: Good points, done for both.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jakepetroules: Done

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jakepetroules: Done!

jakepetroules:
@swift-ci test

jakepetroules:
Thanks @jmschonfeld and @compnerd; I've addressed all your feedback so far.

compnerd:
@jakepetroules any chance that we might be able to pull this into 6.2 as well? 😄

jrflat (approved):
Looks all good for `URL`'s use-case, thanks!

jakepetroules:
> @jakepetroules any chance that we might be able to pull this into 6.2 as well? 😄

Happy to -- @jmschonfeld / @parkera, is your team doing regular merges from main to 6.2 or should I manually bring it over?

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -35,15 +40,42 @@ extension String {
         // leading slash indicates a rooted path on the drive for the current
         // working directory.
         return try Substring(self.utf8.dropFirst(bLeadingSlash ? 1 : 0)).withCString(encodedAs: UTF16.self) { pwszPath in
+            if relative && PathIsRelativeW(pwszPath) {
+                return try body(pwszPath)
+            }
+
             // 1. Normalize the path first.
+            // Contrary to the documentation, this works on long paths independently
+            // of the registry or process setting to enable long paths (but it will also
```
jakepetroules, line 49: Ah, it states the truth on [this page](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats) instead of the docs for GetFullPathName itself:

> You can pass paths of more than MAX_PATH characters to [GetFullPathName](https://learn.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-getfullpathnamea) without \\?\. It supports arbitrary length paths up to the maximum string size that Windows can handle.

jmschonfeld (approved):
LGTM! Thanks!

jmschonfeld:
> > @jakepetroules any chance that we might be able to pull this into 6.2 as well? 😄
> 
> Happy to -- @jmschonfeld / @parkera, is your team doing regular merges from main to 6.2 or should I manually bring it over?

@parkera / @itingliu can confirm but I believe our plan is to have periodic merges from `main` into `release/6.2` for now until later in the release cycle

PR #1258, opened 2025-04-21
Revert strict IPv6 validation in URL

jrflat:
Previously, `URL` and the `URLParser` mistakenly allowed invalid characters in an IP-literal (within the square brackets) if the zone-ID was percent-encoded. This was revealed by testing the `NSURL` and `CFURL` re-core in Swift. My original goal was to make the new implementation match the slightly more strict validation of the old NS/CFURL implementations for compatibility.

However, I misunderstood which characters `CFURL` was actually allowing when validating an IP-literal. Instead of validating hex digits, colons, and periods before a potential zone ID, it was allowing any valid ASCII character anywhere in the host, so malformed URLs like `http://[www.apple.com]` were allowed.

While I think better IPv6 validation would ultimately be good, it could impact bincompat for `URL`, `URLComponents`, and `NSURL` (using the Swift parser) in rare cases when a client passes an invalid IPv6 string. This PR keeps the proper compatibility fix while reverting the stricter IPv6 validation code, which we could revisit at a later time.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLParser.swift
+++ b/Sources/FoundationEssentials/URL/URLParser.swift
@@ -435,41 +435,6 @@ internal struct RFC3986Parser {
         return validate(string: password, component: .password, percentEncodingAllowed: percentEncodingAllowed)
     }
 
-    private static func isIPvFuture(_ innerHost: some StringProtocol) -> Bool {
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1160, opened 2025-02-05
Base64 faster decode

fabianfett:
This PR is here to show further plans and to give a first glimpse into perf numbers that we will see. 
It should land after #1157 has landed, which will make the diff significantly smaller.

PR #1263, opened 2025-04-22
Reapply #1160, Add fix to allow more padding characters than necessary

fabianfett:


itingliu:
@swift-ci please test

parkera (approved):

parkera:
@swift-ci test

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -33,13 +43,10 @@ extension Data {
             return nil
         }
```
itingliu, line 44: Previously in #1160, you removed these `#if FOUNDATION_FRAMEWORK` path so we call into a unified implementation. Is it intentional that you aren't doing that here?

itingliu:
@swift-ci please test

PR #1262, opened 2025-04-22
Update `CMakeLists.txt` with missing `URLTemplate_Expression.swift`

bnbarham:
https://github.com/swiftlang/swift-foundation/pull/1198 added most of the new files, but missed `URLTemplate_Expression.swift`, add it.

PR #1264, opened 2025-04-22
Add swiftRegexBuilder dependency on Windows

Steelskin:
Following #1198 and #1262, this is needed to build on Windows.

### DIFF
--- a/Sources/FoundationInternationalization/CMakeLists.txt
+++ b/Sources/FoundationInternationalization/CMakeLists.txt
@@ -37,6 +37,7 @@ target_compile_options(FoundationInternationalization PRIVATE ${_SwiftFoundation
 target_compile_options(FoundationInternationalization PRIVATE -package-name "SwiftFoundation")
 
 target_link_libraries(FoundationInternationalization PUBLIC
+    $<$<PLATFORM_ID:Windows>:swiftRegexBuilder>
     FoundationEssentials
     _FoundationCShims
     _FoundationICU)


Sources/FoundationInternationalization/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ target_compile_options(FoundationInternationalization PRIVATE ${_SwiftFoundation
 target_compile_options(FoundationInternationalization PRIVATE -package-name "SwiftFoundation")
 
 target_link_libraries(FoundationInternationalization PUBLIC
+    $<$<PLATFORM_ID:Windows>:swiftRegexBuilder>
```
compnerd, line 40: I wonder if this should be universal.

jmschonfeld: I'm not sure if I fully understand why this is required. We don't actually have any direct dependencies on any other swift targets so this one shouldn't be any different - we just get them all from the toolchain, we don't actually depend on cmake targets for any of the stdlib modules. If this is related to the current CI failure where `plutil` fails to link, I have a different fix in-flight that I'm about to push that doesn't require adding this dependency

jmschonfeld: Is there a build failure somewhere that you saw that prompted the need to add this?

jmschonfeld: The `plutil` linking failure related to an implicitly imported symbol from RegexBuilder is being resolved at https://github.com/swiftlang/swift-corelibs-foundation/pull/5204 if that's the failure that prompted this

Steelskin: The link failure happens on Windows when building `FoundationInternationalization`:
```
ICUDateFormatter.swift.obj : error LNK2019: unresolved external symbol __imp_$sSy12RegexBuilderSs11SubSequenceRtzrlE5split9separator9maxSplits25omittingEmptySubsequencesSaySsGSS_SiSbtF referenced in function $s20FoundationEssentials4DateV0A20InternationalizationE11FormatStyleV0C15FieldCollectionV14updateScheduleAD16ICUDateFormatterC0cE4InfoV06UpdateJ0Ovg
ICUDateFormatter.swift.obj : error LNK2019: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc referenced in function $sS2S17_StringProcessing14RegexComponent0C7BuilderWl
Duration+UnitsFormatStyle.swift.obj : error LNK2001: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc
TimeZone_ICU.swift.obj : error LNK2001: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc
bin\FoundationInternationalization.dll : fatal error LNK1120: 2 unresolved externals
```

These changes might not be right right thing to do but they do fix the issue.

Steelskin: Following #1265, there is still one missing symbol:
```
ICUDateFormatter.swift.obj : error LNK2019: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc referenced in function $sS2S17_StringProcessing14RegexComponent0C7BuilderWl
Duration+UnitsFormatStyle.swift.obj : error LNK2001: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc
TimeZone_ICU.swift.obj : error LNK2001: unresolved external symbol __imp_$sSS17_StringProcessing14RegexComponent0C7BuilderMc
bin\FoundationInternationalization.dll : fatal error LNK1120: 1 unresolved externals
```

jmschonfeld: Are there any other linking errors / do you have a full build log I can help look at? It looks like those are just type metadata symbols and I'm not certain why the binary would still be referencing the conformance symbols when it no longer calls the function that requires the conformance (unless there's still another call somewhere in there)

jmschonfeld: Just merged https://github.com/swiftlang/swift-foundation/pull/1268 as well, which should resolve this problem for now. We'll continue to investigate how to land the original change without requiring client modules to link the `RegexBuilder` library

Steelskin: Here is a build on swift CI that hits the issue:
https://ci-external.swift.org/job/swift-corelibs-libdispatch-PR-windows/132/console

Steelskin: > Just merged #1268 as well, which should resolve this problem for now. We'll continue to investigate how to land the original change without requiring client modules to link the `RegexBuilder` library

Thank you, there are other issues on swift-ci that prevent checking for the fix but I can confirm that the issue is fixed with our CI with these changes.

PR #1265, opened 2025-04-22
[FoundationInternationalization] Avoid implicitly linking RegexBuilder symbol

jmschonfeld:
This is a similar fix to https://github.com/swiftlang/swift-corelibs-foundation/pull/5204 but for `FoundationInternationalization` instead of `plutil`. Specifying `Character` explicitly here avoids attempting to link a RegexBuilder symbol instead of the symbol from the standard library.


jmschonfeld:
@swift-ci please test

PR #1267, opened 2025-04-22
Revert url template implementation

itingliu:
Importing RegexBuilder in FoundationEssentials has the side effect of exposing it to all Foundation dependencies. Therefore clients started to pick up `String` overloads defined in RegexBuilder. 

Ideally we would want to adopt `MemberImportVisibility` in Foundation, but until we're confident that wouldn't cause other side effects we need to revert this change.


itingliu:
@swift-ci please test

itingliu:
Will go with https://github.com/swiftlang/swift-foundation/pull/1268

PR #1273, opened 2025-04-25
Move off of RegexBuilder for URL.Template

### LINKED ISSUE #1269: Make URI template's regex parsing available in swift-foundation
We temporarily guarded regex parsing implementation using FOUNDATION_FRAMEWORK in #1268 so that this would not affect downstream projects of FoundationEssentials. 

We'll want to enable that for non-framework builds.

danieleggert:
This fixes #1269

This uses `try! Regex("…")` instead of `RegexBuilder` to create regular expressions used by the URL Template implementation.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLTemplate.swift
+++ b/Sources/FoundationEssentials/URL/URLTemplate.swift
@@ -130,7 +130,7 @@ extension URL {
 }
 
 // MARK: - Parse
-#if FOUNDATION_FRAMEWORK
+
 extension URL.Template {
     /// Creates a new template from its text form.
     ///
@@ -164,7 +164,6 @@ extension URL.Template {
         }
     }
 }
-#endif
 
 // MARK: -
 
--- a/Sources/FoundationEssentials/URL/URLTemplate_Expression.swift
+++ b/Sources/FoundationEssentials/URL/URLTemplate_Expression.swift
@@ -10,7 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-internal import RegexBuilder
 #if canImport(CollectionsInternal)
 internal import CollectionsInternal
 #elseif canImport(OrderedCollections)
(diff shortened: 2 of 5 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/URL/URLTemplate_Expression.swift
```diff
@@ -152,61 +147,26 @@ extension URL.Template {
...
 
         private init() {
-            self.operatorRegex = Regex {
-                Optionally {
-                    Capture {
-                        One(.anyOf("+#./;?&"))
-                    }
-                }
-            }
+            self.operatorRegex = try! Regex(#"([\+#.\/;\?&])?"#)
```
itingliu, line 154: Do you know if it makes a difference whether you create a Regex this way or [the other one that specifies the output type](https://developer.apple.com/documentation/swift/regex/init(_:as:)-5z5nu)?

```swift
init(
    _ pattern: String,
    as outputType: Output.Type = Output.self
) throws
```
For example would it help the performance since we can bypass parsing the result from `AnyOutputRegex`?

itingliu (approved):

danieleggert:
@jrflat Did I merge the changes correctly from your 876aa1fb1aa7babb0bbad9e145095a140460e1f2 ?

Sources/FoundationEssentials/URL/URLTemplate.swift
```diff
@@ -176,6 +177,7 @@ extension URL.Template: CustomStringConvertible {
     }
 }
 
+@available(FoundationPreview 6.2, *)
```
jrflat, line 180: I don't think this annotation is needed since `URL.Template.Element` is an internal type, but otherwise LGTM!

jrflat (approved):

Sources/FoundationEssentials/URL/URLTemplate_Expression.swift
itingliu: I'm ok with this as-is for now. We can always come back if we find this useful later.

itingliu:
@swift-ci please test

itingliu (approved):

PR #1274, opened 2025-04-25
Implementation of #bundle

Tantalum73:
This PR implements `#bundle`, as pitched [here](https://forums.swift.org/t/pitch-introduce-bundle-macro/79288) and proposed [here](https://github.com/swiftlang/swift-foundation/pull/1251/)


Sources/FoundationMacros/BundleMacro.swift
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftSyntax
+import SwiftSyntaxMacros
+import Foundation
```
jmschonfeld, line 15:
```suggestion

```

The macro itself doesn't use Foundation so we should avoid importing it here

Sources/FoundationMacros/BundleMacro.swift
```diff
@@ -0,0 +1,42 @@
...
+
+#if FOUNDATION_FRAMEWORK
+/// Returns the bundle most likely to contain resources for the calling code.
+///
+/// Code in an app, app extension, framework, etc. will return the bundle associated with that target.
+/// Code in a Swift Package target will return the resource bundle associated with that target.
+@available(macOS 10.0, iOS 2.0, tvOS 9.0, watchOS 2.0, *)
+@freestanding(expression)
+public macro bundle() -> Bundle = #externalMacro(module: "FoundationMacros", type: "BundleMacro")
```
jmschonfeld, line 24: This declaration will need to go in one of the source files of the framework itself, rather than the macro

jmschonfeld (approved):

Tantalum73:
@swift-ci test please

jmschonfeld:
@swift-ci please test

Sources/FoundationInternationalization/CurrentBundle.swift
```diff
@@ -0,0 +1,22 @@
...
+//===----------------------------------------------------------------------===//
+
+#if FOUNDATION_FRAMEWORK
+/// Returns the bundle most likely to contain resources for the calling code.
+///
+/// Code in an app, app extension, framework, etc. will return the bundle associated with that target.
+/// Code in a Swift Package target will return the resource bundle associated with that target.
+@available(macOS 10.0, iOS 2.0, tvOS 9.0, watchOS 2.0, *)
+@freestanding(expression)
+public macro bundle() -> Bundle = #externalMacro(module: "FoundationMacros", type: "BundleMacro")
```
matthewseaman, line 20: @Tantalum73 @jmschonfeld should this go somewhere other than `FoundationInternationalization` given that bundles are not just for internationalization? For example someone could easily use this to lookup an asset or some other resource.

jmschonfeld: Right now, `Bundle` does not exist in `FoundationInternationalization` and this is just a `FOUNDATION_FRAMEWORK` piece that will be picked up by the `Foundation.framework` build. For this to be fully usable on non-Darwin, this `macro` would currently need to be declared in the `Foundation` module in the swift-corelibs-foundation repo. Eventually, if we sink `Bundle` down to swift-foundation this would remove the `FOUNDATION_FRAMEWORK` ifdef and would go alongside wherever `Bundle` is defined, but given that `Bundle` does not currently exist in swift-foundation I'm ok leaving this here for now as just a piece for the Foundation.framework build to pickup and moving it whenever we decide where `Bundle` will go later

itingliu: I had the same question as @matthewseaman when I read this. IMO `Bundle` will very likely be in FoundationEssentials if/when we move it, so we'll likely move this there, though like Jeremy said it doesn't really matter now for swift-foundation when building as a package

matthewseaman (approved):

matthewseaman:
FYI here is the SwiftPM implementation of the compile-time conditions: https://github.com/swiftlang/swift-package-manager/pull/8556

jmschonfeld:
@swift-ci please test

Sources/FoundationInternationalization/CurrentBundle.swift
Tantalum73: Agreed. Removed the declaration and now only the implementation in FoundationMacros remains.

itingliu (approved):

PR #1277, opened 2025-05-01
Switch to a safer technique for obtaining the working directory on Windows

jakepetroules:
Instead of looping 8 times to work around the TOCTOU issue with sizing the current directory buffer, instead keep doubling the buffer up until the 32767 character limit until the result fits. This ensures we always get a working directory if GetWorkingDirectoryW didn't return some other error, rather than returning nil in the case of a race condition.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -489,22 +489,13 @@ extension _FileManagerImpl {
     
     var currentDirectoryPath: String? {
 #if os(Windows)
-        var dwLength: DWORD = GetCurrentDirectoryW(0, nil)
-        guard dwLength > 0 else { return nil }
-
-        for _ in 0 ... 8 {
-            if let szCurrentDirectory = withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(dwLength), {
-                let dwResult: DWORD = GetCurrentDirectoryW(dwLength, $0.baseAddress)
-                if dwResult == dwLength - 1 {
-                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
-                }
-                dwLength = dwResult
-                return nil
-            }) {
-                return szCurrentDirectory
-            }
+        // Make an initial call to GetCurrentDirectoryW to get a buffer size estimate.
+        // This is solely to minimize the number of allocations and number of bytes allocated versus starting with a hardcoded value like MAX_PATH.
+        // We should NOT early-return if this returns 0, in order to avoid TOCTOU issues.
+        let dwSize = GetCurrentDirectoryW(0, nil)
+        return try? FillNullTerminatedWideStringBuffer(initialSize: dwSize >= 0 ? dwSize : DWORD(MAX_PATH), maxSize: DWORD(Int16.max)) {
+            GetCurrentDirectoryW(DWORD($0.count), $0.baseAddress)
         }
-        return nil
 #else
         withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
             guard getcwd(buffer.baseAddress!, FileManager.MAX_PATH_SIZE) != nil else {
--- a/Sources/FoundationEssentials/WinSDK+Extensions.swift
+++ b/Sources/FoundationEssentials/WinSDK+Extensions.swift
@@ -81,6 +81,10 @@ package var ERROR_FILENAME_EXCED_RANGE: DWORD {
     DWORD(WinSDK.ERROR_FILENAME_EXCED_RANGE)
 }
 
+package var ERROR_INSUFFICIENT_BUFFER: DWORD {
+    DWORD(WinSDK.ERROR_INSUFFICIENT_BUFFER)
+}
+
 package var ERROR_INVALID_ACCESS: DWORD {
     DWORD(WinSDK.ERROR_INVALID_ACCESS)
 }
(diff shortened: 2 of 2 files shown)


jakepetroules:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -489,22 +489,10 @@ extension _FileManagerImpl {
...
-                let dwResult: DWORD = GetCurrentDirectoryW(dwLength, $0.baseAddress)
-                if dwResult == dwLength - 1 {
-                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
-                }
-                dwLength = dwResult
-                return nil
-            }) {
-                return szCurrentDirectory
-            }
+        let initialSize = GetCurrentDirectoryW(0, nil)
```
compnerd, line 492: I think that we should guard that the returned value is `>= 0`. Also, most of the windows code that works with the system apis follows the windows recommendations (e.g. would name this `dwSize`).

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -489,22 +489,10 @@ extension _FileManagerImpl {
...
+        return try? Win32FillNullTerminatedStringBuffer(initialSize: initialSize != 0 ? initialSize : DWORD(MAX_PATH), maxSize: DWORD(Int16.max)) {
```
compnerd, line 493: I believe that 5000 should be a reasonable upper bound. NT Paths are limited to 4K I believe (page size), and you need the extra 4 characters for the NT path sigil (`\\?\`).

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -288,4 +288,29 @@ internal func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
     return DWORD(hr)
 }
 
+internal func Win32FillNullTerminatedStringBuffer(initialSize: DWORD, maxSize: DWORD, _ attempt: (UnsafeMutableBufferPointer<WCHAR>) throws -> DWORD) throws -> String {
```
compnerd, line 291: I think that we should include `Wide` in the name. I would also be okay with a `W` suffix. I'm not sure that the Win32 prefix matches any of the other helpers that we have though.

I think that `body` is better than `attempt` as a parameter name.

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -288,4 +288,29 @@ internal func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
...
+    var bufferCount = max(1, min(initialSize, maxSize))
+    while bufferCount < maxSize {
+        if let result = try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(bufferCount), { buffer in
+            let count = try attempt(buffer)
+            switch count {
+            case 0:
+                throw Win32Error(GetLastError())
+            case 1..<DWORD(buffer.count):
+                guard let result = String.decodeCString(buffer.baseAddress!, as: UTF16.self)?.result else {
+                    throw Win32Error(DWORD(ERROR_ILLEGAL_CHARACTER))
```
compnerd, line 301: Can we put `ERROR_ILLEGAL_CHARACTER` into WinSDK+Extensions as an overload please?

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -288,4 +288,29 @@ internal func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
...
+                return result
+            default:
+                bufferCount *= 2
+                return nil
+            }
+        }) {
+            return result
+        }
+    }
+    throw Win32Error(DWORD(ERROR_INSUFFICIENT_BUFFER))
```
compnerd, line 313: Likewise for `ERROR_INSUFFICIENT_BUFFER`.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jakepetroules: NT paths aren't limited to 4k. Long paths are [documented as being a maximum of 32,767 UTF-16 code units](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) and it's completely possible to create a path of such a length. Here's my current directory in PowerShell:

```ps1
PS> (Get-Location).Path.Length
32762
```

32762 plus 4 for `\\?\` plus 1 null terminator = 32767.

I'd rather support up to the absolute maximum the OS allows.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jakepetroules: Done.

Sources/FoundationEssentials/WinSDK+Extensions.swift
jakepetroules: Done

Sources/FoundationEssentials/WinSDK+Extensions.swift
jakepetroules: Done

Sources/FoundationEssentials/WinSDK+Extensions.swift
jakepetroules: Done

jakepetroules:
@swift-ci test

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -288,4 +296,29 @@ internal func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
     return DWORD(hr)
 }
 
+internal func FillNullTerminatedWideStringBuffer(initialSize: DWORD, maxSize: DWORD, _ body: (UnsafeMutableBufferPointer<WCHAR>) throws -> DWORD) throws -> String {
+    var bufferCount = max(1, min(initialSize, maxSize))
+    while bufferCount < maxSize {
```
jmschonfeld, line 301: Should this be `bufferCount <= maxSize` so that we actually try max size? (I might be getting confused here as to whether this size includes the null byte or not - perhaps some documentation on this function might help clarify that)

Sources/FoundationEssentials/WinSDK+Extensions.swift
```diff
@@ -288,4 +296,29 @@ internal func WIN32_FROM_HRESULT(_ hr: HRESULT) -> DWORD {
...
+        if let result = try withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(bufferCount), { buffer in
+            let count = try body(buffer)
+            switch count {
+            case 0:
+                throw Win32Error(GetLastError())
+            case 1..<DWORD(buffer.count):
+                guard let result = String.decodeCString(buffer.baseAddress!, as: UTF16.self)?.result else {
```
jmschonfeld, line 308: Since `repairingInvalidCodeUnits` defaults to `true` here, this will never return `nil`. If the API contract is that this will always return valid UTF-16 upon success, should we just use `let result = String(decodingCString: buffer.baseAddress!, as: UTF16.self)` and remove the `guard`? That's what we do elsewhere in Foundation for APIs like this. If you do want to check for invalid UTF-16, then you'll need to pass `repairingInvalidCodeUnits: false` to instruct the function to return `nil` upon failure

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -489,22 +489,10 @@ extension _FileManagerImpl {
...
-                if dwResult == dwLength - 1 {
-                    return String(decodingCString: $0.baseAddress!, as: UTF16.self)
-                }
-                dwLength = dwResult
-                return nil
-            }) {
-                return szCurrentDirectory
-            }
+        let dwSize = GetCurrentDirectoryW(0, nil)
+        return try? FillNullTerminatedWideStringBuffer(initialSize: dwSize >= 0 ? dwSize : DWORD(MAX_PATH), maxSize: DWORD(Int16.max)) {
```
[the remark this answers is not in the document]
jakepetroules: No, that would be a TOCTOU error. Remember that GetCurrentDirectoryW operates on global state, and the documentation doesn't state _why_ it fails.

This initial call is just an optimization to minimize the total number of bytes allocated and the overall number of allocations. We could just as well omit this and start with 1 or MAX_PATH or something else. I'll add some comments to clarify.

Sources/FoundationEssentials/WinSDK+Extensions.swift
jakepetroules: The buffer includes the null byte. Yes, I guess this should be <=, good catch.

jakepetroules: Fixed, added comments

Sources/FoundationEssentials/WinSDK+Extensions.swift
jakepetroules: Opted for the path consistent with the rest of Foundation

jakepetroules:
@swift-ci test

jmschonfeld (approved):

compnerd (approved):

PR #1278, opened 2025-05-01
Restore selective nil path behavior for bridged NSURLs

jrflat:
Fixes a compatibility issue that occurs when a `URL` such as `scheme://?query` is bridged to an `NSURL`. `NSURL` originally expected `.path` to return `nil` for this string, but `URL` and RFC 3986 say that the path *always* exists and is the empty string in this case. `URL` would then store an empty path range instead when bridging to `NSURL`, causing this discrepancy.

This PR fixes the issue during bridging by only storing the path range if `NSURL` would think there's a path, i.e. in the case where 1) the path is not empty, or 2) the path is empty, but we have a non-empty authority component.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1279, opened 2025-05-02
Fix URL.Template availability annotations

jrflat:
All public `URL.Template` extensions need to be marked with the `FoundationPreview 6.2` availability since extensions are separate declarations and availability is per declaration. Extensions don't automatically inherit the availability from the type they're extending.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1280, opened 2025-05-02
Enable MemberImportVisibility upcoming feature in Foundation

parkera:
Enable the MemberImportVisibility compiler flag, which ensures that we don't accidentally depend on implicit imports in our sources.

PR #1281, opened 2025-05-03
Fix @preconcurrency import Android in CocoaError+FilePath.swift

marcprux:
This fixes the Android [build error](https://github.com/swift-everywhere/swift-package-builds/actions/runs/14801693208/job/41561703634#step:32:2259) that I noted at https://github.com/swiftlang/swift-foundation/pull/1280#issuecomment-2848066387:

```
/home/runner/work/swift-package-builds/swift-package-builds/swift-foundation/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift:76:37: error: initializer 'init(rawValue:)' is not available due to missing import of defining module 'Android'
 18 | #elseif canImport(Bionic)
 19 | @preconcurrency import Bionic
 20 | #elseif canImport(Glibc)
    | `- note: add import of module 'Android'
 21 | @preconcurrency import Glibc
 22 | #elseif canImport(Musl)
    :
 74 |         // (130280235) POSIXError.Code does not have a case for EOPNOTSUPP
 75 |         guard errno != EOPNOTSUPP else { return nil }
 76 |         guard let code = POSIXError.Code(rawValue: errno) else {
    |                                     `- error: initializer 'init(rawValue:)' is not available due to missing import of defining module 'Android'
 77 |             fatalError("Invalid posix errno \(errno)")
 78 |         }
```

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -15,8 +15,8 @@ internal import _ForSwiftFoundation
 
 #if canImport(Darwin)
 import Darwin
-#elseif canImport(Bionic)
-@preconcurrency import Bionic
+#elseif canImport(Android)
+@preconcurrency import Android
 #elseif canImport(Glibc)
 @preconcurrency import Glibc
 #elseif canImport(Musl)


finagolfin:
Makes sense, as [I've noted weird import issues with these platform overlays before](https://github.com/apple/swift-certificates/pull/183#discussion_r1730366251) and turning on the MemberImportVisibility feature has required changing the imports like this before, as that feature closes up a well-known Swift import leak.

finagolfin (approved):

marcprux:
Do you think we should just change the remaining 14 instances of `import Bionic` to `import Android`? Or best to leave them alone until they cause problems?

finagolfin:
Ideally, we'd audit all 14 and see what file invokes which APIs and check if they're currently correct, but I'm not too happy with the `Bionic` module we have now, as it only covers a fraction of the Bionic headers.

The current overlays are [just what we could get to work, given the opacity of what's included](https://github.com/swiftlang/swift/pull/72161) and the various import leaks. It will all need to be properly organized and separated out in time.

Given all that, I don't have a good answer: leave it alone and we'll figure it out later?

finagolfin:
@swift-ci test

parkera:
Thanks for taking care of this @marcprux @finagolfin

PR #1284, opened 2025-05-08
fix `Calendar.RecurrenceRule`

### LINKED ISSUE #1283: RecurrenceRule produces incorrect sequences for start dates with non-zero nanosecond components
[SF-0009](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0009-calendar-recurrence-rule.md) states that the first element in a sequence of `Date` occurrences computed from a `RecurrenceRule` is always the start date (see [SF-0009#Usage](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0009-calendar-recurrence-rule.md#usage)):

> A recurrence rule of a given frequency repeats the start date with the interval of that frequency. For example, assume that now it is February 09 2024, 13:43. Creating a daily recurrence would yield a result for each following date at the same time:
> ```swift
> var recurrence = Calendar.RecurrenceRule(calendar: .current, frequency: .daily)
> for date in recurrence.recurrences(of: .now) {
>     // 2024-02-09, 13:43
>     // 2024-02-10, 13:43
>     // 2024-02-11, 13:43
>     // 2024-02-12, 13:43
>     // ...
> }
> ```

However, this currently is only the case for start dates with a 0 nanosecond component. Any other start date (e.g. also `Date.now`, unless you happen to be creating the date exactly at the start of a second, which admittedly is rather unlikely), will result in `recurrences(of:)` returning a sequence that will skip the start date.

The reason for this is that `Calendar.DatesByRecurring.Iterator` internally uses `Calendar._dates(startingAfter:)`, which in turn uses `Calendar._unadjustedDates(after:)`, which only produces `Date`s with whole-integer seconds.

The tests currently don't catch this, since all RecurrenceRule tests only use dates with whole-integer components as input.

lukaskollmer:
This PR attempts to fix #1283, by adjisting `Calendar.DatesByRecurring.Iterator` to maintain nanosecond precision on the dates it produces, if the original start date had a non-zero nanosecond component.

This brings the behaviour of the `Calendar.RecurrenceRule` API in line with how it was defined and described in [the proposal](https://github.com/swiftlang/swift-foundation/blob/main/Proposals/0009-calendar-recurrence-rule.md).

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -93,6 +93,9 @@ extension Calendar {
             /// value is used as a lower bound for ``nextBaseRecurrenceDate()``.
             let rangeLowerBound: Date?
             
+            /// The start date's nanoseconds component
+            let startDateNanoseconds: TimeInterval
+            
             /// How many occurrences have been found so far
             var resultsFound = 0
             
@@ -231,6 +234,8 @@ extension Calendar {
                     case .yearly:   [.second, .minute, .hour, .day, .month, .isLeapMonth]
                 }
                 var componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
+                
+                startDateNanoseconds = start.timeIntervalSince1970.remainder(dividingBy: 1)
                 
                 let expansionChangesDay = dayOfYearAction == .expand || dayOfMonthAction == .expand || weekAction == .expand || weekdayAction == .expand
                 let expansionChangesMonth = dayOfYearAction == .expand || monthAction == .expand || weekAction == .expand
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -232,6 +235,8 @@ extension Calendar {
                 }
                 var componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
                 
+                startDateNanoseconds = start.timeIntervalSince1970.remainder(dividingBy: 1)
```
itingliu, line 238: Would it work if we modify `components` above to include nanoseconds so the information is readily available in `componentsForEnumerating`, so that we don't have to do this manual adjustment here?

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
lukaskollmer: @itingliu that was my first attempt, but it doesn't work.

in my understanding, the issue is that `Calendar.DatesByRecurring`'s underlying `baseRecurrence` (`Calendar.DatesByMatching`) will always return only whole-second-rounded dates (though this isn't documented, afaict), so we'd need to adjust that type's behaviour instead (which i decided against bc i wasn't sure whether this would affect any other dependent code, and i figured instead going with the simplest possible fix would be the best approach)

parkera (approved):
Fix makes sense to me. @hristost can you take a look as well?

PR #1286, opened 2025-05-09
Implement `String.Encoding.ianaName` and `String.Encoding(ianaName:)`.

YOCKOW:
Implementation of the proposal: https://github.com/swiftlang/swift-foundation/pull/1243


YOCKOW:
@swift-ci Please test

YOCKOW:
@swift-ci Please test

YOCKOW:
@itingliu 

I think this PR is almost ready, but I have a few questions:
- Is it okey that these APIs are implemented in `FoundationEssentials` (not in `FoundationInternationalization`)?
- Should we add a script that converts [original data](https://www.iana.org/assignments/character-sets/character-sets.xhtml) into Swift code?
  + Sample script: https://gist.github.com/YOCKOW/2af7a44937536aedea92f11d2c573ec7

YOCKOW:
@swift-ci Please test

YOCKOW:
@swift-ci Please test

YOCKOW:
@swift-ci Please test

YOCKOW:
@swift-ci Please test

YOCKOW:
@itingliu 

I've made this PR ready for review.
I'm happy if you take a look at this. Thank you.


cc: Foundation Workgroup
@adam-fowler @iCharlesHu @Lukasa @designatednerd @jmschonfeld @lorentey @stephentyrone @itingliu @tomerd @parkera

utils/update-iana-charset-names-impl.py
```diff
@@ -0,0 +1,174 @@
...
+## Copyright (c) 2025 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0 with Runtime Library Exception
+##
+## See https://swift.org/LICENSE.txt for license information
+## See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+##
+##===----------------------------------------------------------------------===##
+
+"""
+This is a python script that converts an XML file containing the list of IANA
```
itingliu: I must ask here since we're a Swift project 😄 
Is there a reason why this needs to be a python script? Can't we write this in Swift?

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,265 @@
...
+            return lhs.scalar.value + 0x20 == rhs.scalar.value
+        } else if ("a"..."z").contains(lhs.scalar) {
+            return lhs.scalar.value - 0x20 == rhs.scalar.value
+        }
+        return false
+    }
+}
+
+/// A type to tokenize string for `String.Encoding` names.
+internal protocol StringEncodingNameTokenizer: ~Copyable {
```
itingliu: Are you planning to introduce other conformer for this protocol? I wonder if omitting using this protocol, but use the concrete `ASCIICaseInsensitiveTokenizer` would give us performance gains.

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,265 @@
...
+    }
+}
+
+/// A type that holds a `Unicode.Scalar` where its value is compared case-insensitively with others'
+/// _if the value is within ASCII range_.
+private struct ASCIICaseInsensitiveUnicodeScalar: Equatable,
+                                                  ExpressibleByUnicodeScalarLiteral {
+    typealias UnicodeScalarLiteralType = Unicode.Scalar.UnicodeScalarLiteralType
+
+    let scalar: Unicode.Scalar
```
itingliu: Can we store and operate every string processing logic on `UInt8` instead of `Unicode.Scalar` (and `UTF8View` instead of `UnicodeScalarView` since we expect it to be an ascii character?

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,265 @@
...
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+
+// MARK: - Private extensions for parsing encoding names
+
+private extension Unicode.Scalar {
```
itingliu: This is a general comment for the functions in this file:

Can we store and operate the string processing logic on `UInt8` instead of `Unicode.Scalar` (and `UTF8View` instead of `UnicodeScalarView`) since we expect the names to be ascii?

We do so for quite a few places, such as parsing utilities in [this file](https://github.com/swiftlang/swift-foundation/blob/32fd38c918b8c9b19288435faf27b53302c22ce0/Sources/FoundationEssentials/Formatting/FormatParsingUtilities.swift#L28) IIRC it did give us a better performance than resorting to UnicodeScalarView

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,265 @@
...
+        self.preferredMIMEName = preferredMIMEName
+        self.name = name
+        self.aliases = aliases
+    }
+
+    func matches<T>(
+        _ string: String,
+        tokenizedBy tokenizer: T.Type
+    ) -> Bool where T: StringEncodingNameTokenizer, T: ~Copyable {
+        if let preferredMIMEName = self.preferredMIMEName,
```
itingliu, line 66: I'm not sure I understand why we need a tokenizer. This function looks like it is effectively the same as doing case-insensitive-compare on two strings. In the case of ascii strings, it is also effectively the same as calling `lowercased()` and compare. Is my understanding correct? Did you add a tokenizer to avoid extra allocation from calling `lowercased()`? Are there other things that I'm missing?

utils/update-iana-charset-names-impl.py
YOCKOW: That's just because there are not a few Python scripts in [swiftlang/swift/utils](https://github.com/swiftlang/swift/tree/main/utils).
Of course, we can rewrite it in Swift because we must love Swift more than Python.😁

Sources/FoundationEssentials/String/String+Encoding+Names.swift
YOCKOW: I _was_ certainly planning to be able to support other conformer such as [UTS #22's Charset Alias Matching](https://www.unicode.org/reports/tr22/tr22-8.html#Charset_Alias_Matching) which has been already [removed](https://github.com/swiftlang/swift-foundation/pull/1286/commits/7acaa40d0ed11bd088defe6cafbe47ac6126cb0a#diff-8a0e3e66529a49d5dba0c686f386180c965e8b61b289af5c602680be71cc6917L101-L103).
I should've removed (refactored) code more aggressively.

Sources/FoundationEssentials/String/String+Encoding+Names.swift
YOCKOW: I now come to think `UInt8` is more effective than `Unicode.Scalar`.
I'll retouch functions to use `UInt8`.

Sources/FoundationEssentials/String/String+Encoding+Names.swift
YOCKOW: There are/were two reasons which came to my mind:
1. To adopt "Once and Only Once" principle: This is not applied to current implementation because code for UTS#22 has been removed.
2. To avoid extra allocation from calling `lowercased()` (as you pointed out): I imagined that user input might be like "VeryVeryLongLongLongInput...".

YOCKOW:
@itingliu
Thank you for reviewing. I'm going to modify implementation reflecting your feedback.

YOCKOW:
@swift-ci Please test

parkera:
@swift-ci test

parkera (approved):
Looking good to me after the recent simplifications. Will also want @itingliu approval.

YOCKOW:
Thank you @parkera.
I'm going to wait for @itingliu.

Just in case, what's changed:
- Rewrote util script in Swift
- Simplified the logic to parse names
  * Use `UInt8`(`UTF8.UnitCode`)
  * Remove redundant tokenizers.

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,164 @@
...
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+
+// MARK: - Private extensions for parsing encoding names
+
+private extension UTF8.CodeUnit {
+    func _isASCIICaseinsensitivelyEqual(to other: UTF8.CodeUnit) -> Bool {
```
itingliu: Nit: 
```suggestion
    func _isASCIICaseInsensitivelyEqual(to other: UTF8.CodeUnit) -> Bool {
```

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    /// Creates an instance from the name of the IANA registry "charset".
+    ///
+    /// - Note: The given name is compared to each IANA "charset" name
+    ///         with ASCII case-insensitive collation
+    ///         to determine which encoding is suitable.
+    @available(FoundationPreview 6.3, *)
+    public init?(ianaName charsetName: String) {
+        let possibilities: [String.Encoding] = [
```
itingliu, line 124: Would it work if you write

```
let possibilities: [IANACharset] = {...
```

and below just do 
```
         func __determineEncoding() -> String.Encoding? {
            for ianaCharset in possibilities {
                if ianaCharset.matches(charsetName) {
                    return encoding
                }
            }
            return nil
        }
```

Am I missing anything?

Sources/FoundationEssentials/String/String+Encoding+Names.swift
```diff
@@ -0,0 +1,164 @@
...
+                    continue
+                }
+                if ianaCharset.matches(charsetName) {
+                    return encoding
+                }
+            }
+            return nil
+        }
+
+        guard let encoding = __determineEncoding() else {
```
itingliu: And to simply this even further, it looks like this logic inside `func __determineEncoding()` doesn't even needed to be extracted in a function since it's only used once?

Sources/FoundationEssentials/String/String+Encoding+Names.swift
YOCKOW: Won't we lose `encoding` for `return encoding` in that implementation?

Sources/FoundationEssentials/String/String+Encoding+Names.swift
YOCKOW: ~~The reason why nested function is used is just because of readability since `self = nil; return` is invalid in Swift.~~

Sorry, we can write `self = encoding; return`...
I'm fixing it.

YOCKOW: I've fixed `String.Encoding(ianaName:)`.
I hope that this is in line with your intention.

YOCKOW:
@swift-ci Please test

YOCKOW:
What's changed since [8f84db7](https://github.com/swiftlang/swift-foundation/pull/1286/commits/8f84db7ae8702407842b4325666a77b34f8a796e):
- Fixed functions' spelling
- Eliminated an unnecessary nested function

YOCKOW:
I sincerely appreciate the reviews and approvals. Thank you so much.

PR #1289, opened 2025-05-09
Fix URL.Template benchmark variable initialization

jrflat:
The `URL.Template` benchmark code needs to be updated since we removed `ExpressibleByStringLiteral` from `URL.Template.VariableName`.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1287, opened 2025-05-09
AttributedString Equatable conformance performance improvements

jmschonfeld:
This commit includes a handful of performance improvements for `AttributedString` that provide some great speedups to its `Equatable` conformance and `Runs` view. These are the changes this makes:

`AttributedString` attribute raw value casting: for a little while we've known that comparing `AttributedString` attribute values is more expensive than we'd like. This fixes a few of those issues by:
- Avoiding unnecessarily boxing the existential attribute value in an `Optional` which requires looking up type metadata and allocating space for the optional. This `Optional` in `__equalAttributes` was required previously but is no longer used
- Avoid dynamic casts when casting the attribute values (as existential) to their expected types for fetching values and comparing values. We know that when comparing `_AttributeValue`s that the raw value types must be the same, and when fetching a value from an `AttributedString` we know that the raw values must be of the expected key type. Instead of using `as?` casts (which go through dynamic casting machinery to support bridging types) we can simply use `_identityCast` to perform a much faster cast that just assumes the raw value is the type provided. I added debug-only asserts to help us quickly catch Foundation bugs where this isn't the case

I also noticed that most of the time spent in `AttributedString`'s equatable conformance was in `RangeSet` operations. The way we were iterating the runs view was calling `RangeSet` APIs multiple times for the same information, and since `RangeSet` isn't fully frozen/inlinable, each call requires metadata allocation and extra costs. Instead, I refactored `rangeIdx(containing:)` to provide more info to the caller and provide a public assertion message so that the callers can avoid extra `RangeSet` calls that were already being made by this function.

Overall, these improvements made a 30% improvement in `AttributedString`'s Equatable conformance and a 45% improvement in iterating the runs view via an attribute slice:

```
----------------------------------------------------------------------------------------------------------------------------
enumerateAttributesSlice-swift metrics
----------------------------------------------------------------------------------------------------------------------------

╒══════════════════════════════════════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╤═══════════╕
│          Throughput (# / s) (K)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
╞══════════════════════════════════════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╪═══════════╡
│                   main                   │        82 │        81 │        81 │        80 │        79 │        77 │        77 │        81 │
[...]

PR #1290, opened 2025-05-14
Retrieving attributed string attribute values should not use identity casts

jmschonfeld:
This reverts a small portion of https://github.com/swiftlang/swift-foundation/pull/1287. It's not always possible to use an identity cast here, because in some cases (such as catalyst applications where `AttributedString`s interchangeably use both UIKit `UIColor` and AppKit `NSColor` attributes) we do rely on dynamic casting behavior. The dynamic cast should attempt an identity cast first so this shouldn't have a strong impact on the overall performance win since we still avoid the optional boxing in `__equalAttributes` and retain the performance wins from the runs view.


jmschonfeld:
@swift-ci please test

PR #1293, opened 2025-05-15
[Data] Fix build issue with Static SDK for Linux.

al45tair:
Musl's definition of `S_IFMT` gets imported by the Swift importer, which means Swift can see `S_IFMT` as a `CInt` and also `S_IFMT` from the Swift overlay as a `mode_t`.  The upshot is that attempting to convert it to `mode_t` produces an ambiguity error, as Swift doesn't know which definition to use.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -336,7 +336,7 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
     }
     
     let fileSize = min(Int(clamping: filestat.st_size), maxLength ?? Int.max)
-    let fileType = mode_t(filestat.st_mode) & mode_t(S_IFMT)
+    let fileType = mode_t(filestat.st_mode) & S_IFMT
 #if !NO_FILESYSTEM
     let shouldMap = shouldMapFileDescriptor(fd, path: inPath, options: options)
 #else


Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -336,7 +336,11 @@ internal func readBytesFromFile(path inPath: PathOrURL, reportProgress: Bool, ma
     }
     
     let fileSize = min(Int(clamping: filestat.st_size), maxLength ?? Int.max)
+#if !os(Windows)
```
parkera, line 339: Is there any way at all to structure this so that the `S_IFMT` is cast to a `mode_t` on all platforms, without an ambiguity warning or error?

This kind of ifdef is so unfortunate, because it complicates the code and produces a fork to maintain but has no difference in behavior at all.

jmschonfeld: This code is already in the `#else` block of an `#if os(Windows)` statement above, so this code is already `#if !os(Windows)`. If we need to make this change for all non-windows platforms, I think we can just outright change the existing line since the existing line is already conditionalized

al45tair: The problem is that with Musl, the Clang importer successfully imports the underlying `S_IFMT` (as a `CInt`) and also sees the `S_IFMT` (a `mode_t`) from the overlay. Doing `mode_t(S_IFMT)` is then ambiguous because it doesn't know whether you meant the `CInt` or the `mode_t` constant.

I had thought that the explicit conversion was there because on Windows, the `S_IFMT` constant in the overlay is _not_ a `mode_t`, but you're right, that entire block of code is not used on Windows, so we can just remove the conversion. I'll update the PR.

al45tair:
@swift-ci Please test

parkera (approved):

parkera:
Perfect, thank you.

Issue #1016, opened 2024-10-28
Discrepancy: `String.data(using: .japaneseEUC)` returns `nil` on non-Darwin.


YOCKOW:
Given the following code:

```swift
import Foundation
print("Hello, world!".data(using: .japaneseEUC) as Any)
```

Results are:

| OS     | Swift | Result               |
|--------|-------|----------------------|
| macOS  | 5.10  | "Optional(13 bytes)" |
| Ubuntu | 5.10  | "nil"                |
| macOS  | 6.0   | "Optional(13 bytes)" |
| Ubuntu | 6.0   | "nil"                |

This is not regression but discrepancy which should be also fixed.

parkera:
Actually - @YOCKOW can you verify that?

YOCKOW:
While I [confirmed ISO-2022-JP was OK](https://github.com/swiftlang/swift-foundation/issues/1014#issuecomment-2878333590), EUC-JP seems still broken.
I'm going to look into this sometime.

YOCKOW:
**Follow-up note:**　This issue was truly resolved by #1989 because #1296 had missed adding its necessary files to 'CMakeLists.txt'. (That was my fault.🙇)

PR #1296, opened 2025-05-16
Enable string conversion in EUC-JP.

YOCKOW:
Background: EUC-JP is not supported by OSS CoreFoundation, while it is supported by macOS Foundation Framework.
See https://github.com/swiftlang/swift-foundation/issues/1016

This PR resolves the issue by calling ICU API if necessary.

### DIFF
--- a/Tests/FoundationInternationalizationTests/StringTests+Data.swift
+++ b/Tests/FoundationInternationalizationTests/StringTests+Data.swift
@@ -115,12 +115,27 @@ final class StringConverterTests: XCTestCase {
         )
 
         // Unsupported characters
-        let sushi = "Sushi🍣"
+        let onsen = "Onsen♨" // BMP emoji
+        let sushi = "Sushi🍣" // non-BMP emoji
+        XCTAssertNil(onsen.data(using: String._Encoding.japaneseEUC))
         XCTAssertNil(sushi.data(using: String._Encoding.japaneseEUC))
+        XCTAssertEqual(
+            onsen.data(using: String._Encoding.japaneseEUC, allowLossyConversion: true),
+            "Onsen?".data(using: .utf8)
+        )
+        #if FOUNDATION_FRAMEWORK
+        // NOTE: Foundation framework replaces an unsupported non-BMP character
+        //       with "??"(two question marks).
+        XCTAssertEqual(
+            sushi.data(using: String._Encoding.japaneseEUC, allowLossyConversion: true),
+            "Sushi??".data(using: .utf8)
+        )
+        #else
         XCTAssertEqual(
             sushi.data(using: String._Encoding.japaneseEUC, allowLossyConversion: true),
             "Sushi?".data(using: .utf8)
         )
+        #endif
     }
 }
 


YOCKOW:
@swift-ci Please test

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
```diff
@@ -255,8 +260,12 @@ extension String {
             // Other encodings, defer to the CoreFoundation implementation
             return _ns.data(using: encoding.rawValue, allowLossyConversion: allowLossyConversion)
 #else
-            // Attempt an up-call into swift-corelibs-foundation, which can defer to the CoreFoundation implementation
-            return _cfStringEncodingConvert(string: self, using: encoding.rawValue, allowLossyConversion: allowLossyConversion)
+            return (
+                // Attempt an up-call into swift-corelibs-foundation, which can defer to the CoreFoundation implementation
+                _cfStringEncodingConvert(string: self, using: encoding.rawValue, allowLossyConversion: allowLossyConversion) ??
+                // Or attempt an up-call into ICU via FoundationInternationalization
+                _icuStringEncodingConvert(string: self, using: encoding, allowLossyConversion: allowLossyConversion)
```
parkera, line 267: When do we fall back to the ICU one here? On Darwin this encoding is handled via CF, which itself calls into ICU.

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -0,0 +1,202 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#endif
+internal import _FoundationICU
+
+private extension String.Encoding {
+    var _icuConverterName: String? {
```
parkera, line 19: Should we make sure this returns `nil` for any encoding which we never want to use ICU for? `UTF8`, `ASCII`, `UTF16/32`, etc.

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
parkera: I guess that'll happen if `import Foundation` has not been done (and CF is present), but `import FoundationInternationalization` has. I wonder how we can further unify these paths; on Darwin too...

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -0,0 +1,202 @@
...
+        case .utf32: "UTF-32"
+        case .utf32BigEndian: "UTF-32BE"
+        case .utf32LittleEndian: "UTF-32LE"
+        default: nil
+        }
+    }
+}
+
+extension ICU {
+    final class StringConverter: @unchecked Sendable {
```
jmschonfeld, line 47: nit: does this need the `@unchecked`? I would have assumed that since it's a `final` class with all immutable, `Sendable` properties that the compiler can validate this conformance

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -0,0 +1,202 @@
...
+                bytesNoCopy: dest,
+                count: Int(actualLength),
+                deallocator: .custom({ pointer, _ in pointer.deallocate() })
+            )
+        }
+    }
+}
+
+extension ICU.StringConverter {
+    nonisolated(unsafe) static private var _converters: LockedState<[String.Encoding: ICU.StringConverter]> = .init(initialState: [:])
```
jmschonfeld, line 165: I think this should just be a `static private let` (the immutable-ness affects the `LockedState` itself but not the contents IIRC, and this isn't unsafe because it's guarded by a lock)

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -0,0 +1,202 @@
...
+
+
+@_dynamicReplacement(for: _icuMakeStringFromBytes(_:encoding:))
+func _icuMakeStringFromBytes_impl(_ bytes: UnsafeBufferPointer<UInt8>, encoding: String.Encoding) -> String? {
+    guard let converter = ICU.StringConverter.converter(for: encoding),
+          let pointer = bytes.baseAddress else {
+        return nil
+    }
+    let data =  Data(
+        bytesNoCopy: UnsafeMutableRawPointer(mutating: pointer),
```
jmschonfeld, line 189: Is it worth adding a comment here about this being "safe" because this data neither escapes this function nor is mutated? Since the pointer provided is not guaranteed to live longer than the function or be mutable, this could be a trap for a later addition that doesn't see those two constraints

Sources/FoundationEssentials/String/StringProtocol+Essentials.swift
jmschonfeld: I had a similar question - I believe `_cfStringEncodingConvert` will return `nil` if any of the following are the case:

1. CF isn't loaded
2. CF is loaded but it doesn't support the provided encoding
3. CF is loaded and supports the encoding but the data is malformed

We want the up-call in 2, and I think Tony's question was about 1, but do we want the up call in case 3?

YOCKOW: I now agree that unexpected fallback should be avoided.
Although ICU *can* support other encodings, I shoud've make this change focus on EUC-JP.

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: Certainly.
I'll let them return `nil` here or in `ICU.StringConverter`'s `init`.

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: Thank you.
That's a vestige of my first implementation where `_converter` was a bare pointer...

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: I agree it's better to have some comments. I'm going to add ones.

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: Indeed it should be.
This is also a vestige of my first impl, I guess.

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: Sorry.
`@unchecked` is necessary still because `LockedState` is not `Sendable`...

YOCKOW:
@swift-ci Please test

YOCKOW:
@parkera @jmschonfeld Thank you for reviewing.

In response to your feedback, I've made some changes:
- Made it clear that `ICU.StringConverter` doesn't support `UTF-*` and `US-ASCII`.
- Removed unnecessary `nonisolated(unsafe)` from the static property.
- Added comments in `func _icuMakeStringFromBytes_impl` to clarify the safety of passing the pointer.
- Let it delegate string conversion to ICU only if the encoding is `EUC-JP` and there's no `FOUNDATION_FRAMEWORK` (at this point).

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
jmschonfeld: Ah `LockedState` is `Sendable` but it requires its stored value to be of a `Sendable` type, and I had mistakenly thought `OpaquePointer` was `Sendable` but it is indeed not `Sendable` which is why the `Sendable` conformance isn't available there. This `@unchecked Sendable` makes sense then, since despite the pointer and anything it stores never escapes the lock.

YOCKOW:
Please let me request reviews again as per changes above.

Sources/FoundationEssentials/String/String+IO.swift
```diff
@@ -184,6 +189,17 @@ extension String {
                 }
             }
             self = bytes.withContiguousStorageIfAvailable(buildString) ?? Array(bytes).withUnsafeBufferPointer(buildString)
+        case .japaneseEUC:
+            // Here we catch encodings that are supported by Foundation Framework
+            // but are not supported by corelibs-foundation.
+            // We delegate conversion to ICU.
+            guard let string = (
+                bytes.withContiguousStorageIfAvailable({ _icuMakeStringFromBytes($0, encoding: encoding) }) ??
```
jmschonfeld, line 197: Actually sorry one more thing that I just caught - this `_icuMakeStringFromBytes` function is only defined in `!FOUNDATION_FRAMEWORK` builds so this fails to build with `FOUNDATION_FRAMEWORK` enabled. Could you wrap this case in `#if !FOUNDATION_FRAMEWORK`. For now in Foundation.framework this will fall back to `NSString` below, and we can look at using ICU instead of `NSString` in Foundation.framework in the future (since that may be a behavioral change. Same goes for the case in `StringProtocol+Essentials.swift` as well

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -0,0 +1,204 @@
...
+                $0[encoding] = converter
+                return converter
+            }
+            return nil
+        }
+    }
+}
+
+
+@_dynamicReplacement(for: _icuMakeStringFromBytes(_:encoding:))
```
jmschonfeld, line 176: This replaced function is only defined in `!FOUNDATION_FRAMEWORK` builds, could we wrap this in `#if !FOUNDATION_FRAMEWORK` as well so that it will build successfully in `FOUNDATION_FRAMEWORK` mode?

Sources/FoundationEssentials/String/String+IO.swift
YOCKOW: Thank you for the review. This is indeed wrapped by `#if !FOUNDATION_FRAMEWORK` already:


```swift
#if !FOUNDATION_FRAMEWORK
case .isoLatin1:
  ...
case .macOSRoman:
  ...
case .japaneseEUC:
  ...
#endif
```

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: This is... not wrapped by `#if`. I'm going to fix that. Thank you for the catch.

YOCKOW: Done!

YOCKOW:
@swift-ci Please test

Tests/FoundationInternationalizationTests/StringTests+Data.swift
```diff
@@ -0,0 +1,126 @@
...
+            ]),
+            encoding: .japaneseEUC
+        )
+
+        // Unsupported characters
+        let sushi = "Sushi🍣"
+        XCTAssertNil(sushi.data(using: String._Encoding.japaneseEUC))
+        XCTAssertEqual(
+            sushi.data(using: String._Encoding.japaneseEUC, allowLossyConversion: true),
+            "Sushi?".data(using: .utf8)
```
jmschonfeld, line 136: This test unfortunately fails in the `FOUNDATION_FRAMEWORK` configuration because `Foundation.framework`'s implementation of this conversion replaces both UTF-16 scalars of the emoji with `?` rather than replacing the entire grapheme cluster with a single `?`. I suspect that maybe ICU is replacing only with a single `?` - do you happen to know if that's intentional? If so, for now we might need to vary the expected value to be `Sushi??` in `FOUNDATION_FRAMEWORK` vs. `Sushi?` in `!FOUNDATION_FRAMEWORK` to account for the discrepancy between `Foundation.framework` and ICU (assuming we think this result from ICU is indeed correct and not a bug in how we're calling ICU)

jmschonfeld (commented):
Thanks - that resolves the build failures, but I have a comment about one issue in the tests that I've discovered as well (this should hopefully be the last thing, thanks for bearing with me)

Tests/FoundationInternationalizationTests/StringTests+Data.swift
YOCKOW: Sorry for my overlooking again, but I'm not sure if we should implement conversion just like Foundation framework. 
I mean the behavior of Foundation framework looks buggy because it doesn't seem to handle surrogate pairs correctly. (🍣 is non-BMP but consists of only one scalar `U+1F363`. Does that "buggy" behavior come from the historical reason that `NSString` has been represented as a sequence of UTF–16??)
Of course, we *can imitate* such behavior by replacing `UCNV_FROM_U_CALLBACK_SUBSTITUTE` with a custom callback function.
However, let me point out that [current `swift-foundation`'s implementation for `.ascii`](https://github.com/swiftlang/swift-foundation/blob/1255d73ab06b707558d9ad3cc27999c4c83ace53/Sources/FoundationEssentials/String/StringProtocol%2BEssentials.swift#L104-L114) is code-point-basis, not UTF-16-value-basis:

```swift
import Foundation

let sushiEmoji = "🍣"
print(sushiEmoji.data(using: .ascii, allowLossyConversion: true)!.count)
// -> Prints "1"
```

**Options...?**

- To keep consistent with current (e.g.)`.ascii` implementation:
  + Divide the test case for `FOUNDATION_FRAMEWORK` and `!FOUNDATION_FRAMEWORK`.
- To make it compatible with Foundation framework's behavior:
  + Implement a custom callback for `ucnv_setFromUCallBack`.

jmschonfeld: Yeah I suspect this behavior comes from the fact that `NSString` is natively stored as UTF-16. For now, let's just divide the test case with a `FOUNDATION_FRAMEWORK`/`!FOUNDATION_FRAMEWORK` value for this expectation since I think enabling the conversion at all on non-Darwin is a good step even if the behavior is slightly different than `Foundation.framework`, and I can come back with a separate change to look into making this behavior the same across both platforms (while ensuring we don't break compatibility with existing clients).

YOCKOW:
@jmschonfeld
Thank you for the advice. I revised the test cases.

(I didn't know CI didn't execute test cases with `FOUNDATION_FRAMEWORK`.)

YOCKOW:
@swift-ci Please test

jmschonfeld (approved):

jmschonfeld:
Thanks for all of your updates - looks great to me now!

PR #1302, opened 2025-05-19
Character-only mutations should not overwrite paragraph-bound attributes in paragraph following mutation

jmschonfeld:
When mutating text in an AttributedString that contains paragraph-bound attributes we fix-up the attributes post-mutation by looking at the paragraphs surrounding the mutation. There are up to two applicable paragraphs: one that begins before the mutation and ends somewhere within the mutation (the "starting" paragraph), and one that begins within the mutation and ends after the mutation (the "ending" paragraph). These may be the same paragraph (when the mutation is in the middle of the paragraph), or one or the other may not exist (if the mutation is at the start or the end of a paragraph).

Today, when the mutation is a character-based mutation (i.e. not just setting an attribute value) we always take the attributes from the beginning of the starting paragraph and apply them to the portion that extends into the mutation, and apply the attributes from the beginning of the ending paragraph (within the mutation) and apply them to the rest of the ending paragraph outside the mutation. This is correct for `AttributedString` based replacements (inserting an `AttributedString` into another `AttributedString`) but this isn't quite right for character based replacements (via the character view) where no attributes are supplied. Instead of the inferred attributes of the newly inserted text overwriting the ending paragraph, we should use the attributes from the existing ending paragraph in the inserted/mutated text. This mainly impacts when characters are inserted or mutated at the start of a paragraph and ensures it has the same behavior as when text is inserted anywhere else in the paragraph.

### DIFF
--- a/Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
+++ b/Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
@@ -224,6 +224,10 @@ class TestAttributedStringConstrainingBehavior: XCTestCase {
         verify(string: result, matches: [("HeTestllo, world\n", 1), ("Next Paragraph", 2)], for: \.testParagraphConstrained)
         
         result = str
+        result.characters.insert(contentsOf: "Test", at: result.index(result.startIndex, offsetByCharacters: 13))
+        verify(string: result, matches: [("Hello, world\n", 1), ("TestNext Paragraph", 2)], for: \.testParagraphConstrained)
+        
+        result = str
         result.characters.append(contentsOf: "Test")
         verify(string: result, matches: [("Hello, world\n", 1), ("Next ParagraphTest", 2)], for: \.testParagraphConstrained)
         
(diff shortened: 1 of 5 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/AttributedString/AttributedStringConstrainingBehaviorTests.swift
```diff
@@ -257,7 +261,7 @@ class TestAttributedStringConstrainingBehavior: XCTestCase {
         
         result = str
         result.characters.replaceSubrange(result.index(result.startIndex, offsetByCharacters: 8) ..< result.index(result.startIndex, offsetByCharacters: 15), with: "Test\nReplacement")
-        verify(string: result, matches: [("Hello, wTest\n", 1), ("Replacementxt Paragraph", 1)], for: \.testParagraphConstrained)
+        verify(string: result, matches: [("Hello, wTest\n", 1), ("Replacementxt Paragraph", 2)], for: \.testParagraphConstrained)
```
jmschonfeld, line 264: This is an intentional change in behavior that I feel was previously incorrect. When replacing text across a paragraph boundary, we shouldn't overwrite the second paragraph with the first paragraph's attribute value, instead the two paragraphs should maintain their individual attribute values. This is unlike replacing the subrange with an `AttributedString` where the client supplies attributes that should extend from the replacement into the rest of the ending paragraph

jmschonfeld:
Upon further investigation, I've found that this behavior isn't exactly what we want everywhere so I'm going to close this for now and open a new change in the future if I come up with a more restricted change in the behavior

Issue #717, opened 2023-03-05
Locale.current returns en_US even when the user's locale settings is different on Linux


jasaldivara:
I'm using Swift Language on a Linux system

My OS Distro is Fedora 37 on an ARM64 CPU
I'm using Swift version 5.7.3 (swift-5.7.3-RELEASE) from the Fedora repos

I'm using GNOME desktop and have set my user session language to spanish, and my region to Mexico

When I run the `locale` command on a terminal, I get this output:

```
LANG=es_ES.UTF-8
LC_CTYPE="es_ES.UTF-8"
LC_NUMERIC=es_MX.UTF-8
LC_TIME=es_MX.UTF-8
LC_COLLATE="es_ES.UTF-8"
LC_MONETARY=es_MX.UTF-8
LC_MESSAGES="es_ES.UTF-8"
LC_PAPER=es_MX.UTF-8
LC_NAME="es_ES.UTF-8"
LC_ADDRESS="es_ES.UTF-8"
LC_TELEPHONE="es_ES.UTF-8"
LC_MEASUREMENT=es_MX.UTF-8
LC_IDENTIFICATION="es_ES.UTF-8"
LC_ALL=
```

But when I reference Locale.current, either on a compiled program, or from the Swift REPL, I get a "en_US" Locale. For example, if I run `print (Locale.current.identifier)` on the REPL, i get:

```
en_US
```

I think the expected behavior is that Locale.current returns the same locale from my user settings. Or am I missing something?

Cyberbeni:
With Swift 6.0.1 it is hardcoded to be "en_001": https://github.com/swiftlang/swift-foundation/blob/5141cf69577f26cdbabd65f6b12ff643b93f4685/Sources/FoundationEssentials/Locale/Locale_Cache.swift#L281-L286

There doesn't seem to be a way to change `Locale.current` but you can set a different Locale for stuff like DateFormatter:
```swift
let dateFormatter = DateFormatter()
dateFormatter.dateStyle = .long
dateFormatter.timeStyle = .long
dateFormatter.locale = Locale(identifier: "hu")
```

PR #1308, opened 2025-05-21
Use the LANG environment variable on Linux

Cyberbeni:
Partial fix for #717 

Tried running `swift test --filter LocaleTests` but it seems to be crashing even without my changes, so I couldn't test these changes.


YOCKOW:
💭 I think the difficult problem to solve here is that reading the value from `LANG` will cause another inconsistency with Darwin's behavior. I mean `Locale.current.identifier` is not affected by `LANG` on Darwin.

PR #1312, opened 2025-05-22
build: adjust the autolink library for Windows

compnerd:
When building statically, ensure that we prefix the Swift libraries with the correct prefix to permit static linking of the libraries.

### DIFF
--- a/Sources/FoundationEssentials/CMakeLists.txt
+++ b/Sources/FoundationEssentials/CMakeLists.txt
@@ -99,7 +99,7 @@ if(NOT BUILD_SHARED_LIBS)
     target_compile_options(FoundationEssentials PRIVATE
         "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend $<$<PLATFORM_ID:Windows>:${CMAKE_STATIC_LIBRARY_PREFIX_Swift}>_FoundationCollections>")
     target_compile_options(FoundationEssentials PRIVATE
-        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend swiftSynchronization>")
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/CMakeLists.txt
```diff
@@ -99,7 +99,7 @@ if(NOT BUILD_SHARED_LIBS)
     target_compile_options(FoundationEssentials PRIVATE
         "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend $<$<PLATFORM_ID:Windows>:${CMAKE_STATIC_LIBRARY_PREFIX_Swift}>_FoundationCollections>")
     target_compile_options(FoundationEssentials PRIVATE
-        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend swiftSynchronization>")
+        "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -public-autolink-library -Xfrontend $<$<PLATFORM_ID:Windows>:${CMAKE_STATIC_LIBRARY_PREFIX_Swift}>swiftSynchronization>")
```
jmschonfeld, line 102: Is this missing from the `_FoundationCShims` option above as well?

compnerd: No, it isn't present on `_FoundationCShims` since that is a C library, and we do not currently touch the C library naming on Windows. The Swift libraries are the only ones that we are currently prefixing (in the hopes that we can merge the static and shared directories some day).

jmschonfeld (approved):
Seems fine to me, but before merging could we test the CMake build via a cross-repo test with the swift repo (sadly we don't have the ability to test CMake in this repo) just to make sure we don't miss anything there?

compnerd:
Running the build, but unlikely to catch anything as we are not doing enough of the static library builds.

compnerd:
Yeah, because we weren't building Synchronization. I've added that and am going to re-try.

PR #1313, opened 2025-05-22
Resolve AttributedString attribute storage Sendable warnings

jmschonfeld:
`AttributedString` is `Sendable`, and it requires that all of its attribute values are `Sendable`. However, prior to this decision, we shipped attributes in the Apple SDKs that have non-`Sendable` values. We could not add an `AttributedStringKey.Value : Sendable` constraint because that constraint would cause downstream projects to fail to build when built with concurrency errors enabled, so instead every public `AttributedString` entry point has a `where Key.Value : Sendable` constraint to prevent reading/writing attributes values that are non-`Sendable`.

Unfortunately, there are a few situations where we cannot put this constraint because clients provide an attribute scope rather than an individual attribute key: attribute encoding/decoding and `NSAttributedString` conversion. In these cases, Foundation currently has warnings that indicate the values we're reading/writing may not be `Sendable`. This change effectively silences those warnings as there isn't anything we can do in Foundation about them.

This does leave a `Sendable` "escape hatch" in `AttributedString`: you can have an `NSAttributedString` that contains non-`Sendable` attributes and convert it to an `AttributedString` which stores it in a `Sendable` object. The constraints prevent reading the attribute value from `AttributedString` itself, but you can pass it across an isolation boundary and then convert it back to an `NSAttributedString` at which point you can read the non-`Sendable` value. Unfortunately, there's not much that we can do to prevent this currently, so the best we can do is at least allow Foundation to build without the warnings.

I've put this internal escape hatch in `FOUNDATION_FRAMEWORK` to help prevent accidental uses where it isn't needed.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedStringAttributeStorage.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedStringAttributeStorage.swift
@@ -32,6 +32,21 @@ extension AttributedString {
             inheritedByAddedText = K.inheritedByAddedText
             invalidationConditions = K.invalidationConditions
         }
+        
+        #if FOUNDATION_FRAMEWORK
+        @inline(__always)
+        private static func _unsafeAssumeSendableRawValue<T>(_ value: T) -> RawValue {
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #1314, opened 2025-05-22
[Proposal] Early Data Support in URLSession

guoye-zhang:
Introducing a flag to opt in to HTTP/3 early data in the new HTTP loader on Darwin.

PR: https://github.com/swiftlang/swift-corelibs-foundation/pull/5216


guoye-zhang:
@swift-ci please test

PR #1320, opened 2025-05-26
fix recurrence nanosecond handling for nanosecond values > 0.5

lukaskollmer:
This PR is a follow up to #1284: we missed the fact that `remainder(dividingBy:)` can produce negative values if the input's fractional part is greater than 0.5. (eg: `1.9.remainder(dividingBy: 1)` is -0.1).

Since the dates produced by `Calendar.DatesByRecurring.Iterator` all have their nanosecond component set to 0 (rather than being rounded to the nearest whole number), we need to use `truncatingRemainder(dividingBy:)` instead.

I also changed it to operate on `timeIntervalSinceReferenceDate` instead of `timeIntervalSince1970`, since that saves us on having to perform the `timeIntervalSince1970` computation every time.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -235,7 +235,7 @@ extension Calendar {
                 }
                 var componentsForEnumerating = recurrence.calendar._dateComponents(components, from: start) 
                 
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

parkera (approved):

stephentyrone (approved):

PR #1327, opened 2025-06-02
Remove extra AttributedStringProtocol.range(of:) overload

jmschonfeld:
I suspect this overload was added to our test files to support the tests before `AttributedStringProtocol.range(of:)` was ported to swift-foundation. But today, the `range(of:)` function exists in swift-foundation so we should remove this extra one from our tests to make sure we're testing the real function rather than the one overloaded by the tests.


jmschonfeld:
@swift-ci please test

PR #1328, opened 2025-06-02
Reduce size of `cachedFixedComponentsLocales`

itingliu:
Previously we maintain a cache `private var cachedFixedComponentsLocales: [Locale.Components : any _LocaleProtocol]`, and every call to `Locale(components: Locale.Components)` reads a
nd writes to the cache. This results in a cache with non trivial size of keys.

Optimize this by replacing the key with a String instead of the full `Locale.Components` instance.

Using this as the baseline:

```swift
for identifier in Locale.availableIdentifiers {
    let locale = Locale(identifier: identifier)
}
```

With the following as the testing code:

```swift
for identifier in Locale.availableIdentifiers {
    let components = Locale.Components(identifier: identifier)
    let locale = Locale(components: components)
    let components2 = Locale.Components(locale: locale)
    let locale2 = Locale(components: components2) // cache hit
}
```

Previously, this incurred +2.6 MB of footprint comparing to the above baseline. After the fix, it incurred only +0.9MB increase.

Also benchmarked that we have not regressed the time now that we are using the string-based identifier as the key.

### DIFF
--- a/Benchmarks/Benchmarks/Internationalization/BenchmarkLocale.swift
+++ b/Benchmarks/Benchmarks/Internationalization/BenchmarkLocale.swift
@@ -13,15 +13,20 @@
 import Benchmark
 import func Benchmark.blackHole
 
-#if FOUNDATION_FRAMEWORK // This test uses CFString
+#if os(macOS) && USE_PACKAGE
+import FoundationEssentials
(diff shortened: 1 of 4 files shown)


itingliu:
@swift-ci please test

PR #1330, opened 2025-06-03
Include `<stdnoreturn.h>` last

Steelskin:
Starting with the Windows SDK 10.0.26100.3916, inclusion of `<stdnoreturn.h>` clashes with some of the definitions in other ucrt headers. Move `<stdnoreturn.h>` to be last to work around this issue.

### DIFF
--- a/Sources/_FoundationCShims/include/_CStdlib.h
+++ b/Sources/_FoundationCShims/include/_CStdlib.h
@@ -101,10 +101,6 @@
 #include <stdlib.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


compnerd:
:sigh: this is horrible ...

```
#define noreturn _Noreturn
__declspec(noreturn)
```
is then translated into `__declspec(_Noreturn)` which is not valid :/

compnerd:
@swift-ci please test

compnerd (approved):

compnerd:
Going to merge this - this should be harmless.

PR #1331, opened 2025-06-04
Temporarily make Data.bytes a disfavored overload

glessard:
Make `Data.bytes` a disfavored overload, to resolve issues when a using project defines a property of that name, while also importing FoundationEssentials.


glessard:
@swift-ci please test

PR #1344, opened 2025-06-10
Add an implementation of the UTCClock and associated tests

phausler:


phausler:
@swift-ci please test

Tests/FoundationEssentialsTests/UTCClockTests.swift
```diff
@@ -0,0 +1,95 @@
...
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(TestSupport)
+import TestSupport
+#endif
+
+#if canImport(FoundationEssentials)
+@testable import FoundationEssentials
```
jmschonfeld, line 18: Does this test need a `@testable` import? If not, I think you can remove this. If it does, I think you'll want an `#else` branch that has `@testable import Foundation` to make sure this works for `FOUNDATION_FRAMEWORK` where `FoundationEssentials` doesn't exist

benrimmington:
Can the empty `UTCClock` struct be `@frozen` or `BitwiseCopyable`?

PR #1348, opened 2025-06-11
[wasm] Fix WASI build by using `time_t` for tv_sec

kateinoigakukun:
`time_t` is defined as `Int64` in wasi-libc, not `Int`.
Leave `tv_nsec` as is now because it's defined as `long` 
c.f. https://pubs.opengroup.org/onlinepubs/009695399/basedefs/time.h.html


kateinoigakukun:
@swift-ci test

jmschonfeld (approved):

PR #1350, opened 2025-06-11
Continued swift testing adoption

jmschonfeld:
These are cherry-picked parts of https://github.com/swiftlang/swift-foundation/pull/908. In particular, I pulled in each file that applied cleanly or almost cleanly (i.e. files that haven't really added any or more than 1-2 new tests since my original porting work last Fall) for tests for FoundationEssentials. Since we still have some XCTests present I moved the existing `checkHashable`-related functions aside (prefixed with `XCT` for XCTest) and added the new `checkHashable` functions based on swift-testing. This covers most of the FoundationEssentials tests, but leaves out some suites that have changed a lot (ex. String/Data/FileManager) and some suites that are new.

### DIFF
--- a/Tests/FoundationEssentialsTests/StringTests.swift
+++ b/Tests/FoundationEssentialsTests/StringTests.swift
@@ -1463,7 +1463,7 @@ final class StringTestsStdlib: XCTestCase {
             .ascii,
             .utf8,
         ]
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
@swift-ci please test

PR #1351, opened 2025-06-12
Convert FileManager and URL tests to swift-testing

jmschonfeld:
This updates the `FileManager` and `URL` tests to swift-testing, based on the patch in https://github.com/swiftlang/swift-foundation/pull/908 but updating to accommodate the latest tests. Additionally, this patch takes a different approach to serialization. Rather than creating a parent `Suite` for all tests that change the CWD, it instead defines a `GlobalActor` to represent the process' CWD. All tests that use the helper functionality to change the CWD await the test on this global actor which ensures that each block that requires a specific, consistent CWD will never run in parallel to another such block so that the CWD is globally serialized amongst any test that uses the helper automatically.

### DIFF
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -10,106 +10,102 @@
 //
 //===----------------------------------------------------------------------===//
 
-
-#if canImport(TestSupport)
-import TestSupport
-#endif // canImport(TestSupport)
+import Testing
 
 #if canImport(FoundationEssentials)
 @testable import FoundationEssentials
-#endif
-
-#if FOUNDATION_FRAMEWORK
+#else
 @testable import Foundation
 #endif
 
-private func checkBehavior<T: Equatable>(_ result: T, new: T, old: T, file: StaticString = #filePath, line: UInt = #line) {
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/URLTests.swift
```diff
@@ -237,41 +232,41 @@ final class URLTests : XCTestCase {
...
+                #expect(url.absolutePath() == url.path)
             }
         }
     }
 
-    func testURLPathComponentsPercentEncodedSlash() throws {
-        try XCTSkipIf(!foundation_swift_url_enabled())
+    
+    @Test(.enabled(if: foundation_swift_url_enabled()))
+    func URLPathComponentsPercentEncodedSlash() throws {
```
jrflat: Small nit: can we remove the `URL` to make it `pathComponentsPercentEncodedSlash()` to match the other test names? There's a few more below, too.

jmschonfeld: Yep - I must have missed those, thanks for catching that. Updated!

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1353, opened 2025-06-13
Disable macOS GitHub Actions CI

jmschonfeld:
It turns out that swiftlang's GitHub actions macOS CI doesn't use a nightly snapshot like the linux/windows testing, but instead uses the compiler from the Xcode installed on the device which currently only contains a 6.1 toolchain (which isn't new enough to build swift-foundation). This turns macOS GitHub action CI back off for this repo until we have an action that will use a new enough toolchain.


jmschonfeld:
@swift-ci please test

itingliu (approved):
😔

PR #1357, opened 2025-06-13
Convert String/Predicate/ProcessInfo/RecurrenceRule tests to swift-testing

jmschonfeld:
Applies more patches from https://github.com/swiftlang/swift-foundation/pull/908 to bring a handful more FoundationEssentials test suites to swift-testing (adjusting the patches to account for newly added and modified tests)

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
@@ -542,6 +542,7 @@ private struct GregorianCalendarRecurrenceRuleTests {
     
     @Test func yearlyRecurrenceWithWeekNumberExpansion() {
         var calendar = Calendar(identifier: .gregorian)
+        calendar.timeZone = .gmt
         calendar.firstWeekday = 2 // Week starts on Monday
         calendar.minimumDaysInFirstWeek = 4
         
@@ -569,6 +570,7 @@ private struct GregorianCalendarRecurrenceRuleTests {
     
     @Test func yearlyRecurrenceWithWeekNumberExpansionCountingBack() {
         var calendar = Calendar(identifier: .gregorian)
+        calendar.timeZone = .gmt
         calendar.firstWeekday = 2 // Week starts on Monday
         
         let start = Date(timeIntervalSince1970: 1704117600.0) // 2024-01-01T14:00:00-0000
--- a/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
@@ -10,17 +10,16 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if canImport(TestSupport)
-import TestSupport
+import Testing
+
+#if canImport(FoundationEssentials)
+@testable import FoundationEssentials
+#else
+@testable import Foundation
 #endif
 
-#if FOUNDATION_FRAMEWORK
-@testable import Foundation
-#else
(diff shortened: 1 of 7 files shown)


jmschonfeld:
@swift-ci please test

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -10,35 +10,38 @@
...
         var gregorian = Calendar(identifier: .gregorian)
         gregorian.timeZone = .gmt
         return gregorian
     }()
   
-    func testRoundtripEncoding() throws {
+    @Test func roundtripEncoding() throws {
         // These are not necessarily valid recurrence rule, they are constructed
         // in a way to test all encoding paths
-        var recurrenceRule1 = Calendar.RecurrenceRule(calendar: .current, frequency: .daily)
```
hristost, line 34: If I recall correctly, this was meant to test that the current calendar is encoded as such. We don't have any custom logic in `RecurrenceRule` encoding to handle such cases so it's also okay to encode with other calendars. But it'd be nice to minimize the diff.

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
```diff
@@ -681,12 +689,12 @@ final class GregorianCalendarRecurrenceRuleTests: XCTestCase {
         rule.matchingPolicy = .strict
 
         for _ in rule.recurrences(of: start) {
-            XCTFail("Recurrence rule is not expected to produce results")
+            Issue.record("Recurrence rule is not expected to produce results")
```
hristost, line 692: Would this automatically mark the test as failing?

hristost (commented):
The recurrence rule tests look alright, modulo one comment about `Calendar.current`

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
jmschonfeld: Removing the usage of `.current` here is important because the `.current` calendar also implies things like the current `TimeZone`. Unlike XCTest tests, swift-testing tests run in parallel with each other, including other tests which may change the value of the current calendar/timezone/locale. Tests that use or change the current calendar/timezone/locale need to be serialized to ensure they don't run in parallel like I did with the tests relying on the CWD. I can go ahead and setup the infrastructure for that now (I planned to do that later when refactoring the FoundationInternationalization tests) but if there's no custom logic for the current calendar, to avoid that need for these tests it's likely simpler to just switch to a non-current calendar than switch to an asynchronous task run on an actor.

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
jmschonfeld: Yep, `XCTFail` and `Issue.record` are equivalent - both record an issue that will cause the test to be marked as failed and continue execution to the end of the test to allow for recording additional issues

Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
hristost: I see! It's good with me then. We can also use `GregorianCalendarRecurrenceRuleTests.gregorian`, which is initialized only once.

PR #1359, opened 2025-06-13
Convert FoundationEssentials formatting tests to swift-testing

jmschonfeld:
This converts the newer Formatting tests for FoundationEssentials to swift-testing. It also moves one test to the internationalization test module since it requires ICU information, and removes an extraneous `internal` `TimeZone` initializer in favor of using the API initializer and removing the `@testable` import.


jmschonfeld:
@swift-ci please test

PR #1363, opened 2025-06-17
Convert Data/DataIO Tests to swift-testing

jmschonfeld:
Continued cherry-picks (with edits) of the initial swift-testing adoption for all of our data-related tests


Tests/FoundationEssentialsTests/StringTests.swift
```diff
@@ -893,6 +901,47 @@ private struct StringTests {
             #expect(result == expected)
         }
     }
+    
+    @Test func deletingLastPathComponent() {
```
jrflat, line 905: Thanks for consolidating these!

Tests/FoundationEssentialsTests/DataTests.swift
```diff
@@ -1674,56 +1686,58 @@ class DataTests : XCTestCase {
...
         span = source.mutableSpan
-        let i = try XCTUnwrap(span.indices.randomElement())
-        XCTAssertFalse(span.isEmpty)
-        XCTAssertEqual(span.count, count)
+        let indices = span.indices
+        let i = try #require(indices.randomElement())
+        isEmpty = span.isEmpty
+        #expect(!isEmpty)
+        count = span.count
+        #expect(count == count)
```
jrflat: Maybe a different variable name (or `source.count`) would be good :)

jmschonfeld: Oh good call, thanks this is probably a typo on my part 🙂 - likely a find and replace error while extracting the span calls out of the expectations (which sadly don't work because `#expect` can't capture the non escapable type in its  escaping closure)

jmschonfeld: It turns out that `span.count` in an expectation is fine, so I removed the extra variables. `span.isEmpty` doesn't work, so I've left the `isEmpty` variables and filed https://github.com/swiftlang/swift-testing/issues/1164 about the discrepancy

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1364, opened 2025-06-17
Convert JSON/PropertyList tests to swift-testing

jmschonfeld:
Continued cherry-picks (with edits) of the initial swift-testing adoption for all of our encoder/decoder-related tests

In particular, this change (in addition to adopting swift-testing):

- Enables the exit tests previously if-def'ed out in these files (they now pass)
- Slightly refactors some expectation handling (using `#expect(throws:)` instead of `do`/`catch` with `Issue.record` for example) for older tests written before these types of expectations existed
- Removes trivial uses of `try!`/`as!` to help trigger expectation failures instead of crashes when unit tests fail
- Removes accidental uses of Foundation APIs not present in FoundationEssentials (like `replacingOccurrences(of:with:)` and `range(of:)`)

### DIFF
--- a/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
@@ -1283,7 +1283,9 @@ private struct JSONEncoderTests {
         #expect(try 3.14 == JSONDecoder().decode(Double.self, from: partialData))
     }
 
-    @Test func depthTraversal() {
+    @Test
+    @MainActor // Deeply recursive tests which requires running on the main thread which has a higher stack size limit
+    func depthTraversal() {
         struct SuperNestedArray : Decodable {
             init(from decoder: Decoder) throws {
                 var container = try decoder.unkeyedContainer()
--- a/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
@@ -9,138 +9,145 @@
 // See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
 //
 //===----------------------------------------------------------------------===//
-//
-// RUN: %target-run-simple-swift
-// REQUIRES: executable_test
-// REQUIRES: objc_interop
-// REQUIRES: rdar49634697
-// REQUIRES: rdar55727144
 
-#if canImport(TestSupport)
-import TestSupport
-#endif // canImport(TestSupport)
+import Testing
+
+#if canImport(Darwin)
+import Darwin
+#elseif canImport(Bionic)
+@preconcurrency import Bionic
+#elseif canImport(Glibc)
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif canImport(CRT)
+import CRT
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
 
 #if canImport(FoundationEssentials)
 @_spi(SwiftCorelibsFoundation)
-@testable import FoundationEssentials
+import FoundationEssentials
 #endif
 
 #if FOUNDATION_FRAMEWORK
-@testable import Foundation
+import Foundation
 #endif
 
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
```diff
@@ -951,11 +945,13 @@ final class JSONEncoderTests : XCTestCase {
                 }
             }
         }
-        let json = "[1, 2, 3]".data(using: String._Encoding.utf8)!
-        XCTAssertNoThrow(try JSONDecoder().decode(RepeatNilCheckDecodable.self, from: json))
+        let json = "[1, 2, 3]".data(using: .utf8)!
+        #expect(throws: Never.self) {
+            try JSONDecoder().decode(RepeatNilCheckDecodable.self, from: json)
+        }
```
natecook1000, line 951: Is the idea that this never throws? It might be more clear to mark the function as `throws` and then just have the code be:
```suggestion
        let _ = try JSONDecoder().decode(RepeatNilCheckDecodable.self, from: json)
```

Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
```diff
@@ -1505,86 +1528,87 @@ data1 = <7465
...
         
-        XCTAssertEqual(try PropertyListDecoder().decode([String:String].self, from: acceptable), ["bar":"foo"])
+        #expect(try PropertyListDecoder().decode([String:String].self, from: acceptable) == ["bar":"foo"])
     }
 }
             
 
 // MARK: - Helper Global Functions
-func XCTAssertEqualPaths(_ lhs: [CodingKey], _ rhs: [CodingKey], _ prefix: String) {
+func AssertEqualPaths(_ lhs: [CodingKey], _ rhs: [CodingKey], _ prefix: String) {
```
natecook1000: Would be nice to augment this with a `sourceLocation: SourceLocation = #_sourceLocation`.

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
natecook1000: Ah, I see that this is a specific `Testing` pattern. I don't totally get this distinction, but it sounds like what's there is the right code to use!

> If the intent is for a test to fail when an error is thrown by expression, rather than to explicitly check that an error is not thrown by it, do not use this macro. Instead, simply call the code in question and allow it to throw an error naturally.

natecook1000 (approved):
Looks great! Just a couple notes, but no changes needed 👍🏻

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
jmschonfeld: Yeah my thought was that with `let _ = try JSONDecoder().decode(RepeatNilCheckDecodable.self, from: json)` the intent is a little ambiguous (we don't do anything with the result, so what exactly are we testing?) whereas the `#expect(throws: Never.self)` makes it explicitly clear that we're testing that this doesn't throw (as opposed to doesn't crash, or doesn't produce some other side effect). We do use the `let _ = ...` syntax elsewhere, but for cases where we're just testing things that shouldn't crash (but ideally those would have an expectation on the return value as well).

Additionally, if there is an issue, `let _ = try ...` stops the test immediately and does not attempt the rest of the test, whereas `#expect(throws: Never.self)` will continue to check other expectations even if this throws, which can get you a more complete story of failures when running in remote environments like CI.

Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
jmschonfeld: Good call! I'll update this

jmschonfeld:
@swift-ci please test

kperryua (approved):

jmschonfeld:
@swift-ci please test

PR #1368, opened 2025-06-18
Remove unknown context from PredicateTests.Object type name

jmschonfeld:
Windows swift-ci is currently failing on the swift repo because the name of `Object` produced by `FoundationEssentials` does not match the name produced by the unit tests (the context of the type is different, in `FoundationEssentials` it's unknown but in the unit test file it's based on the file name). To resolve this, I moved `Object` to the global scope and marked it as `internal` so its name is consistently `FoundationEssentialsTests.PredicateTestObject` instead of including a potentially changing intermediate context scope.


jmschonfeld:
@swift-ci please test

PR #1370, opened 2025-06-19
[DRAFT] Migrate to UTF8Span

milseman:
This might be better understood by looking at the 3 commits individually (otherwise it's a big diff).

First commit gets us out of passing down the original `String` value all the way through for the purposes of diagnostics.

Second commit untangles all diagnostic code from the low-level scanner/lexer style code, while keeping the behavior and basic code structure unchanged. This shows us the basic patterns that we could surface as API on our Span-like types and should spark some interesting designs and discussions.

Third commit just migrates over to `UTF8Span`, albeit via a `guard #available else { fatalError() } ` pattern. Not sure if that's the way we want to go.

Note that I haven't tested this fully yet.


milseman:
@swift-ci please test

Issue #1367, opened 2025-06-18
URLs created using an URL at `relativeTo:` which contains either `?` or `#` return truncated `URL.path`


fheidenreich:
**Describe the bug**
Constructing URLs using an URL at `relativeTo:` which contains either ? or # has unexpected behaviour. If any of those two characters is in the directory portion of the URL, `URL.path` truncates at this character.

**To Reproduce**
Sample code to reproduce the behavior.

```swift
@Test func urlRelativeToContainsQuestionMark() {
    let basePath = "/Users/florian/Test1 Test2? Test3/Test4"
    let baseURL = URL(fileURLWithPath: basePath, isDirectory: true)
    let fileURL = URL(fileURLWithPath: "../Test5.txt", isDirectory: false, relativeTo: baseURL)
    
    // Will be `/Users/florian/Test1 Test2` instead of `/Users/florian/Test1 Test2? Test3`
    #expect(fileURL.deletingLastPathComponent().path == baseURL.deletingLastPathComponent().path)
}
```

**Expected behavior**
I expect `URL.path` not to return a truncated path in the above-described cases. 

**Configuration:**
 - Swift Version: Apple Swift version 6.2-dev (LLVM e64ef521c460153, Swift c1db12a0d435cf2)
 - OS: macOS
 - OS Version: Sequoia 15.5, Tahoe 26 Beta 1

**Regression information:**
Xcode 16.4 on macOS Sequoia 15.5

PR #1378, opened 2025-06-24
URL.absoluteString should always use encoded base components

jrflat:
If we aren't in the special `CFURL` case where we need the original absolute string, we should always leave the base components encoded. This ensures we receive properly encoded absolute strings that correctly treat `?` and `#` as encoded path characters if those were previously inputted to the `URL(filePath:)` initializer.

Resolves https://github.com/swiftlang/swift-foundation/issues/1367


jrflat:
@swift-ci please test

parkera (approved):

PR #1381, opened 2025-06-26
Begin converting FoundationInternationalizationTests to swift-testing

jmschonfeld:
This applies changes from https://github.com/swiftlang/swift-foundation/pull/908 to all files for which the changes apply cleanly which covers a significant number of tests. I'll come back with separate PRs to update the patches for files that have changed since the initial audit, but this updates all suites that cleanly apply.

The main change from https://github.com/swiftlang/swift-foundation/pull/908 is how we handle tests that read or mutate the current Locale/TimeZone/Calendar. Instead of using a parent suite to serialize these tests, we use a global actor which tests await work on that depends on these values. The global actor also ensures that the current Calendar/Locale/TimeZone are reset back to their default values after each test to ensure mutations from one test don't have any interactions with future tests. Since XCTest tests all run before swift-testing tests begin, it's ok that XCTests that have yet to be converted might mutate this state while we're working to convert everything over.


jmschonfeld:
@swift-ci please test

Tests/FoundationInternationalizationTests/Formatting/DateIntervalFormatStyleTests.swift
```diff
@@ -5,195 +5,192 @@
...
-        XCTAssertEqual(weekDayStyle.format(date..<date + hour), "Mon")
-        XCTAssertEqual(weekDayStyle.format(date..<date + day), "Mon – Tue")
-        XCTAssertEqual(weekDayStyle.format(date..<date + day * 32), "Mon – Fri")
-
-        // This style doesn't really make sense since the gap between `weekDay` and `hour` makes the result ambiguous. ICU fills the missing pieces on our behalf.
+        #expect(weekDayStyle.format(date..<date + hour) == "Mon")
+        #expect(weekDayStyle.format(date..<date + day) == "Mon – Tue")
+        #expect(weekDayStyle.format(date..<date + day * 32) == "Mon – Fri")
+        
+        // This style doesn't really make sense since the gap between `weekDay` and `hour` makes the result AMbiguous. ICU fills the missing pieces on our behalf.
```
jrflat, line 79: Not sure if intentional but this is kinda punny and I like it 🙂

jmschonfeld: Ha sadly I can't claim credit for this one, no idea how that happened - it must have somehow been one of my regex copy and pastes gone wrong back with the original patch. But I agree the pun is great haha

itingliu: Kudos to @jrflat for finding a punneedle in this haystack of changes

Tests/FoundationInternationalizationTests/GegorianCalendarInternationalizationTests.swift
```diff
@@ -23,7 +23,7 @@ import Testing
 @Suite("Gregorian Calendar (Internationalization)")
 private struct GregorianCalendarInternationalizationTests {
     @Test func copy() {
-        let gregorianCalendar = _CalendarGregorian(identifier: .gregorian, timeZone: nil, locale: nil, firstWeekday: 5, minimumDaysInFirstWeek: 3, gregorianStartDate: nil)
+        let gregorianCalendar = _CalendarGregorian(identifier: .gregorian, timeZone: .gmt, locale: nil, firstWeekday: 5, minimumDaysInFirstWeek: 3, gregorianStartDate: nil)
```
itingliu, line 26: This is a note to myself, but we should probably lift this logic of falling back to the current time zone out of `_CalendarGregorian` :/ Seems like a footgun even just for unit tests

itingliu (approved):

PR #1382, opened 2025-06-26
Ensure that subscripting AttributeSliceX behaves consistently regardless of index location

jmschonfeld:
The `AttributedString.Runs.AttributesSliceX` types represent a run view that is sliced over a specific set of attributes. Iterating this view provides a coalesced range and the value of each attribute across this range. This view can also be subscripted at any `AttributedString.Index` to find the longest effective range of the specified attributes at a given index. The underlying implementation of this subscript had an early-return condition for when the run containing the index happened to be the final run in the slice, in which case the input index was used as the start of the returned range. This isn't correct - the start of this run is dependent upon where the specified attribute value begins and where the prior run boundary is (if applicable). In most situations you'll never hit this early-exit or it happens to be correct, but if you have an attribute slice over a substring and subscript into the middle of the final run, the returned range will be `inputIndex ..< endOfRun` rather than `startOfRun ..< endOfRun`. The added unit test failed due to the mismatch in ranges but by removing this early-return it now passes.

Resolves rdar://154342005

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributedString+Runs.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributedString+Runs.swift
@@ -594,9 +594,6 @@ extension AttributedString.Runs {
         // _strBounds.range(containing:) below validates that i._value is within the bounds of this slice
         precondition(!attributeNames.isEmpty)
         let r = _guts.findRun(at: i._value)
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

PR #1383, opened 2025-06-26
[WIP][Proposal] add is identical methods to concrete types

vanvoorden:
[SE-0494](https://forums.swift.org/t/-/82695)

### DIFF
--- a/Proposals/NNNN-add-is-identical-methods.md
+++ b/Proposals/NNNN-add-is-identical-methods.md
@@ -3,8 +3,8 @@
 * Proposal: [SF-NNNN](NNNN-add-is-identical-methods.md)
 * Authors: [Rick van Voorden](https://github.com/vanvoorden), [Karoy Lorentey](https://github.com/lorentey)
 * Review Manager: TBD
-* Status: **Awaiting implementation**
-* Implementation: TODO
+* Status: **Awaiting review**
+* Implementation: ([AttributedString, AttributedSubstring](https://github.com/swiftlang/swift-foundation/pull/1385)), ([Data](https://github.com/swiftlang/swift-foundation/pull/1384))
 * Review: ([Pre-Pitch](https://forums.swift.org/t/how-to-check-two-array-instances-for-identity-equality-in-constant-time/78792)), ([Pitch #1](https://forums.swift.org/t/pitch-distinguishable-protocol-for-quick-comparisons/79145))
 
 ## Introduction


vanvoorden:
@parkera Would you have any opinions about running this proposal concurrently with our proposal to add these methods to concrete types from stdlib?[^1] Can you think of examples in the past where one evolution proposal opened on stdlib and another proposal on Foundation covered similar designs but just focused on different concrete types?

I'm thinking there might be interesting feedback in one proposal review that we might want to feed back and influence the other proposal review. Or it might be easier to start with stdlib and then consider that completed design the "prior art" for a later proposal on Foundation.

Hmm… any ideas about that?

[^1]: https://github.com/swiftlang/swift-evolution/pull/2875

parkera:
If we run a proposal through swift-evolution and it includes extensions to Foundation API, and the Foundation workgroup doesn't object for some reason, then we can consider it accepted.

parkera:
Maybe we just give interested parties a heads up by creating a thread in the Foundation section, in case they don't follow the other evolution threads on the forums.

vanvoorden:
@parkera Hmm… so it sounds like maybe the idea here is:
* Wait for an LSG review manager to pick up the stdlib proposal.
* Ask the LSG review manager if they are ok with adding these Foundation changes in the stdlib proposal.
* Show the final draft to the Foundation workgroup to ask if they are ok with these changes going along with a stdlib proposal.

For now… I'm still blocked waiting on an LSG review manager for the stdlib proposal. But once that is ready to schedule for a review we can discuss unifying them back together? Does that sounds about right? Thanks!

itingliu:
Either way I can see if the swift-evolution is accepted, I imagine adding it to `AttributedString` will be fairly straightforward because we generally let `AttributedString` follow what `String` has. But I'm not sure if we've had such a clear parallel with `Data` with a stdlib type so perhaps we'll get more feedback in this type.

I see that the [swift-evolution proposal](https://github.com/swiftlang/swift-evolution/pull/2875/files) doesn't currently include Foundation types. If you decide to continue with that version, when it's in review, you can post a pitch about these two proposals for adding it to `Data` and `AttributedString` in the foundation forum, and we can move it into the abbreviated review right away so it runs the same time as the swift-evolution one, so we can get feedback together. The downside is that you'll have to juggle the two review threads at the same time.

If you want to add a section in the swift-evolution proposal to cover `AttributedString` and `Data`, then I think a short message in foundation forums to bring people's attention to the other review thread would be nice. I (or anyone on the foundation workgroup) would also be happy to direct people's attention there if you prefer us to do that. The swift-evolution review is sufficient for foundation workgroup for the purpose of getting feedback, so we don't need another review for swift-foundation. The downside is the review will likely be more complicated since it'd include more types.

vanvoorden:
> But I'm not sure if we've had such a clear parallel with Data with a stdlib type so perhaps we'll get more feedback in this type.

@itingliu My idea here is that engineers reviewing the proposal see these dynamic length copy-on-write data structures being related conceptually. So a type like `Array` functions similar to a type like `Data`: they are collections of elements.

Where things might get a little tricky is what we do about `Data` when elements are small and saved on the stack with `InlineData`. I believe we will not expose an identity operation for `InlineData`: we just return `false` to indicate we do not know if these two are identical. But this might be a better discussion for the diff implementation review on GitHub.[^1] I think we do not have to block the proposal review on this detail.

[^1]: https://github.com/swiftlang/swift-foundation/pull/1384

vanvoorden:
@itingliu ~~I have not heard back from LSG… maybe they are blocked right now on accepting new proposals?~~

~~Maybe it's ok for us to go ahead with this proposal review as a standalone addition to `Foundation`. And then if we hear back from LSG then the stdlib proposal can run another time. How does that idea sound?~~

Actually I think we do have a review manager from LSG now. I can ask them if they want me to merge both proposals together.

itingliu:
I see that SE-0494 is accepted 🎉  Would you want to update this proposal as a follow up for that one?

vanvoorden:
@itingliu Thanks! We also added a lot more types from standard library. It was originally just for 7 but now I think it is up to 20. And we also decided to go with a different name.

I can update this proposal draft for the `Foundation` types. I also can eliminate some of the sections that would just be repeating back what is in the stdlib proposal because I would just be repeating myself from there anyway.

vanvoorden:
@itingliu It actually looks like my schedule might be a little busy for the next few months and I might not have as much time as I thought to take on a new evolution proposal…

I also think maybe I do not need to be the one that authors this new proposal. Most of the "heavy lifting" was done in our standard library proposal that has already been accepted. I think a foundation proposal might be a lot more light weight. This might be a great opportunity for an engineer that is new to proposal evolution to adopt this idea and then lead the evolution on foundation.

I also have these WIP implementation diffs that I think are all mostly correct… a new engineer could use those as a starting point for their own implementations and tests.

What do you think about that? This might be a great opportunity for an engineer that is trying to demonstrate impact shipping with a broader scope and managing ambiguity. Would you have any engineers you could think of that would be a good fit?

PR #1388, opened 2025-06-28
Support for building with 64-bit time_t/off_t on 32-bit platforms

lhoward:
Patch is against release/6.1 but it will apply against main.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -727,7 +727,11 @@ extension _FileManagerImpl {
             let blockSize = UInt64(result.f_bsize)
             #else
             let fsNumber = result.f_fsid
+            #if canImport(Glibc)
+            let blockSize = fsblkcnt_t(result.f_frsize) // support 64-bit block sizes on 32-bit platforms
+            #else
             let blockSize = UInt(result.f_frsize)
+            #endif
             #endif
             var totalSizeBytes = result.f_blocks * blockSize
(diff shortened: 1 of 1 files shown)


lhoward:
If for process reasons you need a PR against main, can do, just lmk once this is reviewed. Sorry to do it back to front.

jessezamora:
@al45tair @MaxDesiatov corresponding PR to https://github.com/swiftlang/swift/pull/82595 for swift-foundation support

MaxDesiatov:
> If for process reasons you need a PR against main, can do, just lmk once this is reviewed. Sorry to do it back to front.

Yes, PRs should be merged first to `main`, and only then reviewed on release branches. This is not entirely about process, but also about consistency in how fixes are, in a way, incubated on `main` to assess the risk before cherry-picking to releases.

lhoward:
> > If for process reasons you need a PR against main, can do, just lmk once this is reviewed. Sorry to do it back to front.
> 
> Yes, PRs should be merged first to `main`, and only then reviewed on release branches. This is not entirely about process, but also about consistency in how fixes are, in a way, incubated on `main` to assess the risk before cherry-picking to releases.

Gotcha. PRs on main filed, please ignore this until those are merged.

al45tair:
> Yes, PRs should be merged first to `main`, and only then reviewed on release branches. This is not entirely about process, but also about consistency in how fixes are, in a way, incubated on `main` to assess the risk before cherry-picking to releases.

It's also to prevent unexpected regressions, where we've merged a fix to a release branch and then it doesn't get merged on main, or gets reverted on main, and then the next release suddenly goes backwards and has a bug that was already fixed.

PR #1389, opened 2025-06-30
Use __BundleLookupHelper.self when building mergable library

Tantalum73:
Updating the `#bundle` macro to check whether `SWIFT_BUNDLE_LOOKUP_HELPER_AVAILABLE` is set. If that is the case, the build system built a mergable library and emitted a class called `__BundleLookupHelper`. 

The linker has a code path that makes `Bundle(forClass: )` work for merged libraries, whereas the DSO-handle is the same for all merged targets.

The build system counterpart has been merged here: https://github.com/swiftlang/swift-build/pull/599


jmschonfeld:
@swift-ci please test

PR #1391, opened 2025-06-30
Continue to move FoundationInternationalization tests to swift-testing

jmschonfeld:
This moves most of the remaining test cases to swift-testing (with a few still remaining that I'm working on)

### DIFF
--- a/Tests/FoundationInternationalizationTests/CalendarTests.swift
+++ b/Tests/FoundationInternationalizationTests/CalendarTests.swift
@@ -202,21 +202,21 @@ private struct CalendarTests {
         #expect(anyHashables[1] == anyHashables[2])
     }
 
-    func decodeHelper(_ l: Calendar) -> Calendar {
+    func decodeHelper(_ l: Calendar) throws -> Calendar {
         let je = JSONEncoder()
-        let data = try! je.encode(l)
+        let data = try je.encode(l)
         let jd = JSONDecoder()
-        return try! jd.decode(Calendar.self, from: data)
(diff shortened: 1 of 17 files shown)


jmschonfeld:
@swift-ci please test

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -351,18 +370,19 @@ final class CalendarTests : XCTestCase {
...
+                    autoreleasepool {
+                        Self.validateOrdinality(expected, calendar: immutableCalendar, date: date)
+                    }
+                }
             }
         }
-        XCTAssertEqual(.success, group.wait(timeout: .now().advanced(by: .seconds(3))))
     }
-    #endif // (arch(x86_64) || arch(arm64)) && FOUNDATION_FRAMEWORK
+    #endif // _pointerBitWidth(_64) && FOUNDATION_FRAMEWORK
```
itingliu, line 383: I think with this change we can remove this `#if FOUNDATION_FRAMEWORK`. I'd be happy to do that in a follow up separately.

jmschonfeld: Sounds good! Yeah I'd prefer to do it in a separate change so that this is as close to behaviorally equivalent as possible in case there's any latent issues that testing on new platforms like Linux/Windows would pick up to reduce the risk of this change

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1392, opened 2025-06-30
Consolidate URL compatibility checks

jrflat:
`URL.compatibility4` is no longer needed and we can just use `URL.compatibility1` for the `url.deletingPathComponent().standardized` compatibility path instead.


jrflat:
@swift-ci please test

parkera (approved):

PR #1394, opened 2025-06-30
Make sure to free the malloced pointer even in case of throwing an error

parkera:
This is detected in our existing tests when running `leaks`.

Resolves rdar://154702045


parkera:
@swift-ci test

PR #1397, opened 2025-07-01
Update availability macros for macOS 15 deployment target

jmschonfeld:
This PR does the following:

- Removes all `FoundationPredicate`/`FoundationPredicateRegex` availability: since these are always available in the package there is no longer a distinction between these and `FoundationPreview` so they are no longer needed
- Replace all `FoundationPreview 0.x` availability with the "real" availability: all of these versions match OS versions lower than the deployment target of the package (macOS 15) so it does not have any impact on the availability of things within the package
- Update `FoundationSpan` to list macOS 26-aligned availability instead of future availability to match the availability of `Span` types in the standard library


jmschonfeld:
@swift-ci please test

PR #1399, opened 2025-07-01
Work towards leap month bug fixes

itingliu:
There is a bug in ICU that returns the wrong value for min/max range. Workaround that here.

Also fix the same issue in Gregorian calendar, where the returned value isn't sensible.

153548677


itingliu:
@swift-ci please test

PR #1402, opened 2025-07-02
Finish FoundationInternationalization swift-testing migration

jmschonfeld:
This finishes the swift-testing migration for FoundationInternationalization (and the entirety of swift-foundation). I'll followup with one final PR to cleanup out TestSupport module and remove our XCTest workarounds.

This also includes one change to the `ProcessInfo` tests that I noticed failed when run in Xcode but passed when run via the CLI since they use different test runner process names.


jmschonfeld:
@swift-ci please test

parkera (approved):

PR #1403, opened 2025-07-03
Cleanup swift-testing migration

jmschonfeld:
This is the final step in our swift-testing migration. Now that all tests use swift-testing we can:

- Remove all remaining imports of `XCTest` and `XCTest`-based helper functions
- Remove `typealias` declarations used to disambiguate between `FoundationEssentials` and `Foundation` now that the ambiguity is gone (due to the lack of an imported Foundation)
- Move the `TestSupport` module to the tests folder and remove its dependencies on `FoundationEssentials`/`FoundationInternationalization` to remove `FoundationEssentialsTests`'s dependency on `FoundationInternationalization`
- Fix or move a few tests in `FoundationEssentials` that depended on `FoundationInternationalization` and add a new test to validate that `FoundationInternationalization` upcalls are not enabled in `FoundationEssentialsTests`


jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -878,130 +878,6 @@ private struct GregorianCalendarTests {
         test(.yearForWeekOfYear, date, expectedStart: nil, end: nil)
     }
 
-    @Test func testDateInterval_DST() {
```
jmschonfeld, line 881: This test actually already existed in the `GregorianCalendarInternationalizationTests` suite, so it's just removed from here since it requires a non-GMT timezone

Tests/TestSupport/Utilities.swift
```diff
@@ -0,0 +1,266 @@
+//===----------------------------------------------------------------------===//
```
jmschonfeld, line 1: This file was just moved from the old `TestSupport` folder, but the XCTest helpers in it were removed

jmschonfeld:
@swift-ci please test

PR #1407, opened 2025-07-07
update expanded calendar support proposal

cjchapman:
renamed isAdhikaDay to isRepeatedDay

added DateComponents.swift section


itingliu (approved):

itingliu (approved):
There were indeed discussions previously about this unconventional name. I think this change and new name makes sense.

itingliu:
@swift-ci please test

PR #1410, opened 2025-07-08
String.pathExtension should allow "..name" file names but not ".."

jrflat:
Minor fix for `String.pathExtension` found when comparing the old `NSString` behavior. `String.pathExtension` returns "" for path extensions appended to "." and ".." file names. This behavior is in part to prevent `.deletingPathExtension()` from returning a path referencing the current or parent directory.

However, `String.pathExtension` currently returns "" for all file names that start with "..", so some valid file paths like "..name.txt" aren't receiving the correct behavior.

This PR makes it so we only disallow extensions that would leave "." and ".." file names when removed. I also renamed a variable to make it more clear we're working with the `UTF8View`.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Path.swift
+++ b/Sources/FoundationEssentials/String/String+Path.swift
@@ -224,14 +224,19 @@ extension String {
     }
 
     internal var pathExtension: String {
-        let lastComponent = lastPathComponent.utf8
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

parkera (approved):

jmschonfeld (approved):

PR #1411, opened 2025-07-09
Bin compat: use Platform.getFullExecutablePath instead of CommandLine.arguments.first for ProcessInfo.processName.

iCharlesHu:
The original `NSProcessInfo` implementation calls `_CFProcessPath` for `processName`, which `Platform.getFullExecutablePath` internally calls. We should maintain the same behavior for binary compatibility.


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu:
@swift-ci please test

PR #1422, opened 2025-07-10
formattedAttributedLeadingDotSyntax() test failed

itingliu:
This test and other tests testing the dot syntax assume that the left hand side of the comparison is using the same locale as the right side, in which case is `Locale.autoupdatingCurrent`. This further assumes that the state of the host machine remains unchanged throughout the tests.

Fix this by specifying the locale of the left and right side so they don't depend on the current locale settings.

Addresses 154994643


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1423, opened 2025-07-10
Limit the use of concurrent scope in unit tests

itingliu:
Many tests for the dot syntax assume that the left hand side of the comparison is using the same locale as the right side, in which case is `Locale.autoupdatingCurrent`. Previously we put them inside `usingCurrentInternationalizationPreferences` to ensure that the current locale isn't mutated concurrently while the tests are running. That works but we can also set the locale of both the left and right side to a specific locale so they don't depend on the current status.

Addresses 155437764


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1424, opened 2025-07-11
Fix or disable tests around accessing current locale

itingliu:
I believe these tests are failing because some other tests are mutating the global cached current locale or calendar.

We've identified most tests whose test expectations depend on the current locale. We've addressed them by either specifying a fixed locale on the tests, or guard them inside `usingCurrentInternationalizationPreferences`. But even if a test doesn't assert the result based on the value of the current locale, the test might still cause the cache to be populated concurrently at the time of accessing autoupdating and current locale, and subsequently break tests that expect a given value of the current locale.
    
Disable the tests for now while investigating a proper way to update the tests. These tests pass when building as part of the Darwin SDK, so leave them enabled there.
    
155476600


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1427, opened 2025-07-11
Update stringshims to always use clocale

jmschonfeld:
When updating the JSON encoder tests to swift-testing, we noticed that on Windows the test that calls `setlocale` (which now runs in parallel to other tests) was interfering with other JSON tests causing rare crashes. This updates our string shims to explicitly use the c locale rather than relying on a `NULL` value (which on Windows used the global current locale)


jmschonfeld:
@swift-ci please test

bnbarham:
Thanks @jmschonfeld!

kperryua (approved):
Thanks!

PR #1428, opened 2025-07-11
Prevent decimal_withFormat_localeDependent from relying on the current locale

jmschonfeld:
Updates this unit test to avoid relying on the current locale so that it won't fail if the locale changes to non-en_US in the middle of the test


jmschonfeld:
@swift-ci please test

itingliu (approved):
Thanks! I was going to clean these up too

PR #1430, opened 2025-07-14
Fix some force unwraps and type conversion issues in OpenStepPlist

parkera:
Fix a few force unwraps and type conversion issues for OpenStepPlist.


parkera:
@swift-ci test

PR #1432, opened 2025-07-14
Enable FoundationMacrosTests on Windows

jmschonfeld:


jmschonfeld:
@swift-ci please test

parkera (approved):

compnerd (approved):
If it builds, why not. However, I'd be curious to know how this was "fixed" (`/alternatename:` doesn't give the desired behaviour AFAIK).

jmschonfeld:
> If it builds, why not. However, I'd be curious to know how this was "fixed" (`/alternatename:` doesn't give the desired behaviour AFAIK).

@daveinglis mentioned that it was fixed by https://github.com/swiftlang/swift-package-manager/pull/8515. I was expecting this to fix the more general problem with test targets that depend on an executable, but I think macro tests are just a specific type of target in this category so I figured I'd give re-enabling this a shot and it looks like it builds

PR #1436, opened 2025-07-15
Update automergers for late release cycle

jmschonfeld:
This adds a new auto merger to the repo that merges forwards from the release/6.2 branch to main. We will use this during the later period of a release cycle (instead of the existing backwards auto merger from main to release/6.2) to include all PRs merged to release branches on main automatically to avoid the need for cherry-picks.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1438, opened 2025-07-16
Reinstate local definitions of `_overrideLifetime()`

glessard:
This partially reverts https://github.com/swiftlang/swift-foundation/pull/1400

By a stroke of bad timing, the Foundation framework cannot build without these because one of the compilers we require did not get updated on the expected schedule. We will remove these at a less critical time.


glessard:
@swift-ci please test

glessard:
We were able to wait until [the ](https://github.com/swiftlang/swift/pull/81224) landed everywhere necessary, making this change unneeded.

PR #1443, opened 2025-07-17
Disable `nsICUNumberFormatterCache` test

itingliu:
This test would fail if the cache is cleared while running the test. It could happen if the number of objects exceeds the internal limit.

For some reason we have not seen test failures when building for FOUNDATION_FRAMEWORK, so still leave it enabled there.

We should come back to re-write the test to be more reliable for all platforms regardless.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1444, opened 2025-07-17
Update current preference-using Date.RelativeFormatStyle tests

jmschonfeld:
`Date.RelativeFormatStyle` uses the current locale and calendar (which uses the default timezone) by default. There are a few unit tests that accidentally used one of these current values, or intentionally used them without guarding via `usingCurrentInternationalizationPreferences`. This updates `dateRelativeFormatConvenience` to be guarded with `usingCurrentInternationalizationPreferences` and explicitly specifies a locale/calendar/timezone in a few other missing places throughout the tests

Resolves rdar://155753882


jmschonfeld:
@swift-ci please test

PR #1445, opened 2025-07-17
Update the proposal for UTCClock to reflect the discussion surrounding it

phausler:
There were a few places where the examples left some lack of clarity of the proposed clock usages and behaviors. This refines those examples to a more concrete use case that avoids any confusion about the role associated with UTCClock and additionally clarifies the stance on the placement/homing of Date, the definition of timezones and the relationship to UTC, the naming of systemEpoch and finally a reflection of the proposal discussion out lasting the cutoff of 6.2.

### DIFF
--- a/Proposals/0027-UTCClock.md
+++ b/Proposals/0027-UTCClock.md
@@ -23,37 +23,37 @@ All three of the aforementioned clocks all have a concept of a starting point of
 
 In short, a new clock will be added: `UTCClock`. This clock will have its `Instant` type defined as `Date`. There will also be affordances added for calculations that account for the edge cases of [leap seconds](https://en.wikipedia.org/wiki/Leap_second) (which currently `Date` on its own does not currently offer any sort of mechanism either on itself or via `Calendar`). `Date` has facilities for expressing a starting point from an epoch, however that mechanism is not shared to `ContinuousClock.Instant` or `SuspendingClock.Instant`. All three types will have an added new static property for fetching the `epoch` - and it is suggested that any adopters of `InstantProtocol` should add a new property to their types to match this convention where it fits.
 
-Usage of this `UTCClock` can be illustrated by awaiting to perform a task at a given time of day. This has a number of interesting wrinkles that the `SuspendingClock` and `ContinousClock` wouldn't be able to handle. Example cases include where the deadline might be beyond a daylight savings time. Since `Date` directly interacts with `Calendar` it then ensures appropriate handling of these edges and is able to respect the user's settings and localization.
+Usage of this `UTCClock` can be illustrated by awaiting to perform a task at a given coordinated time. This has a number of interesting wrinkles that the `SuspendingClock` and `ContinousClock` wouldn't be able to handle. Example usages include server side coordination between machines or processes to execute work at a given time. In the following example a server workload is scheduled to synchronize at 18:00 every Sunday.
 
 ```swift
-let calendar = Calendar.current
-var when = calendar.dateComponents([.day, .month, .year], from: .now)
-when.day = when.day.map { $0 + 1 }
-when.hour = 8
-
-if let tomorrowMorning8AM = calendar.date(from: when) {
-  try await UTCClock().sleep(until: tomorrowMorning8AM)
-  playAlarmSound()
+func synchronization() async throws {
+  while true {
+    let nextSunday = DateComponents(hour: 18, minute: 0, weekday: 1)
+    if let when = Calendar(identifier: .iso8601).nextDate(after: Date(), matching: nextSunday, matchingPolicy: .nextTime) {
+      try await UTCClock().sleep(until: when)
+      try await synchronize()
+    }
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Proposals/0027-UTCClock.md
```diff
@@ -23,37 +23,37 @@ All three of the aforementioned clocks all have a concept of a starting point of
...
 
-The structure of the `UTCClock` is trivially sendable since it houses no specific state and has the defined typealias of its `Instant` as `Date`. The minimum feasible resolution of `Date` is 1 nanosecond (however that may vary from platform to platform where Foundation is implemented).
+The structure of the `UTCClock` is trivially sendable since it houses no specific state and has the defined typealias of its `Instant` as `Date`. The minimum feasible resolution of `Date` is 1 nanosecond. The sleep method may respond to host system time adjustments; e.g. if a system has it's time changed manually or if the system adjusts the time drift via ntp updates so the sleep method, unlike ContinuousClock, is not strictly monotonic. However, since UTC itself is the root definition of timezones[^timezones] there is no daylight savings time adjustment; any calculation for accounting to DST must be done before hand by interacting with the Calendar and deriving a Date accordingly.
```
itingliu, line 47:
```suggestion
The structure of the `UTCClock` is trivially sendable since it houses no specific state and has the defined typealias of its `Instant` as `Date`. The minimum feasible resolution of `Date` is 1 nanosecond. The sleep method may respond to host system time adjustments; e.g. if a system has its time changed manually or if the system adjusts the time drift via ntp updates so the sleep method, unlike ContinuousClock, is not strictly monotonic. However, since UTC itself is the root definition of timezones[^timezones] there is no daylight savings time adjustment; any calculation for accounting to DST must be done before hand by interacting with the Calendar and deriving a Date accordingly.
```

Proposals/0027-UTCClock.md
```diff
@@ -64,13 +64,13 @@ extension UTCClock: Clock {
 The extension of `Date` conforms it to `InstantProtocol` and adds one addition "near miss" of the protocol as an additional function that in practice feels like a default parameter. This `duration(to:includingLeapSeconds:)` function provides the calculation of the duration from one point in time to another and calculates if the span between the two points includes a leap second or not. This calculation can be used for historical astronomical data since the irregularity of the rotation causes variation in the observed solar time. Those points are historically fixed and are a known series of events at specific dates (in UTC)[^utclist].
```
itingliu, line 64: Is this part an outdated version of the previous? I don't see a `duration(to:includingLeapSeconds:)`. Should it just be `func duration(to other: Date)`? 

And the functionality to get the leap seconds between two would be covered by `public static func leapSeconds(from start: Date, to end: Date) -> Duration` below

Proposals/0027-UTCClock.md
```diff
@@ -104,10 +104,11 @@ This is a purely additive set of changes.
...
 
-It was considered to add a new `Instant` type instead of depending on `Date` however this was rejected since it would mean re-implementing a vast swath of the supporting APIs for `Calendar` and such. The advantage of this is minimal and does not counteract the additional API complexity.
+It was considered to add a new `Instant` type instead of depending on `Date` however this was rejected since it would mean re-implementing a vast swath of the supporting APIs for `Calendar` and such. The advantage of this is minimal and does not counteract the additional API complexity. This consideration did delve into the idea of moving Date to a lower scope and also changing it's storage, however due to the complexity of that; it is considered a non-goal of this proposal and squarely in the out of scope territory. The beneifit of scheduling and appropriate calculations with regards to leap seconds are valueable on their own and distinctly seperable tasks to any sort of re-orgnaization of Date et al.
```
itingliu, line 109: typo:
```suggestion
It was considered to add a new `Instant` type instead of depending on `Date` however this was rejected since it would mean re-implementing a vast swath of the supporting APIs for `Calendar` and such. The advantage of this is minimal and does not counteract the additional API complexity. This consideration did delve into the idea of moving Date to a lower scope and also changing its storage, however due to the complexity of that; it is considered a non-goal of this proposal and squarely in the out of scope territory. The benefit of scheduling and appropriate calculations with regards to leap seconds are valuable on their own and distinctly separable tasks to any sort of re-organization of Date et al.
```

itingliu (dismissed):

itingliu:
@swift-ci please test

Proposals/0027-UTCClock.md
```diff
@@ -23,37 +23,37 @@ All three of the aforementioned clocks all have a concept of a starting point of
...
+@available(FoundationPreview 6.3, *)
 public struct UTCClock: Sendable {
   public typealias Instant = Date
   public init()
 }
 
-@available(FoundationPreview 6.2, *)
+@available(FoundationPreview 6.3, *)
 extension UTCClock: Clock {
     public func sleep(until deadline: Date, tolerance: Duration? = nil) async throws
```
itingliu, line 58: Can we spell out the behavior of this function if `deadline < Date.now` in the comment?

PR #1447, opened 2025-07-19
Fallback to `argv[0]` for `ProcessInfo.processName` on platforms without specific APIs.

kateinoigakukun:
WASI does not provide a specific API to retrieve the process name, so we used `CommandLine.arguments.first` for `ProcessInfo.processName`. After #1411, `CommandLine.arguments.first` started using platform-specific APIs to retrieve the process name, but the fallback path with `argv[0]` was removed. This commit restores that fallback for platforms like WASI.


kateinoigakukun:
@swift-ci test

iCharlesHu (approved):

PR #1450, opened 2025-07-22
Handle 1-day weekend properly

itingliu:
India only has Sunday as their weekend; the other places (GB, US, etc) have two days. Update `Calendar.nextWeekend(startingAfter:direction:)` to handle this case.

153114732


itingliu:
@swift-ci please test

parkera (approved):

PR #1451, opened 2025-07-23
Fix Platform.getFullExecutablePath for platforms with no file system

iCharlesHu:
`CommandLine` is not available on platforms with no file system. Return `nil` in these cases

Resolves: rdar://156436329


iCharlesHu:
@swift-ci please test

jmschonfeld (approved):

PR #1459, opened 2025-07-31
Keep executable path

etcwilde:
The Swift driver uses the location of the driver to search for other binaries including the swift-frontend, legacy driver, etc... Dropping the filepath means that we lose this information and search from the current working directory instead of where the driver exists, which then results in the driver failing.

This broke the FreeBSD job: `error: unknown or missing subcommand 'swiftc-legacy-driver'`


etcwilde:
@swift-ci please test

PR #1466, opened 2025-08-12
Update example to assign to deadlineTracker (if intended)

noppefoxwolf:
Hello @chloe-yeo, thank you for your proposal about progress. I really appreciate your work on this.
I may be misunderstanding, but is the second assignment intended to target deadlineTracker rather than overall?


chloe-yeo:
I believe you've caught this correctly, thanks!

chloe-yeo:
@swift-ci please test

chloe-yeo (approved):
Looks good to me! Accepting this

PR #1467, opened 2025-08-12
Prevent encodingDateFormatted() from using the current internationalization preferences

jmschonfeld:
This test may run at the same time as other tests that change the current locale/calendar/timezone. To ensure that doesn't interfere with this test, we should explicitly set the locale/calendar/timezone of the date formatter in use.


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1469, opened 2025-08-12
Cleanup FoundationMacros availability checks

jmschonfeld:
We no longer have any scenarios where we need to build the macro for a host older than macOS 14 so we can cleanup some of the availability checks in the macros target to assume a macOS 14 deployment target


jmschonfeld:
@swift-ci please test

itingliu (approved):

Issue #1455, opened 2025-07-30
Raise Decimal to a power with FoundationEssentials


zcohan:
### The problem
The following math operations can be performed on `Decimal` with FoundationEssentials:

```swift
import FoundationEssentials

let number: Decimal = 2.0
let added = number + number
let subtracted = number - number
let multiplied = number * number
let divided = number / number
```

However, raising a`Decimal` to a power requires Foundation:

```swift
import Foundation

let number: Decimal = 2.0
let raised = pow(number, 2)
```

### The solution

- The power operation should be available in FoundationEssentials alongside the basic decimal arithmetic operations.
- An exponentiation operator could be offered here to be consistent with the existing inline operator support (Python uses `**`, for example).

### Additional context
- My app is particularly sensitive to binary size, so it's unfortunate that I have to import Foundation on non-Darwin platforms just for this _somewhat essential_ function. 
- `func pow(_ x: Decimal, _ y: Int) -> Decimal` in Foundation appears to be a wrapper around `NSDecimalPower()`. 
- The `NSDecimalX()` functions are part of Foundation, not FoundationEssentials.
- `NSDecimalAdd`, `NSDecimalSubtract`, `NSDecimalMultiply`, and `NSDecimalDivide` are available in FoundationEssentials via operators, but an equivalent for `NSDecimalPower()` is not offered.

itingliu:
> `func pow(_ x: Decimal, _ y: Int) -> Decimal` in Foundation appears to be a wrapper around NSDecimalPower().

Where did you see this? I'm seeing that it is implemented natively https://github.com/swiftlang/swift-foundation/blob/a8bee5bfc71210168fa1b973fb1a1deb8bde2047/Sources/FoundationEssentials/Decimal/Decimal%2BCompatibility.swift#L86

`__NSDecimalPower()` also actually calls into the shared `func _power()` https://github.com/swiftlang/swift-foundation/blob/a8bee5bfc71210168fa1b973fb1a1deb8bde2047/Sources/FoundationEssentials/Decimal/Decimal%2BCompatibility.swift#L236


But yes I agree that it's reasonable to expose this in `FoundationEssentials`

itingliu:
Currently [swift-corelibs-foundations defines `public func pow()`](https://github.com/swiftlang/swift-corelibs-foundation/blob/33c045a8f6e7ec7d54d39b39d82341959e9dd1c6/Sources/Foundation/Decimal.swift#L42)

[...]

PR #1471, opened 2025-08-14
Expose Decimal.pow() as a public API on FoundationEssentials

iCharlesHu:
Resolves: https://github.com/swiftlang/swift-foundation/issues/1455

### DIFF
--- a/Sources/FoundationEssentials/Decimal/Decimal+Compatibility.swift
+++ b/Sources/FoundationEssentials/Decimal/Decimal+Compatibility.swift
@@ -80,7 +80,6 @@ extension Decimal : _ObjectiveCBridgeable {
 // MARK: - Bridging code to C functions
 // We have one implementation function for each, and an entry point for both Darwin (cdecl, exported from the framework), and swift-corelibs-foundation (SPI here and available via that package as API)
 
-#if FOUNDATION_FRAMEWORK
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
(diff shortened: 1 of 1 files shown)


jmschonfeld (approved):

Sources/FoundationEssentials/Decimal/Decimal+Compatibility.swift
```diff
@@ -80,23 +80,13 @@ extension Decimal : _ObjectiveCBridgeable {
...
 
-#if FOUNDATION_FRAMEWORK
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 public func pow(_ x: Decimal, _ y: Int) -> Decimal {
     let result = try? x._power(
         exponent: y, roundingMode: .plain
     )
     return result ?? .nan
 }
-#else
```
parkera, line 91: Is `pow` available as API on `Decimal` itself instead of just via the free function?

itingliu: My read is that this has always been a free function, both on Darwin and non-Darwin. This PR just unifies the code and makes it available for FoundationEssentials too.

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Decimal/Decimal+Compatibility.swift
iCharlesHu: Ditto

PR #1472, opened 2025-08-15
Update default attribute scope framework paths

jmschonfeld:
Updates the default attribute scope framework paths to match where the symbols are defined to improve performance of the `dlsym` calls.


jmschonfeld:
@swift-ci please test

PR #1473, opened 2025-08-18
NotificationCenter.messages() should return Sendable type

cthielen:
`NotificationCenter.messages()` returns an opaque type that is Sendable but is not marked as such. This change updates the API to note the Sendable conformance.


cthielen:
@swift-ci please test

FranzBusch (approved):

Issue #1173, opened 2025-02-14
Bug: #Predicate macro incorrectly fails for ‘final’ classes


bdkjones:
# Summary

I've discovered that Foundation's `#Predicate` fails when a class is marked as `final`. Here's a very simple test case:

```
final class Master {
    var id: UUID = UUID()
}

let u = UUID()
let p = #Predicate<Master>{ $0.id == u }
```

That expands to: 

```
Foundation.Predicate<Master>({
    PredicateExpressions.build_Equal(
        lhs: PredicateExpressions.build_KeyPath(
            root: PredicateExpressions.build_Arg($0),
            keyPath: \.id
        ),
        rhs: PredicateExpressions.build_Arg(someID)
    )
})
```

It compiles correctly, but at runtime crashes on the `build_KeyPath` line: `Thread 1: Fatal error: Predicate does not support keypaths with multiple components`

This is obviously not correct; there is only one component in the KeyPath. If you remove `final` from the class declaration, it works fine.

I see where this check exists in Foundation (https://github.com/swiftlang/swift-foundation/blob/2f4f5b80cfb4cd4540e1e75bc07dff4a54a5ea47/Sources/FoundationEssentials/Predicate/KeyPath%2BInspection.swift#L50-L88) but I don't see why `final` should affect whether a `UUID` property has an offset.

## Any Value

The behavior is not limited to `UUID`. I can reproduce it with a `String`, `Int`, and other values.

## Speculation

I can’t find any information on how the presence of `final` affects the memory layout of a class such that `offset(of:)` would fail. I thought maybe the issue affects only types that are capable of storing their value without allocation (i.e. tagged pointers), but again I can’t see how `final` would affect that.

## Intermittent 

IF a class has a certain combination of properties, this issue does not always manifest. For example, Predicates ran just fine here:

```
final class Master
{
    var title: String = “”
    var numbers: [Int] = []
    var is enabled: Bool = false
    var percentage: Double = 57.5
}
```

But add `var id: UUID = UUID()` and that fails until `final` is removed. Given the `offset(of:)` check, I figured the issue might have something to do with the exact way properties are aligned and packed into the storage for the class. But that’s just a guess.

## Environment

Xcode 16.2.

Since `#Predicate` isn’t available in Playgrounds, I created a new project from the “command line application” template in Xcode. I changed no build settings.

jmschonfeld:
[...]

PR #1475, opened 2025-08-19
Fix crash when using keypaths on final classes in Predicates

jmschonfeld:
This predicate code validates that the provided `KeyPath` only contains a single component. Originally, for stored property components, this code checked the payload value to see if it was an inline offset or the out-of-line sentinel defined by the [KeyPath ABI documentation](https://github.com/swiftlang/swift/blob/main/docs/ABI/KeyPaths.md) to determine if the buffer size matches the expected single component buffer size. However, this document was determined to be incorrect and there are actually  [multiple sentinel values](https://github.com/swiftlang/swift/blob/d837d6052cf5403644e14956ee4376454ad633bd/stdlib/public/SwiftShims/swift/shims/KeyPath.h#L70-L77) defined. https://github.com/swiftlang/swift-foundation/pull/216 attempted to correct this by switching to `MemoryLayout.offset(of:)` to check for a single stored property component. However, `MemoryLayout.offset(of:)` is also not an appropriate way to check for a single stored property component as not all single stored components have offsets (for example, a property in a final class).

This changes the implementation back to the original style of check before https://github.com/swiftlang/swift-foundation/pull/216 but with a corrected check for whether the offset is stored inline (based on the existence of multiple sentinel values and the max inline offset defined [here](https://github.com/swiftlang/swift/blob/d837d6052cf5403644e14956ee4376454ad633bd/stdlib/public/SwiftShims/swift/shims/KeyPath.h#L67-L68)). This fixes the regression caused for final class properties while still ensuring that we respect other sentinel values as well.

Resolves https://github.com/swiftlang/swift-foundation/issues/1173

### DIFF
--- a/Sources/FoundationEssentials/Predicate/KeyPath+Inspection.swift
+++ b/Sources/FoundationEssentials/Predicate/KeyPath+Inspection.swift
@@ -19,6 +19,8 @@ extension UInt32 {
     private static var COMPUTED_COMPONENT_PAYLOAD_ARGUMENTS_MASK: UInt32 { 0x0008_0000 }
     private static var COMPUTED_COMPONENT_PAYLOAD_SETTABLE_MASK: UInt32 { 0x0040_0000 }
     
+    private static var STORED_COMPONENT_PAYLOAD_MAXIMUM_INLINE_OFFSET: UInt32 { 0x007F_FFFC }
+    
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

Azoy (approved):

PR #1476, opened 2025-08-20
Add predicate keypath assertion exit tests

jmschonfeld:
As a followup to https://github.com/swiftlang/swift-foundation/pull/1475, this adds exit tests to validate situations when we _should_ be preventing usage of a particular keypath (specifically, when a keypath is multiple components or when it represents optional chaining or subscripts).

Adding these tests uncovered that the default value of `restrictArguments` in `_validatedForPredicateUsage` was incorrect and that we weren't actually validating that subscripts with arguments can't be used. In addition to adding these tests, this fixes the default value of this argument so that in cases where we don't expect arguments we enforce that they cannot be used.


jmschonfeld:
@swift-ci please test

PR #1477, opened 2025-08-21
Improve temporary file location for atomic writes

jmschonfeld:
When `_amkrtemp` is available (`FOUNDATION_FRAMEWORK` cases) we should use this before trying `mktemp` to create the temporary file location to preserve characteristics of the final destination like sandbox extensions. This patch adds a new use of `_amkrtemp` first and falls back to the existing `mktemp` if that fails to preserve existing behavior for failure cases. It also adds some infrastructure to include a debug description in file path errors so we can indicate the original `_amkrtemp` failure if both `_amkrtemp` and `mktemp` fail.


jmschonfeld:
@swift-ci please test

kperryua (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:

PR #1478, opened 2025-08-22
Update status and proposal name

itingliu:


itingliu:
@swift-ci please test

PR #1479, opened 2025-08-22
Fix a corner case where createDirectory would fail on windows

daveinglis:
The //?/ prefix needs to be added to directories if the length is
> MAX_PATH - 12, however the PATHCCH_ALLOW_LONG_PATHS to
PathAllocCanonicalize would only do that if the path was > MAX_PATH. This change uses PATHCCH_ENSURE_IS_EXTENDED_LENGTH_PATH, which forces it on all the time.  Just needed to remove the prefix in a few places as it was making it way out into tests as this is suppose to be transparent to the user.


daveinglis:
@swift-ci please test

daveinglis:
@swift-ci please test

daveinglis:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -271,36 +271,49 @@ extension _FileManagerImpl {
...
-                }
-                switch errorCode {
-                case ERROR_SUCCESS:
-                    if let attributes {
-                        try? fileManager.setAttributes(attributes, ofItemAtPath: path)
+            // Create intermediate directories recursively
+            func createDirectoryRecursively(at directoryPath: String) throws {
+                // Check if directory already exists
+                var isDirectory: Bool = false
+                if fileExists(atPath: directoryPath, isDirectory: &isDirectory) {
```
jmschonfeld: I think this could open us up to TOC-TOU issues. Since we're already handling `ERROR_ALREADY_EXISTS` below when we try to create the directory, can we just elide this check and let the error handling of `CreateDirectoryW` take care of the rest? (I know the error handling below also feels TOC-TOU-like, but given we can't know if the pre-existing result was a directory in the same call, it seems better to have a TOC-TOU issue when deciding whether to throw the error than deciding whether to create the directory to begin with)

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -139,6 +139,9 @@ extension FileManager {
     fileprivate func _shouldRemoveItemAtPath(_ path: String) -> Bool {
         var delegateResponse: Bool?
         if let delegate = self.safeDelegate {
+#if os(Windows)
+            let path = path.removingNTObjectNamespacePrefix()
```
jmschonfeld: This same pattern is used in a lot of delegate calls such as `_shouldCopyItemAtPath`, `_shouldLinkItemAtPath`, and `_shouldMoveItemAtPath`. If it's possible for the file system to contain a file that uses such a path, should we standardize all of these to remove the NT prefix before calling the delegate?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1195,8 +1193,13 @@ private struct FileManagerTests {
 
             #expect((cwd + "/" + dirName + "/" + "lnk").resolvingSymlinksInPath == (cwd + "/" + dirName + "/" + fileName).resolvingSymlinksInPath)
 
-            #expect(throws: Never.self) { try fileManager.createDirectory(at: URL(fileURLWithPath: dirName + "/" + "subdir2"), withIntermediateDirectories: false) }
-            #expect(throws: Never.self) { try fileManager.createDirectory(at: URL(fileURLWithPath: dirName + "/" + "subdir2" + "/" + "subdir3"), withIntermediateDirectories: false) }
+            #expect(throws: Never.self) { try fileManager.createDirectory(at: URL(fileURLWithPath: dirName + "/" + "subdir1"), withIntermediateDirectories: false) }
```
jmschonfeld: This creation of `<dirName>/subdir1` looks like it was moved from the above, but line 1181 still uses this directory. Was this an intentional move, or can this be moved back to ensure that the expectation on line 1181 (changing the CWD to `<dirName>/subdir1`) still runs when subdir1 exists?

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -79,6 +80,32 @@ extension String {
             }
         }
     }
+    /// Removes the Windows NT object namespace prefix if present.
+    /// The \\??\ prefix is used by Windows NT for device paths and may appear
+    /// in paths returned by system APIs. This method provides a clean way to
+    /// normalize such paths to standard format.
+    ///
+    /// - Returns: A string with the NT object namespace prefix removed, or the original string if no prefix is found.
+    package func removingNTObjectNamespacePrefix() -> String {
```
compnerd: I think that we should call this `removingNTDevicePathPrefix`. The NT Object prefix would be `\??\`.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -497,9 +510,14 @@ extension _FileManagerImpl {
         // This is solely to minimize the number of allocations and number of bytes allocated versus starting with a hardcoded value like MAX_PATH.
         // We should NOT early-return if this returns 0, in order to avoid TOCTOU issues.
         let dwSize = GetCurrentDirectoryW(0, nil)
-        return try? FillNullTerminatedWideStringBuffer(initialSize: dwSize >= 0 ? dwSize : DWORD(MAX_PATH), maxSize: DWORD(Int16.max)) {
+        let cwd = try? FillNullTerminatedWideStringBuffer(initialSize: dwSize >= 0 ? dwSize : DWORD(MAX_PATH), maxSize: DWORD(Int16.max)) {
             GetCurrentDirectoryW(DWORD($0.count), $0.baseAddress)
         }
+        
+        // Handle Windows NT object namespace prefix
+        // The \\??\ prefix is used by Windows NT for device paths and may appear
```
jakepetroules: The prefix is spelled `\\?\` (you have an extra question mark)

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: Or maybe just "removingNTPathPrefix", more similar to the underlying API.

If we can trust [Wine's implementation](https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/kernelbase/path.c#L881) to match Windows, it looks like this removes the following prefixes:
- [`\\?\UNC\`](https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/kernelbase/path.c#L144)
- [`\\?\`](https://gitlab.winehq.org/wine/wine/-/blob/master/dlls/kernelbase/path.c#L149)

There's also the device path prefix `\\.\` which PathCchStripPrefix does _NOT_ handle - I checked - and I'm not sure is relevant or would be correct to strip here.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
daveinglis: yep, that makes sense, will update.

Sources/FoundationEssentials/FileManager/FileOperations.swift
daveinglis: ya, I thought about this but I couldn't see how any of the other delegates would get a prefixed path since none of them passes it in, and I didn't want to add any extra unnecessary overhead.  If you feel strongly Im fine with updating the other delegate calls.

Sources/FoundationEssentials/String/String+Internals.swift
daveinglis: will update

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
daveinglis: good catch, I think I move one too many lines, will fix

Sources/FoundationEssentials/FileManager/FileOperations.swift
jmschonfeld: Ah I see - that's one part that I wasn't sure of. In what cases do we create paths with the new prefix? Is it just implementations that happen to call through `withNTPathRepresentation` directly (is that why remove file is different, since it calls the delegate with the NTPath representation instead of the result of `withUnsafeFileSystemRepresentation`)?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
daveinglis: so Im rethinking this....  This first check is recursively walking the parent looking for where to start creating the folders, without it, it would always walk to the root, creating the folders from there, this seems it could be quite the performance hit if this first check is removed, would it not?

Sources/FoundationEssentials/FileManager/FileOperations.swift
daveinglis: ya, looking at this again, I may be able to fix up the caller so the correct path is aways passed.

daveinglis:
@swift-ci please test

daveinglis:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -66,7 +66,8 @@ extension String {
                 // 2. Canonicalize the path.
                 // This will add the \\?\ prefix if needed based on the path's length.
                 var pwszCanonicalPath: LPWSTR?
-                let flags: ULONG = PATHCCH_ALLOW_LONG_PATHS
+                // Alway add the long path prefix since we don't know if this is a directory.
+                let flags: ULONG =  PATHCCH_ENSURE_IS_EXTENDED_LENGTH_PATH
```
jakepetroules, line 70: Do we not also need the PATHCCH_ALLOW_LONG_PATHS flag anymore?

@daveinglis  Also, the same code is copied to:
- swift-system
- swift-build
- swift-subprocess

There are also variations in swift-tools-support-core and swift-testing.

Can you open PRs to make the change there as well?

daveinglis: The PATHCCH_ALLOW_LONG_PATHS is not needed,  PATHCCH_ENSURE_IS_EXTENDED_LENGTH_PATH will always prepend the `//?/`

daveinglis: So I saw the swift-build version, but it's never used in the context of creating a directory so should work fine as is.  I can look at the other packages but if the usage is never used with creating directories do we still need it changed?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jmschonfeld: The previous implementation didn't have this check, right (just confirming it's not a performance regression from a previous release to remove it)? I agree it would be more work, but it seems like that's probably correct work to do since the system doesn't have an API that does this in one call atomically, right? As unfortunate as performing a bit more work is, I think minimizing the potential security issues that typically come with TOC-TOU discrepancies is important here even if it means we need to do more work because we can't check for the optimal case first.

daveinglis:
@swift-ci please test

Sources/FoundationEssentials/String/String+Internals.swift
jakepetroules: > So I saw the swift-build version, but it's never used in the context of creating a directory so should work fine as is. I can look at the other packages but if the usage is never used with creating directories do we still need it changed?

You never know when it might be used in _new_ contexts, and because of cargo-culting I think we want to ensure that the most-correct implementation is kept in sync across all places it's copied to.

Ideally we'd have _one_ function that everything could call into, but not a great place for this sort of SPI...

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
daveinglis: @jmschonfeld Ok, I have updated this to just use the CreateDirectoryW result to determine if it already exists walking done the tree, then back up once it exists or is created.

Sources/FoundationEssentials/String/String+Internals.swift
daveinglis: ok, I will start putting up some PR for the other version so that all match.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
daveinglis: @jmschonfeld Are we good with this now?

jmschonfeld (approved):

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jmschonfeld: Thanks this LGTM now (but would appreciate a double check on this from @compnerd before merging just to make sure everything else looks good)

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -271,36 +271,48 @@ extension _FileManagerImpl {
...
+                    // Create this directory
+                    guard CreateDirectoryW(pwszPath, &saAttributes) else {
+                        let lastError = GetLastError()
+                        if lastError == ERROR_ALREADY_EXISTS {
+                            var isDir: Bool = false
+                            if fileExists(atPath: directoryPath, isDirectory: &isDir), isDir {
+                                return // Directory now exists, success
+                            }
+                        } else if lastError == ERROR_PATH_NOT_FOUND {
+                            let parentPath = directoryPath.deletingLastPathComponent()
```
jrflat, line 289: `.deletingLastPathComponent()` will replace all `\` with `/` in the path, but I think it's OK since the next call to `.withNTPathRepresentation` will call `GetFullPathNameW` and normalize them back to `\`. I was curious if this might cause issues if someone passes a path that already has a `\\?\` prefix to `.createDirectory(atPath:)`, which would then have parent paths of the form `//?/...`, but I ran some simple tests and it still seems to work OK. We could add a `.replacing(._slash, with: ._backslash)` to the end to be safe.

daveinglis:
@swift-ci please test

tshortli:
Reverted by https://github.com/swiftlang/swift-foundation/pull/1512 because this broke the Windows build for Swift CI.

Issue #1482, opened 2025-08-27
`AnyKeyPath._validateForPredicateUsage` should validate the keypath contains no arguments


jmschonfeld:
KeyPaths with computed components can contain arguments if the component is a subscript component. `PredicateExpressions.KeyPath` does not support subscript keypaths (the subscript operator is used instead) and so `_validateForPredicateUsage` should validate that a computed component does not contain arguments. However, arguments are also used to pass generic constraints (in the case where a keypath is rooted on a generic type, for example). When validating arguments don't exist, we need to ensure that we do allow cases where all arguments are generic arguments rather than concrete values passed to a subscript.

This distinction should be able to be determined by examining the witness table pointer in the computed component. For KeyPaths with actual argument values for subscripts, the witness table pointer will point to a concrete witness table for the value(s). However, if only generic arguments are used, the witness table pointer will be the global `swift_keyPathGenericWitnessTable`. We should allow arguments iff. the witness table pointer is NULL or is a pointer to `swift_keyPathGenericWitnessTable`.

PR #1483, opened 2025-08-27
Fix crash when creating predicate with generic root type

jmschonfeld:
`KeyPath`s used in predicates with a computed component can contain two types of arguments:

1. Concrete values passed as parameters to subscripts
2. Generic arguments used to satisfy generic constraints of a property

`Predicate` can support generic arguments, but not concrete values passed to subscripts (the subscript predicate expression should be used instead). When originally writing this code, I had conflated the two concepts as one and introduced the `restrictArguments` flag to work around cases where code accidentally asserted eagerly for only generic arguments. However, I also accidentally defaulted the argument to `false` so the assertion never ran. As part of https://github.com/swiftlang/swift-foundation/pull/1476 I set the default value to `true` correcting this mistake. Unfortunately, since the assertion conflates the two types of arguments, the assertion trips incorrectly in some existing client code (`Predicate`s created where the root type is generic and therefore property access provides a generic argument in the key path). For now, to fix this regression, this removes the `restrictArguments` parameter and removes the assertion altogether. In the long term, we should followup with a fix that adds this assertion back with improved validation logic not based on a `restrictArguments` parameter but rather based on other contents in the keypath to distinguish between the two types of arguments (tracked by https://github.com/swiftlang/swift-foundation/issues/1482 - I think this is something we should do as a followup since its implementation requires more testing and either requires a new C shim or some even more hacky workarounds).

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
+++ b/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
@@ -310,7 +310,7 @@ extension PredicateCodableConfiguration {
             guard root == rootReflectionType.partial, let constructed = constructor(rootReflectionType.genericArguments) else {
                 return nil
             }
-            constructed._validateForPredicateUsage(restrictArguments: false)
+            constructed._validateForPredicateUsage()
             return constructed
         }
     }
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

Azoy (approved):

PR #1484, opened 2025-08-28
FoundationEssentials: try to make the file atomic behaviour more robust

compnerd:
We have observed `ERROR_ACCESS_DENIED` in CI on
`SetRenameInformationFile`. Try to make the path more robust by first performing a kernel based rename with POSIX semantics. This requires Windows 10 1809+. If that is unsuccessful, verify that attributes are not to blame.

A failure may still occur if the file is on a different volume. In such a case, fallback to the `MoveFileExW` operation to perform a copy + delete operation.

Hopefully this should make the implementation more robust to failures.


Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -19,6 +19,8 @@ internal import DarwinPrivate // for VREG
 
 internal import _FoundationCShims
 
+import WinSDK
```
jmschonfeld: This import already exists below in a `os(Windows)` block

parkera: Does this need `#if os(Windows)`?

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -387,44 +389,112 @@ private func writeToFileAux(path inPath: PathOrURL, buffer: UnsafeRawBufferPoint
...
+        }
+        defer {
+            if hFile == INVALID_HANDLE_VALUE { return }
+            _ = CloseHandle(hFile)
         }
 
-        writeExtendedAttributes(fd: fd, attributes: attributes)
+        try inPath.path.withNTPathRepresentation { pwszPath in
+            let cbSize = wcslen(pwszPath) * MemoryLayout<WCHAR>.size
+            let dwSize = DWORD(MemoryLayout<FILE_RENAME_INFO>.size + cbSize)
```
jmschonfeld: Here the allocation size for the name (`cbSize`) doesn't contain space for the terminating null byte, but below when we call `wcscpy_s` into `pInfo?.pointee.FileName` we provide the length `wcslen(pwszPath) + 1` to include the null terminating byte. Should `cbSize` have a `+ 1` to include the null byte in the allocation size? My rough read is that this over-writes the buffer by one scalar for the null byte.

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -387,44 +389,112 @@ private func writeToFileAux(path inPath: PathOrURL, buffer: UnsafeRawBufferPoint
...
-        _close(fd)
-        fd = -1
+                if !SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize) {
+                    let dwError = GetLastError()
+                    if dwError == ERROR_ACCESS_DENIED {
+                        let dwAttributes = GetFileAttributesW(pwszPath)
+                        if dwAttributes > 0,
+                                dwAttributes & (FILE_ATTRIBUTE_READONLY | FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM) != 0 {
+                            // The target file is read-only, hidden or a system file. Remove those attributes and try again.
+                            if SetFileAttributesW(pwszPath, dwAttributes & ~(FILE_ATTRIBUTE_READONLY | FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM)) {
```
jmschonfeld: For my context, could you elaborate a bit more on why this behavior is here? It seems a bit counterintuitive to me that if the client tries to write to a file they don't have access to write to (because it's read-only for example) that we'd change the attributes of the file to make it writeable first. I think I'd expect that if that were the case, we should error and instead require that the client make this file writeable first. Is there a particular scenario we're trying to account for here in which the client perceives the location as writeable even if it isn't?

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -387,44 +389,112 @@ private func writeToFileAux(path inPath: PathOrURL, buffer: UnsafeRawBufferPoint
...
-        try auxPath.withNTPathRepresentation { pwszAuxiliaryPath in
-            guard MoveFileExW(pwszAuxiliaryPath, pwszPath, MOVEFILE_COPY_ALLOWED | MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH) else {
-                let dwError = GetLastError()
-                _ = DeleteFileW(pwszAuxiliaryPath)
-                throw CocoaError.errorWithFilePath(inPath, win32: dwError, reading: false)
+                    _ = CloseHandle(hFile)
+                    hFile = INVALID_HANDLE_VALUE
+
+                    // The move is across volumes.
+                    guard MoveFileExW(pwszAuxiliaryPath, pwszPath, MOVEFILE_COPY_ALLOWED | MOVEFILE_REPLACE_EXISTING) else {
```
jmschonfeld, line 468: The previous version of this call also set `MOVEFILE_WRITE_THROUGH`, should we preserve that here?

jmschonfeld (commented):
Could we also add a unit test for this? Even if the unit test can't reliably cause this issue 100% of the time, the example that @bnbarham mentioned in https://github.com/swiftlang/swift/issues/83606#issuecomment-3229837312 might be good to add as a test (but rewritten to use Swift concurrency instead of dispatch)

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: Yes, it does, but in fact, should not be needed (it is imported below). This was me trying to get something to quickly work.

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: Ah, right, the count is off by 1, thanks!

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: The scenario in my mind is that one of these attributes might be in use as a temporary thing while another thread is performing the operation. I'm happy to drop it and see if it occurs in practice though. It would help simplify the code quite a bit.

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: Right, I had originally put write through, but I'm not sure if that is the right thing anymore. That basically is equivalent to calling `flush` on the file, and waiting for that synchronously. The new variant performs the rename though the file may not be flushed to disk. I'm not entirely sure what the desired behaviour is.

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: I'm dropping this for now, I think that we can get away with it.

compnerd:
How do we want to integrate the test case that @bnbarham and I were using?

Sources/FoundationEssentials/Data/Data+Writing.swift
jmschonfeld: Good question. The Linux/Darwin equivalent that we use as `rename` so ideally we'd opt for whatever behavior is most similar to `rename`. I don't know offhand what the answer is there (I didn't find anything particularly specific in the man page) but I can try to find out what the semantics are of `rename` to match here

bnbarham:
> How do we want to integrate the test case that @bnbarham and I were using?

How about something like the following in `DataIOTests`?
```
  @Test
  func atomicWrite() async throws {
    let data = generateTestData()

    await withThrowingTaskGroup(of: Void.self) { group in
      for _ in 0..<8 {
        group.addTask { [url] in
          #expect(throws: Never.self) {
            try data.write(to: url, options: [.atomic])
          }
        }
      }
    }

    let readData = try Data(contentsOf: url, options: [])
    #expect(readData == data)
  }
```

Sources/FoundationEssentials/Data/Data+Writing.swift
compnerd: From a quick check, `rename` doesn't flush, the data is queued up in the object cache.

bnbarham:
> is that still the desire?

I'd actually like to get this into release/6.2.0 if we think it's safe enough for the initial 6.2 release. But definitely release/6.2 if not. The issue this fixes in swift-driver is actually very likely to happen, assuming that any command in a build ends up needing response files.

compnerd:
Let me cherry-pick this up to the release/6.2 branch.

airspeedswift (approved):

compnerd:
https://github.com/swiftlang/swift-foundation/pull/1485 is the 6.2 cherry-pick; I don't have a strong opinion on which way we handle this (merge this and merge the other one or merge the other one and let the auto-merger handle the rest).

compnerd:
@swift-ci please test

jmschonfeld:
Closing this PR as this has been merged to `main` via a merge forward from `release/6.2`: https://github.com/swiftlang/swift-foundation/pull/1489

PR #1485, opened 2025-08-29
[6.2] Use `SetFileInformationByHandle` for rename in atomic `Data.write` on Windows

compnerd:
We have observed `ERROR_ACCESS_DENIED` in CI on
`SetRenameInformationFile`. Try to make the path more robust by first performing a kernel based rename with POSIX semantics. This requires Windows 10 1809+. If that is unsuccessful, verify that attributes are not to blame.

A failure may still occur if the file is on a different volume. In such a case, fallback to the `MoveFileExW` operation to perform a copy + delete operation.

Hopefully this should make the implementation more robust to failures.


compnerd:
@swift-ci please test

parkera:
This is the branch for 6.2.1 and beyond, correct?

parkera:
Oh, 6.2.2 and beyond - this looks fine to me then.

PR #1490, opened 2025-09-03
[wasm] Use pthread mutex for wasm32-unknown-wasip1-threads target

kateinoigakukun:
`wasm32-unknown-wasip1-threads` target provides access to basic pthread APIs, so we should use it in LockedState


kateinoigakukun:
@swift-ci test

MaxDesiatov (approved):

jmschonfeld (approved):

PR #1491, opened 2025-09-03
TimeZone, Calendar, and Locale disagree on how `.current` is serialized

jmschonfeld:
There's been a handful of changes over the years in this area, but the situation we have today is that `TimeZone`, `Calendar`, and `Locale` all have different behaviors for how they encode `.current`. All 3 mostly agree that `.autoupdatingCurrent` is always encoded as as sentinel value becoming the `.autoupdatingCurrent` at decode time, and that fixed values are decoded simply as fixed values (Calendar doesn't do the latter sometimes). However:

- `TimeZone` always encodes `.current` as a fixed value. When decoded it is the same time zone that was encoded regardless of `.current` at decode time
- `Locale` always encodes `.current` as a sentinel value. When decoded, it takes the value of whatever `.current` is at decode time regardless of the locale that was encoded
- `Calendar` encodes a sentinel current value if the calendar to be encoded is equivalent to the current locale (i.e. if it has the same identifier/properties). This means that `.current` is always encoded as a sentinel value like `Locale`, but some fixed values are also encoded as a `.current` sentinel

The desired behavior is what `TimeZone` does today - `.current` represents a snapshot in time and we should encode that fixed snapshot unless `.autoupdatingCurrent` is used (in which case it should update at decode time). This is trickier for `Locale` because `Locale` contains extra user preferences that are persisted neither in the identifier nor other archive keys (which is why it is not its behavior today). This means encoding `Locale.current` as a fixed locale with just an identifier results in lost information due to lost preferences.

To fix this we do two things:
- `Locale` now persists preferences in the archive (when present) alongside the identifier (to prevent data loss) and `.current` is no longer decoded as a sentinel to match `TimeZone`
- `Calendar` no longer decodes `.current` as a sentinel value to match `TimeZone` (and now `Locale`)

With this change, we always decode a `Calendar` that was encoded as `.current` as a fixed Calendar based on the encoded values. We still encode the `.current` key to preserve the preexisting behavior when decoding on an older runtime, but newer runtimes will ignore the `.current` sentinel and decode the various keys.

This also adds unit tests to each type that validates consistent behavior for all 3 types.

### DIFF
--- a/Sources/FoundationEssentials/Locale/Locale_Preferences.swift
+++ b/Sources/FoundationEssentials/Locale/Locale_Preferences.swift
@@ -453,13 +453,9 @@ extension LocalePreferences: Codable {
             }
             return result as CFDictionary
         }
-        var icuSymbolsAndStrings = try container.decodeIfPresent(ICUSymbolsAndStrings.self, forKey: .icuSymbolsAndStrings)
-        if (icuDateFormats != nil || icuNumberSymbols != nil) && icuSymbolsAndStrings == nil {
-            // Ensure that we have a value to store these in even if the archive didn't contain any serialized info
-            icuSymbolsAndStrings = ICUSymbolsAndStrings()
-        }
-        icuSymbolsAndStrings?.icuDateFormatStrings = icuDateFormats
-        icuSymbolsAndStrings?.icuNumberSymbols = icuNumberSymbols
+        var icuSymbolsAndStrings = try container.decodeIfPresent(ICUSymbolsAndStrings.self, forKey: .icuSymbolsAndStrings) ?? ICUSymbolsAndStrings()
+        icuSymbolsAndStrings.icuDateFormatStrings = icuDateFormats
+        icuSymbolsAndStrings.icuNumberSymbols = icuNumberSymbols
         
         self.init(
             metricUnits: metricUnits,
@@ -475,7 +471,7 @@ extension LocalePreferences: Codable {
             force12Hour: force12Hour,
             numberSymbols: numberSymbols,
             dateFormats: dateFormats,
-            icuSymbolsAndStrings: icuSymbolsAndStrings ?? .init()
+            icuSymbolsAndStrings: icuSymbolsAndStrings
         )
         #else
         self.init(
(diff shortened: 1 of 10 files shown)


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/Locale/Locale_Preferences.swift
```diff
@@ -331,3 +335,196 @@ package struct LocalePreferences: Hashable, Sendable {
...
+        }
+        let icuNumberSymbols = numberSymbols.map {
+            var result = [String : String]()
+            for (rawValue, symbol) in $0 {
+                result["\(rawValue)"] = symbol
+            }
+            return result as CFDictionary
+        }
+        var icuSymbolsAndStrings = try container.decodeIfPresent(ICUSymbolsAndStrings.self, forKey: .icuSymbolsAndStrings)
+        if (icuDateFormats != nil || icuNumberSymbols != nil) && icuSymbolsAndStrings == nil {
```
itingliu: Nit: It looks like we can always create a `icuSymbolsAndStrings` since we're going to always need one below at `init`?

Sources/FoundationEssentials/Locale/Locale_Preferences.swift
```diff
@@ -331,3 +335,196 @@ package struct LocalePreferences: Hashable, Sendable {
...
+    }
+    
+    package init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        
+        self.icuDateTimeSymbols = try container.decodeIfPresent([String : [String]].self, forKey: .icuDateTimeSymbols).map { $0 as CFDictionary }
+        self.icuTimeFormatStrings = try container.decodeIfPresent([String : String].self, forKey: .icuTimeFormatStrings).map { $0 as CFDictionary }
+        self.icuNumberFormatStrings = try container.decodeIfPresent([String : String].self, forKey: .icuNumberFormatStrings).map { $0 as CFDictionary }
+        
+        // Will be filled in by LocalePreferences serialized value
```
itingliu, line 521: Thanks for this comment. I was indeed confused above why these two are handled separately from the others.

itingliu (approved):

jmschonfeld:
@swift-ci please test

PR #1494, opened 2025-09-04
FoundationEssentials: retry with modified permissions

compnerd:
On an access denied, reset the system permissions and retry the rename. The original variant seemed to work, this reduces some of the complexity in the hopes that it is sufficiently robust.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please build toolchain windows platform

jmschonfeld (changes requested):
Discussed with @parkera and we'd like to spend more time investigating this issue to come to a more understood solution. While resetting these file attributes does solve the problem, my understanding is that we don't have a clear picture of why exactly and it feels incorrect to me that we should unconditionally make a file writeable/non-hidden/non-system when writing if we failed to perform the write. If we can figure out more about what's happening here and better understand the root cause, we can confirm if this is the required approach (and perhaps limit its effects) or determine if there are other solutions. I've filed https://github.com/swiftlang/swift-foundation/issues/1507 that we'll use to continue investigation into this issue

Sources/FoundationEssentials/Data/Data+Writing.swift
jakepetroules, line 450: What about adding FILE_RENAME_IGNORE_READONLY_ATTRIBUTE?

jmschonfeld:
We ended up resolving this via a different PR so I'm going to go ahead and close this, but feel free to re-open if anything here is still wanted for merging

PR #1495, opened 2025-09-05
Locale.Region subregion API is misspelled

itingliu:
With @abi we can make an ABI compatible change so existing clients can continue to work on user's 6.2 devices even after they adopt the correct name on 6.2.1 and above.

But only use @abi for FOUNDATION_FRAMEWORK since ABI compatibility isn't a concern in other scenarios.


jmschonfeld (approved):

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #1500, opened 2025-09-06
Add Locale encoding option to prevent .current sentinel

jmschonfeld:
This adds an entry point to allow for encoding a `Locale` without using the compatibility behavior for older runtimes. This allows other `FOUNDATION_FRAMEWORK` types that contain a locale in their archive but previously worked around the unexpected `Locale` behavior to adopt this implementation without adopting the compatibility behavior.


jmschonfeld:
@swift-ci please test

itingliu (approved):
Awesome thanks

PR #1509, opened 2025-09-12
Improve URL component validation performance

jrflat:
During URL parsing, much of the time is spent validating the components in `UInt8.allowedURLComponents.getter`. We can make this faster by using allowed bit masks per component rather than relying on a switch statement for the `UInt8` code units.

The added test checks the validation behavior for all `UInt8` code units in each URL component. The test passes before and after this change.

20-25% speed up for parsing valid ASCII URLs:

URLValidationPerformance

### DIFF
--- a/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
+++ b/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
@@ -32,38 +32,38 @@ let benchmarks = {
 
     // MARK: - String Parsing
 
-    Benchmark("URL-ParseValidASCII") { benchmark in
+    Benchmark("URL.ParseValidASCII") { benchmark in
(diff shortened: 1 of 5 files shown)


Benchmarks/Benchmarks/URL/BenchmarkURL.swift
```diff
@@ -32,38 +32,38 @@ let benchmarks = {
 
     // MARK: - String Parsing
 
-    Benchmark("URL-ParseValidASCII") { benchmark in
+    Benchmark("URL.ParseValidASCII") { benchmark in
```
jrflat, line 35: Changed the benchmark names to use `.` so they display better in the JMH visualizer.

jrflat:
@swift-ci please test

PR #1511, opened 2025-09-16
[Testing] Don't strip Windows NT prefix from cwd

jrflat:
I suspect stripping the NT prefix from `FileManager.currentDirectoryPath` might be causing test failures in swift-corelibs-foundation.

```
TestFoundation/TestURL.swift:512: error: TestURL.test_fileURLWithPath_isDirectory : XCTAssertEqual failed: ("41") is not equal to ("36") - fileSystemRepresentation was too short
```

The file system representation of a file `URL` initialized with `FileManager.default.currentDirectoryPath` appears to be 5 characters too short. Will add better logging to the tests in swift-corelibs-foundation and test with this PR to see if this resolves it and to get more info.


jrflat:
Well this speculative fix looks to be causing more test failures here, I'll just open up a revert to test instead

PR #1514, opened 2025-09-18
Accept SF-0032 and SF-0033

itingliu:


itingliu:
@swift-ci please test

PR #1468, opened 2025-08-12
ProgressManager v6 implementation

chloe-yeo:
Latest ProgressManager implementation.

PR #1519, opened 2025-09-24
[Proposal] `ProgressManager` Post-Review Update

chloe-yeo:
Proposal change to reflect minor implementation change due to performance concerns of previously-proposed version of `ProgressManager` API. 

Updated implementation found here: https://github.com/swiftlang/swift-foundation/pull/1468


chloe-yeo:
@swift-ci please test

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
 ### Allow declared custom additional property to be any type that can be casted as `any Sendable`  
 We initially allowed the full flexibility of allowing developers to declare `ProgressManager.Property` types to be of any type, including structs. However, we realized that this has a severely negative impact on performance of the API. Thus, for now, we allow developers to only declare `ProgressManager.Property` with only certain `Value` and `Summary` types. 
 
+### Introduce `withProperties` closure as an entry point to atomically mutate multiple properties
+We initially considered introducing a withProperties closure as an entry point to atomically mutate multiple properties. This approach would have used a dedicated ProgressManager.Values struct with `@dynamicMemberLookup` to provide a unified interface for modifying all properties within a single atomic operation.
```
itingliu:
```suggestion
We initially considered introducing a `withProperties` closure as an entry point to atomically mutate multiple properties. This approach would have used a dedicated `ProgressManager.Values` struct with `@dynamicMemberLookup` to provide a unified interface for modifying all properties within a single atomic operation. 
```

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
 ### Allow declared custom additional property to be any type that can be casted as `any Sendable`  
 We initially allowed the full flexibility of allowing developers to declare `ProgressManager.Property` types to be of any type, including structs. However, we realized that this has a severely negative impact on performance of the API. Thus, for now, we allow developers to only declare `ProgressManager.Property` with only certain `Value` and `Summary` types. 
 
+### Introduce `withProperties` closure as an entry point to atomically mutate multiple properties
+We initially considered introducing a withProperties closure as an entry point to atomically mutate multiple properties. This approach would have used a dedicated ProgressManager.Values struct with `@dynamicMemberLookup` to provide a unified interface for modifying all properties within a single atomic operation. 
+
+However, this design had several drawbacks:
+- Limited atomicity: While it allowed atomic mutation of `totalCount` and `completedCount`, it could not achieve the same atomicity for additional properties due to their differences in backing storage
```
itingliu: I would add a few sentences here about why backing storage (an implementation detail) would affect this change of design. Why does it matter whether they're all stored into one single storage or multiple separate ones?

Either way, is it correct to say that if we are to enforce a whole atomicity for all of the additional properties, we would have to synchronize the access with a big lock for all of the properties, and would hinder performances?

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+- Less intuitive API: The withProperties closure promoted a less natural pattern for accessing and mutating additional properties, particularly when combined with `@dynamicMemberLookup` on the nested `ProgressManager.Values` struct
```
itingliu:
```suggestion
- Less intuitive API: The `withProperties` closure promoted a less natural pattern for accessing and mutating additional properties, particularly when combined with `@dynamicMemberLookup` on the nested `ProgressManager.Values` struct
```

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+- Inconsistent access patterns: We would need to use different patterns for accessing fraction-related properties and additional properties
```
itingliu: Were there discussions or feedback in the current version that was supportive of this pattern? What was the justification back then, and are they no longer valid?

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+
+We ultimately decided to replace this approach with a more streamlined design:
+- `setCounts` closure: Atomically update `totalCount` and `completedCount` 
+- `ProgressManager` directly using `@dynamicMemberLookup`: Applied directly to the ProgressManager class, enabling intuitive dot-syntax access for all additional properties (e.g., manager.totalFileCount = 100)
```
itingliu:
```suggestion
- `ProgressManager` directly using `@dynamicMemberLookup`: Applied directly to the ProgressManager class, enabling intuitive dot-syntax access for all additional properties (e.g., `manager.totalFileCount = 100`)
```

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+### Introduce `withProperties` closure as an entry point to atomically mutate multiple properties
+We initially considered introducing a withProperties closure as an entry point to atomically mutate multiple properties. This approach would have used a dedicated ProgressManager.Values struct with `@dynamicMemberLookup` to provide a unified interface for modifying all properties within a single atomic operation. 
+
+However, this design had several drawbacks:
+- Limited atomicity: While it allowed atomic mutation of `totalCount` and `completedCount`, it could not achieve the same atomicity for additional properties due to their differences in backing storage
+- Less intuitive API: The withProperties closure promoted a less natural pattern for accessing and mutating additional properties, particularly when combined with `@dynamicMemberLookup` on the nested `ProgressManager.Values` struct
+- Inconsistent access patterns: We would need to use different patterns for accessing fraction-related properties and additional properties
+
+We ultimately decided to replace this approach with a more streamlined design:
+- `setCounts` closure: Atomically update `totalCount` and `completedCount`
```
itingliu: Can you clarify why only `totalCount` and `completedCount` are considered here, and why do you think that there needs to be a closure to atomically update both of them?

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+
+We ultimately decided to replace this approach with a more streamlined design:
+- `setCounts` closure: Atomically update `totalCount` and `completedCount` 
+- `ProgressManager` directly using `@dynamicMemberLookup`: Applied directly to the ProgressManager class, enabling intuitive dot-syntax access for all additional properties (e.g., manager.totalFileCount = 100)
+- Consistent access pattern: Both accessing and mutating additional properties now use the same dot-syntax pattern, making the API more discoverable and natural to use
+
+With this change, we can atomically update `totalCount` and `completedCount` via the `setCounts` closure, and access and mutate additional properties via dot syntax. 
+
+### Minimal requirements for `ProgressManager.Property` protocol 
+We initially considered a simpler version of the ProgressManager.Property protocol that required only two components:
```
itingliu:
```suggestion
We initially considered a simpler version of the `ProgressManager.Property` protocol that had two requirements:
```
I think we just call them requirements

Proposals/0023-progress-manager.md
```diff
@@ -1434,6 +1432,43 @@ There were discussions about representing indeterminate state in `ProgressManage
...
+
+- `Value` - The type for individual property values
+- `defaultValue` - A default value when the property isn't explicitly set
+
+In this simplified approach, the framework would automatically:
```
itingliu: Can you clarify what "the framework" refers to? Did you mean Foundation would do the below?

Proposals/0023-progress-manager.md
chloe-yeo: Yep, sounds good!

Proposals/0023-progress-manager.md
chloe-yeo: I'll change this!

Proposals/0023-progress-manager.md
chloe-yeo: Yeah, but let me reword this!

Proposals/0023-progress-manager.md
chloe-yeo: I think in one of the discussions that was brought up was we mentioned that it's easy to forget that `withProperties` is the entry point to access and mutate additional properties of a `ProgressManager`, so I phrased it this way, but perhaps there's a better way to phrase this?

chloe-yeo: Actually the discussion that I had in mind was covered under less intuitive API, this is not needed.

Proposals/0023-progress-manager.md
chloe-yeo: For reasons related to backing storage, my line of thinking is that If we were to allow `totalFileCount` and `completedFileCount` to be stored in a struct the way `totalCount` and `completedCount` are, we should then also allow developers to do the same. But this will then then mean we have to allow developers to declare any `ProgressManager.Property`, making us store `(any Sendable)`. But for performance reasons, we do restrict them to only certain types such as `Int`, `UInt64` etc. I think it would be more about consistency between what our pre-declared custom properties can do vs what developers' custom properties can do.

chloe-yeo:
@swift-ci please test

Proposals/0023-progress-manager.md
itingliu: Why should an API user care about how `totalFileCount` and `completedFileCount` are stored, whether they're stored in a struct or not 😉 ? It could also be guarded by a semaphore for concurrent access. Would that allow atomically mutating everything?

chloe-yeo: Hmm you're right, those are implementation details, I think the thing that would matter most to developers, especially UI developers, will be the performance cost that comes with having `withProperties`. Because `withProperties` acquires the lock and does the access and mutation of additional properties after, and releasing the lock before returning the values, the way we make these properties `Observable` is much less performant. We have to register all access and mutations of additional properties onto only one keypath because the `access` and `withMutation` calls on the `ObservationRegistrar`, which may cause unexpected and unnecessary redraws of the UI, putting performance cost on developers who opt to report additional properties with their progress via `withProperties`.

chloe-yeo:
@swift-ci please test

chloe-yeo:
@swift-ci please test

PR #1523, opened 2025-09-25
Prevent nil force unwrap in JSONEncoder test

jrflat:
Prevent a nil force unwrap from crashing this test while we investigate the cause.


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1526, opened 2025-09-27
Fix JSONEncoder key conversion for CodingKeyRepresentable dictionary keys

ShivaHuang:
# Summary
Fixes an issue where `JSONEncoder` incorrectly applied key encoding strategies (like `.convertToSnakeCase`) to dictionary keys that implement `CodingKeyRepresentable`, when these keys should preserve their semantic meaning and remain unchanged.

# Problem
When encoding dictionaries with `CodingKeyRepresentable` keys (e.g., custom enum or struct keys), `JSONEncoder` was treating them as regular coding key values and applying key conversion strategies. This caused keys like "`leaveMeAlone`" to be transformed to "`leave_me_alone`" even when the key itself represents a semantic identifier that should remain unchanged.

# Solution
  - Expands the existing `_JSONStringDictionaryEncodableMarker` protocol to cover any `CodingKeyRepresentable`-keyed  dictionaries, not just `String`-keyed ones.
  - Renames the protocol to `_JSONCodingKeyRepresentableDictionaryEncodableMarker` to reflect its broader scope.
  - Updates the encoding logic to convert `CodingKeyRepresentable` keys directly to their string representation without applying encoding transformations.

# Test Plan
Added test case `encodingDictionaryCodingKeyRepresentableKeyConversionUntouched()` that verifies:
  - Dictionary with `CodingKeyRepresentable` keys preserves original key names.
  - Key encoding strategies like `.convertToSnakeCase` are not applied to the dictionary keys.
  - The fix works with custom structs implementing `CodingKeyRepresentable`.

# Checklist
  - [x] Added test coverage for the fix
  - [x] All existing tests pass
  - [x] Code follows Swift Foundation conventions
  - [x] Changes are backward compatible

  This change ensures that dictionaries with semantic keys maintain their intended structure while still allowing key
  encoding strategies to work correctly for object properties during regular encoding.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONEncoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONEncoder.swift
@@ -1184,14 +1184,16 @@ private extension __JSONEncoder {
         }
     }
 
-    func wrap(_ dict: [String : Encodable], for additionalKey: (some CodingKey)? = _CodingKey?.none) throws -> JSONEncoderValue? {
+    func wrap(_ dict: _JSONCodingKeyRepresentableDictionaryEncodableMarker, for additionalKey: (some CodingKey)? = _CodingKey?.none) throws -> JSONEncoderValue? {
+        let dict = dict as! [AnyHashable: Encodable]
         var result = [String: JSONEncoderValue]()
         result.reserveCapacity(dict.count)
 
         let encoder = __JSONEncoder(options: self.options, ownerEncoder: self)
         for (key, value) in dict {
-            encoder.codingKey = _CodingKey(stringValue: key)
-            result[key] = try encoder.wrap(value)
+            let stringKey = (key.base as! CodingKeyRepresentable).codingKey.stringValue
+            encoder.codingKey = _CodingKey(stringValue: stringKey)
+            result[stringKey] = try encoder.wrap(value)
         }
 
         return .object(result)
@@ -1214,8 +1216,8 @@ private extension __JSONEncoder {
             return self.wrap(url.absoluteString)
         } else if let decimal = value as? Decimal {
             return .number(decimal.description)
-        } else if let encodable = value as? _JSONStringDictionaryEncodableMarker {
-            return try self.wrap(encodable as! [String:Encodable], for: additionalKey)
+        } else if let encodable = value as? _JSONCodingKeyRepresentableDictionaryEncodableMarker {
+            return try self.wrap(encodable, for: additionalKey)
         } else if let array = value as? _JSONDirectArrayEncodable {
             if options.outputFormatting.contains(.prettyPrinted) {
                 let (bytes, lengths) = try array.individualElementRepresentation(encoder: self, additionalKey)
@@ -1362,11 +1364,11 @@ extension JSONEncoder : @unchecked Sendable {}
 // Special-casing Support
 //===----------------------------------------------------------------------===//
 
-/// A marker protocol used to determine whether a value is a `String`-keyed `Dictionary`
+/// A marker protocol used to determine whether a value is a `CodingKeyRepresentable`-keyed `Dictionary`
 /// containing `Encodable` values (in which case it should be exempt from key conversion strategies).
-private protocol _JSONStringDictionaryEncodableMarker { }
+private protocol _JSONCodingKeyRepresentableDictionaryEncodableMarker { }
 
-extension Dictionary : _JSONStringDictionaryEncodableMarker where Key == String, Value: Encodable { }
+extension Dictionary : _JSONCodingKeyRepresentableDictionaryEncodableMarker where Key: CodingKeyRepresentable, Value: Encodable { }
 
 /// A protocol used to determine whether a value is an `Array` containing values that allow
 /// us to bypass UnkeyedEncodingContainer overhead by directly encoding the contents as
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

ShivaHuang:
Hi @itingliu, I rebased the PR and seems the CI work needs to be run again, can you help?

ShivaHuang:
Would anyone be able to lend a hand with running the test and reviewing this?

ShivaHuang:
@jmschonfeld can you help reviewing this PR?

jmschonfeld:
Thanks for the ping, @kperryua has worked in this area before so I think he'd be best to review this change. But I can go ahead and kick off CI for you to get some CI results in the meantime

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/JSON/JSONEncoder.swift
```diff
@@ -1184,14 +1184,16 @@ private extension __JSONEncoder {
...
+    func wrap(_ dict: _JSONCodingKeyRepresentableDictionaryEncodableMarker, for additionalKey: (some CodingKey)? = _CodingKey?.none) throws -> JSONEncoderValue? {
+        let dict = dict as! [AnyHashable: Encodable]
         var result = [String: JSONEncoderValue]()
         result.reserveCapacity(dict.count)
 
         let encoder = __JSONEncoder(options: self.options, ownerEncoder: self)
         for (key, value) in dict {
-            encoder.codingKey = _CodingKey(stringValue: key)
-            result[key] = try encoder.wrap(value)
+            let stringKey = (key.base as! CodingKeyRepresentable).codingKey.stringValue
```
kperryua: I'm concerned about the additional uses of `as!` casting, and the hidden costs of `AnyHashable` any `any CodingKeyRepresentable` boxing that's happening here. Have you done any benchmarking before vs after of dictionaries that do and do not have `CodingKeyRepresentable` keys?

kperryua (commented):
Can you comment on whether a similar (or perhaps reverse) problem exists for the Decodable side of this? If a key is snake case and a decoder with `.convertFromSnakeCase` is used to decode a `CodingKeyRepresentable` dictionary key, does it convert the string to camel case first?

Sources/FoundationEssentials/JSON/JSONEncoder.swift
ShivaHuang: Thanks for the insight.
                                                                                                                                
I confirmed that the same bug exists on the decoder side and implemented a fix. Please review when you have a chance.

Regarding the AnyHashable boxing concern: I created a simple benchmark that confirmed a measurable performance impact. I refactored the approach to avoid AnyHashable boxing entirely, which resolves the performance issue.

One observation about test coverage: the current test cases use a simple structure:
```json
{"leave_me_alone": "test"}
```

This verifies that dictionary keys bypass the key encoding/decoding strategy, but doesn't verify that nested values still correctly apply the strategy. For example, with a [CustomKey: Person] dictionary where Person has properties like firstName, we should verify that:
1. Dictionary keys bypass the strategy (correct)
2. Nested struct properties still use the strategy (firstName ↔ first_name)

Would you prefer I add more comprehensive test cases in this PR, or should that be a separate PR?

kperryua: Thank you. This looks much better.

> Would you prefer I add more comprehensive test cases in this PR, or should that be a separate PR?

I'm fine with adding these additional tests to this PR.

ShivaHuang: Thanks for the feedback. I've updated the test cases to be more comprehensive - they now use a nested `Codable` struct as the dictionary value to verify that key encoding/decoding strategies are still correctly applied to nested properties, while dictionary keys remain exempt.

I also noticed the existing String-keyed dictionary tests (`decodingDictionaryStringKeyConversionUntouched` and `encodingDictionaryStringKeyConversionUntouched`) have the same gap - they only verify that keys are preserved, but don't check that nested values still apply the strategy. That's beyond the scope of this PR, so I'll address it separately.

ShivaHuang: Hi @kperryua, a separate PR for improving String-keyed dictionary tests has been created at  https://github.com/swiftlang/swift-foundation/pull/1777.
Please verify the modifications.

kperryua:
@swift-ci please test

ShivaHuang:
Hi @kperryua, just a gentle ping on this PR - all checks are passing. Would you have a chance to take a look when you get a moment?
Thanks!

ShivaHuang:
Hi @kperryua, just another gentle ping. 🙋🏻‍♂️

kperryua:
Apologies Shiva. On my todo list for today 😀

kperryua:
@ShivaHuang , the diff looks fine, but the commit history with the merge of `main` is causing some issues with the Foundation.framework version. Would you please force squash-rebase this PR's history into a single commit?

ShivaHuang:
Thanks @kperryua!
Just finished the cleanup. Please re-run the tests and see if there any problems.

Also updated the original string-key dictionary tests at https://github.com/swiftlang/swift-foundation/pull/1777 by the way.

itingliu:
@swift-ci please test

PR #1530, opened 2025-10-01
Data.mutableSpan and Data.mutableBytes do not perform uniqueness check

jmschonfeld:
Unfortunately the `mutableSpan` and `mutableBytes` accessors on `Data` do not currently perform a uniqueness check to trigger CoW. This means that you can mutate a `Data` via its mutable span even if it references a shared storage. This updates the implementation to trigger CoW for the shared storage representations when accessing the mutable spans as well as a unit test to ensure a copy occurs when not uniquely referenced (and that no copy occurs if the storage is uniquely referenced).

### DIFF
--- a/Tests/FoundationEssentialsTests/DataTests.swift
+++ b/Tests/FoundationEssentialsTests/DataTests.swift
@@ -1439,6 +1439,50 @@ private final class DataTests {
             #expect(data.count == 0)
         }
     }
+    
+    @Test func validateMutation_cow_mutableBytes() {
+        var data = Data(Array(repeating: 0, count: 32))
+        holdReference(data) {
+            var bytes = data.mutableBytes
+            bytes.storeBytes(of: 1, toByteOffset: 0, as: UInt8.self)
+            
+            #expect(data[0] == 1)
+            #expect(heldData?[0] == 0)
+        }
+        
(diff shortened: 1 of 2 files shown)


jmschonfeld:
@swift-ci please test

kperryua (approved):

Tests/FoundationEssentialsTests/DataTests.swift
```diff
@@ -1439,6 +1439,50 @@ private final class DataTests {
             #expect(data.count == 0)
         }
     }
+    
+    @Test func validateMutation_cow_mutableBytes() {
+        var data = Data(Array(repeating: 0, count: 32))
```
glessard: Don't we have `Data(count: 32)` to get a zero-initialized instance?

glessard (approved):
👍 We might want to also add a test to exercise the large slice code path.

jmschonfeld:
> 👍 We might want to also add a test to exercise the large slice code path.

Good call, added tests for the large slice too!

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1533, opened 2025-10-02
Double-double Date

stephentyrone:
Adapts Date to use a double-Double representation (the underlying representation is the sum of two double-precision values, giving about 106 bits of precision). 

Previously Date was backed by a single Double measuring time since Jan 1 2001 in seconds. Because Double's precision is non-uniform, this means that times within a hundred days of the epoch are represented with approximately nanosecond precision, but as you get farther away from that date the precision decreases. For times close to today, it has been reduced to about 100ns.
     
The obvious thing would be to adopt an integer-based representation similar to C's timespec (32b nanoseconds, 64b seconds) or Swift's Duration (128b attoseconds). These representations suffer from a few difficulties:
     
 - Existing API on Date takes and produces `TimeInterval` (aka Double). Making Date use an integer representation internally would mean that existing users of the public API would suddently start getting different results for computations that were previously exact; even though we could add new API, and the overall system would be more precise, this would be a surprisingly subtle change for users to navigate.
     
- We have been told that some software interprets the raw bytes of Date as a Double for the purposes of fast serialization. These packages formally violate Foundation's API boundaries, but that doesn't help users of those packages who would abruptly be broken by switching to an integer representation.
     
Using DoubleDouble instead navigates these problems fairly elegantly.
     
- Because DoubleDouble is still a floating-point type, it still suffers from non-uniform precision. However, because DoubleDouble is so fantastically precise, it can represent dates out to ±2.5 quadrillion years at ~nanosecond or better precision, so in practice this won't be much of an issue.
     
- Existing API on Date will produce exactly the same result as it did previously in cases where those results were exact, minimizing surprises. In cases where the existing API was not exact, it will produce much more accurate results, even if users do not adopt new API, because its internal calculations are now more precise.
     
- Software that (incorrectly) interprets the raw bytes of Date as a Double will get at least as accurate of a value as it did previously (and often a more accurate value).                             

DateInterval gets the same treatment, so that it can benefit from more accurate internal computations as well. Follow-on work may adapt/add-to Date's API to make it easier to specify Dates with high precision.

### DIFF
--- a/Sources/FoundationEssentials/DateInterval.swift
+++ b/Sources/FoundationEssentials/DateInterval.swift
@@ -42,7 +42,7 @@ public struct DateInterval: Comparable, Hashable, Codable, Sendable {
         }
         set {
             precondition(newValue >= 0, "Negative durations are not allowed")
-            _duration = DoubleDouble(head: newValue, tail: 0)
+            _duration = DoubleDouble(uncheckedHead: newValue, tail: 0)
         }
     }
 
@@ -68,7 +68,7 @@ public struct DateInterval: Comparable, Hashable, Codable, Sendable {
     public init(start: Date, duration: TimeInterval) {
         precondition(duration >= 0, "Negative durations are not allowed")
         self.start = start
-        _duration = DoubleDouble(head: duration, tail: 0)
+        _duration = DoubleDouble(uncheckedHead: duration, tail: 0)
     }
 
     /**
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -34,16 +34,64 @@ public typealias TimeInterval = Double
  A `Date` is independent of a particular calendar or time zone. To represent a `Date` to a user, you must interpret it in the context of a `Calendar`.
 */
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
-public struct Date : Comparable, Hashable, Equatable, Sendable {
-
+public struct Date: Comparable, Hashable, Equatable, Sendable {
+    /* Date is internally represented as a sum of two Doubles.
+    
+     Previously Date was backed by a single Double measuring time since
+     Jan 1 2001 in seconds. Because Double's precision is non-uniform, this
+     means that times within a hundred days of the epoch are represented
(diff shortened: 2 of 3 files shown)


stephentyrone:
Converted to draft as I want to add some additional comments and tests before this moves forward, but the basic shape of the change is correct, I think.

parkera:
@swift-ci test

stephentyrone:
@swift-ci test

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -131,7 +131,10 @@ extension Calendar {
                 }
                 self.recurrence = recurrence
                 
-                self.start = start
+                // round start down to whole seconds, set aside fraction.
+                let wholeSeconds = start._time.floor()
+                fractionalSeconds = (start._time - wholeSeconds).head
```
itingliu, line 136: Note for myself: we should/could probably add a helper method somewhere to decompose the floating point second into the whole number and the fractional part. IIRC we have quite a few places where we're doing it

Sources/FoundationEssentials/DateInterval.swift
```diff
@@ -162,6 +169,24 @@ public struct DateInterval : Comparable, Hashable, Codable, Sendable {
     public static func <(lhs: DateInterval, rhs: DateInterval) -> Bool {
         return lhs.compare(rhs) == .orderedAscending
     }
+    
+    enum CodingKeys: String, CodingKey {
+        case start = "start"
+        case duration = "duration"
+    }
+    
+    public init(from decoder: Decoder) throws {
```
itingliu, line 178: This is effectively identical to the current auto-generated implementation, right?

stephentyrone: That's the idea!

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
stephentyrone: I'll take a note to do this as part of the follow-on work.

sebsto:
@stephentyrone FYI https://github.com/swiftlang/swift-foundation/issues/1561

PR #1536, opened 2025-10-06
Julian year 4713 BC is -4712, because year 0 = 1 BC

UInt2048:
Companion PR to (currently unmerged) https://github.com/swiftlang/swift-corelibs-foundation/pull/5274

This is a minor comment-only fix. The minimum of the range (`-211845067200.0`) does indeed represent Julian day 0 (proleptic Gregorian `-4713-11-24 12:00:00 +0000` = November 24, 4714 BC), but it is documented incorrectly as proleptic Julian `-4713-01-01 12:00:00 +0000` which is 4714 BC instead of the correct year -4712 = 4713 BC


itingliu:
@swift-ci please test

PR #1539, opened 2025-10-09
CMake: modulemap override in build interface

etcwilde:
The modulemap override should only be added in the build interface, where we don't copy the modulemap and header files. The install interface installs the modulemap and header files to a location where the compiler will see it, so passing the override results in the compiler seeing both modulemaps defining the same module and then failing. This moves the override to only be part of the build interface.


etcwilde:
@swift-ci please test

PR #1550, opened 2025-10-15
Micro Optimization: Elide overflow checking in Data.bytes

jmschonfeld:
This is a tiny optimization in conversion from `Data` to a `RawSpan` that elides unnecessary overflow checks when performing arithmetic on the bounds/offsets of the bytes pointer. Given the constraints that the `Range` type guarantees (`lowerBound <= upperBound`) and the constraint that all `Data` indices are non-negative, we can avoid performing overflow checking on these operations since the possible inputs can never overflow. Runtime-wise the other parts of this conversion (copying the representation, switching over the cases, etc.) outweigh these overflow checks. However, I did find that removing these checks resulted in a 15% reduction in code size in the caller and as we remove the other heavy pieces these will become more prominent. As we heavily promote using `Span`s in API this conversion will become much more common and I think is worthy of a micro optimization like this to ensure our best performance.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data.swift
+++ b/Sources/FoundationEssentials/Data/Data.swift
@@ -207,7 +207,7 @@ internal final class __DataStorage : @unchecked Sendable {
 
     @inlinable // This is @inlinable as trivially computable.
     var mutableBytes: UnsafeMutableRawPointer? {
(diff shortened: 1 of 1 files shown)


kperryua (approved):

glessard (commented):
Thanks

PR #1565, opened 2025-10-30
Generalize ContiguousBytes to be noncopyable and nonescapable for spans

DougGregor:
The ContiguousBytes protocol covers various types that can provide unsafe access to their stored bytes. Generalize the ContiguousBytes protocol by making it non-copyable and non-escapable, so that the Span family of types and InlineArray can conform to it.

This pull request also generalizes the ContiguousBytes APIs for typed throws and introduces a new `withBytes` operation as a safe counterpart to `withUnsafeBytes` that provides a `RawSpan`.

Fixes rdar://163716671.

### DIFF
--- a/Sources/FoundationEssentials/Data/ContiguousBytes.swift
+++ b/Sources/FoundationEssentials/Data/ContiguousBytes.swift
@@ -71,6 +71,7 @@ extension ContiguousBytes where Self: ~Escapable, Self: ~Copyable {
     /// - note: Calling `withBytes` multiple times does not guarantee that
     ///         the same span will be passed in every time.
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
 #if !hasFeature(Embedded)
         do {
@@ -96,6 +97,7 @@ extension ContiguousBytes where Self: ~Escapable, Self: ~Copyable {
 extension Array : ContiguousBytes where Element == UInt8 {
     // FIXME: Generalize to R: ~Copyable when withUnsafeBufferPointer does
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         try withUnsafeBufferPointer { (buffer) throws(E) in
             try body(buffer.span.bytes)
@@ -108,6 +110,7 @@ extension Array : ContiguousBytes where Element == UInt8 {
 extension ArraySlice : ContiguousBytes where Element == UInt8 {
     // FIXME: Generalize to R: ~Copyable when withUnsafeBufferPointer does
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         try withUnsafeBufferPointer { (buffer) throws(E) in
             try body(buffer.span.bytes)
@@ -119,6 +122,7 @@ extension ArraySlice : ContiguousBytes where Element == UInt8 {
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 extension ContiguousArray : ContiguousBytes where Element == UInt8 {
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(span.bytes)
     }
@@ -146,6 +150,7 @@ extension UnsafeRawBufferPointer : ContiguousBytes {
     }
 
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(bytes)
     }
@@ -168,6 +173,7 @@ extension UnsafeMutableRawBufferPointer : ContiguousBytes {
     }
 
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(bytes)
     }
@@ -190,6 +196,7 @@ extension UnsafeBufferPointer : ContiguousBytes where Element == UInt8 {
     }
 
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(span.bytes)
     }
@@ -212,6 +219,7 @@ extension UnsafeMutableBufferPointer : ContiguousBytes where Element == UInt8 {
     }
 
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(span.bytes)
     }
@@ -234,6 +242,7 @@ extension EmptyCollection : ContiguousBytes where Element == UInt8 {
     }
 
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         return try body(RawSpan())
     }
@@ -263,6 +272,7 @@ extension CollectionOfOne : ContiguousBytes where Element == UInt8 {
 
     // FIXME: Generalize to R: ~Copyable when withUnsafeBufferPointer does
     @_alwaysEmitIntoClient
+    @available(macOS 10.14.4, iOS 12.2, watchOS 5.2, tvOS 12.2, visionOS 1.0, *)
     public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
         let element = self.first!
         return try Swift.withUnsafeBytes(of: element) { (buffer) throws(E) in


DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -12,6 +12,20 @@
 
 //===--- ContiguousBytes --------------------------------------------------===//
 
+#if compiler(>=6.2)
```
parkera: Foundation only compiles with the most recent released compiler + main, so I don't think this check is required.

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -109,3 +124,34 @@ extension Slice : ContiguousBytes where Base : ContiguousBytes {
         }
     }
 }
+
+#if compiler(>=6.2)
+
+//===--- Span Conformances -----------------------------------------===//
+
+@available(FoundationPreview 6.3, *)
```
parkera: This is a public conformance, right? We would need an API review.

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    @available(macOS 26.0, iOS 26.0, tvOS 26.0, watchOS 26.0, visionOS 26.0, *)
+    @Test func span() throws {
+        if #available(FoundationPreview 6.3, *) {
+            var bytes: [UInt8] = [1, 2, 3]
+            acceptContiguousBytes(bytes.span)
+            acceptContiguousBytes(bytes.mutableSpan)
+            acceptContiguousBytes(bytes.span.bytes)
+
+            let ms = bytes.mutableSpan
+            acceptContiguousBytes(ms.bytes)
```
glessard, line 56:
```suggestion
            acceptContiguousBytes(ms.mutableBytes)
```

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -12,6 +12,20 @@
 
 //===--- ContiguousBytes --------------------------------------------------===//
 
+#if compiler(>=6.2)
+/// Indicates that the conforming type is a contiguous collection of raw bytes
+/// whose underlying storage is directly accessible by withUnsafeBytes.
+@available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
+public protocol ContiguousBytes: ~Escapable, ~Copyable {
```
jmschonfeld, line 33: Just confirming, are there any ABI implications on this change?

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -109,3 +124,34 @@ extension Slice : ContiguousBytes where Base : ContiguousBytes {
...
+@available(FoundationPreview 6.3, *)
+extension Span: ContiguousBytes where Element == UInt8 {
+}
+
+@available(FoundationPreview 6.3, *)
+extension MutableSpan: ContiguousBytes where Element == UInt8 {
+}
+
+@available(FoundationPreview 6.3, *)
+@available(macOS 26.0, iOS 26.0, tvOS 26.0, watchOS 26.0, visionOS 26.0, *)
```
jmschonfeld: The double availability attribute seems a bit suspicious here - each declaration should only have a single availability. Is the problem here the macOS back deployment build of the package? If so, we'll need to create a new availability macro (something like `FoundationInlineArray`) that maps to macOS 26-aligned in the package manifest / CMake files

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
```diff
@@ -0,0 +1,41 @@
...
+
+#if compiler(>=6.2)
+
+func acceptContiguousBytes(_ bytes: borrowing some ContiguousBytes & ~Escapable & ~Copyable) { }
+
+@Suite("ContiguousBytesTests")
+private struct ContiguousBytesTests {
+    @available(macOS 26.0, iOS 26.0, tvOS 26.0, watchOS 26.0, visionOS 26.0, *)
+    @Test func span() throws {
+        if #available(FoundationPreview 6.3, *) {
```
jmschonfeld: Since the function is already annotated as only available on macOS 26+, I don't think we need this `#available` check here. In the end what we probably want is for this function to be annotated as `@available(FoundationInlineArray 6.3, *)`

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -12,6 +12,20 @@
...
+/// whose underlying storage is directly accessible by withUnsafeBytes.
+@available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
+public protocol ContiguousBytes: ~Escapable, ~Copyable {
+    /// Calls the given closure with the contents of underlying storage.
+    ///
+    /// - note: Calling `withUnsafeBytes` multiple times does not guarantee that
+    ///         the same buffer pointer will be passed in every time.
+    /// - warning: The buffer argument to the body should not be stored or used
+    ///            outside of the lifetime of the call to the closure.
+    func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R
```
jmschonfeld, line 45: There's probably still work to be done on this protocol to update it (it doesn't allow for producing a span, it hasn't adopted typed throws, etc.) but this seems like a reasonable step in a good direction that we can do without fully fleshing out those changes at the moment if we want to update this incrementally

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
DougGregor: I actually want to do both, thanks!

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: No, it is not ABI-breaking (or source-breaking) to make this change.

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: This worked out, thank you for the suggestion

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: Right. The best we can do with the protocol as-is (without breaking existing clients) would be to add a `withSpan`, because that's implementable by going through `withUnsafeBytes`.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: Yes. Here's the proposal PR: https://github.com/swiftlang/swift-foundation/pull/1582

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,16 +13,57 @@
...
+        }
+#else
+        return try withUnsafeBytes { (buffer) throws(E) in
+            try body(buffer.bytes)
+        }
+#endif
+    }
 }
 
 //===--- Collection Conformances ------------------------------------------===//
```
benrimmington, line 93: Should the [array types][] implement `withBytes(_:)` instead of using the default implementation?

[array types]: <https://github.com/swiftlang/swift-foundation/blob/swift-6.2.1-RELEASE/Sources/FoundationEssentials/Data/ContiguousBytes.swift#L28-L40>

DougGregor:
@swift-ci please test

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
DougGregor: I was able to clean this up with your prior suggestion

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: Yes, that's reasonable. For Array and ArraySlice, it's just a matter of completeness right now, because we still have to go through `withUnsafeBufferPointer`. For `ContiguousArray`, we can use `span` because it back-deploys.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+    ///
+    /// - note: Calling `withUnsafeBytes` multiple times does not guarantee that
+    ///         the same buffer pointer will be passed in every time.
+    /// - warning: The buffer argument to the body should not be stored or used
+    ///            outside of the lifetime of the call to the closure.
+    ///
+    /// Clients should prefer the `withBytes` function to this one, because
+    /// `withBytes` uses the non-escapable type `RawSpan` to ensure that the
+    /// buffer argument is not stored outside of the closure.
+    func withUnsafeBytes<R, E>(_ body: (UnsafeRawBufferPointer) throws(E) -> R) throws(E) -> R
```
itingliu, line 57: I thought adopting typed throw for `rethrow`ing closures is source compatible, like @glessard's PR: https://github.com/swiftlang/swift-foundation/pull/1649/changes

In that case we probably don't need to conditionally do it `#if !hasFeature(Embedded)`. Or am I missing other considerations?

DougGregor: It's source-compatible, but not ABI-compatible. We need the `rethrows` version to maintain the ABI.

itingliu: ah yes of course. In that case would it make sense to @_alwaysEmitIntoClient for ABI everywhere?

DougGregor: It's reasonable to use `@_alwaysEmitIntoClient` for concrete APIs to help them back-deploy. You can't use it on a protocol requirement, though.

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+    // Historical ABI
+    @usableFromInline
+    @abi(func withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R)
+    func __abi__withUnsafeBytes<R>(_ body: (UnsafeRawBufferPointer) throws -> R) rethrows -> R {
         return try body(self)
     }
+    #endif
+
+    @_alwaysEmitIntoClient
+    public func withUnsafeBytes<R, E>(_ body: (UnsafeRawBufferPointer) throws(E) -> R) throws(E) -> R {
```
jmschonfeld, line 148: Should this (and the one below on `UnsafeMutableRawBufferPointer`) be generalized to `R: ~Copyable`?

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+
+    @_alwaysEmitIntoClient
+    public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
+        return try body(span.bytes)
+    }
 }
 
 // FIXME: When possible, expand conformance to `where Element : Trivial`.
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 extension EmptyCollection : ContiguousBytes where Element == UInt8 {
```
jmschonfeld, line 230: Should this and the extension on `CollectionOfOne` get concrete implementations of the span variant like the other types above did?

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        // Ensure that we can use withBytes with typed throws.
+        return try bytes.withBytes { (buffer) throws(HomeworkError) in
+            if buffer.isEmpty {
+                throw .dogAteIt
+            }
+
+            return buffer.byteCount
+        }
+    } catch let error {
+        precondition(error == .dogAteIt)
```
jmschonfeld, line 38: Could we update this to an `#expect`? If something like this fails, it's a lot easier to diagnose expectation failures in CI instead of crashes.

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
jmschonfeld: You can move this to an `@available` on the function instead of an `#available` in the body. And you'll want to change this to `FoundationInlineArray`

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+    @_alwaysEmitIntoClient
+    public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
+        try body(span.bytes)
+    }
+}
+
+@available(FoundationInlineArray 6.3, *)
+extension InlineArray: ContiguousBytes where Element == UInt8 { }
+
+@available(FoundationInlineArray 6.2, *)
```
jmschonfeld: Hmm this is tricky - 6.2 (or 6.3 or 6.4) isn't actually correct here because this function shouldn't come with availability. However, the availability is needed in the CI environment because CI devices are too old. Can you surround this `@available` annotation (just the annotation) with `#if os(macOS) && !FOUNDATION_FRAMEWORK`? That way it only applies to the CI environment.

Sources/FoundationEssentials/Data/ContiguousBytes.swift
jmschonfeld: All of these annotations should be updated to `FoundationPreview 6.4`

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+    @_alwaysEmitIntoClient
+    public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
+        try body(bytes)
+    }
+}
+
+@available(FoundationInlineArray 6.3, *)
+extension UTF8Span: ContiguousBytes { }
+
+@available(FoundationInlineArray 6.2, *)
```
jmschonfeld: Same comment here as the inline array one below

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,389 @@
...
+/// suppressing the implicit `Copyable` and `Escapable` requirements:
+///
+///     func encrypt<Bytes: ContiguousBytes>(_ bytes: Bytes) -> [UInt8]
+///             where Bytes: ~Copyable, Bytes: ~Escapable {
+///         return bytes.withBytes { rawSpan in ... }
+///     }
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
-public protocol ContiguousBytes {
+public protocol ContiguousBytes: ~Escapable, ~Copyable {
+#if !hasFeature(Embedded)
```
jmschonfeld, line 34: Do we actually want to include this in this change? I didn't see the API proposal mention this and it's also unclear to me what the guarantees are of Embedded Swift (do we need to require source stability between versions? Is it expected that clients who want to use typed throws a the call site need to use `#if hasFeature(Embedded)` too since we can't make it typed throws everywhere?). Maybe we should make this a followup change (perhaps with an API proposal depending on whether @itingliu thinks that's needed here). As it is today, Foundation doesn't build for embedded yet (and other conformances in Foundation like `Data`s conformance to this haven't adopted typed throws)

itingliu: I mentioned in our workgroup thread about this proposal that I'm ok doing it. My reason is that given FoundationEssentials doesn't build for embedded at all currently, I'm ok taking it as a build fix.

Although I see your point since @parkera also had some questions about this. While I'm convinced by Doug's explanation that this is a ok change (I forgot where he explained it unfortunately), splitting it into a separate PR would help minimizing the confusion, and allow Doug to clarify any questions in a centralized place. @DougGregor what do you think?

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
glessard, line 38:
```suggestion
        #expect(error == .dogAteIt)
```

glessard:
@swift-ci please test

glessard:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
glessard: It seems like we could have `R: ~Copyable` for all the concrete implementations here. We should do that in a separate PR since it wasn't stated in the proposal.

glessard:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
glessard: Yes, added.

Sources/FoundationEssentials/Data/ContiguousBytes.swift
jmschonfeld: Ah ok makes sense, in that case no concerns here since we don't really need to worry about compatibility there

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,404 @@
...
+    }
+
+    @_alwaysEmitIntoClient
+    public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
+        try body(bytes)
+    }
+}
+
+#if os(macOS) && !FOUNDATION_FRAMEWORK
+@available(FoundationInlineArray 6.3, *)
```
jmschonfeld: This availability should be 6.4, and also shouldn't be wrapped in an ifdef (conformances can't be back deployed)

Sources/FoundationEssentials/Data/ContiguousBytes.swift
jmschonfeld: Can we update this to 6.3 (I know it doesn't matter in practice, but for consistency)

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,99 +13,404 @@
...
+    }
+
+    @_alwaysEmitIntoClient
+    public func withBytes<R: ~Copyable, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
+        try body(span.bytes)
+    }
+}
+
+#if os(macOS) && !FOUNDATION_FRAMEWORK
+@available(FoundationInlineArray 6.3, *)
```
jmschonfeld: Same comment as above, can we update this to 6.4 and remove the ifdefs on this one?

Sources/FoundationEssentials/Data/ContiguousBytes.swift
jmschonfeld: Same comment as above: 6.2 --> 6.4

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
```diff
@@ -0,0 +1,66 @@
...
+@testable import FoundationEssentials
+#else
+@testable import Foundation
+#endif
+
+enum HomeworkError: Error {
+    case dogAteIt
+}
+
+@available(FoundationPreview 6.3, *)
```
jmschonfeld: Is this availability required?

Tests/FoundationEssentialsTests/ContiguousBytesTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        #expect(error == .dogAteIt)
+        return -1
+    }
+}
+
+struct NC: ~Copyable { }
+
+@Suite("ContiguousBytesTests")
+private struct ContiguousBytesTests {
+    @available(FoundationInlineArray 6.3, *)
```
jmschonfeld: 6.3 --> 6.4

glessard:
@swift-ci please test

CMakeLists.txt
jmschonfeld, line 104: Sorry for one more followup - I missed this before: the `6.4` version needs to be added to this list. Otherwise this all LGTM!

Sources/FoundationEssentials/Data/ContiguousBytes.swift
DougGregor: Yes, will do.

DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

glessard:
@swift-ci please test

DougGregor:
@swift-ci please test

Sources/FoundationEssentials/Data/ContiguousBytes.swift
```diff
@@ -13,31 +13,116 @@
...
+    func withBytes<R, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R
+}
+
+extension ContiguousBytes where Self: ~Escapable, Self: ~Copyable {
+    /// Calls the given closure with the contents of underlying storage.
+    ///
+    /// - note: Calling `withBytes` multiple times does not guarantee that
+    ///         the same span will be passed in every time.
+    @_alwaysEmitIntoClient
+    public func withBytes<R, E>(_ body: (RawSpan) throws(E) -> R) throws(E) -> R {
```
jmschonfeld, line 75: For this and all applicable functions below, I think we need to add Span's availability to the declaration. `ContiguousBytes` is available back to macOS 10.10 but `RawSpan` is only available in macOS 10.14.4 and newer so when building in an environment where availability is important this fails to build because `buffer.bytes` below isn't available

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

PR #1570, opened 2025-11-03
 Use a new TimeZone ICU API instead of directing time zone through ucal

itingliu:
Switch to use the uatimezone API when we only need timezone information. This is not only faster but also allows us to not have to lock around the shared `UCalendar`.


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1573, opened 2025-11-03
Prevent CalendarTests.datesMatching_simpleExample() from using the current TimeZone

jmschonfeld:
This test inadvertently used the current timezone, which can impact results if that timezone changes during the test. This sets the timezone to GMT.

Resolves rdar://163960766


jmschonfeld:
@swift-ci please test

itingliu (approved):
Thanks!

PR #1576, opened 2025-11-06
Adopt span in Data IO implementation

jmschonfeld:
This adopts Span types throughout the Data reading/writing implementation. There are still a few cases where we use unsafe buffers directly (getting unsafe buffers from spans directly at the call site for an unsafe C API, and storing unsafe pointers that are the result of `mmap`-like calls) but this removes the majority of unsafe buffers passed around in our code and adopts span as the internal currency type wherever possible. Additionally, this allows us to adopt Span types in the API surface in the future now that the internal implementations are based upon span. I ran our benchmarks before and after this change and found no difference in performance when switching from unsafe buffers to spans.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -490,7 +490,9 @@ private func readBytesFromFileDescriptor(_ fd: Int32, path: PathOrURL, buffer in
 #else
                 numBytesRead = CInt(read(fd, buffer.baseAddress!.advanced(by: initializedCount), Int(numBytesRequested)))
 #endif
-                initializedCount += Int(clamping: numBytesRead)
+                if numBytesRead >= 0 {
+                    initializedCount += Int(clamping: numBytesRead)
+                }
             }
         } while numBytesRead < 0 && errno == EINTR
         


jmschonfeld:
Moving to draft to sort out a few miscellaneous test failures

parkera:
Looks good to me in principle, although I didn't try to figure out what the test failures are about.

jmschonfeld:
> Looks good to me in principle, although I didn't try to figure out what the test failures are about.

Found it - an accidental "while not empty" instead of "while not full" when reading bytes from a file into a buffer 🙂

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -193,33 +193,33 @@ struct ReadBytesResult {
...
 
         let dwBytesToRead: DWORD =
-            DWORD(clamping: DWORD(min(DWORD(dwChunk), dwBytesRemaining)))
+            DWORD(clamping: min(dwChunk, pBuffer.freeCapacity))
+        
         var dwBytesRead: DWORD = 0
-        if !ReadFile(hFile, pBuffer, dwBytesToRead, &dwBytesRead, nil) {
-            throw CocoaError.errorWithFilePath(path, win32: GetLastError(), reading: true)
+        try pBuffer.withUnsafeMutableBytes { bytes, initializedCount in
+            if !ReadFile(hFile, bytes.baseAddress!.advanced(by: initializedCount), dwBytesToRead, &dwBytesRead, nil) {
```
glessard, line 213: This makes me wonder if we want a variant of `withUnsafeMutableBytes` that gives you only the uninitialized portion of the `OutputRawSpan`.

glessard (approved):

Sources/FoundationEssentials/Data/Data+Reading.swift
jmschonfeld: Seems like it could be beneficial for C interop where you repeatedly call a function that fills in a buffer!

PR #1579, opened 2025-11-06
Move Data Hashable/Equatable conformance back to struct

jmschonfeld:
Moving the conformance to an `extension` proved to be source breaking for rare cases where clients had an `internal` import of Foundation but used the conformance publicly. This moves the conformance back to the struct declaration to work around the issue


itingliu (approved):

itingliu:
@swift-ci please test

PR #1582, opened 2025-11-07
Add proposal for generalizing ContiguousBytes to support Span

DougGregor:
API proposal corresponding to https://github.com/swiftlang/swift-foundation/pull/1565

### DIFF
--- a/Proposals/nnnn-ContiguousBytes-Span.md
+++ b/Proposals/nnnn-ContiguousBytes-Span.md
@@ -149,10 +149,6 @@ This means that the `encrypt` function will not be usable with non-copyable type
 
 Ideally, Swift code bases using `ContiguousBytes` would move from using `withUnsafeBytes` to using the safer `withBytes` introduced by this proposal. This can be helped somewhat by the opt-in strict memory safety mode introduced in [SE-0458](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0458-strict-memory-safety.md), which will identify uses of unsafe buffer pointers and require them to be marked `unsafe`.
 
-## Future directions
-
-In the future, we could consider deprecating `withUnsafeBytes` in favor of `withBytes`: aside from source compatibility, there is no reason to use `withUnsafeBytes` instead of `withBytes`, and the latter is safer.
-
 ## Alternatives considered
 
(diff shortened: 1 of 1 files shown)


Proposals/nnnn-ContiguousBytes-Span.md
```diff
@@ -0,0 +1,176 @@
...
+
+### Replace `withBytes` with a property
+
+An alternative to the new `withBytes` requirement of `ContiguousBytes` is to provide a property
+
+```swift
+var bytes: RawSpan { get }
+```
+
+in the protocol. However, not all types that currently conform to the `ContiguousBytes` protocol can provide a `bytes` property that satisfies this. For example, a type that needs to materialize data into a buffer to pass to the closure provided to `with(Unsafe)Bytes` would not be able to implement this property, which depends on having the lifetime of the resulting `RawSpan` tied to that of its enclosing type. [SE-0456](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0456-stdlib-span-properties.md) describes some of the changes required in the implementations of `String.UTF8View` and `Array` that were needed to provide `span` properties and which might not be possible for other types. Therefore, while adding this property would provide more ergonomic access to the contiguous bytes of a type, doing so necessary breaks source compatibility.
```
Jnosh:
```suggestion
in the protocol. However, not all types that currently conform to the `ContiguousBytes` protocol can provide a `bytes` property that satisfies this. For example, a type that needs to materialize data into a buffer to pass to the closure provided to `with(Unsafe)Bytes` would not be able to implement this property, which depends on having the lifetime of the resulting `RawSpan` tied to that of its enclosing type. [SE-0456](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0456-stdlib-span-properties.md) describes some of the changes required in the implementations of `String.UTF8View` and `Array` that were needed to provide `span` properties and which might not be possible for other types. Therefore, while adding this property would provide more ergonomic access to the contiguous bytes of a type, doing so necessarily breaks source compatibility.
```

DougGregor:
@swift-ci please test

itingliu (approved):

itingliu:
@swift-ci please test

PR #1594, opened 2025-11-12
Fix Calendar Typos in Foundation Essentials

MojtabaHs:
All changes are local and do not affect the public API. 
Some changes are made to local scoped variables, while others are made to comments. Therefore, all of these changes appear to be backward compatible.

Please review.
Thanks

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -2395,7 +2395,7 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable 
                 let newOffset = timeZone.secondsFromGMT(for: newDateInWholeSecond)
                 let prevOffset = timeZone.secondsFromGMT(for: date)
 
-                // No need for normal gmt-offset adjustment because the revelant bits are handled above individually
+                // No need for normal gmt-offset adjustment because the relevant bits are handled above individually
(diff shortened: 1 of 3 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -846,15 +846,15 @@ internal final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable
         if startAtUnit == .day || startAtUnit == .weekday || startAtUnit == .weekdayOrdinal {
             let targetDay = dateComponent(.day, from: updatedDate)
             var currentDay = targetDay
-            var udate = updatedDate
+            var update = updatedDate
             var prev: Date
             repeat {
-                prev = udate
```
itingliu, line 852: I personally prefer not having a particular commit just to change internal variable names, even if they're not perfect. I'd only do so when there's a strong reason for changing them, or do it as part of a bug fix or feature work. This way we still have access to the commit history.

MojtabaHs: I actually combined multiple commits into a single PR to address this concern. These are all typos in **Calendar** within **FoundationEssentials** that Xcode underlines. Please let me know if you’d like me to drop any of these comments from the PR or rebase the result.

Thanks!

P.S. There is a type named "UDate"Format, so this particular typo might cause some confusion for readers.

PR #1596, opened 2025-11-13
Update automerger to merge from 6.3 to main

itingliu:
Now that we have branched for 6.3, update automerger to merge from 6.3 to main.

While we are not currently running automerger to merge into the release branch, update the version in `automerge_to_release.yml` as well for consistency.

As part of this, we will also disable merge -> future workflow from github action.

### DIFF
--- a/.github/workflows/automerge_to_main.yml
+++ b/.github/workflows/automerge_to_main.yml
@@ -13,7 +13,7 @@ jobs:
     uses: swiftlang/github-workflows/.github/workflows/create_automerge_pr.yml@main
     with:
       base_branch: main
-      head_branch: release/6.2
+      head_branch: release/6.3
     permissions:
       contents: write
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

jmschonfeld:
I think we'll want to target this at the `main` branch even though it's related to the release branch (since that's where GitHub will pick up the configuration for auto mergers that run on a schedule)

itingliu:
> I think we'll want to target this at the `main` branch even though it's related to the release branch (since that's where GitHub will pick up the configuration for auto mergers that run on a schedule)

Sure. I was just thinking that 6.3 will get forwarded to main, so it'll end up there anyways... but I updated it!

itingliu:
@swift-ci please test

PR #1597, opened 2025-11-13
Fix crashes caused by a race condition in icu::OlsonTimeZone::initTransitionRules

itingliu:
ICU's `UTimeZone` isn't actually thread safe. Lock around the access.

I was able to reproduce the UAF crash by calling uatimezone API concurrently. With this fix, I no longer reproduce the crash.

164388304


itingliu:
@swift-ci please test

jmschonfeld (approved):

itingliu:
@swift-ci please test

jmschonfeld:

Issue #1604, opened 2025-11-17
`JSONDecoder` throws a broken error when parsing decimal value as `Int`


hiragram:
**Describe the bug**

`JSONDecoder` throws an error when the value is decimal but the expected type is `Int`. In the final release of Swift 5, the error represents the situation correctly, but not in Swift 6.

The behavior is slightly different between macOS and Linux.

**To Reproduce**

**Minimal code:**

```swift
import Foundation

struct Object: Decodable {
    var foo: Int
}

let json = """
{
  "foo": 123.45
}
""".data(using: .utf8)!

do {
    _ = try JSONDecoder().decode(Object.self, from: json)
} catch let error {
    print(error)
}
```

**Result:**

|Target|Swift|Result|`codingPath`|`debugDescription`|
|:---:|:---:|:---|:---:|:---:|
|Linux|6.2.1|`dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: nil))`|:x:|:x:|
|Linux|6.0|`dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: nil))`|:x:|:x:|
|Linux|5.10.1|`dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "foo", intValue: nil)], debugDescription: "Parsed JSON number <123.45> does not fit in Int.", underlyingError: nil))`|:white_check_mark:|:white_check_mark:|
|macOS|6.2|`dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "The given data was not valid JSON.", underlyingError: Optional(Error Domain=NSCocoaErrorDomain Code=3840 "Number 123.45 is not representable in Swift." UserInfo={NSDebugDescription=Number 123.45 is not representable in Swift.})))`|:x:|:question:<br>available in `underlyingError`|

**Expected behavior**

- `codingPath` should be correctly provided.
- `debugDescription` should tell us what is actually happening.
- ❓ On macOS, `underlyingError` provides more detailed information, but `debugDescription does not`, although I believe it should.

**Configuration (please complete the following information):**

See the table above.

**Regression information:**

- On Linux:
  - Swift 5.10.1 works fine.
  - Swift 6.0 is the first broken version.
  - Swift 6.2.1 is the latest release and still broken.
- On macOS:
  - Swift 6.2 is broken.

[...]

PR #1606, opened 2025-11-17
Fix broken errors thrown by `JSONDecoder` when parsing decimal as `Int`

hiragram:
fixes: #1604 

This PR fixes a regression in `JSONDecoder` where attempting to decode incompatible numeric values (e.g., decimal `123.45` as `Int`, or negative `-123` as `UInt`) returns incorrect error information. The bug affects both the `codingPath` (empty instead of correct path) and `debugDescription` (generic message instead of specific error details).

See more detail: #1604

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONDecoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONDecoder.swift
@@ -1000,13 +1000,29 @@ extension JSONDecoderImpl: Decoder {
     }
 
     static private func _slowpath_unwrapFixedWidthInteger<T: FixedWidthInteger>(as type: T.Type, json5: Bool, numberBuffer: BufferView<UInt8>, fullSource: BufferView<UInt8>, digitBeginning: BufferViewIndex<UInt8>, for codingPathNode: _CodingPathNode, _ additionalKey: (some CodingKey)?) throws -> T {
+        // Helper function to create number conversion error
+        func createNumberConversionError() -> DecodingError {
+            #if FOUNDATION_FRAMEWORK
+            let underlyingError: Error? = JSONError.numberIsNotRepresentableInSwift(
+                parsed: String(decoding: numberBuffer, as: UTF8.self)
+            ).nsError
+            #else
+            let underlyingError: Error? = nil
+            #endif
+            return DecodingError.dataCorrupted(DecodingError.Context(
+                codingPath: codingPathNode.path(byAppending: additionalKey),
+                debugDescription: "Parsed JSON number <\(String(decoding: numberBuffer, as: UTF8.self))> does not fit in \(T.self).",
+                underlyingError: underlyingError
+            ))
+        }
+
         // This is the slow path... If the fast path has failed. For example for "34.0" as an integer, we try to parse as either a Decimal or a Double and then convert back, losslessly.
         if let double = Double(prevalidatedBuffer: numberBuffer) {
             // T.init(exactly:) guards against non-integer Double(s), but the parser may
             // have already transformed the non-integer "1.0000000000000001" into 1, etc.
             // Proper lossless behavior should be implemented by the parser.
             guard let value = T(exactly: double) else {
-                throw JSONError.numberIsNotRepresentableInSwift(parsed: String(decoding: numberBuffer, as: UTF8.self))
+                throw createNumberConversionError()
             }
 
             // The distance between Double(s) is >=2 from ±2^53.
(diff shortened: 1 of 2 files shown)


omochi:
@swift-ci please test

Sources/FoundationEssentials/JSON/JSONDecoder.swift
```diff
@@ -1000,13 +1000,29 @@ extension JSONDecoderImpl: Decoder {
     }
 
     static private func _slowpath_unwrapFixedWidthInteger<T: FixedWidthInteger>(as type: T.Type, json5: Bool, numberBuffer: BufferView<UInt8>, fullSource: BufferView<UInt8>, digitBeginning: BufferViewIndex<UInt8>, for codingPathNode: _CodingPathNode, _ additionalKey: (some CodingKey)?) throws -> T {
+        // Helper function to create number conversion error
+        func createNumberConversionError() -> DecodingError {
+            #if FOUNDATION_FRAMEWORK
```
itingliu, line 1005: I don't think we need to preserve this existing (broken) behavior for FOUNDATION_FRAMEWORK as long as `underlyingError`'s error code is untouched... @kperryua what do you think?

itingliu:
@swift-ci please test

itingliu:
Sorry, I meant to click "comment" but clicked "close" accidentally.

itingliu:
@swift-ci please test

Sources/FoundationEssentials/JSON/JSONDecoder.swift
hiragram: @itingliu Thanks for your review. Just to confirm, are you saying that `underlyingError` should always be
  nil on all platforms in this case?

There is existing code that switches the presence of `underlyingError` based on the `FOUNDATION_FRAMEWORK` flag, which I used as a reference. I don't fully understand the intent behind it, but I think it would be better to keep the behavior consistent.

https://github.com/hiragram/swift-foundation/blob/fix-jsondecoder/Sources/FoundationEssentials/JSON/JSONDecoder.swift#L388-L393

omochi:
@swift-ci please test windows platform

hiragram:
@itingliu anything to do before this getting merged?

hiragram:
@itingliu ping 😄

Sources/FoundationEssentials/JSON/JSONDecoder.swift
kperryua: In retrospect, a `JSONError` seems incorrect here to begin with, as we've already pre-validated the buffer and successfully parsed some a number. Hence, it's valid JSON. (If we fail to parse, we call `validateNumber` afterwards, which gives us a JSONError that tells us exactly what's wrong.)

We should just be throwing the `dataCorrupted` error with no underlying error at all, on all platforms.

hiragram: @kperryua Should that be handled in this Pull Request? Personally, I think it's better to first fix the specific broken behaviors and have the discussion about what the "grand design" should look like in a separate thread.

kperryua: I wouldn’t consider this a “grand design” change. It’s a simplification of this PR by virtue of realigning this particular error throwing instance with the correct layering of errors. Nothing about the PR needs to change except eliding the creation of the underlying error (and the corresponding test code).

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
```diff
@@ -901,6 +901,122 @@ private struct JSONEncoderTests {
...
+                "debugDescription should mention the number 123.45, got: \(context.debugDescription)"
+            )
+            #expect(
+                context.debugDescription.contains("Int") || context.debugDescription.contains("fit"),
+                "debugDescription should mention Int or fit, got: \(context.debugDescription)"
+            )
+
+            // Verification 3 (macOS only): Check if underlyingError is provided
+            #if FOUNDATION_FRAMEWORK
+            #expect(context.underlyingError != nil, "macOS should provide underlyingError, got nil")
```
kperryua, line 945: This is a bit of a nitpick, but calling `FOUNDATION_FRAMEWORK`-specific code "macOS-only" is inaccurate, as this code will apply to any Apple platform using Apple SDKs.

Tests/FoundationEssentialsTests/JSONEncoderTests.swift
```diff
@@ -901,6 +901,122 @@ private struct JSONEncoderTests {
...
+        let json = """
+        {
+          "foo": 123.45
+        }
+        """.data(using: .utf8)!
+
+        do {
+            _ = try JSONDecoder().decode(Object.self, from: json)
+            Issue.record("Expected decoding error but decoding succeeded")
+        } catch let error as DecodingError {
```
kperryua, line 919: `#expect(throws:)` ?

PR #1609, opened 2025-11-19
Change return type of gid function from uid_t to gid_t

LumioseSil:
This seems to be a typo. In practice on macOS at least, they both alias to the same type, but still.


jmschonfeld:
@swift-ci please test

AZero13:
Yay! Can we please merge? @jmschonfeld

PR #1616, opened 2025-12-02
Use the canonicalized name to create ICU timezone object

itingliu:
ICU expects time zone identifiers to be IANA identifiers, such as "America/Los_Angeles". Hence names such as "UTC+9" aren't officially supported.

There is a canonicalization function available, but I did not remember to use it when switching to the new ICU `uatimezone` SPI.

Fix this by always using the canonicalized version to initiate an ICU timezone object.

164490980


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1618, opened 2025-12-03
Fix a corner case where createDirectory would fail on windows

daveinglis:
The //?/ prefix needs to be added to directories if the length is greater than MAX_PATH - 12, however the PATHCCH_ALLOW_LONG_PATHS to PathAllocCanonicalize would only do that if the path was greater than MAX_PATH. This change uses PATHCCH_ENSURE_IS_EXTENDED_LENGTH_PATH, which forces it on all the time.  Just needed to remove the prefix in a few places as it was making it way out into tests as this is suppose to be transparent to the user.

This also fixes createDirectory() "withIntermediateDirectories: true" for long paths on Windows.


daveinglis:
@swift-ci test

PR #1623, opened 2025-12-05
Fix FileManager's extended attribute symlink handling

LumioseSil:
The if statement appears inverted:  setxattr follows simlinks: lsetxattr does not.

Additionally, _extendedAttributes was ignoring simlinks entirely.

Both of these issues have been addressed.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -522,21 +522,21 @@ extension _FileManagerImpl {
     
     private func _extendedAttributes(at path: UnsafePointer<CChar>, followSymlinks: Bool) throws -> [String : Data]? {
         #if canImport(Darwin)
-        var size = listxattr(path, nil, 0, 0)
+        var size = listxattr(path, nil, 0, followSymlinks ? 0 : XATTR_NOFOLLOW)
         #elseif os(FreeBSD)
         var size = (followSymlinks ? extattr_list_file : extattr_list_link)(path, EXTATTR_NAMESPACE_USER, nil, 0)
         #else
-        var size = listxattr(path, nil, 0)
+        var size = followSymlinks ? listxattr(path, nil, 0) : llistxattr(path, nil, 0)
         #endif
         guard size > 0 else { return nil }
         let keyList = UnsafeMutableBufferPointer<CChar>.allocate(capacity: size)
         defer { keyList.deallocate() }
         #if canImport(Darwin)
-        size = listxattr(path, keyList.baseAddress!, size, 0)
+        size = listxattr(path, keyList.baseAddress!, size, followSymlinks ? 0 : XATTR_NOFOLLOW)
         #elseif os(FreeBSD)
-        size = (followSymlinks ? extattr_list_file : extattr_list_link)(path, EXTATTR_NAMESPACE_USER, nil, 0)
+        size = (followSymlinks ? extattr_list_file : extattr_list_link)(path, EXTATTR_NAMESPACE_USER, keyList.baseAddress!, size)
         #else
-        size = listxattr(path, keyList.baseAddress!, size)
+        size = followSymlinks ? listxattr(path, keyList.baseAddress!, size) : llistxattr(path, keyList.baseAddress!, size)
         #endif
         guard size > 0 else { return nil }
         
@@ -553,7 +553,7 @@ extension _FileManagerImpl {
             }
             #endif
             
-            if let value = try _extendedAttribute(current, at: path, followSymlinks: false) {
+            if let value = try _extendedAttribute(current, at: path, followSymlinks: followSymlinks) {
                 extendedAttrs[currentKey] = value
             }
         }
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -193,15 +193,15 @@ extension _FileManagerImpl {
             #else
             var result: Int32
             if followSymLinks {
-                result = lsetxattr(path, key, buffer.baseAddress!, buffer.count, 0)
-            } else {
                 result = setxattr(path, key, buffer.baseAddress!, buffer.count, 0)
+            } else {
+                result = lsetxattr(path, key, buffer.baseAddress!, buffer.count, 0)
             }
             #endif
 
             #if os(macOS) && FOUNDATION_FRAMEWORK
-            // if setxaddr failed and its a permission error for a sandbox app trying to set quaratine attribute, ignore it since its not
-            // permitted, the attribute will be put on the file by the quaratine MAC hook
+            // if setxattr failed and its a permission error for a sandbox app trying to set quarantine attribute, ignore it since its not
+            // permitted, the attribute will be put on the file by the quarantine MAC hook
             if result == -1 && errno == EPERM && _xpc_runtime_is_app_sandboxed() && strcmp(key, "com.apple.quarantine") == 0 {
                 return
             }
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -1046,6 +1046,60 @@ private struct FileManagerTests {
         }
     }
 
+    // Extended attributes are not supported on all platforms
+    #if !os(Windows) && !os(WASI) && !os(OpenBSD) && !canImport(Android)
+    @Test func extendedAttributesOnSymlinks() async throws {
+        let xattrKey = FileAttributeKey("NSFileExtendedAttributes")
+        #if os(Linux)
+        // Linux requires the user.* namespace prefix for regular files
+        let attrName = "user.swift.foundation.symlinktest"
+        #else
+        let attrName = "org.swift.foundation.symlinktest"
+        #endif
+        let attrValue = Data([0xAA, 0xBB, 0xCC])
+        let attrValue2 = Data([0xDD, 0xEE, 0xFF])
+
+        try await FilePlayground {
+            File("target", contents: Data("payload".utf8))
+            SymbolicLink("link", destination: "target")
+        }.test { fileManager in
+            // First, validate that reading the attribute from the link does not read the value from the target
+            do {
+                try fileManager.setAttributes([xattrKey: [attrName: attrValue]], ofItemAtPath: "target")
+                let targetAttrs = try fileManager.attributesOfItem(atPath: "target")
+                let targetXAttrs = targetAttrs[xattrKey] as? [String: Data]
+                #expect(targetXAttrs?[attrName] == attrValue)
+                
+                let linkAttrs = try fileManager.attributesOfItem(atPath: "link")
+                let linkXAttrs = linkAttrs[xattrKey] as? [String: Data]
+                #expect(linkXAttrs?[attrName] == nil)
+            }
+
+            // Attempt to set xattrs on the symlink
+            #if os(Linux)
+            // On Linux, user xattrs cannot be set on symlinks
+            #expect(throws: CocoaError.self) {
+                try fileManager.setAttributes([xattrKey: [attrName: attrValue2]], ofItemAtPath: "link")
+            }
+            let expectedValue: Data? = nil
+            #else
+            try fileManager.setAttributes([xattrKey: [attrName: attrValue2]], ofItemAtPath: "link")
+            let expectedValue: Data? = attrValue2
+            #endif
+            
+            // Ensure that reading back the xattr of the link produces the expected value
+            let linkAttrs = try fileManager.attributesOfItem(atPath: "link")
+            let linkXAttrs = linkAttrs[xattrKey] as? [String: Data]
+            #expect(linkXAttrs?[attrName] == expectedValue)
+            
+            // Ensure that setting the xattr on the link did not set the xattr on the target
+            let targetAttrs = try fileManager.attributesOfItem(atPath: "target")
+            let targetXAttrs = targetAttrs[xattrKey] as? [String: Data]
+            #expect(targetXAttrs?[attrName] == attrValue)
+        }
+    }
+    #endif
+
     #if !canImport(Darwin) || os(macOS)
     @Test func currentUserHomeDirectory() async throws {
         let userName = ProcessInfo.processInfo.userName


Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -193,15 +193,15 @@ extension _FileManagerImpl {
             #else
             var result: Int32
             if followSymLinks {
-                result = lsetxattr(path, key, buffer.baseAddress!, buffer.count, 0)
-            } else {
                 result = setxattr(path, key, buffer.baseAddress!, buffer.count, 0)
+            } else {
+                result = lsetxattr(path, key, buffer.baseAddress!, buffer.count, 0)
```
jmschonfeld, line 198: Is there a unit test that could be added to validate this behavior to confirm that the change works as intended and it doesn't regress in the future? The `FileManager` test suite has some scaffolding to support creating a test directory with symlinks and calling file manager operations on files

AZero13: Yeah I'll add it

AZero13:
@jmschonfeld ping?

AZero13:
@jmschonfeld @compnerd ping?

jmschonfeld:
> @jmschonfeld @compnerd ping?

It's on my list to circle back and review when I have a moment - no need to ping 🙂. Will be working my way through the handful of PRs open to review

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,43 @@ private struct FileManagerTests {
         }
     }
 
+    @Test func extendedAttributesDoNotFollowSymlinksWhenSetting() async throws {
+#if os(Windows) || os(WASI) || os(OpenBSD) || canImport(Android)
+        throw Skip("Extended attributes not supported on this platform")
```
jmschonfeld: I didn't recognize this API and it looks like it's indeed not an API that exists in swift-testing based on CI:

```
error: cannot find 'Skip' in scope
```

You can look at some of the other tests in this file for the pattern for skipping platform-dependent tests - usually we just wrap the test function in the `#if` conditional since there's not a great way to skip tests on a platform-by-platform basis with swift-testing

AZero13:
I fixed it, but also noticed we weren't passing followSymlinks to _extendedAttributes when the test failed on Darwin so. @compnerd

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,70 @@ private struct FileManagerTests {
...
+                let linkAttrs = try? fileManager.attributesOfItem(atPath: "link"),
+                let targetAttrs = try? fileManager.attributesOfItem(atPath: "target")
+            else { return }
+
+            let linkXattrs = linkAttrs[xattrKey] as? [String: Data]
+            let targetXattrs = targetAttrs[xattrKey] as? [String: Data]
+
+            // If neither reports xattrs, treat as unsupported on this FS/runtime.
+            if linkXattrs == nil && targetXattrs == nil {
+                return
```
jmschonfeld: I think in this case and the case above when calling `attributesOfItem` we likely don't want to just `return` if the call fails / the value is `nil`. If we successfully set the extended attributes above (i.e. `setAttributes` didn't throw) we should expect that we can then read that extended attribute since any system that doesn't support this operation would have already exited the test. Do you see another reason to allow for the test to pass when writing succeeds but reading fails? Or instead, should we just call the API with `try` and `#require` the attribute values so that if we succeed writing the attributes but fail reading them will cause the test to fail and report the regression

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,67 @@ private struct FileManagerTests {
...
+
+            let linkAttrs = try fileManager.attributesOfItem(atPath: "link")
+            let targetAttrs = try fileManager.attributesOfItem(atPath: "target")
+
+            let linkXattrs = linkAttrs[xattrKey] as? [String: Data]
+            let targetXattrs = targetAttrs[xattrKey] as? [String: Data]
+
+            try #require(
+                linkXattrs != nil || targetXattrs != nil,
+                "Expected extended attributes after successful setAttributes call")
```
jmschonfeld: Typically we use `#require` to unwrap values:

```suggestion
            let linkXattrs = try #require(linkAttrs[xattrKey] as? [String: Data])
            let targetXattrs = try #require(targetAttrs[xattrKey] as? [String: Data])
```

That way you can remove the `if` check below because the `else` branch will never run, but using `#require` to unwrap values enforces this at compile time instead of runtime.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -522,19 +522,23 @@ extension _FileManagerImpl {
...
     private func _extendedAttributes(at path: UnsafePointer<CChar>, followSymlinks: Bool) throws -> [String : Data]? {
         #if canImport(Darwin)
-        var size = listxattr(path, nil, 0, 0)
+        var size = listxattr(path, nil, 0, followSymlinks ? 0 : XATTR_NOFOLLOW)
         #elseif os(FreeBSD)
         var size = (followSymlinks ? extattr_list_file : extattr_list_link)(path, EXTATTR_NAMESPACE_USER, nil, 0)
+        #elseif os(Linux)
+        var size = followSymlinks ? listxattr(path, nil, 0) : llistxattr(path, nil, 0)
         #else
         var size = listxattr(path, nil, 0)
```
jmschonfeld: With the added `os(Linux)` branch above, what platforms is this line applicable to? And do they not have an `llistxattr` version?

AZero13: llistxattr is a linux extension in particular.

jmschonfeld: In specific, what platform is this `#else` branch for that's not a Linux platform with `llistxattr` or the other excluded platforms?

AZero13: I have no idea: it's just in case.

jmschonfeld: I think if we're going to leave this code in here then that needs investigation. It seems to go against the spirit of this change (which is to respect the `followSymlinks` flag) to leave a case where we don't respect the flag.

If there is a platform that needs this because we can't represent that behavior we should document that with a comment, or if there aren't platforms that need this then we should remove this (to enforce this at compile time). That being said, given that `_setAttributes` does not require this branch (the `#else` branch just uses `lsetxattr` outright) I suspect there is not a platform that actually needs this so removing it is likely fine, but I have not verified this.

AZero13: True. Okay, well, then llistxattr it is!

jmschonfeld:
@swift-ci please test

AZero13:
@jmschonfeld Can you try one more time? TestPM failed and I addressed a typo.

AZero13:
Fixed error about ENOSUP not being in Linux and refactored so it uses guard statements now. @jmschonfeld So the tests should pass now.

AZero13:
@jmschonfeld Can we please test and merge?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
```diff
@@ -522,27 +522,27 @@ extension _FileManagerImpl {
...
         #else
-        size = listxattr(path, keyList.baseAddress!, size)
+        size = followSymlinks ? listxattr(path, keyList.baseAddress!, size) : llistxattr(path, keyList.baseAddress!, size)
         #endif
         guard size > 0 else { return nil }
         
         var extendedAttrs: [String : Data] = [:]
         var current = keyList.baseAddress!
-        let end = keyList.baseAddress!.advanced(by: keyList.count)
+        let end = keyList.baseAddress!.advanced(by: size)
```
jmschonfeld: It looks like this was changed in a recent commit (hard to determine with force pushes rather than follow-on commits which are easier to review). Is there a reason you're switching to use the passed in `size` here rather than using the size of the allocation returned by the `allocate` function?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,57 @@ private struct FileManagerTests {
...
+        }.test { fileManager in
+            // Attempt to set xattrs on the symlink; if unsupported, bail quietly.
+            do {
+                try fileManager.setAttributes(
+                    [xattrKey: [attrName: attrValue]], ofItemAtPath: "link")
+            } catch let error as CocoaError {
+                if error.code == .featureUnsupported { return }
+                guard let posix = error.underlying as? POSIXError else { throw error }
+                #if canImport(Darwin)
+                guard posix.code == .ENOTSUP || posix.code == .EOPNOTSUPP else { throw error }
```
jmschonfeld: Is there a reason why `ENOTSUP` is guarded for `Darwin` only here?

AZero13: Because ENOTSUP is not an error code on Linux.

I tried and it didn't compile.

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
AZero13: Yes — the size variable gets updated by the second listxattr/extattr_list_* call to reflect the actual number of bytes written.

jmschonfeld: In that case if `size` is potentially different (if the file system changes and the number of attributes are removed) can we add an assertion that `size <= keyList.count`? In most cases I assume this shouldn't be an issue, but I'm wary that trusting this value could in some edge case inadvertently introduce a buffer overread here if `size` somehow becomes something larger than our allocation after it is overwritten.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jmschonfeld: `ENOTSUP` is an error code on Linux as far as I can tell if I'm reading the man pages correctly, but it looks like it has the same value as `EOPNOTSUPP` so there's no distinction. It looks like we also don't have a `POSIXErrorCode.EOPNOTSUP` for it which is likely the compilation error you were seeing (either intentionally because it's not necessary or just because nobody has ended up needing it). If that's the case, can you leave a comment about why this distinction is here so it's clear that `ENOTSUP` is not distinct from `EOPNOTSUPP` on non-Darwin platforms?

Sources/FoundationEssentials/FileManager/FileManager+Files.swift
AZero13: If anything I'd say the opposite is more likely to happen: if keyList.count changes between allocation and read then this can happen.

jmschonfeld: I don't think I'm fully following - in what situation would `keyList.count` change?

AZero13: Maybe this is best for another PR. I'll revert it.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
AZero13: Done

AZero13:
I think this is as heavy as change now as it should get. We should merge this

I'll save the count and size replacement for another PR. That has been reverted back @jmschonfeld

jmschonfeld:
@swift-ci test

AZero13:
Test failed because of a permission error on linux. @jmschonfeld what to do? How am I supposed to fix that in the test

AZero13:
Ugh I had to fix the tests to use rawvalue. The previous version passed locally, but not on the CI. My bad...

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,85 @@ private struct FileManagerTests {
...
+            SymbolicLink("link", destination: "target")
+        }.test { fileManager in
+            // First, prove that this environment supports xattrs on regular files and that we
+            // have permission to set them. If this fails, the symlink behavior isn't meaningful.
+            do {
+                try fileManager.setAttributes([xattrKey: [probeName: probeValue]], ofItemAtPath: "target")
+            } catch let error as CocoaError {
+                if error.code == .featureUnsupported { return }
+                guard let posix = error.underlying as? POSIXError else { throw error }
+                #if canImport(Darwin)
```
jmschonfeld: On the `ENOTSUP` vs. `EOPNOTSUPP` here, if we're just using the raw values instead I think you can switch this to just

```swift
posix.code.rawValue == EOPNOTSUPP || posix.code.rawValue == ENOTSUP
```

for all platforms and call it good since the raw values exist on both platforms, which would simplify this code a lot. Is that correct?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1046,6 +1046,85 @@ private struct FileManagerTests {
...
+                guard posix.code == .ENOTSUP || posix.code.rawValue == EOPNOTSUPP || posix.code == .EPERM else { throw error }
```
jmschonfeld: On the `EPERM` issue - yes this seems like it somewhat defeats the purpose of the test, since if permission is always denied then we're always skipping this test effectively. Have you investigated why the `EPERM` error is occurring or do we know why we're getting this error? Perhaps there's something about the test file that should be updated to allow this operation, or does this signify that this operation in general isn't the right direction?

AZero13: Yeah, I think we should just use the raw values. That is what I did.

AZero13: How about having it just for the probe though?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
AZero13: Yeah

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jmschonfeld: I'm referring to the need to check for `EPERM` at all here. Do you know why we're failing with an `EPERM` error when trying to set these attributes? I'd like to better understand in which situations this is actually possible and which situations it is expected to fail for one reason or another. We need to be careful here to make sure that introducing this change does not cause compatibility issues with clients expecting these calls to succeed if the calls begin to fail because the operation is no longer possible.

AZero13: The operation itself is not reliably possible across platforms/filesystems.

AZero13: they already can fail with EPERM depending on filesystem/platform because the implementation calls lsetxattr() for symlinks.

AZero13: Is the CI running in a sandbox?

jmschonfeld: > they already can fail with EPERM depending on filesystem/platform because the implementation calls lsetxattr() for symlinks.

Right what I'm driving towards is figuring out _why_ the can fail with that message. Is it similar to the `EOPNOTSUPP` case where the file system or operating system does not permit the operation on a symlink, and if so which are those? We've already excluded some operating systems for this reason, are there more? Is it a misconfiguration in the `FileManagerPlayground` setup that creates the files incorrectly? The problem with ignoring `EPERM` without understanding why is that we could be simply avoiding testing this behavior across all platforms leaving this change untested and I want to confirm where we expect this to succeed and ensure that tests are running there.

> Is the CI running in a sandbox?

I'm not certain - it's possible that it is, or possible that it varies based on platform. That being said, the process running the tests creates the test files themselves and I don't think a sandbox would permit us to create files that we later would be unable to access but I have not validated that.

AZero13: We still check that the target is untouched, which is the core behavioral guarantee. We are not disabling the test across all platforms.

jmschonfeld: That's not the behavior of the code as-is today. I think the checks for the various error codes are complicating the test here a bit. I discussed with @kperryua and @jrflat and after some research and experimentation we determined that:

- The `EOPNOTSUPP` shouldn't be relevant here as it's only relevant for usage with sockets
- The `ENOTSUP` shouldn't happen here, since this test is only running on platforms where this operation should be supported
- The `EPERM` error seems a bit weird at first, but it seems to be an artifact of the constraint that Linux does not support user-defined xattrs on symlinks themselves

Based on the above, I think we should update the unit test to:

- No longer account for `EOPNOTSUPP` and `ENOTSUP` as they are not relevant here
- Test that we fail due to the `EPERM` error only on Linux
- Still validate the reading side even if the setting side fails

I took a moment to write up my suggestion for how to write this unit test based on the above plus cleaning it up just a little bit to simplify the branches and rename it to match what we're testing:

```swift
    // Extended attributes are not supported on all platforms
    #if !os(Windows) && !os(WASI) && !os(OpenBSD) && !canImport(Android)
    @Test func extendedAttributesOnSymlinks() async throws {
        let xattrKey = FileAttributeKey("NSFileExtendedAttributes")
        #if os(Linux)
        // Linux requires the user.* namespace prefix for regular files
        let attrName = "user.swift.foundation.symlinktest"
        #else
        let attrName = "org.swift.foundation.symlinktest"
        #endif
        let attrValue = Data([0xAA, 0xBB, 0xCC])
        let attrValue2 = Data([0xDD, 0xEE, 0xFF])

        try await FilePlayground {
            File("target", contents: Data("payload".utf8))
            SymbolicLink("link", destination: "target")
        }.test { fileManager in
            // First, validate that reading the attribute from the link does not read the value from the target
            do {
                try fileManager.setAttributes([xattrKey: [attrName: attrValue]], ofItemAtPath: "target")
                let targetAttrs = try fileManager.attributesOfItem(atPath: "target")
                let targetXAttrs = targetAttrs[xattrKey] as? [String: Data]
                #expect(targetXAttrs?[attrName] == attrValue)
                
                let linkAttrs = try fileManager.attributesOfItem(atPath: "link")
                let linkXAttrs = linkAttrs[xattrKey] as? [String: Data]
                #expect(linkXAttrs?[attrName] == nil)
            }

            // Attempt to set xattrs on the symlink
            #if os(Linux)
            // On Linux, user xattrs cannot be set on symlinks
            #expect(throws: CocoaError.self) {
                try fileManager.setAttributes([xattrKey: [attrName: attrValue2]], ofItemAtPath: "link")
            }
            let expectedValue: Data? = nil
            #else
            try fileManager.setAttributes([xattrKey: [attrName: attrValue2]], ofItemAtPath: "link")
            let expectedValue: Data? = attrValue2
            #endif
            
            // Ensure that reading back the xattr of the link produces the expected value
            let linkAttrs = try fileManager.attributesOfItem(atPath: "link")
            let linkXAttrs = linkAttrs[xattrKey] as? [String: Data]
            #expect(linkXAttrs?[attrName] == expectedValue)
            
            // Ensure that setting the xattr on the link did not set the xattr on the target
            let targetAttrs = try fileManager.attributesOfItem(atPath: "target")
            let targetXAttrs = targetAttrs[xattrKey] as? [String: Data]
            #expect(targetXAttrs?[attrName] == attrValue)
        }
    }
    #endif
```

What do you think about that? That way we always test that the reading side reads from the symlink and not the target and on the setting side it fails on Linux but otherwise succeeds and that on all platforms writing to the symlink doesn't write over the target. Does that unit test sound good to you?

AZero13: Yes!

jmschonfeld:
@swift-ci please test

AZero13:
It passed @jmschonfeld

AZero13:
@jmschonfeld Is this good to go?

jmschonfeld:
Apologies for the delay here. I was thinking on this behavior a bit more and I realized I may have misunderstood the base of the problem here. Today it looks like the behavior is reading xattrs always follows symlinks and setting xattrs does not follow symlinks on Darwin but does follow symlinks on Linux, all because it either ignored `followSymlinks:` or inverted it. Your change originally seemed reasonable which updates the functions to respect the `followSymlinks:` parameter which is always passed in as `false` meaning that these will now always not follow symlinks.

However, the rest of setAttributes/attributesForItem where these are called from will always traverse symlinks. In general, most of the FileManager APIs traverse symlinks (compared to URL APIs which typically do not traverse symlinks). So I think while this change is headed in the right direction by respecting the `followSymlinks:` flag, I think it uncovers another issue: that we are passing `followSymlinks: false`. This corrects the behavior of this function, but due to this I think it will introduce an inconsistency in APIs like setAttributes/attributesForItem.

What do you think about changing the call sites to `followSymlinks: true` now that the function behaves correctly? That means the behavior will be consistent with the rest of the FileManager APIs, but will require updating the tests. Thoughts?

AZero13:
I am for it

PR #1624, opened 2025-12-05
[FileManager] Fix off by one error when checking error from read in FileManager+Basics

LumioseSil:
Check for readBytes being equal to -1, not less than -1.

It returns -1 when there's an error.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
@@ -277,7 +277,7 @@ internal struct _FileManagerImpl {
                                     return false
                                 }
                             }
-                            if readBytes < -1 { return false }
+                            if readBytes == -1 { return false }
                             return true
                         }
                     }


AZero13:
@itingliu Hi, could you please check this

AZero13:
@compnerd

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
```diff
@@ -277,7 +277,7 @@ internal struct _FileManagerImpl {
                                     return false
                                 }
                             }
-                            if readBytes < -1 { return false }
+                            if readBytes < 0 { return false }
```
jmschonfeld: Is there a unit test that could be added to validate this behavior to confirm that the change works as intended and it doesn't regress in the future?

AZero13: How am I supposed to make a unit test that causes read, an OS syscall, to fail reliably?

jmschonfeld: There are a number of ways that `read` can fail which we can test for, but it looks like in this particular API since we already have a `stat` call on this file earlier there aren't many reliable ways that `read` could fail that wouldn't have already caused the `stat` call to fail (or to return information that results in us not taking this code path). This scenario likely only happens if `read` fails with something like `EINTR` which we can't test without a potential race condition. Based on that, it's probably fine to leave without a specific test

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileManager+Basics.swift
glessard: This should be `readBytes == -1`
"Upon successful completion, these functions shall return a non-negative integer indicating the number of bytes actually read. Otherwise, the functions shall return -1 and set errno to indicate the error." (POSIX specifications issue 8 – similar on Darwin.)

jmschonfeld:
@swift-ci please test

glessard:
Thanks for finding this, @AZero13!

PR #1625, opened 2025-12-05
Ensure there is a null byte at the end of the JSON

LumioseSil:
In JSON5 mode, a top-level number (e.g. "123") is left without a guaranteed terminator.

Do what JSONScanner does in this case.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSON5Scanner.swift
+++ b/Sources/FoundationEssentials/JSON/JSON5Scanner.swift
@@ -123,7 +123,14 @@ internal struct JSON5Scanner {
             throw JSONError.unexpectedCharacter(context: "after top-level value", ascii: char, location: reader.sourceLocation)
         }
 
-        return JSONMap(mapBuffer: partialMap.mapData, dataBuffer: self.reader.bytes)
+        let map = JSONMap(mapBuffer: partialMap.mapData, dataBuffer: self.reader.bytes)
+
+        // If the input contains only a number, ensure a trailing NUL byte is available for strtod/strtof parsing.
+        if case .number = map.loadValue(at: 0)! {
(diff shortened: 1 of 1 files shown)


kperryua:
@swift-ci please test

parkera:
Can we add a unit test please?

Sources/FoundationEssentials/JSON/JSON5Scanner.swift
```diff
@@ -123,7 +123,14 @@ internal struct JSON5Scanner {
             throw JSONError.unexpectedCharacter(context: "after top-level value", ascii: char, location: reader.sourceLocation)
         }
 
-        return JSONMap(mapBuffer: partialMap.mapData, dataBuffer: self.reader.bytes)
+        let map = JSONMap(mapBuffer: partialMap.mapData, dataBuffer: self.reader.bytes)
+
+        // If the input contains only a number, ensure a trailing NUL byte is available for strtod/strtof parsing.
+        if case .number = map.loadValue(at: 0)! {
```
parkera, line 129: Is it ok to force unwrap here?

kperryua: Yes. The "root" value is always at this index in the map, and if there wasn't any valid value, we would have thrown an error before this.

itingliu:
@swift-ci please test

jmschonfeld:

jmschonfeld:
https://github.com/swiftlang/swift-foundation/pull/1706 causes this bug to trigger in the `json5Numbers` unit test (because it removes extra zero padding after the end of the top level number bytes) so if we merge that then this will be unit tested in `json5Numbers` if we don't have a specific test for this PR

PR #1626, opened 2025-12-05
SequenceAllSatisfy is mapped to SequenceContainsWhere, and not SequenceAllSatisfy

LumioseSil:
This is a typo. SequenceAllSatisfy should map to SequenceAllSatisfy.

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
+++ b/Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
@@ -391,7 +391,7 @@ extension PredicateCodableConfiguration {
         configuration.allowPartialType(PredicateExpressions.Range<PredicateExpressions.Value<Int>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.Range")
         configuration.allowPartialType(PredicateExpressions.SequenceContains<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.SequenceContains")
         configuration.allowPartialType(PredicateExpressions.SequenceContainsWhere<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceContainsWhere")
-        configuration.allowPartialType(PredicateExpressions.SequenceContainsWhere<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceAllSatisfy")
+        configuration.allowPartialType(PredicateExpressions.SequenceAllSatisfy<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceAllSatisfy")
         configuration.allowPartialType(PredicateExpressions.SequenceStartsWith<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<[Int]>>.self, identifier: "PredicateExpressions.SequenceStartsWith")
         configuration.allowPartialType(PredicateExpressions.SequenceMaximum<PredicateExpressions.Value<[Int]>>.self, identifier: "PredicateExpressions.SequenceMaximum")
         configuration.allowPartialType(PredicateExpressions.SequenceMinimum<PredicateExpressions.Value<[Int]>>.self, identifier: "PredicateExpressions.SequenceMinimum")
--- a/Tests/FoundationEssentialsTests/PredicateCodableTests.swift
+++ b/Tests/FoundationEssentialsTests/PredicateCodableTests.swift
@@ -213,6 +213,19 @@ private struct PredicateCodableTests {
         }
     }
     
+    @Test func sequenceAllSatisfyRoundTrip() throws {
+        let predicate = #Predicate<[Int]> {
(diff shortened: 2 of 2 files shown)


AZero13:
@jmschonfeld

Tests/FoundationEssentialsTests/PredicateCodableTests.swift
```diff
@@ -213,6 +213,19 @@ private struct PredicateCodableTests {
...
+        let predicate = #Predicate<[Int]> {
+            $0.allSatisfy { $0 > 0 }
+        }
+        
+        let decoded = try _encodeDecode(predicate, for: StandardConfig.self)
+        
+        #expect(try predicate.evaluate([1, 2, 3]))
+        #expect(try decoded.evaluate([1, 2, 3]))
+        #expect(try predicate.evaluate([-1, 2, 3]) == decoded.evaluate([-1, 2, 3]))
+        #expect(try decoded.evaluate([-1, 2, 3]) == false)
```
jmschonfeld: minor nit: I think these two lines are effectively just trying to confirm that:

```swift
#expect(try !predicate.evaluate([-1, 2, 3]))
#expect(try !decode.evaluate([-1, 2, 3]))
```

Is that correct? That might be a more succinct and understandable way to write it (I was trying to figure out how these two assertions differed from the first 2 since the approach/style was different

jmschonfeld:
@swift-ci please test

AZero13:
> @swift-ci please test

Thank you. I fixed it. Can you run it one more time.

Sources/FoundationEssentials/Predicate/Archiving/PredicateCodableConfiguration.swift
```diff
@@ -391,7 +391,7 @@ extension PredicateCodableConfiguration {
         configuration.allowPartialType(PredicateExpressions.Range<PredicateExpressions.Value<Int>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.Range")
         configuration.allowPartialType(PredicateExpressions.SequenceContains<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Int>>.self, identifier: "PredicateExpressions.SequenceContains")
         configuration.allowPartialType(PredicateExpressions.SequenceContainsWhere<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceContainsWhere")
-        configuration.allowPartialType(PredicateExpressions.SequenceContainsWhere<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceAllSatisfy")
+        configuration.allowPartialType(PredicateExpressions.SequenceAllSatisfy<PredicateExpressions.Value<[Int]>, PredicateExpressions.Value<Bool>>.self, identifier: "PredicateExpressions.SequenceAllSatisfy")
```
jmschonfeld, line 394: Hmm, we might have a compatibility issue here that we need to address. I believe as it stands today, this registers `SequenceContainsWhere` with both the `PredicateExpressions.SequenceContainsWhere` and `PredicateExpressions.SwquenceAllSatisfy` identifiers. `PredicateCodableConfiguration` will take the last identifier registered when the same type is registered with multiple identifiers, so today I think we've been encoding/decoding `PredicateExpressions.SequenceContainsWhere` with the `PredicateExpressions.SequenceAllSatisfy` identifier. Changing this would break any existing archives that already have a serialized `SequenceContainsWhere` stored under the `SequenceAllSatisfy` identifier. It would also break sending a serialized archive from a newer runtime to an application that deserializes it with an older runtime.

I'm not sure how many of those archives exist today in the wild. We can try to quantify that, but we might need to instead think about a compatible way to move forward even if these identifiers are incorrect so that we can solve the problem (allowing serialization of SequenceAllSatisfy) without breaking existing serializations of `SequenceContainsWhere`. Perhaps we leave a comment here about the unfortunate identifier mismatch and use a new identifier for the real `SequenceAllSatisfy` type?

AZero13: Are you okay with the new identifier route

jmschonfeld: Yeah that's probably the best way to go. Let's update the `SequenceContainsWhere` line above to use the incorrect identifier (with a comment explaining why the identifier doesn't match) and then for `SequenceAllSatisfy` we can come up with a new identifier to use.

In terms of a new identifier since the appropriate one is "taken" in existing archives, maybe something like `PredicateExpressions.SequenceAllSatisfy.corrected`? @parkera @itingliu do you have any suggestions? I'm not sure if we've had a pattern before around naming and evolving archive keys that were incorrect historically that we should follow here.

itingliu: I think `PredicateExpressions.SequenceAllSatisfy.corrected` is fine. I think sometimes we also add the version number into the coding key as a way to document when the new key goes into effect, like `PredicateExpressions.SequenceAllSatisfy_6_4`. But I don't have strong opinions on which we choose

jmschonfeld: Sounds good - @AZero13 how about we go with `PredicateExpressions.SequenceAllSatisfy.corrected` as the identifier then? And let's leave a comment here explaining why that's necessary

PR #1628, opened 2025-12-05
Fix BufferReader.lineNumber to correctly count newlines

LumioseSil:
The `BufferReader.lineNumber` property had a bug where it used `fullBuffer[offset: 1]` to check for newline characters. Since `offset:` is relative to the start of the buffer, this always checked the same byte (the second byte) regardless of the current cursor position `p`.

This caused incorrect line counting:
- If the second byte happened to be `\n`, it would increment the count on every iteration, overcounting lines
- If the second byte was not `\n`, standalone newlines (without preceding `\r`) would never be counted, undercounting lines.


glessard (approved):
Thanks for finding this!

glessard:
@swift-ci please test

PR #1629, opened 2025-12-05
Fix OptionalFlatMap evaluation for non-optional transform results

LumioseSil:
The value should be wrapped (or overloads split) instead of force-cast.

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Expressions/Optional.swift
+++ b/Sources/FoundationEssentials/Predicate/Expressions/Optional.swift
@@ -39,11 +39,22 @@ extension PredicateExpressions {
             self.transform = builder(variable)
         }
         
-        public func evaluate(_ bindings: PredicateBindings) throws -> Output {
+        // When the transform produces a non-optional Result, wrap it.
+        public func evaluate(_ bindings: PredicateBindings) throws -> Output where RHS.Output == Result {
             var mutableBindings = bindings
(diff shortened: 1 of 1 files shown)


Sources/FoundationEssentials/Predicate/Expressions/Optional.swift
```diff
@@ -39,11 +39,22 @@ extension PredicateExpressions {
             self.transform = builder(variable)
         }
         
-        public func evaluate(_ bindings: PredicateBindings) throws -> Output {
+        // When the transform produces a non-optional Result, wrap it.
+        public func evaluate(_ bindings: PredicateBindings) throws -> Output where RHS.Output == Result {
```
jmschonfeld, line 43: Can you elaborate more on the intent of this change? The existing `public func evaluate` is a witness to a protocol requirement from `PredicateExpression`. `PredicateExpression` requires that all expressions must be able to be evaluated without additional constraints, so the two new public APIs that you've added no longer satisfy this protocol requirement and I don't expect that it would build successfully. IIRC the split initializers (not protocol requirements) were necessary to "simulate" the compiler's implicit optional promotion when returning a non-optional value from a function with an optional return type which does not apply to generic signature requirements in the same way. The existing force cast in the `evaluate` function performs that optional promotion at runtime like the compiler would have inserted if this were a "real" closure.

jmschonfeld (changes requested):

Issue #1630, opened 2025-12-05
Inconsistent behaviour of CodableConfiguration with optional types


TomaszLizer:
**Describe the bug**
Default behaviour of Decodable types with optionals is to treat missing key or `null` value as expected cases which are parsed to nil without error.
This behaviour seems to be expected for `DecodableWithConfiguration` types as per implementation for Optionals (See [extension Optional : DecodableWithConfiguration where Wrapped : DecodableWithConfiguration](https://github.com/swiftlang/swift-foundation/blob/439b692c331e52d7195b1a37652205a8df0858f8/Sources/FoundationEssentials/CodableWithConfiguration.swift#L227-L237C2))

CodableConfiguration decoding behaviour for Optional types is inconsistent with both `Codable` as well as `DecodableWithConfiguration` types for which `null` value for a key is treated as decoding error.

**To Reproduce**
Sample code to reproduce the behavior.

CodableConfiguration protocol requirements implementation:
```swift
import Foundation

public struct NonCodableType {
    public let value: String

    public init(value: String) {
        self.value = value
    }
}

/// Helper object allowing to decode Date using ISO8601 scheme
public struct CustomConfig: DecodingConfigurationProviding, EncodingConfigurationProviding, Sendable {
    public static let encodingConfiguration = CustomConfig()
    public static let decodingConfiguration = CustomConfig()

    func decode(from decoder: any Decoder) throws -> NonCodableType {
        let container = try decoder.singleValueContainer()
        let value = try container.decode(String.self)
        return NonCodableType(value: value)
    }

    func encode(object: NonCodableType, to encoder: Encoder) throws {
        var container = encoder.singleValueContainer()
        try container.encode(object.value)
    }
}

extension NonCodableType: CodableWithConfiguration {
    public init(from decoder: any Decoder, configuration: CustomConfig) throws {
        self = try configuration.decode(from: decoder)
    }

    public func encode(to encoder: any Encoder, configuration: CustomConfig) throws {
        try configuration.encode(object: self, to: encoder)
    }
}
```

Example code surfacing issue
```swift
import Foundation

struct TestModel: Codable {
    @CodableConfiguration(wrappedValue: nil, from: CustomConfig.self)
    var testObject: NonCodableType?
}

let json = "{\"testObject\": null}"
let jsonData = json.data(using: .utf8)!

let decoder = JSONDecoder()
// Fails with valueNotFound
let sut = try decoder.decode(TestModel.self, from: jsonData)
```

[...]

PR #1631, opened 2025-12-05
Fix codable configuration optionals decoding

TomaszLizer:
This PR handles issue with Optional types decoding using `CodableConfiguration`.
- Adds tests for validation of decoding behaviour of optional type
- Fixes edge case for `null` value for a key

## Context

Expected behaviour based on other examples - like Decodable types decoding is to treat missing key or `null` value as correct and decoding to nil.

Current implementation of `CodableConfiguration` treats `null` value for key as decoding error.

## Note
I was also considering fixing issue by minimally altering original extension by adding ` try !self.decodeNil(forKey: key)` to check however approach in PR seems cleaner to me.
Other considered fix:
```swift
@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
public extension KeyedDecodingContainer {
    func decode<T, C>(_: CodableConfiguration<T?, C>.Type, forKey key: Self.Key) throws -> CodableConfiguration<T?, C> {
        if self.contains(key), try !self.decodeNil(forKey: key) {
            let wrapper = try self.decode(CodableConfiguration<T, C>.self, forKey: key)
            return CodableConfiguration<T?, C>(wrappedValue: wrapper.wrappedValue)
        } else {
            return CodableConfiguration<T?, C>(wrappedValue: nil)
        }
    }
}

```

Closes Issue #1630

### DIFF
--- a/Sources/FoundationEssentials/CodableWithConfiguration.swift
+++ b/Sources/FoundationEssentials/CodableWithConfiguration.swift
@@ -53,14 +53,9 @@ public extension KeyedEncodingContainer {
 @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
 public extension KeyedDecodingContainer {
     func decode<T, C>(_: CodableConfiguration<T?, C>.Type, forKey key: Self.Key) throws -> CodableConfiguration<T?, C> {
-        if self.contains(key) {
-            let wrapper = try self.decode(CodableConfiguration<T, C>.self, forKey: key)
-            return CodableConfiguration<T?, C>(wrappedValue: wrapper.wrappedValue)
-        } else {
-            return CodableConfiguration<T?, C>(wrappedValue: nil)
-        }
+        let wrapper = try self.decodeIfPresent(CodableConfiguration<T?,C>.self, forKey: key)
+        return CodableConfiguration<T?, C>(wrappedValue: wrapper?.wrappedValue)
     }
-
 }
 
 


jmschonfeld (dismissed):

jmschonfeld:
@swift-ci please test

TomaszLizer:
@jmschonfeld Sure - will do that today :)  
Sorry for missing that, but I was unable to build in foundation framework mode at that time hence I guess that is the reason for that.
Will update if any issues faced.

**[EDIT]**: Updated - should be fine now. 
QQ however - is there I way I could simply build and test [`FOUNDATION_FRAMEWORK`](https://github.com/swiftlang/swift-foundation/blob/main/Foundation_Build_Process.md#foundation_framework-build) mode? I experimented a bit with adding flags during initial implementation, but only encountered failures. Actually at that time I was only able to build via `swift test` 🤔

TomaszLizer:
@swift-ci please test

**[EDIT]:** Expected it won't work but had to try :D

Sources/FoundationEssentials/CodableWithConfiguration.swift
```diff
@@ -53,14 +53,9 @@ public extension KeyedEncodingContainer {
...
 public extension KeyedDecodingContainer {
     func decode<T, C>(_: CodableConfiguration<T?, C>.Type, forKey key: Self.Key) throws -> CodableConfiguration<T?, C> {
-        if self.contains(key) {
-            let wrapper = try self.decode(CodableConfiguration<T, C>.self, forKey: key)
-            return CodableConfiguration<T?, C>(wrappedValue: wrapper.wrappedValue)
-        } else {
-            return CodableConfiguration<T?, C>(wrappedValue: nil)
-        }
+        let wrapper = try self.decodeIfPresent(CodableConfiguration<T?,C>.self, forKey: key)
+        return CodableConfiguration<T?, C>(wrappedValue: wrapper?.wrappedValue)
```
jmschonfeld, line 57: Sorry if I missed this before, but now isn't `wrapper` a `CodableConfiguration<T?, C>`? If so, is there any reason not to just return that instead of creating a new `CodableConfiguration` from the wrapped value?

TomaszLizer: Nah - type of `wrapper` is `Optional<CodableConfiguration<T?, C>>` hence it needs to be wrapped some more :D

jmschonfeld:
> QQ however - is there I way I could simply build and test [FOUNDATION_FRAMEWORK](https://github.com/swiftlang/swift-foundation/blob/main/Foundation_Build_Process.md#foundation_framework-build) mode? I experimented a bit with adding flags during initial implementation, but only encountered failures. Actually at that time I was only able to build via swift test 🤔

At this time not currently, the `FOUNDATION_FRAMEWORK` build is a configuration used when building the full `Foundation.framework` and will fail to build without other pieces there. In general, building without that is sufficient (and is the only configuration needed on platforms like Linux + Windows) but every once in a while we just need a small tweak to something like this to get it to work. I'll make sure that's all sorted out, thanks for updating!

Sources/FoundationEssentials/CodableWithConfiguration.swift
jmschonfeld: Ah right yep that makes sense, thanks for explaining 🙂

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

Issue #1560, opened 2025-10-24
Inconsistent handling of IDN vs non-IDN hosts in URL()


nreilly:
**Describe the bug**
URLs created with long hostnames has inconsistent behaviour between IDN and non-IDN hostnames. For hostnames longer than 63 characters, non-IDN hostnames work fine, but IDNs result in URL returning Nil.

**To Reproduce**
Here's some sample code to show the issue:
```swift
import Foundation

func letters(_ count: Int) -> String {
  return String(repeating: "a", count: count)
}

func extraLetters(_ count: Int) -> String {
  let face = "🧐"  //  = xn--9u9h
  switch count {
  case 8: return face
  case 0...12: return letters(count)
  case 13...36: return letters(count - 10) + face
  default: return letters(count - 11) + face
  }
}

func host(_ host: String) -> String {
  return URL(string: "https://\(host)/")?.host(percentEncoded: false)
    ?? "<No URL>"
}

func printComparison(_ count: Int) {
  let prefix = String(format: "%03d: ", count)
  print("\(prefix)\(host(letters(count)))")
  print("\(prefix)\(host(extraLetters(count)))")
}

for comparison in [8, 30, 63, 64, 128, 256] {
  printComparison(comparison)
}
```

And here's the output:
```
008: aaaaaaaa
008: xn--9u9h
030: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
030: xn--aaaaaaaaaaaaaaaaaaaa-uv13t
063: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
063: xn--aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-r3017c
064: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
064: <No URL>
128: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
128: <No URL>
[...]

PR #1632, opened 2025-12-06
Improve URL host IDNA-encoding performance and compatibility

jrflat:
Improves IDNA encoding performance by converting UTF8 to UTF16 in Swift before calling the UTF16-native ICU functions. Improves compatibility by allowing specific UIDNA errors during `nameToASCII`.

### Motivation:

Performance: ICU `uidna_nameToASCII_UTF8` and `uidna_nameToUnicodeUTF8` are just convenience wrappers around the UTF16-native functions. Performing the conversions to and from UTF16 ourselves in Swift is faster than having ICU do it for us, and we can also use the fact that `nameToASCII` produces ASCII on success to efficiently truncate the returned `UInt16` buffer.

Compatibility: Resolves https://github.com/swiftlang/swift-foundation/issues/1560. As the issue describes, URL handling of ASCII and IDNA-encoded hosts is inconsistent. For instance, if an IDNA-encoded host has a domain label longer than 63 bytes or the entire domain is longer than 255 bytes, `URL` previously returned `nil` because the `uidna` functions indicated the respective non-fatal errors. Hosts without IDNA-encoding don't see this limitation. Allowing these errors also aligns our behavior with Safari and other WHATWG URL parsers.

### Modifications:

- In cases where we would previously call the UTF8 `uidna` functions, this PR instead performs the UTF8 to UTF16 transcoding in Swift before passing the UTF16 buffer to the ICU function. On `nameToASCII` success, we truncate the returned `UInt16` ASCII elements to `UInt8` and initialize the resulting `String`. On `nameToUnicode` success, we create the `String` from UTF16, which performs the UTF16 to UTF8 transcoding.

- Use the same allowed errors for `nameToASCII` that are currently allowed for `nameToUnicode`: `UIDNA_ERROR_EMPTY_LABEL | UIDNA_ERROR_LABEL_TOO_LONG | UIDNA_ERROR_DOMAIN_NAME_TOO_LONG | UIDNA_ERROR_LEADING_HYPHEN | UIDNA_ERROR_TRAILING_HYPHEN | UIDNA_ERROR_HYPHEN_3_4`

### Result:

IDNA Encoding Performance

- ~10% speedup for IDNA encoding and decoding.

- Compatibility with WHATWG parsers/browsers and consistent behavior for IDN and non-IDN host name lengths.

### Testing:

- Added benchmarks for IDNA encoding and decoding.

- Added unit test for allowed `nameToASCII` errors.

### DIFF
--- a/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
+++ b/Benchmarks/Benchmarks/URL/BenchmarkURL.swift
@@ -279,4 +279,42 @@ let benchmarks = {
             }
         }
     }
+
+    // MARK: - IDNA Encoding
+
+    var idnaComp = URLComponents()
+    idnaComp.scheme = "https"
+
+    Benchmark("URL.IDNAEncodeHost.Simple") { benchmark in
+        for _ in benchmark.scaledIterations {
+            idnaComp.host = "bücher.example"
+            blackHole(idnaComp)
+            idnaComp.host = nil
+        }
+    }
+
+    Benchmark("URL.IDNAEncodeHost.Complex") { benchmark in
+        for _ in benchmark.scaledIterations {
+            idnaComp.host = "😂😂😂.example"
+            blackHole(idnaComp)
+            idnaComp.host = nil
+        }
+    }
+
+    let idnaCompWithSimpleHost = URLComponents(string: "https://bücher.example")!
+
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

Issue #1633, opened 2025-12-06
`Calendar.RecurrenceRule.afterOccurrences` is applied relative to range start instead of startDate


baekteun:
**Describe the bug**
When using `Calendar.RecurrenceRule.End.afterOccurrences` with
`recurrences(of:in:)`, the occurrence count appears to be applied relative to
the `range.lowerBound` instead of the original `startDate` of the recurrence.

**To Reproduce**
Sample code to reproduce the behavior.

```swift
import Foundation

let recurrenceRule = Calendar.RecurrenceRule.init(
    calendar: .current,
    frequency: .daily,
    end: .afterOccurrences(3)
)

let startDate = DateComponents(calendar: .current, year: 2025, month: 12, day: 1).date!

let rangeStartDate = DateComponents(calendar: .current, year: 2025, month: 12, day: 10).date!
let rangeEndDate = DateComponents(calendar: .current, year: 2025, month: 12, day: 20).date!

let result = Array(recurrenceRule.recurrences(of: startDate, in: rangeStartDate..<rangeEndDate))

print(result) 
// Expect : []
// Actual : [2025-12-10 00:00:00, 2025-12-11 00:00:00, 2025-12-12 00:00:00]
```

**Expected behavior**
Based on the example code, an empty array is expected to be returned.
The `afterOccurrences` limit should be calculated from the `startDate` passed to  
`recurrences(of:in:)`, rather than from the start of the queried range.

**Configuration (please complete the following information):**
 - Swift Version: Swift 6.2
 - OS: iOS, macOS
 - OS Version: macOS 26.1, iOS 26.1, iOS 2.6.0.1

**Regression information:**
-

**Additional context**
-

itingliu:
I'm not sure I agree that this should return an empty array. I would actually expect this to return the current behavior `[2025-12-10 00:00:00, 2025-12-11 00:00:00, 2025-12-12 00:00:00]`

Based on the proposal, the `start` only defines/modifies the starting point of the **recurrence rule**, not the date search range; similarly, `afterOccurrence` defines how many times a **recurrence rule** should be applied, and is independent of the `startDate`:

```swift      
        /// Find recurrences of the given date
        ///
        /// The calculations are implemented according to RFC-5545 and RFC-7529.
        ///
        /// - Parameter start: the date which defines the starting point for the
        ///   recurrence rule.
        /// - Parameter range: a range of dates which to search for recurrences.
        ///   If `nil`, return all recurrences of the event.
[...]

PR #1634, opened 2025-12-06
Fix `afterOccurrences` to be calculated from startDate in RecurrenceRule

baekteun:
Fix `Calendar.RecurrenceRule.afterOccurrences` to be calculated from `startDate` in `recurrences(of:in:)`

### Motivation:

When using `Calendar.RecurrenceRule.End.afterOccurrences` with
`recurrences(of:in:)`, the occurrence count is currently applied relative to
the start of the queried range, rather than the original `startDate` of the
recurrence.

This leads to unexpected results when the query range begins after all
occurrences allowed by `afterOccurrences` have already been exhausted, causing
occurrences to be returned even though the recurrence should have ended.

Fixes: #1633 

### Modifications:

- Updated the recurrence enumeration logic so that the `afterOccurrences` limit
is applied globally from the startDate passed to `recurrences(of:in:)`,
rather than being reset relative to the query range.
- Ensured that occurrences falling before the queried range still contribute
to the total occurrence count.
- Added new tests to validate the corrected behavior and prevent regressions.

### Result:

After this change, `recurrences(of:in:)` no longer returns occurrences when the
query range starts after all occurrences permitted by `afterOccurrences` have
been reached.

### Testing:

- Added new tests covering `afterOccurrences` behavior when the query range
  starts after all allowed occurrences, as well as when the range intersects
  partially with the allowed occurrence set.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
@@ -373,6 +373,11 @@ extension Calendar {
                     // If a range has been specified, we should skip a few extra 
                     // occurrences until we reach the start date
                     if let baseRecurrenceLowerBound, nextDate < baseRecurrenceLowerBound {
+                        resultsFound += 1
+                        if let limit = recurrence.end.occurrences, resultsFound >= limit {
+                            finished = true
+                            return
+                        }
                         continue
                     }
                     anchor = nextDate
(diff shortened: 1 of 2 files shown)


itingliu:
@swift-ci please test

itingliu:
I had a comment in the original issue that I'm not sure if this is really an issue. @baekteun do you want to take a look there?

PR #1635, opened 2025-12-09
Change condition to be < 0

LumioseSil:
This is so it matches the documentation of fcopyfile

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -1083,7 +1083,7 @@ enum _FileOperations {
         defer { close(dstFD) }
         
         #if canImport(Darwin)
-        if fcopyfile(srcFD, dstFD, nil, copyfile_flags_t(COPYFILE_METADATA | COPYFILE_NOFOLLOW | extraFlags)) != 0 {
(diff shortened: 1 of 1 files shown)


jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -1083,7 +1083,7 @@ enum _FileOperations {
         defer { close(dstFD) }
         
         #if canImport(Darwin)
-        if fcopyfile(srcFD, dstFD, nil, copyfile_flags_t(COPYFILE_METADATA | COPYFILE_NOFOLLOW | extraFlags)) != 0 {
+        if fcopyfile(srcFD, dstFD, nil, copyfile_flags_t(COPYFILE_METADATA | COPYFILE_NOFOLLOW | extraFlags)) < 0 {
```
glessard, line 1086: The documentation is that `fcopyfile` returns "less than 0 on error, and 0 on success", so the test already matched the documentation. If we're going to be this nit-picky, we should also assert that the return value is not greater than zero.

jmschonfeld:
There's still an outstanding question on this PR and there have been no recent updates so I'm going to close this - if you circle back to this later and address the comment above feel free to re-open/re-request review

PR #1639, opened 2025-12-10
__DataInvokeDeallocatorUnmap should always call unmap

LumioseSil:
Not sure if this is a typo, nor could I see any rationale for this being free.


jmschonfeld (changes requested):
This change is not correct on it's own and without further investigation. There are certain environments where this code runs that have currently needed to use `free` here. We may be able to update this, but this needs further investigation and testing with a more wholistic look at the unmapping deallocator.

PR #1640, opened 2025-12-10
Fix bounds ordering for captured Range values in Predicate to NSPredicate conversion

LumioseSil:
We seem to be passing upper to lower and lower to upper when converting. That does not seem right.

### DIFF
--- a/Tests/FoundationEssentialsTests/PredicateConversionTests.swift
+++ b/Tests/FoundationEssentialsTests/PredicateConversionTests.swift
@@ -177,6 +177,32 @@ private struct NSPredicateConversionTests {
         #expect(!converted.evaluate(with: ObjCObject()))
     }
     
+    @Test func rangesDistinctBounds() throws {
+        // Test that captured ranges have correct bounds order in NSPredicate conversion
+        // This tests the _RangeValue branch in RangeExpressionContains.convert
+        let lowerDate = Date(timeIntervalSinceReferenceDate: 100)
+        let upperDate = Date(timeIntervalSinceReferenceDate: 200)
+        let range = lowerDate ..< upperDate
+        
+        let predicate = #Predicate<ObjCObject> {
+            range.contains($0.i)
+        }
+        let converted = try #require(convert(predicate))
+        
+        // Verify the bounds are in the correct order (lower, upper)
+        #expect(converted == NSPredicate(format: "i >= %@ AND i < %@", lowerDate as NSDate, upperDate as NSDate))
+        
+        // Create an object with a date in the range
+        let objInRange = ObjCObject()
+        objInRange.i = Date(timeIntervalSinceReferenceDate: 150)
+        #expect(converted.evaluate(with: objInRange), "Date in range should match")
+        
+        // Create an object with a date outside the range
+        let objOutOfRange = ObjCObject()
+        objOutOfRange.i = Date(timeIntervalSinceReferenceDate: 250)
+        #expect(!converted.evaluate(with: objOutOfRange), "Date outside range should not match")
+    }
+    
     @Test func nonObjC() throws {
         let predicate = #Predicate<ObjCObject> {
             $0.nonObjCKeypath == 2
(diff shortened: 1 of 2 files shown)


Tests/FoundationEssentialsTests/PredicateConversionTests.swift
```diff
@@ -177,6 +177,17 @@ private struct NSPredicateConversionTests {
         #expect(!converted.evaluate(with: ObjCObject()))
     }
     
+    @Test func rangesDistinctBounds() throws {
+        let lower = Date.distantPast
+        let upper = Date.distantFuture
+        let predicate = #Predicate<ObjCObject> {
+            (lower ..< upper).contains($0.i)
+        }
```
jmschonfeld: I don't believe this unit test ends up actually testing the changed code. The use of `..<` in the predicate will turn it into a `PredicateExpressions.Range` expression which will use the `as? _RangeOperator` branch above your code change. Your code change is in the path for ranges that are captured by a predicate, so you'll need to capture the range in the predicate to hit this code path

```suggestion
        let lower = Date.distantPast
        let upper = Date.distantFuture
        let range = lower ..< upper
        let predicate = #Predicate<ObjCObject> {
            range.contains($0.i)
        }
```

Tests/FoundationEssentialsTests/PredicateConversionTests.swift
```diff
@@ -177,6 +177,17 @@ private struct NSPredicateConversionTests {
...
+        let converted = try #require(convert(predicate))
+        #expect(converted == NSPredicate(format: "i >= %@ AND i < %@", lower as NSDate, upper as NSDate))
+        #expect(converted.evaluate(with: ObjCObject()))
```
jmschonfeld: `ObjCObject.i` is set to `Date.distantFuture`, which is not actually in the range you created so I suspect this predicate should evaluate to `false`, not `true.
```suggestion
        #expect(!converted.evaluate(with: ObjCObject()))
```

Tests/FoundationEssentialsTests/PredicateConversionTests.swift
```diff
@@ -177,6 +177,37 @@ private struct NSPredicateConversionTests {
...
+        
+        // Create an object with a date outside the range
+        let objOutOfRange = ObjCObject()
+        objOutOfRange.i = Date(timeIntervalSinceReferenceDate: 250)
+        #expect(!converted.evaluate(with: objOutOfRange), "Date outside range should not match")
+        
+        // Verify that if bounds were swapped, the predicate would be wrong
+        // (This would be i >= upperDate AND i < lowerDate, which is always false)
+        let wrongPredicate = NSPredicate(format: "i >= %@ AND i < %@", upperDate as NSDate, lowerDate as NSDate)
+        #expect(!wrongPredicate.evaluate(with: objInRange), "Swapped bounds would incorrectly reject valid date")
```
jmschonfeld: What benefit did you find with adding this additional test? Is it testing `Predicate` to `NSPredicate` conversion in some way?

AZero13: It proves the ordering of bounds matters.

jmschonfeld: Can you elaborate on what you mean by that? I don't fully understand what this is proving here regarding the conversion code that you updated. It doesn't look like this code goes through the `Predicate`/`NSPredicate` conversion code but instead looks like it's just testing `NSPredicate` itself

AZero13: It exercises the _RangeValue conversion path for captured ranges.

AZero13: It checks that when extracting bounds from a captured Range (via range.lowerBound and range.upperBound at line 326), they are used in the correct order when building the NSPredicate (lower for >=, upper for <).

AZero13: But yeah the last two parts are redundant.

jmschonfeld:
@swift-ci please test

PR #1642, opened 2025-12-10
GetUserNameExW returns 0 on error, non-zero on success

LumioseSil:
I know Microsoft is not consistent on this, but GetUserNameExW returns 0 on error.

### DIFF
--- a/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
+++ b/Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
@@ -229,7 +229,7 @@ final class _ProcessInfo: Sendable {
         _ = GetUserNameExW(NameDisplay, nil, &ulLength)
 
         return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(ulLength)) { wszBuffer in
-            guard GetUserNameExW(NameDisplay, wszBuffer.baseAddress!, &ulLength) == 0 else {
+            guard GetUserNameExW(NameDisplay, wszBuffer.baseAddress!, &ulLength) != 0 else {
                 return ""
             }
             return String(decoding: wszBuffer.prefix(Int(ulLength)), as: UTF16.self)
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/ProcessInfo/ProcessInfo.swift
```diff
@@ -229,7 +229,7 @@ final class _ProcessInfo: Sendable {
         _ = GetUserNameExW(NameDisplay, nil, &ulLength)
 
         return withUnsafeTemporaryAllocation(of: WCHAR.self, capacity: Int(ulLength)) { wszBuffer in
-            guard GetUserNameExW(NameDisplay, wszBuffer.baseAddress!, &ulLength) == 0 else {
+            guard GetUserNameExW(NameDisplay, wszBuffer.baseAddress!, &ulLength) != 0 else {
```
jmschonfeld, line 232: It looks like we have a unit test that validates that the user name is not empty but it only runs on macOS. Should we also run it on Windows to validate this behavior?

AZero13: Sure!

Tests/FoundationEssentialsTests/ProcessInfoTests.swift
```diff
@@ -147,7 +147,7 @@ private struct ProcessInfoTests {
         "ProcessInfo thinks we are on System 70")
     }
 
-#if os(macOS)
+#if canImport(Darwin) || os(Windows)
```
jmschonfeld, line 150: It looks like when adding the Windows conditional here you also changed the macOS conditional to `canImport(Darwin)` - was there a reason for that? This will fail to build on platforms like iOS because `ProcessInfo.userName` is unavailable on those platforms

itingliu (changes requested):
Unfortunately we don't have an IOS CI, but I agree with Jeremy that the test won't work.

Issue #1548, opened 2025-10-15
URL.appending(queryItems:) adds a `?` when passing in an empty array of query items


BrentMifsud:
**Describe the bug**
Passing in an empty array of query items to `URL.appending(queryItems: [URLQueryItem])` adds a `?` to the end of the url. Resulting in a potentially invalid URL.

**To Reproduce**

```swift
let url = "www.google.com"
let updatedUrl = url.appending(queryItems: [])
print(updatedUrl) // "www.google.com?"
```

**Expected behavior**
If no query items are provided, the url should not be changed.

**Configuration (please complete the following information):**
 - Swift Version: 6.2 (Xcode 26.1 beta 2)
 - OS: iOS
 - OS Version: iOS 26.0.1

**Regression information:**
n/a

**Additional context**

Probably a minor issue, but I am currently building out an HTTP Client package for my current job.

I had to write an extension on url to get around this issue:

```swift
public extension HTTPRequest {
    private func constructURL(from baseURL: URL) -> URL {
        var url = baseURL.appending(path: path)

        if !queryItems.isEmpty {
            url = url.appending(queryItems: queryItems)
        }

        return url
    }
}
```

instead of being able to rely on:

```swift
var url = baseURL.appending(path: path).appending(queryItems: queryItems)
```

EngOmarElsayed:
I will be very happy to work on this issue can anyone assign this to me @itingliu @jmschonfeld @parkera

EngOmarElsayed:
Or should I just start right away without being assigned ?

EngOmarElsayed:
I have already started a PR 😅 https://github.com/swiftlang/swift-foundation/pull/1645

jrflat:
Thanks for opening the PR! Contributions are always welcome 😁 

I left some comments in the PR--I think there's some compatibility concerns we'll need to consider for this change. Happy to continue the discussion there!

EngOmarElsayed:
> ```
> 
> url.append(queryItems: [])
> 
> print(url.query!) // Crash
> 
> ```

I think the most logical behavior would be for the current implementation to crash when an empty array is passed, since there's no value to unwrap.

Alternatively, we could maintain the existing append(queryItems:) behavior but add a guard to prevent URL modification when the query array is empty. This would resolve the issue without introducing compatibility risks.

[...]

PR #1645, opened 2025-12-11
Fix url appending(queryItems:) method to prevent it from adding empty queryItems

EngOmarElsayed:
_Prevent appending `?` to URL when query items array is empty_

### Motivation:

When calling `URL.appending(queryItems:)` with an empty array, the current implementation appends a trailing `?` to the URL (e.g., `"www.google.com"` becomes `"www.google.com?"`). This produces potentially invalid URLs and forces users to add defensive checks before calling the method.

Related to this issue #1548

### Modifications:

Added an early return guard in `appending(queryItems:)` that returns `self` unchanged when the provided `queryItems` array is empty.

### Result:

Calling `url.appending(queryItems: [])` now returns the original URL without any modifications, allowing developers to chain URL construction methods without defensive empty-array checks:
```swift
// Before: "www.google.com?"
// After:  "www.google.com"
let url = URL(string: "www.google.com")!.appending(queryItems: [])
```

### Testing:

- Added a test case verifying that appending an empty query items array returns the original URL unchanged
- Verified existing tests pass to ensure no regression in behavior when non-empty query items are provided

### DIFF
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -666,6 +666,12 @@ private struct URLTests {
         )
 
         // Append queryItems
+        let emptyQueryItems: [URLQueryItem] = []
+        #expect(
+            base.appending(queryItems: emptyQueryItems).absoluteString ==
+            "https://www.example.com"
+        )
+
         let queryItems = [
             URLQueryItem(name: "id", value: "42"),
             URLQueryItem(name: "color", value: "blue")


jrflat:
Thanks for opening this PR up for discussion! I think there's some app compatibility risk with this change we'll need to consider. For instance, an app might assume that a URL they append query items to will always have a query:

```
url.append(queryItems: queryItems)
print(url.query!) // Crash
```

It's a bit of a contrived example, but unexpectedly `nil` properties can often lead to compatibility issues. (It might also trigger different code paths for an `if let query = ...` block, for example.)  Another argument against changing the current behavior is that other `URL` and `URLComponents` methods behave similarly, e.g:

```
var comp = URLComponents(string: "/path")!
comp.queryItems = []
print(comp) // "/path?" -- appends "?"

let url = URL(string: "/path")!
print(url.appending(path: "")) // "/path/" -- appends "/"
print(url.appendingPathComponent("")) // "/path/" -- appends "/"

let hostURL = URL(string: "https://example.com")!
print(hostURL.appending(path: "")) // "https://example.com/" -- appends "/"
```

I'll admit that `URLComponents.queryItems` isn't exactly the same as `URL.appending(queryItems:)` since you can set `comp.queryItems = nil` to remove the query and "?" entirely, but it's pretty similar.

Overall, I would lean towards keeping the current behavior, but I'm happy to hear a counterargument! :)

EngOmarElsayed:
Replied in the issue page: https://github.com/swiftlang/swift-foundation/issues/1548

EngOmarElsayed:
Better solution to prevent any compatibility risk https://github.com/swiftlang/swift-foundation/pull/1650 @jrflat

PR #1647, opened 2025-12-11
Remove dead, useless storage count check

LumioseSil:
Storage count is always equal to or less than the max, so this is useless dead code.

### DIFF
--- a/Sources/FoundationEssentials/NotificationCenter/NotificationCenter.swift
+++ b/Sources/FoundationEssentials/NotificationCenter/NotificationCenter.swift
@@ -32,8 +32,6 @@ private struct AutoDictionary<Value> {
     
     // Effectively O(1), worst case O(n)
     mutating func insert(_ value: Value) -> UInt64 {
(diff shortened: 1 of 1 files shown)


itingliu:
@swift-ci please test

Sources/FoundationEssentials/NotificationCenter/NotificationCenter.swift
```diff
@@ -32,8 +32,6 @@ private struct AutoDictionary<Value> {
     
     // Effectively O(1), worst case O(n)
     mutating func insert(_ value: Value) -> UInt64 {
-        guard storage.count <= UInt64.max else { fatalError("Exceeded maximum storage size") }
```
cthielen, line 35: This check is necessary to prevent an infinite loop a few lines down on the `while(storage[nextKey] != nil)` line.

Since `storage.count` returns an Int, on 64-bit systems UInt64.max probably won't be reached.

This can be improved but the `fatalError` is necessary.

cthielen (changes requested):
See comment about why the check is necessary.

PR #1648, opened 2025-12-11
Use st_birthtimespec instead of st_ctim.tv_sec on Darwin

LumioseSil:
st_ctim records the last time the inode changed. Closest thing on linux, but st_birthtimespec is what we actually want here.

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -1093,6 +1093,24 @@ private struct FileManagerTests {
     }
     #endif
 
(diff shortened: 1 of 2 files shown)


Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1093,6 +1093,24 @@ private struct FileManagerTests {
     }
     #endif
 
+    @Test(.enabled(if: isDarwin, "Birth time is only exposed on Darwin"))
```
jmschonfeld, line 1096: Is there a reason that we should limit this test to Darwin platforms? I know Darwin platforms are the ones that don't pass this test today, but we should expect platforms like Linux and Windows to pass this test as well, right?

PR #1654, opened 2025-12-17
_amkrtemp error string includes "Optional"

jmschonfeld:
Updates error handling logic to avoid printing "Optional" in an error description

### Motivation:

Updates the error message to no longer include the "Optional" string to make it clearer

### Modifications:

Exits earlier if the error code is `nil` indicating we've already hit an error we can't recover from

### Result:

The error message is now only provided when the error code is non-optional (otherwise a better error is thrown)

### Testing:

Validated in the IDE that the interpolated variable is no longer optional (this is nontrivial to add a unit test for because it requires us to be unable to create a file system representation for this specific path)


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1655, opened 2025-12-18
conversationalDayPeriodsOverride() fails with ICU 78

itingliu:
ICU78's update will cause this test to fail. ICU has merged a follow up fix, which should be available soon. Disable the test for now; will re-enable the test once the fix lands.

166656959


itingliu:
@swift-ci please test

PR #1663, opened 2025-12-25
[Android] Add shims for `getgrnam_r` and `getgrgid_r` for API 23

madsodgaard:
Adds simple shims for these two functions, such that `swift-foundation` can compile on API 23.

### Motivation:

Currently, `swift-foundation` does not compile on Android API 23, because of `getgrnam_r` and `getgrgid_r`, which is only available on API 24. 

This is the only thing that was missing to make this repo compile on API 23. 

An alternative solution, was guarding the calls to these with `if #available(Android 24, *)`, but I am not sure if we should "fail" silently in that case, and just not append the GID to the file on API 23. So, I guess this solution is at least better. Also, the shims are quite simple, since Foundation only needs the name and GID.

### DIFF
--- a/Sources/_FoundationCShims/include/filemanager_shims.h
+++ b/Sources/_FoundationCShims/include/filemanager_shims.h
@@ -121,7 +121,7 @@ static inline int _filemanager_shims_getgrnam_r(const char *name, struct group *
     return 0;
 }
 
-#elif !TARGET_OS_WINDOWS && !TARGET_OS_WASI
+#elif __has_include(<grp.h>)
 #include <grp.h>
 
 static inline int _filemanager_shims_getgrgid_r(gid_t gid, struct group *grp,


finagolfin:
This isn't a polyfill, because it will be compiled in _if_ the target API when compiling the SDK itself is 23 or earlier, regardless of the eventual Android runtime environment. These group functions seem pretty much useless on Android though, so I'd have been fine if you had just stubbed them out when compiling against such an old target API, but now that you've done this work, seems fine to include it, :+1: if the swift-foundation reviewers approve.

finagolfin:
@swift-ci test

finagolfin:
@parkera, could you have someone take a look?

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -46,4 +47,86 @@
 extern int _mkpath_np(const char *path, mode_t omode, const char **firstdir);
 #endif
 
+#if TARGET_OS_ANDROID && __ANDROID_API__ <= 23
+#include <grp.h>
+#include <sys/types.h>
+#include <string.h>
+#include <errno.h>
+
+static inline int _filemanager_shims_getgrgid_r(gid_t gid, struct group *grp,
```
itingliu, line 56: Any reason why we can't do call these code directly in Swift? Yes, there will be some boilerplates, but I'm not sure if we should be adding code to these

jmschonfeld: Yeah, in general this patch seems ok to me, but the goal of this project is to write these implementations in Swift rather than C, so unless we have a good reason to write this in C (because Swift cannot call these functions for some reason) then writing the shim in Swift seems like a better approach to me

finagolfin: How do you propose we write this in Swift? This was written in C because these APIs are only unavailable for Android API 23 and before, so either you'd have to mimic the `#if TARGET_OS_ANDROID && __ANDROID_API__ <= 23` guard above with a compile-time API version check in Swift, which the language does not provide so we'd have to pass it in manually through the `Package.swift` manifest and CMake config, or use the new `#available(Android, )` feature, which we can't yet because [it requires NDK 28 or later](https://github.com/swiftlang/swift-testing/pull/1373#issuecomment-3507830154).

I believe Mads added these as C shims for those reasons and because he noticed these C shims here already. What we could do is get this in for now, then revisit later in Swift once `#available(Android, )` is working with a new NDK on the official Android CI.

finagolfin: Would be good to get this small pull in, let me know what you think, @itingliu.

itingliu: That sounds reasonable. Thanks

itingliu:
@swift-ci please test

Sources/_FoundationCShims/include/filemanager_shims.h
itingliu: Would you mind filing an issue for revisiting this to use `#available(Android, )` when it's ready?

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -46,4 +47,86 @@
...
+        return errno;
+    }
+
+    size_t name_len = strlen(p->gr_name) + 1;
+    if (name_len > buflen) {
+        *result = NULL;
+        return ERANGE;
+    }
+
+    strcpy(buf, p->gr_name);
```
jmschonfeld: nit: I think this with the checks above could be written as a call to `strlcpy` right? Is it worth rewriting to that? It's effectively the same thing I think, this `strcpy` without a limit just caught my eye as something that has caused vulnerabilities in the past when not appropriately validated.

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -46,4 +47,86 @@
...
+    grp->gr_name = buf;
+    grp->gr_gid = p->gr_gid;
+
+    grp->gr_passwd = (char *)"";
+    grp->gr_mem = NULL;
```
jmschonfeld: I know we don't use it right now, but should these two be copied over for consistency? Or does android define these as never set for one reason or another?

Sources/_FoundationCShims/include/filemanager_shims.h
madsodgaard: yup, good catch

Sources/_FoundationCShims/include/filemanager_shims.h
madsodgaard: The official Bionic impl does not set these fields. For passwd, its basically a null-pointer and for the group member list, its just a "synthetic list" of the group name and a null pointer

https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/bionic/grp_pwd.cpp#83

The passwd would be simple enough to copy as well, but the group members would require a bit more complex logic to deep copy all the strings. So, since Foundation does not use these and Bionic does not, I think leaving them out, seems like the best option.

I added a comment in the code.

madsodgaard:
@swift-ci please test

shahmishal:
@swift-ci please test macOS

Sources/_FoundationCShims/include/filemanager_shims.h
jmschonfeld: Thanks, the added comment is helpful to clarify that. It looks like you changed the returned `gr_passwd` to an empty string "for safety" - could you elaborate on what safety you're trying to guarantee here? I know we don't use it right now, but returning the group password as an empty string seems like it would be prone to issues if we did decide to use it (I wouldn't want us to end up performing any form of string comparison against that empty string for example). It seems better to me to leave this as setting it to `NULL` to preserve what the OS does instead of providing a fake value that doesn't match the OS behavior.

madsodgaard: Agreed. I've updated the shim to explicitly set gr_passwd to NULL

finagolfin:
@swift-ci test

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -46,4 +47,92 @@
...
+    // Android Bionic generates a synthetic list ["groupname", NULL].
+    // Replicating that here would require deep copying the strings array.
+    // Foundation does not use this either, so NULL is sufficient and avoids complexity.
+    grp->gr_mem = NULL;
+
+    *result = grp;
+    return 0;
+}
+
+#elif !TARGET_OS_WINDOWS && !TARGET_OS_WASI
```
jmschonfeld: @madsodgaard would you be able to add a `__has_include(<grp.h>)` to this condition (or use that instead if neither Windows nor WASI has the header)? That would be a little more robust and ensures that this doesn't trip up anything in special circumstances like WASI where the API doesn't exist and is already excluded on the calling side?

madsodgaard: Sure! I replaced the conditions by the `_has_include`. Seems neither WASM nor Windows has the header

madsodgaard:
@swift-ci test

jmschonfeld (approved):

Sources/_FoundationCShims/include/filemanager_shims.h
madsodgaard: Sorry for the delay, I filed it here https://github.com/swiftlang/swift-foundation/issues/1769

PR #1664, opened 2025-12-26
Compare only the bytes read, not the entire buffer

LumioseSil:
Otherwise we can get false negatives


jmschonfeld:
@swift-ci please test

PR #1668, opened 2026-01-03
Use hash(into) instead of hashValue

LumioseSil:
hashValue is deprecated.

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError.swift
@@ -27,8 +27,8 @@ public struct CocoaError : _BridgedStoredNSError {
     
     public static var errorDomain: String { return NSCocoaErrorDomain }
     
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Error/CocoaError.swift
```diff
@@ -27,8 +27,8 @@ public struct CocoaError : _BridgedStoredNSError {
     
     public static var errorDomain: String { return NSCocoaErrorDomain }
     
-    public var hashValue: Int {
-        return _nsError.hashValue
+    public func hash(into hasher: inout Hasher) {
```
jmschonfeld, line 30: This changes public API and is an ABI breaking change because it removes an existing public entry point and adds a new one without availability. Can you elaborate on what deprecation warning you're seeing with this code?

jmschonfeld (changes requested):

PR #1669, opened 2026-01-03
Typo fix: check for intValue should be &&, not ||

LumioseSil:
We are checking between 48 and 57 inclusive, so it should be &&.


jmschonfeld:
Since there hasn't been an update here recently, I'm going to go ahead and close this out but if you have any additional updates based on the feedback above feel free to add changes and re-open

PR #1671, opened 2026-01-03
Fix typo in code: Self.compatibiltity2 -> compatibility2

LumioseSil:


itingliu (commented):
`String.compatibiltity2` is defined somewhere else that is not accessible from this repo. That means this code would stop to compile if we only change the typo here without changing the declaration. So while I understand this typo is an eyesore, I don't think it's correct to "fix" it here like this.

PR #1675, opened 2026-01-08
CFURL re-core in Swift

jrflat:
This PR provides a Swift implementation for CFURL on Darwin.

### Motivation:

The previous approach of re-coring CF/NSURL using a Swift subclass of NSURL ran into some limitations due to the hierarchy of NSURL calling CFURL functions and the "get" semantics of commonly used CF functions such as `CFURLGetString`. This approach instead implements CFURL from Swift, so we no longer need to inverse the logic.

This PR has no impact on Swift `URL` or `URLComponents` at this time and is guarded within `FOUNDATION_FRAMEWORK`. The new Swift code will easily adapt to these types in the future so we can consolidate implementations, but that change should be landed separately in the interest of minimizing risk while we perform more compatibility testing.

#### **Removed**

**`URL_ObjC.swift`**: Removed the `_NSSwiftURL` subclass and related implementation details.

#### **Added**

**`URL_Parsing.swift`**: parsing and encoding full URL strings, storing the flags and component ranges. Implemented with a `_URLParseable` protocol so we can use it for `URL/URLComponents` in the future.

**`URL_Resolution.swift`**: path, dot dot, and absolute URL resolution. Implemented with a `_URLHeader` protocol so we can use it for `URL` in the future.

**`URL_Validation.swift`**: optimized component validation with lookup table.

**`URLEncoder.swift`**: percent-encoding/decoding with options to specify string encodings, characters to leave escaped, additional characters to encode, and whether to skip already valid escape sequences. These `CFURL` functionalities are a superset of `URL/URLComponents` use-cases, so we can consolidate them in the future.

**`URL_C+(...).swift`**: various `NSURL` extensions to provide a bridge into the Swift implementation.

### Result:

No effect outside of `FOUNDATION_FRAMEWORK`. Allows testing of the `CFURL` Swift implementation on Darwin.

### Performance:

CFURL Swift Recore Performance

### Testing:

- Verified `swift-foundation` unit tests for `URL/URLComponents`
- Added additional tests for `NSURL` and `CFURL`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -191,7 +191,7 @@ let package = Package(
             cSettings: wasiLibcCSettings,
             swiftSettings: [
                 .enableExperimentalFeature("AccessLevelOnImport"),
-                .enableExperimentalFeature("Lifetimes"),
+                .enableExperimentalFeature("LifetimeDependence"),
             ] + availabilityMacros + featureSettings
         ),
         
--- a/Sources/FoundationInternationalization/CMakeLists.txt
+++ b/Sources/FoundationInternationalization/CMakeLists.txt
@@ -30,7 +30,7 @@ add_subdirectory(TimeZone)
 
 target_compile_options(FoundationInternationalization PRIVATE
     "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend AccessLevelOnImport>"
-    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend Lifetimes>"
+    "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend LifetimeDependence>"
     "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend ImportMacroAliases>"
     "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-experimental-feature -Xfrontend StrictConcurrency>"
     "SHELL:$<$<COMPILE_LANGUAGE:Swift>:-Xfrontend -enable-upcoming-feature -Xfrontend InferSendableFromCaptures>"


jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL_Swift.swift
```diff
@@ -1274,5 +1278,134 @@ extension _SwiftURL {
...
+            if let base = _baseURL as CFURL? {
+                impl._header._base = Unmanaged.passRetained(base)
+            }
+            return withUnsafePointer(to: impl) { bigImpl in
+                return _CFURLCreateWithBigImpl(bigImpl)
+            }
+        }
+    }
+
+    @_specialize(where Impl == __CFSmallURLImpl)
```
itingliu, line 1326: Why wouldn't we want to also have `@_specialize(where Impl == __CFBigURLImpl)`? Are there downsides for using `@_specialize`?

Sources/FoundationEssentials/URL/URL_Parsing.swift
```diff
@@ -0,0 +1,908 @@
...
+    }
+}
+
+// MARK: - Parsing, validation, and encoding (main entry point)
+
+internal func parse<T: _URLEncoding, Impl: _URLParseable>(
+    _ type: T.Type,
+    span: borrowing Span<T.CodeUnit>,
+    flags: inout __CFURLFlags,
+    into impl: UnsafeMutablePointer<Impl>,
```
itingliu, line 532: Why don't we want to use `OutputSpan` here?

Sources/FoundationEssentials/URL/URL_Parsing.swift
```diff
@@ -0,0 +1,908 @@
...
+    // The maximum 9 * buffer.count is way overkill in the vast majority of
+    // cases, since usually very few characters need encoding, or we're ASCII,
+    // so start with 2 * buffer.count and re-allocate if needed.
+
+    assert(!span.isEmpty)
+    let encoded: String? = withUnsafeTemporaryAllocation(of: UInt8.self, capacity: 2 * span.count) {
+        var os = OutputSpan(buffer: $0, initializedCount: 0)
+        guard encode(into: &os) else { return nil }
+        let outputLength = os.finalize(for: $0)
+        return String(decoding: $0[..<outputLength], as: UTF8.self)
```
itingliu, line 858: Can we not use this because we want the String initializer to validate its UTF8 content and fail when it can't
```
let encoded = String(unsafeUninitializedCapacity: 2 * span.count) {
    var os = OutputSpan(buffer: $0, initializedCount)
    ...
    return os.finalize(for:$0)
}
```

Otherwise I thought this would allow us to avoid the temporary allocation altogether by writing directly into String's storage?

Sources/FoundationEssentials/URL/URL_Swift.swift
jrflat: The case where we need a `__CFBigURLImpl` is exceedingly rare (the URL string needs to be longer than 65,535 characters), so I didn't include it to (possibly) reduce code size. The compiler might specialize it anyway, though. I'm not certain I need to specialize for `__CFSmallURLImpl` here either since the compiler probably already does it, but this should guarantee specialization for the 99.99% code path.

Sources/FoundationEssentials/URL/URL_Parsing.swift
jrflat: `Impl` is the actual URL struct that we're filling out, e.g.
```
struct __CFSmallURLImpl {
  var schemeRange: Range<Int>
  var userRange: Range<Int>
  ...
}
```
so we're using `UnsafeMutablePointer` for easier interop with the C struct pointers passed from `CFURL`. For Swift `URL` backing structs in the future, we can just pass `&myURLImpl` to the function for the same effect.

Sources/FoundationEssentials/URL/URL_Parsing.swift
jrflat: I originally had the code using `String(unsafeUnitializedCapacity:)` like you show there but ran into 2 issues:
1. IIUC, that code will always allocate `2 * span.count` for the String even if (in most cases) we use much less. I saw a memory and slight performance improvement by switching to the temporary allocation + `String(decoding:as:)`, I think partially because the temporary allocation can often be stack allocated.
2. The encoding code first tries `2 * span.count`, which should almost always work, then tries `3 * T._maxUTF8BytesPerCharacter * span.count` for the worst case after. Representing failure in the first case by returning `0` in the `String` initializer and checking for an empty String felt semantically awkward (though it would work, since we will always be encoding a non-empty URL string here).

jrflat: Oh and to better answer your first question, we don't need `String` to validate its UTF8 contents here since it's guaranteed ASCII, but I believe both `String(decoding:as:)` and `String(unsafeUninitializedCapacity:)` do validate their contents and replace invalid sequences with the replacement character. However, they always return non-null, which is what we want.

itingliu: Thanks for clarifying. I guess `guard encode(into: &os) else { return nil }` in a way is us doing our own validation before copying into String, so a temporary allocated buffer makes sense.

Sources/FoundationEssentials/URL/URLEncoder.swift
```diff
@@ -0,0 +1,620 @@
...
+// encoding before appending to the result."
+
+// Many CFURL percent-encoding use-cases require skipping valid percent-escapes
+// that are already in the string, e.g. "%20 %20" becomes "%20%20%20". However,
+// some use-cases, such as encoding a file path or path component to append,
+// will encode everything, including the "%" in a valid percent-escape.
+
+extension URLComponentAllowedMask {
+    // Used to encode ";" in a path component
+    static let pathNoSemicolon = Self(rawValue: 0x47fffffe87ffffff27ffffd200000000)
```
itingliu, line 32: How are these rawValues determined?

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
[59edb1c](https://github.com/swiftlang/swift-foundation/pull/1675/commits/59edb1c2c043222c898235895d7b7ab2cb4604ef) added a `Span`-based IDNA-encoding wrapper that requires the `LifetimeDependence` experimental feature to be enabled in FoundationInternationalization (already enabled in FoundationEssentials). I enabled the experimental feature rather than just `#if FOUNDATION_FRAMEWORK`-ing the new function since we'll want it in swift-foundation eventually, and I'm guessing we could adopt `Span` for more use-cases elsewhere in FoundationInternationalization, too.

jrflat:
@swift-ci please test

jrflat:
Tried to use the "Lifetimes" feature instead but realized we'd need to wholesale update all of Foundation to use it and change all instances of `@lifetime` to `@_lifetime`. Keeping "LifetimeDependence" for now.

jrflat:
@swift-ci please test

Sources/FoundationInternationalization/URLParser+ICU.swift
```diff
@@ -218,4 +218,67 @@ struct UIDNAHookICU: UIDNAHook {
...
+                    transcoder.idnaTranscoder,
+                    inBufferPtr,
+                    Int32(inBuffer.count),
+                    outBufferPtr,
+                    Int32(outBuffer.count),
+                    &processingDetails,
+                    &error
+                )
+
+                guard U_SUCCESS(error.rawValue) && shouldAllow(processingDetails.errors, encodeToASCII: true) && convertedLength > 0 else {
```
itingliu, line 269: This is good, but just so you need it next time: we have a wrapper for ICU error in ICUError.swift so you can use it like `error.isSuccess` which I find nice

itingliu (approved):

Sources/FoundationEssentials/URL/URLEncoder.swift
jrflat: In the original `URLComponentAllowedMask` definition there's a comment better explaining how they're generated programmatically using the grammar/sets from RFC 3986. This mask in particular was generated by taking the original `.path` allowed mask raw value and removing the `;` bit.

Issue #1676, opened 2025-12-21
Swift SDK installation fails on official Swift 6.2.3 Docker image with `NSCocoaErrorDomain Code=512`


bc-lee:
### Description

Running `swift sdk install` on the official Swift 6.2.3 Docker image (`swift:6.2.3-jammy`) fails with an `NSCocoaErrorDomain Code=512` error during SDK installation.
Although the SDK appears in `swift sdk list` after the command, the installation is incomplete and the SDK is not usable for building Swift packages.

Specifically, the expected `swift-sdk.json` file is missing from the installed SDK directory.

### Reproduction

```bash
$ podman run -it swift:6.2.3-jammy /bin/bash
# Now inside the docker/podman
$ swift sdk install https://download.swift.org/swift-6.2.3-release/static-sdk/swift-6.2.3-RELEASE/swift-6.2.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz --checksum f30ec724d824ef43b5546e02ca06a8682dafab4b26a99fbb0e858c347e507a2c
Downloading a Swift SDK bundle archive from `https://download.swift.org/swift-6.2.3-release/static-sdk/swift-6.2.3-RELEASE/swift-6.2.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz`...
                                                                          Downloading
100% [============================================================================================================================================]
Downloading swift-6.2.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz

Swift SDK bundle archive successfully downloaded from `https://download.swift.org/swift-6.2.3-release/static-sdk/swift-6.2.3-RELEASE/swift-6.2.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz`.
Verifying if checksum of the downloaded archive is valid...
Downloaded archive has a valid checksum.
Swift SDK bundle at `https://download.swift.org/swift-6.2.3-release/static-sdk/swift-6.2.3-RELEASE/swift-6.2.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz` is assumed to be an archive, unpacking...
[...]

PR #1677, opened 2026-01-08
Ignore EOPNOTSUPP error when copying directory xattrs

jmschonfeld:
Avoids throwing an error when copying directories that contain xattrs that the system does not support copying to the destination directory.

### Motivation:

When SELinux security xattrs are applied to a directory, they cannot be copied to the destination directory when copying (and `cp` does not copy them either). Instead of throwing an error when encountering these attributes, we should ignore them.

Resolves https://github.com/swiftlang/swift-foundation/issues/1676

### Modifications:

Avoids throwing an error when setting xattrs when the error code is `EOPNOTSUPP`

### Result:

Directories with xattrs that cannot be copied will still be copied successfully

### Testing:

I was unable to add a unit test since it is not possible for us to set an SELinux attribute on this file in a unit test (doing so requires calling a function like `setfilecon` which requires the host build machine to have the selinux dev package installed which is not guaranteed), but we can validate this by building a toolchain and ensuring the reported issue is resolved

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -1017,7 +1017,10 @@ enum _FileOperations {
                                 valueSize = fgetxattr(srcFD, current, valueBuffer.baseAddress!, valueSize)
                                 if valueSize >= 0 {
                                     if fsetxattr(dstFD, current, valueBuffer.baseAddress!, valueSize, 0) != 0 {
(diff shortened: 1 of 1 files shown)


itingliu (approved):

MaxDesiatov:
@swift-ci test

parkera (approved):

jmschonfeld:
The android build failure is an infrastructure issue being fixed by https://github.com/swiftlang/github-workflows/pull/221

PR #1680, opened 2026-01-11
Implement locale canonicalization for non darwin platforms

### LINKED ISSUE #45: Implement `Locale.canonicalIdentifier`

IvanBl:
Implemented locale canonicalization for non darwin platforms.
 
### Motivation:
`Locale.canonicalIdentifier()` and `Locale.canonicalLanguageIdentifier()`  were not implemented for non-Darwin platforms (Linux), returning input unchanged instead of canonicalizing locale identifiers according to unicode standards.

Fixes #45

### Modifications:
- [x] Added ICU-based implementation using `uloc_canonicalize` in `Locale_ICU.swift`
- [x] Implemented dynamic replacement pattern to provide ICU implementation for non darwin platforms
- [x] Updated `Locale.swift` to call platform specific implementations via dynamic replacement hooks
- [x] Re-enabled 3 previously disabled test functions
- [x] Added 2 new test functions covering common canonicalization scenarios 

### Result:
Locale canonicalization now works correctly on all platforms. Non darwin platforms properly normalize locale identifiers according to unicode standards, matching darwin behavior.

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale_ICU.swift
@@ -1760,14 +1760,119 @@ extension Locale {
 }
 
 // MARK: - Locale Canonicalization
+
+private let legacyLanguageMappings: [String: String] = [
+    "no": "nb",   // Norwegian
+    "iw": "he",   // Hebrew
+    "in": "id",   // Indonesian
+    "ji": "yi",   // Yiddish
+    "jw": "jv",   // Javanese
+    "jaw": "jv",  // Javanese alternate
+]
+
 internal func canonicalizeLocaleIdentifier(_ identifier: String) -> String {
     guard !identifier.isEmpty else { return identifier }
+    let isBCP47 = identifier.contains("-")
 
-    return _withFixedCharBuffer { buffer, size, status in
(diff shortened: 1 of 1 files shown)


itingliu:
@swift-ci please test

IvanBl:
Added implementation has several conversion steps that might be too complex:
1. ICU's `uloc_canonicalize` doesn't handle some deprecated codes (`no`, `iw`, `in`, `ji`, `jw`, `jaw`) and they have to be mapped manually.
2. I tried to use `uloc_minimizeSubtags` but it removes too much, `zh-Hant-HK` becomes `zh-HK`.
3. Conversions to match darwin 

Is there something I might have missed? Or should Linux/Windows use a simpler implementation that doesn't match darwin's exact format?

Issue #284, opened 2023-10-05
Make String.compare with locale argument available in FoundationInternationalization


parkera:
Port the `String.compare(_:options:locale:)` implementation to FoundationInternationalization. We currently have the unlocalized compare but not the localized one.

IvanBl:
Hi @itingliu,

Sorry I didn't realize this was assigned when I started working on it. 
Just opened PR #1683 with implementation. If you are already working on this, feel free to close my PR.
Otherwise would love to hear your feedback. 

Thanks!

itingliu:
@IvanBl Thank you for picking it up! 

Some of the challenges that blocked me from doing this are:

1. Darwin compatibility issue: Darwin foundation has a custom implementation that is inconsistent with ICU's collator strengths.  Darwin Foundation's implementation is also incorrect in some cases. But I don't know how destructive it'd be if we fix the bugs, let alone changing the behavior, without testing the changes extensively. Also ICU's collator isn't the most performant. If we are to adopt ICU universally, we'd need to benchmark and do some perf work around this area.
2. Limitation of ICU collator: As you probably had noticed in your implementation, ICU collator configuration doesn't support all combinations of the options of this API. 
3. Platform parity: one of the goals of swift-foundation is that clients get consistent behavior across platforms. Forking the implementation break this commitment.
4. Ambition to provide a better API: There are improvements that this API can use. For example the options defined here don't all work well with each other: `regex` option overrides all other options; `forceOrdering` option doesn't specify how the ordering is resolved; `literal` is an objective-c concept, and doesn't sit great in Swift String. 

So to revisit our options
1. Bring Darwin Foundation implementation over to swift-foundation: 
    - What works: (1)-(3)
    - Downside/Effort: brings existing bugs over; requires a huge effort; and misses the chance of using ICU to get the correct behavior
2. Use ICU everywhere: 
    - What works: (3)
    - Downside/Effort: everything else until we invest time in tuning and triaging behavioral differences 
3. Deprecate the existing API and introduce a replacement that follows ICU collator rule
    - What works: All
    - Downside/Effort: needs new API design and performance tuning if implemented using ICU collator 
4. Fork the implementation to use ICU for linux only (your approach in #1683)
    - What works: (1)
    - Downside: I think the biggest one for me is lack of platform parity (3)

itingliu:
[...]

PR #1683, opened 2026-01-12
Implement String.compare with locale for non darwin platforms

IvanBl:
Implemented locale string comparison for non darwin platforms using ICU.

### Motivation:
Issue #284 requests `String.compare(_:options:locale:)` for Linux/Windows. Currently this API only exists on Darwin, preventing proper internationalization for server side Swift applications.

### Modifications:
- [x] Added optional `locale` parameter to `StringProtocol.compare`
- [x] Implemented ICU-based comparison
- [x] Added support for `.caseInsensitive`, `.diacriticInsensitive` and `.numeric` 

### Result:
String comparison now respects locale on Linux/Windows

### Testing:
Added unit tests that cover various locales and options.

### **Note:**
Please note that this PR implements the most commonly used options
```
.caseInsensitive
.diacriticInsensitive
.numeric
```

Additional options like `.widthInsensitive`, `.backwards` and others could also be added later if needed.

### DIFF
--- a/Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
+++ b/Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
@@ -31,29 +31,17 @@ internal func compareStringsWithLocale(_ string1: Substring, _ string2: Substrin
     }
 
     guard let collator = collator, status.rawValue <= U_ZERO_ERROR.rawValue else {
-        if string1 < string2 {
-            return .orderedAscending
-        } else if string1 > string2 {
-            return .orderedDescending
-        } else {
-            return .orderedSame
-        }
+        return ComparisonResult(string1, string2)
     }
 
     configureCollator(collator, options: options, status: &status)
     guard status.rawValue <= U_ZERO_ERROR.rawValue else {
-        if string1 < string2 {
-            return .orderedAscending
-        } else if string1 > string2 {
-            return .orderedDescending
-        } else {
-            return .orderedSame
-        }
+        return ComparisonResult(string1, string2)
     }
 
-    let result = string1.withCString(encodedAs: UTF16.self) { str1Ptr in
-        string2.withCString(encodedAs: UTF16.self) { str2Ptr in
-            ucol_strcoll(collator, str1Ptr, -1, str2Ptr, -1)
+    let result = string1.withCString { str1Ptr in
+        string2.withCString { str2Ptr in
+            ucol_strcollUTF8(collator, str1Ptr, -1, str2Ptr, -1, &status)
         }
     }
 
--- a/Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
+++ b/Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
@@ -21,165 +21,171 @@ import Testing
 
 #if FOUNDATION_FRAMEWORK || FOUNDATION_ICU_STRING_COMPARE
 
-@Test func caseInsensitiveComparison() {
-    let locale = Locale(identifier: "en_US")
+@Suite("String Comparison with Locale")
+private struct StringComparisonLocaleTests {
 
-    let result1 = "Hello".compare("hello", options: .caseInsensitive, locale: locale)
-    #expect(result1 == .orderedSame)
+    @Test func caseInsensitiveComparison() {
+        let locale = Locale(identifier: "en_US")
 
-    let turkishLocale = Locale(identifier: "tr_TR")
-    let result2 = "i".compare("I", options: .caseInsensitive, locale: turkishLocale)
-    #expect(result2 == .orderedDescending)
-}
+        let result1 = "Hello".compare("hello", options: .caseInsensitive, locale: locale)
+        #expect(result1 == .orderedSame)
 
-@Test func diacriticInsensitiveComparison() {
-    let locale = Locale(identifier: "en_US")
+        let turkishLocale = Locale(identifier: "tr_TR")
+        let result2 = "i".compare("I", options: .caseInsensitive, locale: turkishLocale)
+        #expect(result2 == .orderedDescending)
+    }
 
-    let result1 = "café".compare("cafe", options: .diacriticInsensitive, locale: locale)
-    #expect(result1 == .orderedSame)
+    @Test func diacriticInsensitiveComparison() {
+        let locale = Locale(identifier: "en_US")
 
-    let result2 = "Café".compare("cafe", options: .diacriticInsensitive, locale: locale)
-    #expect(result2 == .orderedDescending)
-}
+        let result1 = "café".compare("cafe", options: .diacriticInsensitive, locale: locale)
+        #expect(result1 == .orderedSame)
 
-@Test func caseAndDiacriticInsensitive() {
-    let locale = Locale(identifier: "en_US")
+        let result2 = "Café".compare("cafe", options: .diacriticInsensitive, locale: locale)
+        #expect(result2 == .orderedDescending)
+    }
 
-    let result = "Café".compare("cafe", options: [.caseInsensitive, .diacriticInsensitive], locale: locale)
-    #expect(result == .orderedSame)
-}
+    @Test func caseAndDiacriticInsensitive() {
+        let locale = Locale(identifier: "en_US")
 
-@Test func caseInsensitivePreservesDiacritics() {
-    let locale = Locale(identifier: "en_US")
+        let result = "Café".compare("cafe", options: [.caseInsensitive, .diacriticInsensitive], locale: locale)
+        #expect(result == .orderedSame)
+    }
 
-    let result1 = "Café".compare("Cafe", options: .caseInsensitive, locale: locale)
-    #expect(result1 == .orderedDescending)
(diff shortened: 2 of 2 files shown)


Sources/FoundationEssentials/String/StringProtocol+Stub.swift
```diff
@@ -15,14 +15,22 @@
 extension StringProtocol {
     /// Compares the string using the specified options and
     /// returns the lexical ordering for the range.
-    package func compare<T : StringProtocol>(_ aString: T, options mask: String.CompareOptions = [], range: Range<Index>? = nil) -> ComparisonResult {
-        // TODO: This method is modified from `public func compare<T : StringProtocol>(_ aString: T, options mask: String.CompareOptions = [], range: Range<Index>? = nil, locale: Locale? = nil) -> ComparisonResult`. Move that method here once `Locale` can be staged in `FoundationEssentials`.
+    public func compare<T : StringProtocol>(_ aString: T, options mask: String.CompareOptions = [], range: Range<Index>? = nil, locale: Locale? = nil) -> ComparisonResult {
```
itingliu, line 27: I think this `public func` should also be guarded with the flag. Having `locale` as an argument makes it look like localized compare is supported, but in fact it just calls into `unlocalizedCompare` currently, which wouldn't be truthful...

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if !FOUNDATION_FRAMEWORK
+@_implementationOnly import FoundationEssentials
```
itingliu: We don't want to use `@ _implementationOnly` here. See more https://github.com/swiftlang/swift-evolution/blob/d89189e7b438d90fd121ead562530d13962f8a37/proposals/0409-access-level-on-imports.md

```suggestion
#if canImport(FoundationEssentials)
import FoundationEssentials
#endif
```

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+//
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if !FOUNDATION_FRAMEWORK
```
itingliu: I'm thinking maybe we can just do this
```suggestion
#if FOUNDATION_ICU_STRING_COMPARE
```

And we can decide what this resolves to independent of what `FOUNDATION_FRAMEWORK` is

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+@_implementationOnly import FoundationEssentials
+internal import _FoundationICU
+
+internal func compareStringsWithLocale(_ string1: Substring, _ string2: Substring, options: String.CompareOptions, locale: Locale) -> ComparisonResult {
+    if options.contains(.literal) {
+        if string1 < string2 {
```
itingliu: Comparing strings like this doesn't really take the locale into account, and ignores other options as well. Is this correct?

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+            return .orderedAscending
+        } else if string1 > string2 {
+            return .orderedDescending
+        } else {
+            return .orderedSame
+        }
+    }
+
+    defer { ucol_close(collator) }
```
itingliu: I'm not sure if `collator` is always nil when there's an error. Can we always close it regardless of what `status` is?

Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
```diff
@@ -0,0 +1,189 @@
...
+// Copyright (c) 2014 - 2017 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(TestSupport)
+import TestSupport
```
itingliu: What's this needed for?

itingliu:
@swift-ci please test

Sources/FoundationEssentials/String/StringProtocol+Stub.swift
IvanBl: Thanks for pointing this out! You're right. I have updated it so we have the original func compare without locale always available, and the public func guarded by feature flag.

Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
IvanBl: you are right, I removed it, since not used. I just saw initially that other files contain it and thought it would be needed for the tests.

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
IvanBl: Good point, I have moved `defer` before the guard so we always close it

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+        } else if string1 > string2 {
+            return .orderedDescending
+        } else {
+            return .orderedSame
+        }
+    }
+
+    let result = string1.withCString(encodedAs: UTF16.self) { str1Ptr in
+        string2.withCString(encodedAs: UTF16.self) { str2Ptr in
+            ucol_strcoll(collator, str1Ptr, -1, str2Ptr, -1)
```
itingliu: This is fine, but ICU also has a UTF8 version `ucol_strcollUTF8` that is supposed to be better for Swift strings

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+    })
+
+    defer {
+        if let collator = collator {
+            ucol_close(collator)
+        }
+    }
+
+    guard let collator = collator, status.rawValue <= U_ZERO_ERROR.rawValue else {
+        if string1 < string2 {
```
itingliu: There's an extension in `String+Comparison.swift` that should allow you to use `return ComparisonResult(string1, string2)` for this blob of code

Sources/FoundationInternationalization/String/String+Comparison_ICU.swift
```diff
@@ -0,0 +1,96 @@
...
+        } else if string1 > string2 {
+            return .orderedDescending
+        } else {
+            return .orderedSame
+        }
+    }
+
+    configureCollator(collator, options: options, status: &status)
+    guard status.rawValue <= U_ZERO_ERROR.rawValue else {
+        if string1 < string2 {
```
itingliu: Ditto here

Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
```diff
@@ -0,0 +1,185 @@
...
+@testable import Foundation
+#else
+@testable import FoundationEssentials
+@testable import FoundationInternationalization
+#endif // FOUNDATION_FRAMEWORK
+
+import Testing
+
+#if FOUNDATION_FRAMEWORK || FOUNDATION_ICU_STRING_COMPARE
+
```
itingliu, line 23: Can we group these tests with a [`Suite`](https://developer.apple.com/documentation/testing/organizingtests)?

Tests/FoundationInternationalizationTests/StringComparisonLocaleTests.swift
```diff
@@ -0,0 +1,185 @@
...
+    #expect(result3 == .orderedAscending)
+}
+
+@Test func literalComparison() {
+    let locale = Locale(identifier: "en_US")
+
+    let result1 = "café".compare("cafe", options: .literal, locale: locale)
+    #expect(result1 == .orderedDescending)
+
+    let result2 = "Test".compare("test", options: .literal, locale: locale)
```
itingliu: I don't think using `literal` or normalization or not makes a difference when the test strings are only different in casing. Could we update this test to be more relevant?

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #1684, opened 2026-01-13
Staging for new default AttributeScope lookup approach

jmschonfeld:
Stages new declarations to improve the default `AttributeScope` lookup performance

### Motivation:

Looking up the default attribute scope currently goes through `dlopen` to find all necessary higher level frameworks. `dlopen` is much more expensive than `objc_getClass`, so this stages in a class that higher level frameworks can subclass which we can lookup by name to find scopes that need to be included rather than using `dlopen`

### Modifications:

Add a new class for higher level frameworks to subclass that Foundation will lookup by name (the latter half will come in a later change)

### Result:

The new class is available for subclassing

### Testing:

This does not enable the new behavior yet, so there is no change to validate via testing (aside from ensuring this still builds)


jmschonfeld:
@swift-ci please test

PR #1686, opened 2026-01-14
Update 0034 review period

itingliu:


itingliu:
@swift-ci please test

PR #1687, opened 2026-01-14
Reading extended attributes can read uninitialized memory

jmschonfeld:
When `FileManager` reads the list of extended attributes, there is a TOCTOU issue that may cause it to read an uninitialized portion of a buffer

rdar://168089283

### Motivation:

`FileManager` calls `listxattr` twice - once to get the size of the buffer and another to populate the buffer. If the file system is modified between the two calls, the populated portion of the buffer may be smaller than the original allocation size if an xattr is removed from the file between the calls (it cannot be larger than the buffer because otherwise the second xattr call may fail). `FileManager` currently reads from the full allocation rather than just the populated section which can result in a buffer over-read if the non-populated portion of the buffer is not zeroed.

### Modifications:

Ensure that the calculated end index for what we read is based on the result of the second `listxattr` call rather than the first.

### Result:

`FileManager` now only reads from the initialized portion of the buffer

### Testing:

We can't write a unit test for this since it's triggered by a race between two different operations, but I have high confidence this should resolve the issue based on code inspection. I did, however, add a simple read/write test of the xattr values (which passes today) because we did not already have one).

### DIFF
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -216,6 +216,14 @@ private struct FileManagerTests {
         #endif
     }
     
+    private static var isLinux: Bool {
+        #if os(Linux)
+        true
+        #else
+        false
+        #endif
+    }
+    
     private func randomData(count: Int = 10000) -> Data {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

itingliu (approved):

jmschonfeld:
@swift-ci please test

PR #1691, opened 2026-01-21
[CI] Update the macOS version to Tahoe

shahmishal:


.github/workflows/pull_request.yml
```diff
@@ -13,13 +13,13 @@ on:
...
     name: Test (SwiftPM)
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.5
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.6
     with:
       linux_swift_versions: '["nightly-main"]'
       windows_swift_versions: '["nightly-main"]'
       enable_macos_checks: false
       macos_xcode_versions: '["16.3"]'
-      macos_versions: '["sequoia"]'
+      macos_versions: '["tahoe"]'
```
jmschonfeld, line 22: FYI: we don't currently enable Github Action macOS testing because it does not support building with nightly toolchains. Seems fine to update this if you want, but it won't have any effect due to the `enable_macos_checks: false` above

shahmishal: Let's update it for now, so we don't forget to do this once we enable it.

jmschonfeld:
@swift-ci please test

PR #1657, opened 2025-12-19
Use `_TimeZoneGMTICU` for timezones whose identifier take the form of "GMT/UTC+<offset>" 

itingliu:
Use `_TimeZoneGMTICU` for timezones whose identifiers take the form of "GMT/UTC+<offset>" such as `"GMT+8"` for performance reasons.

Also update to uatimezone API for localizing names.

### Result:

Various improvements in calling time zone API on `TimeZone(identifier: GMT+8)`

Screenshot 2025-12-16 at 09 57 13

### Testing:

Added tests and benchmarks.

PR #1695, opened 2026-01-23
Use _TimeZoneGMTICU instead of using _TimeZoneICU for timezones such as "GMT+8" for bridged timezone

itingliu:
This is a follow up for #1657. In that change, we switched to use `_TimeZoneGMTICU` instead of `_TimeZoneICU` for timezones whose identifier is a fixed offset, e.g. "GMT+8". This applies the same logic to bridged timezone.

I did not add new tests because we have plenty of tests for `NSTimeZone` and `CFTimeZone` in Objective-c that covers this code path.

167791762


itingliu:
@swift-ci please test

PR #1696, opened 2026-01-24
Only calculate needed values for _CalendarGregorian.date(_:from:)

itingliu:
The existing `_CalendarGregorian.date(_:from:)` implementation unconditionally calculates week-related fields regardless what the passed-in `DateComponents` contains. We should not do that unless they are requested.

166835028

### Result:

Performance improvements
Screenshot 2026-01-02 at 10 56 46 am

### Testing:

Added benchmark


itingliu:
@swift-ci please test

PR #1697, opened 2026-01-26
ICU.StringConverter: Avoid memory leaks after encoding failures.

YOCKOW:
Current implementation has a bug that doesn't free an allocated `UnsafeMutableRawPointer` when encoding is failed.
This PR fixes it.

### DIFF
--- a/Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
+++ b/Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Swift.org open source project
 //
-// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Copyright (c) 2025-2026 Apple Inc. and the Swift project authors
 // Licensed under Apache License v2.0 with Runtime Library Exception
 //
 // See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


YOCKOW:
@swift-ci Please test

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
```diff
@@ -103,6 +103,12 @@ extension ICU.StringConverter {
                 byteCount: capacity,
                 alignment: MemoryLayout<CChar>.alignment
             )
+            var shouldFreeDest = true
```
jmschonfeld, line 106: In the long term, we probably want to rewrite this to have `Data` allocate the storage and let us write into it directly (via an `OutputRawSpan`) rather than creating storage ourselves and conditionally handing it off to `Data` or free-ing it, but since we don't have the right initializer for `Data` in-place yet this seems like a good short term resolution. Thanks!

jmschonfeld:
@swift-ci please test

Sources/FoundationInternationalization/ICU/ICU+StringConverter.swift
YOCKOW: Thank you for the comment.
I also hope `Data` would get more friendly with `Span` family. :)

PR #1700, opened 2026-01-28
Specify lifetime for protocol requirement

glessard:
A protocol requirement with a non-escapable parameter should specify that parameter's lifetime.

The existing conformance to this protocol already specifies this lifetime.


glessard:
@swift-ci please test

PR #1701, opened 2026-01-28
Guarantee specialization of String decomposition implementation

jmschonfeld:
Requires the compiler to specialize the core String decomposition implementation used for creating file system representations.

### Motivation:

In many cases the compiler already chooses to specialize this, but it does not always do so. When it does not specialize this code, the unspecialized generics introduce significant overhead to this important funnel point.

### Modifications:

- Use `@specialize` to require that the compiler specialize this for UTF-8 (and UTF-16 in `FOUNDATION_FRAMEWORK` where we may be given a UTF-16 buffer from an `NSString`)
- use `@inline(__always)` to always inline the trivially forwarding generic wrapper around this function to avoid any overhead from that function

### Result:

This function should no longer have any unspecialized generics

### Testing:

Existing unit tests cover usage of this function, and I've manually validated that the compiler emits specialized versions of this function.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -262,7 +262,8 @@ extension UnsafeBufferPointer {
         case illegalScalar
         case decodingError
     }
-    
+
+    @inline(__always) // Trivially forwarding and generic, inlining ensures calls are fully specialized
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1702, opened 2026-01-28
Fix resolution for fragment-only relative NSURLs

jrflat:
Fix absolute URL resolution for fragment-only relative NS/CFURLs when their Swift implementation is enabled.

### Motivation:

Bug fix to preserve the fragment from fragment-only URLs like `#fragment` when resolving against a base URL.

### Modifications:

Ensure that we always copy the relative fragment; this logic was missing in the case where the relative URL is only a fragment and the base URL does not have a query.

### Result:

A relative URL like `#fragment` will resolve correctly agains a URL like `https://example.com/path` to create an absolute URL `https://example.com/path#fragment`.

### Testing:

Internal unit tests added for `NSURL` and `CFURL` when their Swift implementation is enabled.


jrflat:
@swift-ci please test

PR #1703, opened 2026-01-30
Guard against integer overflow

itingliu:
In a recent change we changed to use integer math for better performance. However, this approach doesn't work if the date offset of the input Date cannot be represented by a Int, which would trigger a swift runtime error.

### Modifications:

Fix this by checking if the offset is representable as an Int, and fallback to the previous implementation with floating point math if it isn't.

### Testing:

Added a test that would trigger a runtime error before the fix, but passes after the fix.

168600685


itingliu:
@swift-ci please test

PR #1709, opened 2026-02-02
Compatibility resolution behavior for Swift CFURL

jrflat:
Implement RFC 1808 URL resolution as a compatibility path for `CFURL`.

### Motivation:

`CFURL` uses the absolute URL resolution algorithm from RFC 1808, which differs notably from the modern RFC 3986 and WHATWG behavior by 1) not stripping leading `"../"` or `"/../"` segments and 2) therefore requiring us to check if `<segment>` in `"<segment>/../"` is equal to `".."` before removing it. `CFURL` also does not resolve dot segments if the relative path itself is absolute.

While switching to the modern behavior isn't likely to affect real use-cases, there is some compatibility risk of breaking others' unit tests or having code that compares resolved paths for equality, so we can provide the RFC 1808 path for now and test migrating to the new resolution behavior separately from the whole re-core.

### Modifications:

This PR implements a compatibility path for RFC 1808 resolution behavior that preserves leading `".."` segments.

### Result:

`CFURL` will see the same absolute URL resolution behavior with the Swift implementation.

### Testing:

Verified unit tests for `NSURL` and `CFURL` with the Swift implementation enabled.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_C+Resolution.swift
+++ b/Sources/FoundationEssentials/URL/URL_C+Resolution.swift
@@ -18,7 +18,7 @@ internal import _ForSwiftFoundation
 
 @objc
 extension NSURL {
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

PR #1711, opened 2026-02-03
Simplify #bundle expansion with build configuration

jmschonfeld:
Uses the new build configuration information provided to the macro during expansion to conditionalize the output of the `#bundle` macro.

### Motivation:

Today, `#bundle` expands to a series of `#if`/`#elseif`/`#else` directives to adjust expansion depending on the build configuration at the invocation site. Now, swift-syntax provides the macro expansion with the list of defined build conditions at the location of expansion. We can use this to have `#bundle` expand to just the appropriate initializer rather than the `#if` directives. This makes the expanded form of `#bundle` more readable, provides better diagnostics, and should reduce post-expansion parse time.

### Modifications:

Updates `#bundle` to only emit the appropriate code evaluating the build conditions during expansion

### Result:

Now `#bundle` expands to a single expression (i.e. `Bundle.module` or `Bundle(_dsoHandle: #dsoHandle)`) depending on the build configuration rather than expanding to a conditional.

### Testing:

Macro unit tests are updated to validate the new expansion across a variety of build configurations (and to validate the diagnostic behavior on the failure paths).

### DIFF
--- a/Sources/FoundationMacros/BundleMacro.swift
+++ b/Sources/FoundationMacros/BundleMacro.swift
@@ -13,21 +13,45 @@
 import SwiftSyntax
 import SwiftSyntaxMacros
 internal import SwiftSyntaxBuilder
+internal import SwiftDiagnostics
+internal import SwiftIfConfig
+
+private struct BundleExpansionDiagnostic: DiagnosticMessage {
+    let message: String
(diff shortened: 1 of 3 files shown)


jmschonfeld:
@swift-ci please test

PR #1714, opened 2026-02-03
Turn PathOrURL type into a protocol, allowing for future noncopyable file path types

parkera:
Turn PathOrURL type into a protocol, allowing for future noncopyable file path types.

### Motivation:

Instead of using an enum for `PathOrURL`, turn it into a protocol. This allows for both `String` and `URL` to conform, as well as opening us up to allow for other types that can provide a file system representation.

### Modifications:

Rename `PathOrURL.swift`, and its contents.

### Result:

No functional change, code only.

### Testing:

Existing testing covers these paths.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -194,7 +194,7 @@ struct ReadBytesResult {
 
 #if os(Windows)
 @lifetime(pBuffer: copy pBuffer)
-private func read(from hFile: HANDLE, at path: FileSystemRepresentable,
+private func read(from hFile: HANDLE, at path: borrowing some FileSystemRepresentable & ~Copyable,
                   into pBuffer: inout OutputRawSpan,
                   chunkSize dwChunk: Int = 4096, progress bProgress: Bool)
         throws {
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -81,16 +81,12 @@ extension POSIXError {
 }
 
 extension CocoaError {
-    static func errorWithFilePath(_ pathOrURL: borrowing some FileSystemRepresentable & ~Copyable, errno: Int32, reading: Bool, variant: String? = nil, source: String? = nil, destination: String? = nil, debugDescription: String? = nil) -> CocoaError {
-        pathOrURL
-            .errorWithFilePath(
-                errno: errno,
-                reading: reading,
-                variant: variant,
-                source: source,
-                destination: destination,
-                debugDescription: debugDescription,
-            )
+    static func errorWithFilePath(_ path: borrowing some FileSystemRepresentable & ~Copyable, errno: Int32, reading: Bool, variant: String? = nil, source: String? = nil, destination: String? = nil, debugDescription: String? = nil) -> CocoaError {
+        if let url = path.urlForError {
+            return CocoaError(Code(fileErrno: errno, reading: reading), url: url, underlying: POSIXError(errno: errno), variant: variant, source: source, destination: destination, debugDescription: debugDescription)
+        } else {
+            return CocoaError(Code(fileErrno: errno, reading: reading), path: path.path, underlying: POSIXError(errno: errno), variant: variant, source: source, destination: destination, debugDescription: debugDescription)
+        }
     }
 }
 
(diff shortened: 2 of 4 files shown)


parkera:
@swift-ci test

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -194,7 +194,7 @@ struct ReadBytesResult {
 
 #if os(Windows)
 @lifetime(pBuffer: copy pBuffer)
-private func read(from hFile: HANDLE, at path: PathOrURL,
+private func read(from hFile: HANDLE, at path: FileSystemRepresentable,
```
jmschonfeld:
```suggestion
private func read(from hFile: HANDLE, at path: borrowing some FileSystemRepresentable & ~Copyable,
```

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -144,7 +139,7 @@ extension CocoaError.Code {
 }
 
 extension CocoaError {
-    static func errorWithFilePath(_ path: PathOrURL, win32 dwError: DWORD, reading: Bool, debugDescription: String? = nil) -> CocoaError {
+    static func errorWithFilePath(_ path: FileSystemRepresentable, win32 dwError: DWORD, reading: Bool, debugDescription: String? = nil) -> CocoaError {
         switch path {
```
jmschonfeld: I think this likely needs to be updated since it's no longer an `enum`. Also should this be `borrowing some FileSystemRepresentable & ~Copyable` too?

Sources/FoundationEssentials/String/String+IO.swift
```diff
@@ -250,15 +250,15 @@ extension String {
...
     }
 
     /// Produces a string created by reading data from a given URL and returns by reference the encoding used to interpret the data.
     public init(contentsOf url: __shared URL, usedEncoding: inout Encoding) throws {
-        self = try String(contentsOfFileOrPath: .url(url), usedEncoding: &usedEncoding)
+        self = try String(contentsOfFileOrPath: url, usedEncoding: &usedEncoding)
     }
     
-    internal init(contentsOfFileOrPath path: PathOrURL, usedEncoding: inout Encoding) throws {
+    internal init(contentsOfFileOrPath path: borrowing FileSystemRepresentable, usedEncoding: inout Encoding) throws {
```
jmschonfeld:
```suggestion
    internal init(contentsOfFileOrPath path: borrowing some FileSystemRepresentable & ~Copyable, usedEncoding: inout Encoding) throws {
```

Sources/FoundationEssentials/Data/Data+Reading.swift
kperryua: `borrowing some FileSystemRepresentable & ~Copyable` ?

parkera:
@swift-ci test

Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -144,14 +135,8 @@ extension CocoaError.Code {
...
-    static func errorWithFilePath(_ path: PathOrURL, win32 dwError: DWORD, reading: Bool, debugDescription: String? = nil) -> CocoaError {
-        switch path {
-        case let .path(path):
-            return CocoaError(.init(win32: dwError, reading: reading, emptyPath: path.isEmpty), path: path, underlying: Win32Error(dwError), debugDescription: debugDescription)
-        case let .url(url):
-            let pathStr = url.withUnsafeFileSystemRepresentation { String(cString: $0!) }
-            return CocoaError(.init(win32: dwError, reading: reading, emptyPath: pathStr.isEmpty), path: pathStr, url: url, underlying: Win32Error(dwError), debugDescription: debugDescription)
-        }
+    static func errorWithFilePath(_ path: borrowing some FileSystemRepresentable & ~Copyable, win32 dwError: DWORD, reading: Bool, debugDescription: String? = nil) -> CocoaError {
+        return CocoaError(.init(win32: dwError, reading: reading, emptyPath: path.isEmpty), path: path.path, underlying: Win32Error(dwError), debugDescription: debugDescription)
```
jmschonfeld, line 139: Tests pass so I assume the answer is probably, but is `.path` guaranteed to be the same as `withUnsafeFileSystemRepresentation { String(cString: $0!) }` on Windows? Not sure if one or the other has any special behavior w.r.t. drive letters or something similar that we'd be losing here.

parkera: If anything I think it's probably a better way to get the path for the error.

PR #1715, opened 2026-02-03
Cache the result of getting likely subtags in Locale

parkera:
Cache the result of getting likely subtags in Locale

### Motivation:

Getting the list of likely subtags is fairly expensive, and we can cache the result that we get from ICU.

### Modifications:

Add a cache to `Locale_Cache`. This moves the check of FoundationICU availability into the cache instead of `NSLocale`.

This only affects the framework version of Foundation.

### Result:

Faster likely subtag lookup, when we know the result.

### Testing:

Existing tests cover this functionality.


parkera:
@swift-ci test

parkera:
@swift-ci test

PR #1716, opened 2026-02-04
Disable large data tests on watchOS

jmschonfeld:
Disables the `LargeDataTests` suite on watchOS

### Motivation:

These tests are isolated because they allocate very large amounts of memory to test `Data`'s large slice paths (~2GB of memory). In environments like watchOS, this can cause the test runner to be terminated due to memory growth.

### Modifications:

On watchOS, these tests are marked as disabled so that they will be skipped to avoid terminating the test runner.

### Result:

The large data tests are skipped on watchOS but continue running on all platforms

### Testing:

This is a unit-test only change


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1717, opened 2026-02-04
Swift implementation for ICU UATimeZone and Calendar API used by TimeZone

itingliu:
Introduced two TimeZone classes that implements core time zone functionalities without calling into ICU's time zone class.
    
Currently we still use ICU's packaged time zone data to create the time zone.

The two new types:

- `_TimeZoneICUResource`: like ICU's OlsonTimeZone, this class has knowledge of all historical time zone transitions of a given time zone identifier.
- `_TimeZoneSingleDSTRule`: like ICU's SimpleTimeZone class, this class owns knowledge of a single time zone rule.

The general DST lookup logic works like the following:

- If the query Date is further than the last known time zone rule change, forward the call to `_TimeZoneSingleDSTRule`.
- Else, `_TimeZoneICUResource` finds the rule used around Date and do the calculation accordingly.

Also add a wrapper `ICU.ResourceBundle` to help reading time zone data from ICU resources.

167626187

### DIFF
--- a/Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
+++ b/Sources/FoundationInternationalization/TimeZone/TimeZone_ICU.swift
@@ -44,9 +44,8 @@ private func _timeZoneIdentifier_ICU(forWindowsIdentifier windowsIdentifier: Str
 internal import _ForSwiftFoundation
 
 internal func foundation_swift_ICUResourceTimeZone_feature_enabled() -> Bool {
-    // TODO: fetch this from feature flag
-    return false
-}
+     _foundation_swift_ICUResourceTimeZone_feature_enabled()
+ }
 #else
 internal func foundation_swift_ICUResourceTimeZone_feature_enabled() -> Bool { return false }
 #endif


Sources/FoundationInternationalization/ICU/ICU+ResourceBundle.swift
```diff
@@ -0,0 +1,128 @@
...
+            self.bundle = existing
+        }
+
+        deinit {
+            ures_close(bundle)
+        }
+
+        func resourceBundle(forKey key: String) throws(ICUError) -> ResourceBundle? {
+            var status: UErrorCode = U_ZERO_ERROR
+            let subBundle = ures_getByKey(bundle, key, nil, &status)
```
parkera: Presumably this is thread safe?

Sources/FoundationInternationalization/ICU/ICU+ResourceBundle.swift
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        func asIntegers() throws(ICUError) -> [Int32] {
+            var length: Int32 = 0
+            var status: UErrorCode = U_ZERO_ERROR
+
+            let vector = ures_getIntVector(bundle, &length, &status)
+            try status.checkSuccess()
+
+            return Array(UnsafeBufferPointer(start: vector!, count: Int(length)))
```
parkera: If this is a permanent allocation, does it make sense to return a `Span<Int32>` instead?

parkera: Even if later it is turned into an array, it pushes the allocation up the stack to the place that really needs it.

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
```diff
@@ -0,0 +1,729 @@
...
+            throw ICUError(code: U_INVALID_FORMAT_ERROR)
+        }
+
+        guard let ruleBundle = findRuleByName(rulesBundle: rulesBundle, ruleName: ruleName) else {
+            throw ICUError(code: U_INVALID_FORMAT_ERROR)
+        }
+
+        // ICU DST rules are stored as integer vectors with the following format:
+        // [month, dayOfWeekInMonth, dayOfWeek, time, timeMode, dstSavings, ...]
+        let ruleData = try ruleBundle.asIntegers()
```
parkera, line 161: Related to my earlier comment about span - this would save a temporary allocation of the array of integers since we just fetch them out of the list here.

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
```diff
@@ -0,0 +1,729 @@
...
+            lastMid = mid
+            
+            do {
+                guard let subBundle = try? bundle.resourceBundle(forIndex:mid) else {
+                    break
+                }
+                let idString = try subBundle.asString()
+                
+                let comparison = identifier.compare(idString)
+                if comparison == .orderedSame {
```
parkera, line 293: So the list is binary sorted by `idString`, and we're sure that it uses the same order as we do for `compare`?

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
```diff
@@ -0,0 +1,729 @@
...
+            return firstTransition
+        } else {
+            // No DST transitions in final zone, use standard time
+            return finalStartDate
+        }
+    }
+
+    // Find next historical transition
+    // Follows ICU's backward search algorithm for now.
+    // TODO: This can be optimized with binary search
```
parkera, line 582: Probably a pretty big perf impact!

Sources/FoundationInternationalization/TimeZone/TimeZone_SingleDSTRule.swift
```diff
@@ -0,0 +1,635 @@
...
+    private func previousMonthLength(year: Int, month: Int) -> Int {
+        month == 0 ? 31 : monthLength(year: year, month: month - 1)
+    }
+
+    private func isLeapYear(_ year: Int) -> Bool {
+        (year % 4 == 0) && ((year % 100 != 0) || (year % 400 == 0))
+    }
+    
+    // Returns the day number from 1970, with day 0 == Jan 1 1970
+    private func daysSince1970(year: Int, month: Int, dayOfMonth: Int) -> Int {
```
parkera, line 388: I don't suppose we have some overlap with `Calendar_Gregorian` logic for these kinds of calendar math?

itingliu: We do have a few, but they're all slightly different from each other. But I'd like to refactor out common logic in a follow up.

Sources/FoundationInternationalization/ICU/ICU+ResourceBundle.swift
itingliu: I believe the `ures_get...` ones are, but on a second thought we probably shouldn't be relying on this assumption. Also while we aren't using any mutating functions now, we might inadvertently use those in the future. 

I'll add a lock

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
itingliu: Yes. This function is only called by `openOlsonTimeZoneResource(identifier:)`, and the bundle loaded inside that function is the "Names" array in zoneinfo64.txt, which is sorted in ascending order. I renamed this function to be more specific.

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
itingliu: Added a comment to do it as a followup

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
itingliu: I'll do it as a followup as well. I want to add a generic binary search helper function for us to share

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #1719, opened 2026-02-05
NSData to Data bridging should avoid eagerly copying bytes when possible

jmschonfeld:
Avoids copying bytes when bridging `NSData` to `Data` when it's possible to maintain compatibility

### Motivation:

If we can avoid copying bytes during `NSData` bridging, we can significantly reduce the amount of time and memory bridging takes. Always avoiding copying is riskier because it changes the semantics of indices when bridging (due to self slicing) but when indices start at 0, we can avoid copying without changing client behavior

### Modifications:

Avoids copying a round tripped bridged `Data` when dealing with a non-slice (or a prefix)

### Result:

Roundtrip bridging a `Data` is now done without copying the underlying byte allocation which is great win for performance

### Testing:

This code only impacts Foundation.framework code which tests this bridging

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Bridging.swift
+++ b/Sources/FoundationEssentials/Data/Data+Bridging.swift
@@ -103,15 +103,25 @@ extension Data {
         let range = swiftData._range!
         let originalBacking = swiftData._backing!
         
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

PR #1720, opened 2026-02-05
Add a UTF8Span FileSystemRepresentation helper

parkera:
Add an internal extension to `UTF8Span` to emit its file system representation into an `OutputBuffer`.

### Motivation:

This part of #1714 which will let us provide more flexible types for file system representations, including noncopyable types. This improves performance when dealing with file paths. 

### Modifications:

Adds the internal extension, but does not use it from anywhere in actual code just yet.

### Testing:

For now, internal only with no additional testing.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Internals.swift
+++ b/Sources/FoundationEssentials/String/String+Internals.swift
@@ -413,7 +413,7 @@ extension UTF8Span {
         var seenNullIdx: Int? = nil
         var iterator = self.makeUnicodeScalarIterator()
         
-        guard span.capacity > 0 else {
+        guard !span.isFull else {
             if !nullTerminated && iterator.next() == nil {
                 // No bytes to write, so an empty buffer is OK
                 return


parkera:
Will require macOS 26.

parkera:
@swift-ci test

parkera:
@swift-ci test

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -400,6 +400,120 @@ extension UnsafeBufferPointer {
     }
 }
 
+extension UTF8Span {
+    internal func fileSystemRepresentation(into span: inout OutputSpan<UInt8>, nullTerminated: Bool) throws {
+        try _decomposed(.hfsPlus, into: &span, nullTerminated: nullTerminated)
+    }
+    
+    fileprivate func _decomposed(_ type: String._NormalizationType, into span: inout OutputSpan<UInt8>, nullTerminated: Bool = false) throws {
```
jmschonfeld, line 408: It'd be interesting to benchmark whether a `self.isKnownASCII` fast path is helpful here. We still need to copy to the output span and scan for embedded null bytes as we do so, so I expect probably not really, but perhaps there's overhead from the unicode scalar decoding that we'd avoid. Also fine to do that as a followup.

Sources/FoundationEssentials/String/String+Internals.swift
jmschonfeld, line 175: Spoke about this offline but noting here, at some point it'd be great to go ahead and update this to `utf8Span._decomposed(...)` so that things like `FileManager` start using the new `UTF8Span` based implementation, but fine for a followup change

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -400,6 +400,120 @@ extension UnsafeBufferPointer {
...
+        let scalarSet = BuiltInUnicodeScalarSet(type: type.setType)
+        // Start at the end of the already-filled part of the span
+        let start = span.count
+        var sortBuffer: [UnicodeScalar] = []
+        var seenNullIdx: Int? = nil
+        var iterator = self.makeUnicodeScalarIterator()
+        
+        guard span.capacity > 0 else {
```
jmschonfeld: I think you meant `freeCapacity` instead of `capacity` here, but using `isFull` seems even more meaningful

```suggestion
        guard !span.isFull else {
```

Sources/FoundationEssentials/String/String+Internals.swift
```diff
@@ -400,6 +400,120 @@ extension UnsafeBufferPointer {
...
+extension UTF8Span {
+    internal func fileSystemRepresentation(into span: inout OutputSpan<UInt8>, nullTerminated: Bool) throws {
+        try _decomposed(.hfsPlus, into: &span, nullTerminated: nullTerminated)
+    }
+    
+    fileprivate func _decomposed(_ type: String._NormalizationType, into span: inout OutputSpan<UInt8>, nullTerminated: Bool = false) throws {
+        let scalarSet = BuiltInUnicodeScalarSet(type: type.setType)
+        // Start at the end of the already-filled part of the span
+        let start = span.count
+        var sortBuffer: [UnicodeScalar] = []
```
jmschonfeld, line 412: Future note: at some point we probably want to use a `UniqueArray` here to remove the CoW checks for this to shave off a bit of the overhead here

jmschonfeld (commented):
A few notes but really just 1 main change suggested for this PR

Sources/FoundationEssentials/String/String+Internals.swift
parkera: Great suggestion, thanks.

parkera:
@swift-ci test

Sources/FoundationEssentials/String/String+Internals.swift
parkera: Agreed - we can update the other version of this too, perhaps when we unify them as suggested above.

parkera:
@swift-ci test

parkera:
Depends on https://github.com/swiftlang/swift-foundation/pull/1725

parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

PR #1721, opened 2026-02-05
Add span functions for binary plist parsing

parkera:
Add some span accessors for `BPlistMap`, which sets us up for future performance improvements when we need to parse a bplist.

### Motivation:

With a span accessor for strings, we can compare strings to the binary contents of a bplist, especially when it is known to be an ASCII string.

### Modifications:

Add new internal functions for Span. Requires macOS 26 or Swift 6.2 or later.

Also changes an internal enum from using `indirect` to... not using it, which results in a performance improvement by avoiding retain/release overhead.

### Testing:

Internal only for now, no additional testing.

### DIFF
--- a/Sources/FoundationEssentials/JSON/BufferView.swift
+++ b/Sources/FoundationEssentials/JSON/BufferView.swift
@@ -44,7 +44,7 @@ internal struct BufferView<Element> {
     }
     
     var bytes: RawSpan {
-        @lifetime(borrow self)
+        @_lifetime(borrow self)
         borrowing get {
             let buffer = UnsafeRawBufferPointer(start: baseAddress, count: count)
             let span = unsafe RawSpan(_unsafeBytes: buffer)
@@ -53,9 +53,10 @@ internal struct BufferView<Element> {
     }
 }
 
-extension BufferView where Element: BitwiseCopyable {
+// TODO: This can be improved once stdlib API for safe conversions for fully-inhabited types is ready
+extension BufferView where Element == UInt8 {
     var span: Span<Element> {
-        @lifetime(borrow self)
+        @_lifetime(borrow self)
         borrowing get {
             let span = unsafe bytes._unsafeView(as: Element.self)
             return _overrideLifetime(span, borrowing: self)
--- a/Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
+++ b/Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
@@ -180,26 +180,18 @@ class BPlistMap : PlistDecodingMap {
     
     /// Fetch an index as an ASCII span. This avoids overhead of intermediate creation of "value" types for a common case of looking for an ASCII key string.
     /// Returns an empty span if the value was not an ASCII string (either not a string at all, or UTF16BE string).
-    @_lifetime(self)
-    func asciiString(at idx: BPlistObjectIndex) throws -> Span<UInt8> {
-        let view: BufferView? = try dataLock.withLockUnchecked { state in
+    func withASCIIString<T>(at idx: BPlistObjectIndex, body: (Span<UInt8>) throws -> T) rethrows -> T {
+        try dataLock.withLockUnchecked { state in
             guard Int(idx) < objectOffsets.count else {
                 throw BPlistError.corruptedValue("object index")
             }
             let offset = objectOffsets[Int(idx)]
             let scanInfo = BPlistScanner(buffer: state.buffer, trailer: trailer)
             if let region = try scanInfo.scanASCIIStringRegion(at: offset) {
-                return state.buffer[region]
+                return try body(state.buffer[region].span)
             } else {
-                return nil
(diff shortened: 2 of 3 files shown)


Sources/FoundationEssentials/JSON/BufferView.swift
```diff
@@ -42,6 +42,25 @@ internal struct BufferView<Element> {
...
+        @lifetime(borrow self)
+        borrowing get {
+            let buffer = UnsafeRawBufferPointer(start: baseAddress, count: count)
+            let span = unsafe RawSpan(_unsafeBytes: buffer)
+            return unsafe _overrideLifetime(span, borrowing: self)
+        }
+    }
+}
+
+extension BufferView where Element: BitwiseCopyable {
```
jmschonfeld: Is it worth leaving a comment here to indicate this should only be used on types that can be safely loaded (since not all `BitwiseCopyable` types are) and that this should be replaced by uses of @glessard new APIs/protocol once they're available?

Sources/FoundationEssentials/JSON/BufferView.swift
```diff
@@ -42,6 +42,25 @@ internal struct BufferView<Element> {
...
+    var span: Span<Element> {
+        @lifetime(borrow self)
```
jmschonfeld: Should this use the `@_lifetime` syntax to be consistent?

Package.swift
```diff
@@ -149,6 +149,7 @@ let package = Package(
           swiftSettings: [
             .enableExperimentalFeature("VariadicGenerics"),
             .enableExperimentalFeature("LifetimeDependence"),
+            .enableExperimentalFeature("Lifetimes"),
```
jmschonfeld, line 152: These will need to be updated in the CMake files too

Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
```diff
@@ -72,7 +72,7 @@ private enum BPlistTypeMarker: UInt8 {
 }
 
 class BPlistMap : PlistDecodingMap {
-    internal indirect enum Value {
+    internal enum Value {
```
kperryua, line 75: This is a subtle change that's probably a big overall perf win.

Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
```diff
@@ -134,6 +134,16 @@ class BPlistMap : PlistDecodingMap {
...
     }
+    
+    @_lifetime(self)
+    @inline(__always)
+    func span(for region: Region) -> Span<UInt8> {
+        let escaped = dataLock.withLock {
+            $0.buffer[region]
+        }
+        // We're just going to assume that the data itself is going to stay alive as long as this map does.
+        return _overrideLifetime(escaped.span, borrowing: self)
```
jmschonfeld, line 145: Just confirming, is this safe? Since `BPlistMap` is a reference type, I think it's possible that you could have:

```
let mapA = /* some BPlistMap */
let mapB = mapA
let span = mapB.span
// span borrows mapB but not mapA which can be accessed concurrently with span
```

Not sure if it's possible to mutate these maps but the combination of escaping the pointer out of the lock + the fact that this is a reference type worries me a little bit 😅 but hopefully there's a detail here that prevents an issue there?

Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
```diff
@@ -134,6 +134,16 @@ class BPlistMap : PlistDecodingMap {
...
         }
     }
+    
+    @_lifetime(self)
+    @inline(__always)
+    func span(for region: Region) -> Span<UInt8> {
+        let escaped = dataLock.withLock {
+            $0.buffer[region]
+        }
+        // We're just going to assume that the data itself is going to stay alive as long as this map does.
```
kperryua, line 144: We need to be careful, but this isn't technically true, as `copyInBuffer` will replace the buffer pointer if/when it gets called. But as long as these spans don't escape out of the actual `.decode` call that they're created in, we're fine.

Package.swift
parkera: Yup, thanks.

Sources/FoundationEssentials/JSON/BufferView.swift
parkera: I think I only need this for `UInt8` now, so I'll just limit it to that instead of attempting to be more generic. Plus the comment.

Sources/FoundationEssentials/PropertyList/BPlistScanner.swift
parkera: I'm going to refactor this to be closure based instead.

parkera:
Depends on https://github.com/swiftlang/swift-foundation/pull/1725

parkera:
@swift-ci test

parkera:
@swift-ci test

PR #1722, opened 2026-02-05
Simplify use of dyld SPI in ObjC-only ProcessInfo extensions

parkera:
Simplify use of dyld SPI in ObjC-only ProcessInfo extensions. This affects the framework side of the build only.


parkera:
@swift-ci test

PR #1723, opened 2026-02-05
Strip trailing slashes when appending to a file CFURL

jrflat:
Match a trailing slash edge case behavior from the old `CFURL` implementation for compatibility.

### Motivation:

The old `CFURL` implementation strips all trailing slashes from an inputted path component or extension before appending, but only for file URLs. Match this behavior for compatibility.

### Modifications:

Strip trailing slashes in `__copySwiftEncodedFSRPath(_:)`, which encodes a path component or extension when appending them to a file `CFURL`. 

### Result:

`NSURL` and `CFURL` methods for appending a path component or extension will maintain their old behavior for file URLs.

### Testing:

Added large unit test for `NSURL` appending edge cases.


jrflat:
@swift-ci please test

jmschonfeld (approved):

Issue #1724, opened 2026-02-06
FileManager.default.attributesOfFileSystem crashes on 32-bit systems with filesystem larger than 4GB


xtremekforever:
It was noticed that a call to `FileManager.attributesOfFileSystem(forPath: path)` would crash on armv7/armhf (which is a 32-bit platform) whenever the filesystem is larger than 4GB: https://github.com/swift-embedded-linux/armhf-debian/issues/10

**To Reproduce**

Here's some example code that was tried:

```swift
        if let attributes = try? FileManager.default.attributesOfFileSystem(forPath: "/") {
            print("File system attributes: \(attributes)")
        }
```

When this is compiled for and run on the Raspberry Pi 2 on armv7, this crashes immediately:

```bash
xtremek@rpi2-trixie32:~ $ ./swift-statvfs-issue 
Hello, world!
Illegal instruction
```

The backtrace reveals that some calculations in the `attributesOfFileSystem` method in swift-foundation are overflowing, causing the crash: [foundation-attributesOfFileSystem-bt.txt](https://github.com/user-attachments/files/25131179/foundation-attributesOfFileSystem-bt.txt)

**Expected behavior**
We would expect that the call to `attributesOfFileSystem` method would succeed on a 32-bit system as it does on 64-bit systems with filesystems larger than 4GB.

**Configuration:**
 - Swift Version: 6.2.3
 - OS: Debian/Raspberry Pi OS
 - OS Version: Trixie (armhf)

**Regression information:**
This seems to be present since the days of Swift 6.0 since presumably 32-bit architectures were not tested.

**Additional context**
I did a little bit of investigation, and looks like the issue is in these calculations in FileManager+Files.swift: https://github.com/swiftlang/swift-foundation/blob/8e3b765425ad0c2ea68506278c6bd9270cb34d85/Sources/FoundationEssentials/FileManager/FileManager%2BFiles.swift#L743-L749

A quick fix I found was to simply make the calculations UInt64, like this:

```swift
        let fsNumber = result.f_fsid
        let blockSize = UInt64(result.f_frsize)
        #endif
        var totalSizeBytes = UInt64(result.f_blocks) * blockSize
        var availSizeBytes = UInt64(result.f_bavail) * blockSize
```

Perhaps there are other solutions, but it seems like an oversight to ensure that these values are always at least 64-bit unsigned even on 32-bit systems.

jmschonfeld:
[...]

PR #1726, opened 2026-02-06
#1724: Convert to UInt64 in attributesOfFileSystem to fix 32-bit Linux platforms

xtremekforever:
Add the proposed fix from #1724 to ensure that filesystem size calculations do not crash on 32-bit platforms for Linux.

### Motivation:

Resolves the issue described in #1724.

### Modifications:

Cast `result.f_frsize`, `result.f_blocks`, and `result.f_bavail` to `UInt64` in `attributesOfFileSystem(forPath: String)` so that filesystem size calculations do not crash on 32-bit Linux platforms.

### Result:

When this method is called on an armv7/armhf platform in Linux, the call no longer crashes with a segfault or illegal instruction.

### Testing:

The entire `FoundationEssentialsTests` suite passes with this change when compiled with `6.3-snapshot-2026-01-29`.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Files.swift
@@ -741,10 +741,10 @@ extension _FileManagerImpl {
             let blockSize = UInt64(result.f_bsize)
             #else
             let fsNumber = result.f_fsid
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

PR #1729, opened 2026-02-09
Enable macOS checks in pull request workflow

shahmishal:
Enable macOS checks and update Xcode version.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,8 +17,8 @@ jobs:
     with:
       linux_swift_versions: '["nightly-main"]'
       windows_swift_versions: '["nightly-main"]'
-      enable_macos_checks: false
-      macos_xcode_versions: '["16.3"]'
+      enable_macos_checks: true
+      macos_xcode_versions: '["26.2"]'
       macos_versions: '["tahoe"]'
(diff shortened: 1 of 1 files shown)


parkera (approved):

.github/workflows/pull_request.yml
```diff
@@ -17,8 +17,8 @@ jobs:
     with:
       linux_swift_versions: '["nightly-main"]'
       windows_swift_versions: '["nightly-main"]'
-      enable_macos_checks: false
-      macos_xcode_versions: '["16.3"]'
+      enable_macos_checks: true
```
jmschonfeld, line 20: Does this now use the nightly toolchains instead of an Xcode-provided toolchain? Nightly toolchains are required to build swift-foundation

shahmishal: Going to move this PR to draft until we have nightly support.

jmschonfeld:
```
Run xcrun swift --version
Apple Swift version 6.2.3 (swiftlang-6.2.3.3.21 clang-1700.6.3.2)
```

It looks like the answer is no - @shahmishal I don't think we can adopt this yet (indeed, the build failed)

PR #1730, opened 2026-02-10
Remove a file that is no longer used per #1714

parkera:
In #1714 I renamed PathOrURL.swift, but forgot to delete the old file.


parkera:
@swift-ci test

PR #1731, opened 2026-02-10
Add DataBenchmarks suite

jmschonfeld:
Separates benchmarks for `Data` into their own suite and adds a variety of new benchmarks on all variants of `Data`

### Motivation:

When making changes to `Data`, it can impact the performance of different variants in different ways. `Data` benchmarks need to measure performance for each variant across a variety of common operations.

### Modifications:

- Moves existing `Data` benchmarks into their own suite
- Adds utilities to benchmark each variant of `Data`

### Result:

Benchmarks are provided on a wider set of operations and for all applicable variants.

### Testing:

N/A - benchmarks do not have tests, but I've validated locally that the benchmarks run


jmschonfeld:
@swift-ci please test

PR #1732, opened 2026-02-10
Remove unsafe span API

parkera:
I updated #1721 to use a "with" function style but didn't remove the unsafe Span API that it replaced. This fixes that.


parkera:
@swift-ci test

jmschonfeld (approved):

PR #1734, opened 2026-02-11
Remove incorrect logic in `Data.replaceSubrange(_:with:UnsafeBufferPointer<T>)`

jmschonfeld:
Removes incorrect logic in `Data.replaceSubrange(_:with:UnsafeBufferPointer<T>)` when provided with an empty buffer

### Motivation:

Replacing a subrange with an empty buffer is not a no-op, rather it should remove that subrange.

### Modifications:

Update `Data.replaceSubrange(_:with:UnsafeBufferPointer<T>)` to trivially forward to the funnel point which handles empty buffers appropriately rather than using separate (incorrect) logic.

### Result:

Replacing a subrange with an empty buffer now removes that subrange rather than doing nothing

### Testing:

Added an additional unit test which previously failed but now passes


jmschonfeld:
@swift-ci please test

PR #1735, opened 2026-02-11
[Proposal] Support for nil comparisons without Equatable conformance

MatthewTurk247:
This proposal includes two new overloads for both `PredicateExpressions.build_Equal(lhs:rhs:)` and `PredicateExpressions.build_NotEqual(lhs:rhs:)`. The goal is to facilitate checking whether a variable is `nil` without requiring `Equatable` conformance.


parkera:
@swift-ci test

parkera:
running the tests as a way to get the checks green even though this is just a proposal

jmschonfeld:
@swift-ci please test

PR #1736, opened 2026-02-11
When possible, clone files on Linux and FreeBSD.

### LINKED ISSUE #1727: `FileManager` does not support file cloning on Linux/FreeBSD
https://github.com/swiftlang/swift-foundation/blob/75d20dd712d99122ae60059dd17ec8c5e44916ed/Sources/FoundationEssentials/FileManager/FileOperations.swift#L941

`FileManager` can use `ioctl(dst, FICLONE, src)` on Linux and can use `copy_file_range(dst, nil, src, nil, size_t(SSIZE_MAX), COPY_FILE_RANGE_CLONE)` on FreeBSD (≥15.0) to clone a file instead of making a byte-by-byte copy, as we do on Darwin. If either call fails, we can then fall back to the existing operation.

grynspan:
Resolves #1727.

This change adds support for file cloning on Linux and FreeBSD where supported by the underlying OS and filesystem.

### Motivation:

Cloning is generally faster than block-based or bit-by-bit copying, so if a platform supports it we should try to leverage it to improve overall I/O performance.

### Modifications:

Added calls to the relevant API on Linux and FreeBSD. FreeBSD's API is new in 15.0, so there's a version check.

### Result:

Ideally, faster file copies on these platforms!

### Testing:

I don't have a great story for unit testing here as if either call fails, we'll fall back to block-based copying which, in the general case, will appear to the caller like a successful copy. Furthermore, our CI systems aren't configured with file systems that support file cloning anyway. I'm open to suggestions!

### DIFF
--- a/Sources/_FoundationCShims/include/filemanager_shims.h
+++ b/Sources/_FoundationCShims/include/filemanager_shims.h
@@ -13,7 +13,6 @@
 #ifndef CSHIMS_FILEMANAGER_H
 #define CSHIMS_FILEMANAGER_H
 
-#include "_CShimsTargetConditionals.h"
 #include "_CShimsMacros.h"
 #include "_CShimsTargetConditionals.h"
 


grynspan:
@swift-ci test

grynspan:
@michael-yuji Mind taking a look?

Sources/_FoundationCShims/include/platform_shims.h
```diff
@@ -109,4 +109,16 @@ static inline int32_t _platform_shims_O_NOFOLLOW(void) { return O_NOFOLLOW; }
...
 #endif
 
+#if TARGET_OS_BSD
+static inline unsigned int _platform_shims_COPY_FILE_RANGE_CLONE(void) {
+#if defined(COPY_FILE_RANGE_CLONE)
+  return COPY_FILE_RANGE_CLONE;
+#else
+  // Compiled against an older unistd.h, but presumably running on FreeBSD 15.0
+  // or newer. SEE: https://github.com/freebsd/freebsd-src/blob/main/sys/sys/unistd.h
+  return 0x00800000;
```
grynspan: We're building toolchains against FreeBSD 14.3's SDK right now, so the constant isn't defined yet.

jmschonfeld:
Re. testing - is it possible to easily check in a unit test whether or not cloning is supported and whether or not two files are clones? I suspect the latter might be feasible but the former is probably non trivial, and if none of our CI environments have file systems that actually support this then the point is probably moot.

Were you able to test locally on Linux and FreeBSD that it correctly creates a clone when running in an environment that supports it just to confirm we've tested it somewhere?

grynspan:
I've tested the equivalent code in Swift Testing on local VMs. I'm not certain how to go about doing that for Foundation but maybe it's straightforward to do it as a package dependency?

grynspan:
Verified in a VM that `FileManagerTests.c` clones files successfully:

```
% filefrag -e ../dir/foo
Filesystem type is: 58465342
File size of ../dir/foo is 10000 (3 blocks of 4096 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..       2:      57613..     57615:      3:             last,shared,eof
../dir/foo: 1 extent found
% filefrag -e ../dir2/foo
Filesystem type is: 58465342
File size of ../dir2/foo is 10000 (3 blocks of 4096 bytes)
 ext:     logical_offset:        physical_offset: length:   expected: flags:
   0:        0..       2:      57613..     57615:      3:             last,shared,eof
../dir2/foo: 1 extent found
```

Note **`shared`** on both files indicating the extents are shared (not necessarily between each other but, yes, between each other).

grynspan:
I can't figure out how to check reflinked extents on FreeBSD but the call succeeds at least. It's [documented](https://man.freebsd.org/cgi/man.cgi?copy_file_range) to fail with `EOPNOTSUPP` if cloning isn't supported, so… good enough signal.

grynspan:
@swift-ci test

Sources/_FoundationCShims/include/platform_shims.h
```diff
@@ -109,4 +109,22 @@ static inline int32_t _platform_shims_O_NOFOLLOW(void) { return O_NOFOLLOW; }
 
 #endif
 
+#if TARGET_OS_LINUX
+#include <linux/fs.h>
+
+static inline unsigned long _platform_shims_FICLONE(void) { return FICLONE; }
```
jmschonfeld: Could we move these two shims to `filemanager_shims.h`? I know `platform_shims.h` and `filemanager_shims.h` are somewhat similar in purpose but it'd be nice to keep these with the other shims for file manager functionality

grynspan: Sure. I wasn't sure where to put them.

grynspan: Done!

grynspan:
@swift-ci test

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -13,6 +13,7 @@
 #ifndef CSHIMS_FILEMANAGER_H
 #define CSHIMS_FILEMANAGER_H
 
+#include "_CShimsTargetConditionals.h"
```
jmschonfeld: This might have been an issue when you merged `main` down into this branch, but it looks like we now have two `#include`s of this header file - should we remove the one you added?

grynspan: Yeah it's because of the Android change that _just_ landed.

grynspan:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -906,8 +910,25 @@ enum _FileOperations {
...
+
         let total: Int = Int(fileInfo.st_size)
+
+        // Attempt to clone the file using platform-specific API. If this operation fails, don't throw
+        // an error and just fall back to chunked writes.
+        #if os(Linux)
+        if ioctl(dstfd, _filemanager_shims_FICLONE(), srcfd) != -1 {
+            return
+        }
+        #elseif os(FreeBSD)
```
michael-yuji, line 922: technically we can use `__FreeBSD_version` macro (in sys/param.h)
https://docs.freebsd.org/en/books/porters-handbook/versions/

michael-yuji: oh wait, this is swift

grynspan: It's also a runtime check on purpose :)

PR #1737, opened 2026-02-11
[6.3] Add shims for `getgrnam_r` and `getgrgid_r` for API 23

madsodgaard:
- **Explanation**:
Adds shims for the missing Android group ID functions to 6.3, such that swift-foundation compiles for API 23.
- **Scope**:
Should only effect Android API 23 and below.
- **Issues**:
It is related to this forums thread https://forums.swift.org/t/android-api-minimum-for-the-swift-sdk-for-android/82874
- **Original PRs**:
#1663 
- **Risk**:
The risk of this is low, it only affects Android and devices below API 23. The `#else` path in the shim, just delegates to the actual C functions for all other targets 
- **Testing**:
The original PR passed all CI, including Android.
I have tested these APIs on API 23 devices and they work as expected.

- **Reviewers**:
 @jmschonfeld approved the original PR

### DIFF
--- a/Sources/_FoundationCShims/include/filemanager_shims.h
+++ b/Sources/_FoundationCShims/include/filemanager_shims.h
@@ -14,6 +14,7 @@
 #define CSHIMS_FILEMANAGER_H
 
 #include "_CShimsMacros.h"
+#include "_CShimsTargetConditionals.h"
 
 #if __has_include(<sys/param.h>)
 #include <sys/param.h>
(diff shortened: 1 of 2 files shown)


finagolfin:
@swift-ci test

Sources/_FoundationCShims/include/filemanager_shims.h
```diff
@@ -46,4 +47,92 @@
...
+#include <errno.h>
+
+static inline int _filemanager_shims_getgrgid_r(gid_t gid, struct group *grp,
+                                                char *buf, size_t buflen, struct group **result) {
+    errno = 0;
+
+    // Call the non-reentrant version.
+    // On Android, this uses Thread Local Storage (TLS),
+    // so it is safe from race conditions with other threads.
+    struct group *p = getgrgid(gid);
```
grynspan, line 63: This is naïve of me, but if the function uses thread-local storage, why not just call it directly on Android?

PR #1739, opened 2026-02-12
Add back-deployment for `Date.now`

atifdev10:
Add back deployment for `Date.now`

### Motivation:

Making the `Date.now` property avaiable on older OS versions.

### Modifications:

Dropped the `@availability` attribute (lowered the OS version required) and added the `@backDeployed` attribute.

### Result:

`Date.now` property will now be avaiable on older OS versions.

### Testing:

How this will impact ABI stability.

### DIFF
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -164,7 +164,7 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
     public static let distantPast = Date(timeIntervalSinceReferenceDate: -63114076800.0)
 
     /// Returns a `Date` initialized to the current date and time.
-    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    @backDeployed(before: macOS 12, iOS 15, tvOS 15, watchOS 8)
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

shahmishal:
@swift-ci please test windows

atifdev10:
Is the Windows testing platform frozen? It's been a while and I didn't see any other activity in ci-external.

glessard:
@atifdev10 yes it was frozen for a while. It should work now (after we kickstart it.)

glessard:
@swift-ci please test

grynspan:
I'm not objecting to the change but does it need to go through Swift Evolution?

atifdev10:
Personally, I don't think it needs evolution since the change is so small (a single line), and it doesn't really add anything new. It just makes a convenience "initializer" available on older devices. But I can open a proposal if it's needed.

itingliu:
> I'm not objecting to the change but does it need to go through Swift Evolution?

I don't think we need to do that. It's already API

itingliu (approved):

PR #1746, opened 2026-02-13
FoundationEssentials: Force unwrap the result of a `realloc()` when reading Data

finagolfin:
The new safe C interop mode added in swiftlang/swift#87131 found this when cross-compiling to Android now.

On Android, the declaration is `void* _Nullable realloc(void* _Nullable __ptr, size_t __byte_count) __BIONIC_ALLOC_SIZE(2) __wur;` but Swift long and seemingly wrongly accepted this assignment, with it [starting to fail](https://ci.swift.org/job/swift-PR-swift-sdk-for-android/8/console) on [the Android CI lately](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/253/console).

@hnrklssn, could you confirm that your safe Interop pull corrected this?

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Reading.swift
+++ b/Sources/FoundationEssentials/Data/Data+Reading.swift
@@ -381,7 +381,7 @@ internal func readBytesFromFile(path inPath: borrowing some FileSystemRepresenta
             if length != chunkSize {
                 break
             }
-            ptr = realloc(ptr, totalRead + chunkSize)
+            ptr = realloc(ptr, totalRead + chunkSize)!
         }
         result = ReadBytesResult(bytes: ptr, length: totalRead, deallocator: .free)
         #else


finagolfin:
@swift-ci test

hnrklssn:
Yes, there's been a long time bug in clang where nullability annotations are dropped from return values on redeclarations. I.e. if you write this, clang will ignore the nullability annotation:
```
void * foo();
void * _Nullable foo();
```
The underlying bug is not fixed yet, as it was discovered quite recently during qualification of the safe interop feature. Safe interop interacts with this bug since it enables parsing of bounds attributes in clang, and clang is better at merging attributes for functions with bounds attributes. Clang infers a `__sized_by_or_null(2)` attribute on the return type from `__BIONIC_ALLOC_SIZE(2)`, triggering attribute merging between the declarations.

Sources/FoundationEssentials/Data/Data+Reading.swift
```diff
@@ -381,7 +381,7 @@ internal func readBytesFromFile(path inPath: borrowing some FileSystemRepresenta
             if length != chunkSize {
                 break
             }
-            ptr = realloc(ptr, totalRead + chunkSize)
+            ptr = realloc(ptr, totalRead + chunkSize)!
```
jmschonfeld, line 384: How does this impact other platforms where the nullability annotation is ignored? Force unwrapping a non-optional value is an error - is it possible that this produces an error on Linux? It looks like CI shows that it does not, but I do not fully understand why - was the clang importer behavior on Linux changed too?

hnrklssn: Clang has a builtin declaration for `realloc` that implicitly adds the `alloc_size` attribute, so at least for C this should be equivalent across platforms, I think. It *might* behave differently for C++ interop though. The safest workaround, to make sure this compiles both with old and new compilers, would be do something like this:
```
let tmp: UnsafeRawPointer? = realloc(...)
ptr = tmp!
```

jmschonfeld: Doing that here might be good to mitigate the risk of this failing to build in a scenario where it isn't perceived as nullable. @finagolfin it might help me to understand what scenarios exactly this is failing with - it looks like this project is still building with the latest nightly toolchain. Is the expectation that this project will fail to build with an upcoming nightly toolchain where this feature has been enabled by default? If so, I'm surprised the feature was able to be enabled given that CI on the Swift repo should fail if it has broken Foundation

finagolfin: > Force unwrapping a non-optional value is an error

You sure about this? I thought force unwraps fail at runtime if the optional turns out to be nil. Since a non-optional will always be non-nil, my understanding is that force-unwrapping this on platforms where it may not be nullable, like linux, is always fine, and [already used extensively in this repo](https://github.com/swiftlang/swift-foundation/blob/45268026c236513b3458936062c62a095185f226/Sources/FoundationEssentials/Data/Data%2BReading.swift#L372) and in Swift too.

> Is the expectation that this project will fail to build with an upcoming nightly toolchain where this feature has been enabled by default?

Yes, the official Android CI is building the trunk compiler from source with the latest trunk compiler commits and notifying us of this in the linked CI logs above.

> that CI on the Swift repo should fail if it has broken Foundation

It has, but only on Android, where the declaration is nullable, as shown. Older glibc on linux doesn't have nullability annotations, whereas we may not be building CI yet on newer distros that started adding them, swiftlang/swift#70647.

finagolfin:
@jrflat, I'd like to get this in and fix the Android CI before the weekend, let me know if you can review this one-liner.

Sources/FoundationEssentials/Data/Data+Reading.swift
hnrklssn: It depends on whether it is imported as an implicitly unwrapped optional (corresponds to unknown nullability), or a non-optional type (corresponds to nonnull). See example here: https://swiftfiddle.com/3zq3tyncrjgjnkyr5aiuszmefq

finagolfin: @hnrklssn, OK, I guess these C imports are always imported as implicitly optional then, unless otherwise explicitly specified as on Android?  In that case too, this compiles and works fine, and should with previous toolchains too, as I noted that this pattern is used extensively already in this very codebase, but a few like this one were simply missed before because of the type-checking bug you highlight.

jmschonfeld: So just to make sure my understanding is correct here:
- There was a compiler change made to `main` that causes the compiler to import nullability annotations on the API that it didn't before
- With older compilers, `realloc` is imported with an implicitly unwrapped optional return value which is why the optional handling is not required previously but optional handling is valid
- With newer compilers, `realloc` on non-Android the return value is still implicitly unwrapped on other platforms (so optional handling is ok but not required) but on Android it is now explicitly optional which requires optional handling
- We don't build for Android in CI on the Swift repo PRs which was why this change was able to be merged and break the build without being prevented by failing CI

Do I have all that right? I don't see any other patterns in this repo where we force unwrap `realloc` like this like you mentioned (but we do handle `malloc` similarly so it's in line with that approach) and Foundation builds in a large number of configurations so I want to make sure we fully understand the story here, but assuming the above is all correct this sounds good

finagolfin: > With newer compilers, realloc on non-Android the return value is still implicitly unwrapped on other platforms (so optional handling is ok but not required) but on Android it is now explicitly optional which requires optional handling

I think you got all these points correct, but I'd like to emphasize that Android labeled the return value as `_Nullable` for years now, so nothing changed there, but only in the Swift compiler with Henrik's pull. So if other libc's start adding these nullability annotations, they will need these force-unwraps also.

> I don't see any other patterns in this repo where we force unwrap realloc like this like you mentioned (but we do handle malloc similarly so it's in line with that approach)

I never said the pattern was to force unwrap `realloc` calls, rather I explicitly linked to a `malloc` call in the same method that force unwrapped. That force unwrapping of C functions is done throughout this codebase: I know because I added some of them as Android's libc added nullability annotations.

If you or one of the Foundation people would review, we can get this in.

jmschonfeld: Thank you for confirming my understanding.

I understand that nothing has changed in the Android SDK here, but the compiler change - one that I was not familiar with prior to this review - is relevant and has the potential to change behavior for not just Android but all platforms, as you mention. Foundation builds in a wide variety of different environments that are validated and tested in various different places. It's important that we have a full understanding of changes to make educated decisions about the best course of action for addressing build issues to make sure that fixing one build does not break others (as I'm sure you've experienced with other fixes breaking the Android build). The PR review process is an important part of that to ensure code owners can fully understand and stand by changes they approve. I understand your urgency here in wanting to unblock the Android build, but that urgency should not suppress the need for an appropriate code review (and when urgency is crucial, the typical course of action is to revert the causing change rather than to force through changes in clients). I've done a bit more testing to validate this works across all the build configs I'm aware of and with the confirmed understanding of the above this change looks good to me. I'm adding my approval now that we've done our due diligence to review this change and confirmed it won't break at desk workflows, CI, or routine builds.

I'm still a bit wary that this compiler change will be a source breaking change on some platforms for other clients like it was for Foundation. @hnrklssn have you thought about this at all? Should this instead be staged in via a language mode, or is it an explicit choice to accept a few source breaks in this release?

itingliu: I also want to ask why this isn't caught by the CI run in that compiler change. Probably a question for @hnrklssn; can you confirm this?

> We don't build for Android in CI on the Swift repo PRs which was why this change was able to be merged and break the build without being prevented by failing CI

If this is indeed the reason why, should we consider adding one?

hnrklssn: We have thought about it and decided it was fine since the previous behaviour was due to a clang bug, and being relatively rare since it requires *redeclaring* a function with different nullability on the return value than initially declared *while also* having bounds information. When enabling this internally at Apple it only affected a single project, fwiw.

hnrklssn: > > We don't build for Android in CI on the Swift repo PRs which was why this change was able to be merged and break the build without being prevented by failing CI

> If this is indeed the reason why, should we consider adding one?

This is correct. I was recently informed that it's possible to manually trigger Android CI on compiler PRs, which probably should have been done for a change of this impact. I agree that it would be nice if it were automatic though.

finagolfin: > I understand your urgency here in wanting to unblock the Android build, but that urgency should not suppress the need for an appropriate code review

Understood, but other than my note to @jrflat before the weekend, I have not expressed any urgency today, as the review resumed. I am surprised that this single force unwrap is being made a big deal of, considering this code base is full of such force unwraps of such C functions already, so maybe my comments came across that way.

> If this is indeed the reason why, should we consider adding one?

As Henrik notes, [a manual trigger for Android on the compiler repo was recently added](https://forums.swift.org/t/swift-android-workgroup-meeting-notes-february-11th-2026/84711), which [I applied for one of his other pulls](https://github.com/swiftlang/swift/pull/86866#issuecomment-3893147763). I have asked the CI team to [make that Android job mandatory to run but not pass for compiler pulls](https://github.com/swiftlang/swift-docker/pull/528#issue-3946916046), let's see what they say.

> We have thought about it and decided it was fine since the previous behaviour was due to a clang bug

I agree, I have seen only this single build failure so far on Android also.

finagolfin:
Thanks for the quick review, [the Android CI went green](https://ci.swift.org/job/oss-swift-package-swift-sdk-for-android/265/) again.

PR #1747, opened 2026-02-13
Remove ContiguousBytes.withBytes protocol requirement

jmschonfeld:
Removes the `withBytes` requirement on `ContiguousBytes` due to source breaking changes in clients with `MemberImportVisibility` enabled

### Motivation:

Clients like swift-crypto fail to build with this requirement. In swift-crypto's case, there is a type (`MD5Digest`) defined in a file that does not import `FoundationEssentials`. It conforms to a protocol (`DigestPrivate: Digest`) which refines `ContiguousBytes` and is defined in a separate file that does import `FoundationEssentials`. In that scenario, `MD5Digest` conforms to `ContiguousBytes` but does not see the default implementation member of the protocol requirement where the type is declared.

### Modifications:

Removes the protocol requirement until this issue is resolved (but leaves the default implementation and implementations on concrete types so it can still be used without overriding the requirement).

### Result:

swift-crypto builds successfully

### Testing:

Locally confirmed that swift crypto builds successfully

### DIFF
--- a/Sources/FoundationEssentials/Data/ContiguousBytes.swift
+++ b/Sources/FoundationEssentials/Data/ContiguousBytes.swift
@@ -57,12 +57,15 @@ public protocol ContiguousBytes: ~Escapable, ~Copyable {
     func withUnsafeBytes<R, E>(_ body: (UnsafeRawBufferPointer) throws(E) -> R) throws(E) -> R
 #endif
 
+    /*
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

glessard (approved):

jmschonfeld:
GitHub issue filed for `MemberImportVisibility` at https://github.com/swiftlang/swift/issues/87227

PR #1751, opened 2026-02-16
Add CharacterSet Benchmarks & Updates to BuiltInUnicodeScalarSet

chloe-yeo:
This PR adds benchmarks for `CharacterSet` and adds cases to `BuiltInUnicodeScalarSet` to account for types that were previously not accounted for. 

### Motivation:

This is to lay the groundwork for adding a Swift-native `CharacterSet` implementation in this repository. 

### Modifications:

Added benchmarks for `CharacterSet` and modified the `BuiltInUnicodeScalarSet` to account for additional built-in sets. 

### Result:

No behavior changes. 

### Testing:

Building and unit testing to confirm functions are unused/behavior is the same.

### DIFF
--- a/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
+++ b/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
@@ -412,17 +412,9 @@ internal struct BuiltInUnicodeScalarSet {
                 // Add or remove newline characters
                 for newlineChar in newlines {
                     if isInverted {
-                        #if FOUNDATION_FRAMEWORK
                         _CharacterSet.modifyBitmap(.remove, char: newlineChar, mutableSpan: &bitmapMutableSpan)
-                        #else
-                        modifyBitmap(.remove, char: newlineChar, mutableSpan: &bitmapMutableSpan)
-                        #endif
                     } else {
-                        #if FOUNDATION_FRAMEWORK
                         _CharacterSet.modifyBitmap(.add, char: newlineChar, mutableSpan: &bitmapMutableSpan)
-                        #else
-                        modifyBitmap(.add, char: newlineChar, mutableSpan: &bitmapMutableSpan)
-                        #endif
                     }
                 }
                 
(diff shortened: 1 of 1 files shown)


chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
```diff
@@ -11,79 +11,471 @@
...
+            }
+        } else if charset == .whitespace || charset == .whitespaceAndNewline || charset == .newline {
+            return plane == 0 || isInverted
+        } else if charset == .illegal {
+            if isInverted {
+                if plane < 3 || plane > 13 {
+                    return true
+                } else {
+                    return false
+                }
```
itingliu, line 182: This is correct but I believe we also have a followup to improve this to 

```suggestion
               return plane < 3 || plane > 13 
```
right?

I think this is less error prone given we've seen bugs caused by copy-paste errors when we manually unroll the boolean checks like this...

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
```diff
@@ -11,79 +11,471 @@
...
+    
+    // __CFCSetGetBitmap
+    internal func getBitmap(isInverted: Bool) -> Data {
+        let (result, bitmapData) = bitmap(forPlane: 0, isInverted: isInverted)
+        
+        switch result {
+        case .bitmapEmpty:
+            // For empty result, return the appropriate fill pattern
+            if isInverted {
+                return Data(repeating: 0x00, count: 65536 / 8)
```
itingliu, line 208: I also believe we have a followup task to return a static `Data` for these ones right?

itingliu (approved):

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
chloe-yeo: Yep added as follow-up!

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
chloe-yeo: Added as follow-up!

chloe-yeo:
@swift-ci please test

PR #1752, opened 2026-02-16
Fully qualify PredicateExpression function calls in #Predicate expansion

jmschonfeld:
Uses a newly-available syntax to fully qualify calls to `PredicateExpressions` functions with module names in `#Predicate`/`#Expression` expansions.

### Motivation:

With [SE-0491](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0491-module-selectors.md), you can now qualify function calls defined in extensions with a module name. Doing so limits the scope in which the compiler performs name lookup for these calls. This reduces the chance for ambiguity (in case any clients define overloads in this namespace that may become ambiguous - unlikely but possible) but, more importantly, I expect this should improve type checking performance since the module names are explicitly specified instead of inferred.

### Modifications:

- Every `PredicateExpressions` function call (or type reference) emitted by `#Predicate`/`#Expression` is now qualified with `Foundation::` or `FoundationEssentials::` accordingly
- Localization-specific APIs are now fully guarded by `FOUNDATION_FRAMEWORK` in the macro instead of just in the library (which ensures that emitting `FoundationInternationalization::` is not necessary at the moment)
- Unit tests are updated accordingly

### Result:

All existing unit tests in the library pass, and updated unit tests in the macro all pass

### Testing:

All macro unit tests have been updated accordingly, and existing library unit tests still verify the macro via the compiler.

Resolves rdar://165800976

### DIFF
--- a/Sources/FoundationMacros/PredicateMacro.swift
+++ b/Sources/FoundationMacros/PredicateMacro.swift
@@ -29,27 +29,36 @@ private let _knownSupportedFunctions: Set<FunctionStructure> = [
     FunctionStructure("starts", arguments: ["with"]),
     FunctionStructure("min", arguments: []),
     FunctionStructure("max", arguments: []),
-    FunctionStructure("localizedStandardContains", arguments: [.unlabeled]),
-    FunctionStructure("localizedCompare", arguments: [.unlabeled]),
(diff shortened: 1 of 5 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):
Makes sense to me!

PR #1753, opened 2026-02-17
Fix BenchmarkTimeZone build failure

itingliu:
There's a build failure because `locale` is declared twice.


itingliu:
@swift-ci please test

PR #1754, opened 2026-02-17
Decimal fix allocations on the critical path 5x faster + benchmarks

blindspotbounty:
Basic Decimal Benchmarks

### Motivation:

We noticed that initialization and division of Decimals may take too much time.
So decided to write benchmarks that would allow to fix it eventually.

image

Related to https://github.com/swiftlang/swift-foundation/issues/1755

### Modifications:

Pure adding Benchmark.

### Result:

Before:
image

After:
image

New benchmark is added.

Results:
```
Decimal init from Double(1):
Time (wall clock):
Before:
#[Mean    = 3117416448.000, StdDeviation   =        0.000]
After:
#[Mean    =  591790080.000, StdDeviation   =  1355819.310]

Malloc (total):
Before:
#[Mean    = 39993344.000, StdDeviation   =        0.000]
After:
#[Mean    =        0.000, StdDeviation   =        0.000]

Decimal init from Double.pi
Time (wall clock):
Before:
#[Mean    = 204183961.600, StdDeviation   =  4627806.902]
After:
#[Mean    =  80170188.800, StdDeviation   =   578501.384]

Decimal divide:

Time (wall clock):
Before:
#[Mean    = 4787798016.000, StdDeviation   =        0.000]
After:
#[Mean    = 1509949440.000, StdDeviation   =  1572864.000]

Decimal multiply:

Time (wall clock):
Before:
#[Mean    = 416887603.200, StdDeviation   =  1361130.814]
After:
#[Mean    = 173539328.000, StdDeviation   =   536435.279]

Decimal add:

Time (wall clock):
Before:
#[Mean    = 391992661.333, StdDeviation   =  9314313.206]
After:
#[Mean    =  80212131.840, StdDeviation   =   190915.716]
```

### Testing:

Actually benchmark tests + run unit tests.

### DIFF
--- a/Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
+++ b/Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
@@ -24,12 +24,22 @@ let benchmarks: @Sendable () -> Void = {
     Benchmark.defaultConfiguration.maxDuration = .seconds(2)
     Benchmark.defaultConfiguration.maxIterations = .count(100_000_000)
 
-    Benchmark("Decimal init from Double") { benchmark in
-        let double = Double.pi
+    Benchmark("Decimal init from Double(1)") { benchmark in
+        let value = Double(1)
 
         benchmark.startMeasurement()
         for _ in benchmark.scaledIterations {
-            blackHole(Decimal(double))
+            blackHole(Decimal.init(value))
+        }
+        benchmark.stopMeasurement()
+    }
+
+    Benchmark("Decimal init from Double.pi") { benchmark in
+        let value = Double.pi
+
+        benchmark.startMeasurement()
+        for _ in benchmark.scaledIterations {
+            blackHole(Decimal.init(value))
         }
         benchmark.stopMeasurement()
     }


blindspotbounty:
I decided to push changes for using InlineArray instead of plain Array for intermediate operations.

Now it looks like this, has no allocations and about 5x faster.

image

Let me know if it is better to commit separately instead...

itingliu:
@swift-ci please test

blindspotbounty:
@itingliu seems CI was never triggered due to conflicts appeared after merge of other PRs...

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -747,29 +747,106 @@ extension Decimal {
...
-            return [a0, a1, a2, a3]
-        case let (a0, a1, a2, a3, a4, 0, 0, 0):
-            return [a0, a1, a2, a3, a4]
-        case let (a0, a1, a2, a3, a4, a5, 0, 0):
-            return [a0, a1, a2, a3, a4, a5]
-        case let (a0, a1, a2, a3, a4, a5, a6, 0):
-            return [a0, a1, a2, a3, a4, a5, a6]
-        case let (a0, a1, a2, a3, a4, a5, a6, a7):
-            return [a0, a1, a2, a3, a4, a5, a6, a7]
+        @usableFromInline
```
iCharlesHu: All these are internal functions so I don't think `@useableFromInline` and `@inlinable` are needed.

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -747,29 +747,106 @@ extension Decimal {
 
 // MARK: - Integer Mathmatics
 extension Decimal {
-    typealias VariableLengthInteger = [UInt16]
+    /// Fixed-capacity inline buffer for intermediate integer arithmetic,
+    /// replacing `[UInt16]` to eliminate heap allocations on the critical path.
+    struct VariableLengthInteger: ExpressibleByArrayLiteral, Sendable {
+        // 2 x Mantissa is 16 + 1 for carry
+        @usableFromInline
+        static let maxCapacity = 17
```
iCharlesHu: You should be able to just use `storage.count` here.

Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
```diff
@@ -0,0 +1,42 @@
...
+#endif
+
+let benchmarks: @Sendable () -> Void = {
+    Benchmark.defaultConfiguration.scalingFactor = .mega
+    Benchmark.defaultConfiguration.maxDuration = .seconds(2)
+    Benchmark.defaultConfiguration.maxIterations = .count(100_000_000)
+
+    Benchmark("Decimal init from Double") { benchmark in
+        for i in benchmark.scaledIterations {
+            let double = Double(i)
```
iCharlesHu: This line should be moved outside of `scaledIterations` so the `Double` initializer doesn't affect the results as well.

Also since this is a benchmark, we want to use the same number in all iterations so the result is repeatable.

Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
```diff
@@ -0,0 +1,42 @@
...
+            let result = Decimal(double)
+            blackHole(result)
+        }
+    }
+
+    Benchmark("Decimal divide") { benchmark in
+        let divisor = Decimal(10_000)
+        for i in benchmark.scaledIterations {
+            let result = Decimal(i) / divisor
```
iCharlesHu: Same here. If the goal of this test is to test division then the Decimal initializer should also be moved out of the for loop like `divisor`. We should also just use the same number.

Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
blindspotbounty: Fixed!

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
blindspotbounty: Right

blindspotbounty:
I've addressed the review comments and also added couple of more benchmarks to cover other cases while here (add/multiply).
According to measurements, all operations benefit from the change.

itingliu:
@swift-ci please test

Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
```diff
@@ -0,0 +1,79 @@
...
+
+let benchmarks: @Sendable () -> Void = {
+    Benchmark.defaultConfiguration.scalingFactor = .mega
+    Benchmark.defaultConfiguration.maxDuration = .seconds(2)
+    Benchmark.defaultConfiguration.maxIterations = .count(100_000_000)
+
+    Benchmark("Decimal init from Double(1)") { benchmark in
+        let value = Double(1)
+
+        benchmark.startMeasurement()
```
itingliu, line 30: I didn't know we can do this! I've been sprinkling variables that I didn't want to include in the benchmark in the outer scope and polluting the whole scope. This is much nicer!

itingliu:
@swift-ci please test

Benchmarks/Benchmarks/Decimal/BenchmarkDecimal.swift
hassila: There are also some init/teardown closures that you might find useful depending on your needs - check the docs on swift package index!

hassila: https://swiftpackageindex.com/ordo-one/package-benchmark/1.29.11/documentation/benchmark/writingbenchmarks

 Search for and check out:

"specific setup closure can return an optional value of any type"

blindspotbounty:
Btw, I don't have write permissions to this repo, so someone will need to tap merge :)

PR #1757, opened 2026-02-18
Improve performance of AttributedString.runs getter

jmschonfeld:
Significantly improves the performance of creating an `AttributedString.Runs` from a concrete `AttributedString`

### Motivation:

Creating an `AttributedString.Runs` from an `AttributedString` should be fairly trivial. `AttributedString.Runs` mostly just wraps the `AttributedString.Guts`, but it also records some information about index bounds. Since `AttributedString.Runs` can be discontiguous, these index calculations can be complex. However, in the case of `AttributedString` itself, we know that it is both not discontiguous and it represents the entirety of the `AttributedString.Guts`. In these cases we can eliminate a lot of overhead. `AttributedString.Runs` is used both directly by clients as well as indirectly via APIs like equality and hashing, so this has some substantial improvements for clients.

### Modifications:

- Add a new `AttributedString.Runs` initializer that does not accept a `Range` to slice with
- Use this new initializer from `AttributedString.runs`, `AttributedString.==`, `AttributedString.hash(into:)`, and `AttributedString.description` (leaving the `DIscontiguousAttributedSubstring` and `AttribtuedSubstring` types as-is).

### Result:

`AttributedString.runs` is more than 900% faster than before. When content is actually equal, the cost of verifying content equality outweighs the improvements here in `AttributedString.==`, but when contents are unequal (and the overhead of setup is much more noticeable) this also has significant improvements to `AttributedString.==` (`equalityDifferingCharacters` improved by 670%)

### Testing:

These changes are covered by existing unit tests. I've added new benchmarks to cover `AttributedString.runs` specifically

### DIFF
--- a/Benchmarks/Benchmarks/AttributedString/BenchmarkAttributedString.swift
+++ b/Benchmarks/Benchmarks/AttributedString/BenchmarkAttributedString.swift
@@ -513,6 +513,12 @@ let benchmarks = {
     Benchmark("range(of:)") { benchmark in
         longString.range(of: "cccc")
     }
+
+    Benchmark("runs") {benchmark in
+        for _ in benchmark.scaledIterations {
+            blackHole(manyAttributesString.runs)
+        }
+    }
 }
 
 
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

PR #1760, opened 2026-02-18
Only attempt encoding once when updating NSURL ranges in place

jrflat:
Prevent a crash in the new `NSURL` implementation for certain inputs that require encoding.

### Motivation:

There is a flaw in the new encoding logic for `NSURL`, which updates its component ranges in place during encoding (a huge performance win). The problem is that we attempt encoding twice in some cases, first with `2 * span.count` allocated for the output, then with the maximum possible size (often `9 * span.count`) if we hit the limit on the first attempt. This means that ranges we updated in a first attempt are now invalid for the input string during a second attempt.

This can cause a crash in `span.extracting(range)` for certain inputs to `+[NSURL URLWithString:]` if the updated range is larger than the input span. Specifically, if one component is encoded to size `x` where `span.count < x <= 2 * span.count`, then a subsequent component pushes past the `2 * span.count` limit, causing the first attempt to fail, we will crash in the second attempt.

### Modifications:

As a low-risk fix, this PR skips the first `2 * span.count` attempt and only tries the maximum allocation if `updateRanges` is `true`. It also ensures that we don't try to allocate more than `Impl.maxStringLength`, since at that point we risk assigning truncated values to smaller `UInt16` ranges (which doesn't cause issues since we will just return `nil` later if the string is too big, but is good for defense-in-depth).

### Result:

No behavior change in swift-foundation. When the Swift implementation is enabled, `NSURL` no longer crashes for specific inputs that match the above requirements.

### Testing:

Unit tests for `NSURL` that reproduce the crash before the fix.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Parsing.swift
+++ b/Sources/FoundationEssentials/URL/URL_Parsing.swift
@@ -862,17 +862,32 @@ private func encode<T: _URLEncoding, Impl: _URLParseable>(
     // cases, since usually very few characters need encoding, or we're ASCII,
     // so start with 2 * buffer.count and re-allocate if needed.
 
+    // However, if updating ranges in place, only do so with the second
+    // maximum allocation, or else we could push past the 2 * span.count
+    // limit in the first allocation and be left using an invalid range.
+
     assert(!span.isEmpty)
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jrflat:
Yes, that's correct. Eventually `URL` should call into this path, too (and that will get some nice perf wins), but that requires a good amount of refactoring.

PR #1761, opened 2026-02-18
NSURL should reject invalid ASCII in host when IDNA-encoding

jrflat:
Reject invalid ASCII characters such as `"{"` that appear in the IDNA-encodable host of a `NSURL`.

### Motivation:

For an `NSURL` such as `"https://{example"`, `"{"` is considered an invalid character in the host according to RFC 3986. We do identify the host is invalid and needs encoding, and for `https` URLs, we use IDNA-encoding, which delegates to `uidna_nameToASCII`.

That particular function, however, only encodes non-ASCII Unicode characters, and otherwise leaves ASCII alone. This means that `uidna_nameToASCII` returns the original ASCII string, including the invalid `"{"`. The parser for `URLComponents`, which also IDNA-encodes, validates the host string once more after encoding to prevent this, but I missed that check in the Swift parser for `NSURL`.

### Modifications:

In the case where host validation fails and we're planning to IDNA-encode the host, we can first check if the host contains any invalid ASCII characters. If it does, we can short-circuit the parsing to return `nil` instead of continuing through all the encoding flow.

### Result:

No behavior change in swift-foundation. When the Swift implementation for `NSURL` is enabled, this maintains prior behavior of rejecting invalid ASCII in the host.

### Testing:

Unit tests for `NSURL` to confirm hosts with invalid ASCII are rejected whether or not they contain non-ASCII Unicode characters.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Validation.swift
+++ b/Sources/FoundationEssentials/URL/URL_Validation.swift
@@ -38,6 +38,19 @@ internal func validate<T: UnsignedInteger & FixedWidthInteger>(
     return true
 }
 
+internal func containsInvalidASCII<T: UnsignedInteger & FixedWidthInteger>(
+    host: borrowing Span<T>
+) -> Bool {
+    let allowedSet = URLComponentAllowedSet.host
+    for i in host.indices {
+        let codeUnit = host[i]
+        if codeUnit < 128 && !allowedSet.contains(codeUnit) {
+            return true
+        }
+    }
+    return false
+}
+
 private func isHexDigit<T: UnsignedInteger & FixedWidthInteger>(
     _ codeUnit: T
 ) -> Bool {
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL_Validation.swift
```diff
@@ -38,6 +38,19 @@ internal func validate<T: UnsignedInteger & FixedWidthInteger>(
     return true
 }
 
+internal func containsInvalidASCII<T: UnsignedInteger & FixedWidthInteger>(
```
jmschonfeld, line 41: We should make sure the compiler is always specializing this function as needed so we never hit a generic implementation (via `@inline(__always)` or `@specialize`) but that can be done as a followup if needed

jmschonfeld (approved):

Sources/FoundationEssentials/URL/URL_Validation.swift
jrflat: Good call, I didn't notice a perf hit but will follow-up to make sure this is being inlined (and likely will just add `@inline(__always)` to be sure).

PR #1762, opened 2026-02-18
Revert `ContiguousBytes` Additions

jmschonfeld:
Reverts prior changes to `ContiguousBytes` because it is causing failures in downstream projects

### Motivation:

Some downstream projects are failing to build due to breaking swift modules and ambiguity introduced by the new concrete implementations of `withUnsafeBytes`. While we investigate, we're reverting these changes.

### Modifications:

Reverts #1747 and #1565 

### Result:

`ContiguousBytes` is left the same as it was previously

### Testing:

Tests are reverted as well


jmschonfeld:
@swift-ci please test

PR #1768, opened 2026-02-20
Add Data.init(_: Data) fast path

jmschonfeld:
Adds a fast path when initializing `Data` with another `Data`

### Motivation:

When initializing a `Data` with another `Data`, we currently eagerly copy the bytes. However, if we know the source is a `Data` instead of copying we can just adopt the underlying storage (and defer the copy to the next write). This should significantly improve performance when initializing a `Data` from a `Data`

### Modifications:

- Add a specialized `Data.init(_: Data)` entry point that is always inlined for clients to use when the compiler knows statically that the provided value is a `Data`
- Add specialize fast paths to the existing `Data` initializers that call to the new entry point when we can quickly determine at runtime that a `Data` was provided (but not known at compile time)

Note: The `Data.init(_: Data)` initializer still performs an eager copy when the start index is not 0 or it was created via the `bytesNoCopy` initializer. This is to guarantee that:

- Indexing semantics on the created data behave the same (otherwise, the created `Data` would have previously had a `startIndex` of 0 but now it could be higher, which has broken clients in the past)
- The lifetime of the created data is the same (otherwise, we could be inheriting a limited lifetime where previously we created an owned lifetime which may result in a `Data` outliving its referenced allocation when previously it would not)

### Result:

Initializing a `Data` with a `Data` now no longer performs a copy (when possible)

### Testing:

Added a unit test to verify the copy occurs when possible and verify the copy still does not occur in riskier situations

### DIFF
--- a/Sources/FoundationEssentials/Data/Data.swift
+++ b/Sources/FoundationEssentials/Data/Data.swift
@@ -329,7 +329,7 @@ public struct Data : RandomAccessCollection, MutableCollection, RangeReplaceable
         // We still check for fast paths here on ABI stable platforms (withContiguousStorageIfAvailable and ContiguousBytes) because older SDKs did not contain always-inline fast paths, so some callers may still be using this ABI entrypoint with values that have fast paths
 
         if let data = _specialize(elements, for: Data.self) {
-            self.init(data)
+            self = Data(data) // If we already have a Data, call the specialized entrypoint
             return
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1775, opened 2026-02-23
Re-land ContiguousBytes changes without source-breaking issues

jmschonfeld:
Re-applies the changes reverted by #1762 and temporarily removes the pieces that caused source breaks in clients until we resolve the source breaking issues

### Motivation:

We reverted the changes to `ContiguousBytes` because it caused a number of source breaks in clients. This re-lands the work that did not cause source breaking changes, which will allow us to land the other changes piecewise as source breaking issues are resolved (either by compiler fixes or by changes in known clients).

### Modifications:

- Re-applies the commit reverted by #1762
- Removes the `~Escapable`/`~Copyable` refinement as this temporarily caused breakage in some builds - I will add this back shortly (along with the Span conformances) when this is resolved
- Removes the typed throws adoption in the concrete `withUnsafeBytes` implementations (this caused a source break in clients and we are investigating whether a compiler fix is needed or if clients need to update)
- Updates the comment w.r.t. the removal of the `ContiguousBytes.withBytes` protocol requirement (previously removed due to https://github.com/swiftlang/swift/issues/87227

### Result:

- `ContiguousBytes` now has a default implementation of `withBytes`
- All concrete types that conform to `ContiguousBytes` (and soon-to-conform types like the span types) now have implementations of `withBytes`
- Soon-to-conform types like span also now have concrete implementations of `withUnsafeBytes` (if not previously there)

### Testing:

Covered by existing unit tests. I will be performing separate testing to ensure this avoids the source breaking issues

### DIFF
--- a/Sources/FoundationEssentials/Data/ContiguousBytes.swift
+++ b/Sources/FoundationEssentials/Data/ContiguousBytes.swift
@@ -140,17 +140,9 @@ extension Data : ContiguousBytes { }
 
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 extension UnsafeRawBufferPointer : ContiguousBytes {
-    #if !hasFeature(Embedded)
-    // Historical ABI
-    @usableFromInline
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

PR #1776, opened 2026-02-24
Safer Span-based parsing for Swift CFURL file paths

jrflat:
Refactor the `NS/CFURL` file path initializer code to use `Span`, separate concerns, and better document preconditions.

### Motivation:

Passing a `CFString` of only null characters, e.g. `CFSTR("\0\0\0")`, with `isDirectory: true` to the file path initializer can cause a crash when the new Swift implementation is enabled. We can isolate the mutable buffer trailing-slash handling and other unsafe API usage while favoring `Span` where possible to prevent this and future unknown crashes.

### Modifications:

- Split trailing-slash handling into a standalone `finalPathLength` function that returns the adjusted length
- The adjusted length is used to create a `Span` representing the finalized file system representation to parse
- Extracted `memcmp`/`memcpy` operations to private extensions on `Span<UInt8>` and `UnsafeMutableBufferPointer<UInt8>` that take a `StaticString`, making correctness easier to verify
- `CFURL` Windows path style now rejects embedded null bytes, matching POSIX path style behavior
- Minor behavior fixes for deprecated HFS path style.

### Result:

No behavior change for swift-foundation. Prevents crashes involving strings with explicit null bytes for `NS/CFURL` when the new Swift implementation is enabled.

### Testing:

Unit testing for `NS/CFURL` file path initializers with various null-byte edge cases.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_C+File.swift
+++ b/Sources/FoundationEssentials/URL/URL_C+File.swift
@@ -91,22 +91,28 @@ extension NSURL {
     }
 }
 
-/// Call this function when `pathBuffer` has already been allocated with enough
-/// space for a trailing slash and we can modify the buffer directly instead of
-/// making a separate allocation.
+/// Updates `pathBuffer` for directory trailing-slash handling, returning the
+/// new path length. May append a trailing slash for directories or trim trailing
+/// slashes for non-directories.
 ///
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1778, opened 2026-02-24
Adopt Span in ICU.ResourceBundle

itingliu:
Adopt Span in `ICU.ResourceBundle`. This allows us to delay creating an Array until when we really need it and save intermediate allocations.

Also remove an intermediate struct `FixedOffsetTimeZoneRule`. It only wraps two variables, and removing it allows reducing one wrapper.

### Testing:

No particular tests were added because this change shouldn't change any behavior. The profiling result does show that the intermediate allocation calls are gone after the change.


itingliu:
@swift-ci please test

glessard:
@swift-ci please test

itingliu:
@swift-ci please test windows

itingliu:
@swift-ci please test windows

Issue #1779, opened 2026-02-24
Apparent typo in OpenStepPlist.swift


slavapestov:
**Describe the bug**
This code appears in OpenStepPlist.swift:
```
private func lineNumberStrings(_ pInfo: _ParseInfo) -> Int {
    var p = pInfo.utf16.startIndex
    var count = 1
    while p < pInfo.utf16.endIndex && p < pInfo.curr {
        if pInfo.utf16[p] == UInt16(ascii: "\r") {
            count += 1
    
            let nextIdx = pInfo.utf16.index(after: p)
            if nextIdx < pInfo.utf16.endIndex && nextIdx < pInfo.curr && pInfo.utf16[nextIdx] == UInt16("\n") {
                p = nextIdx
            }
        } else if pInfo.utf16[p] == UInt16(ascii: "\n") {
            count += 1
        }
        p = pInfo.utf16.index(after: p)
    } 
    return count
}
```
Notice how there are two calls to `UInt16(ascii:)` and one call to `UInt16(_:)`. This actually calls the `(String) -> UInt16?` overload which returns nil. I think this should be changed to `UInt16(ascii: "\n")`.

**To Reproduce**
I'm not familiar enough to come up with one because I found this while investigating a compiler bug, but it should either be trivial to either find one, or perhaps, prove that this check is dead either way.

PR #1780, opened 2026-02-24
Change `UInt16(_:)` to `UInt16(ascii:)` in OpenStepPlist.swift

slavapestov:
This was calling the failable (String) -> UInt16? init, which always returns nil when given "\n". This was not intended.

Found by inspection while testing some Swift type checker changes.

Thanks to @parkera for the test case!

Fixes https://github.com/swiftlang/swift-foundation/issues/1779.

### DIFF
--- a/Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
@@ -872,6 +872,23 @@ private struct PropertyListEncoderTests {
         }
     }
 
+    @Test func oldStylePlist_errors() {
+        let oldStyleWithErrors = "{ hello = there; }\r\nextra stuff at end"
+
+        let data = oldStyleWithErrors.data(using: .ascii)!
+
+        struct Thing : Decodable {
+            var hello: String
+        }
+        let plistDecoder = PropertyListDecoder()
+
+        #expect {
+            try plistDecoder.decode(Thing.self, from: data)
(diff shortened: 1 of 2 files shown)


xedin:
cc @jmschonfeld

slavapestov:
@swift-ci Please test

Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
```diff
@@ -872,6 +872,26 @@ private struct PropertyListEncoderTests {
...
+
+        let data = oldStyleWithErrors.data(using: .ascii)!
+
+        struct Thing : Decodable {
+            var hello: String
+        }
+        let plistDecoder = PropertyListDecoder()
+
+        do {
+            let _ = try plistDecoder.decode(Thing.self, from: data)
```
jmschonfeld: If you want, swift-testing has an `#expect` API for testing error types that we use in many of our other tests:

```swift
#expect {
    try plistDecoder.decode(Thing.self, from: data)
} throws: {
    #expect(String(describing: $0).contains("Junk after plist at line 2"))
}
```

slavapestov:
@swift-ci Please test

Tests/FoundationEssentialsTests/PropertyListEncoderTests.swift
slavapestov: Thanks!

Issue #1764, opened 2026-02-19
Data: `endIndex` documentation is incorrect


florianpircher:
The [documentation of `endIndex` of `Data`](https://github.com/swiftlang/swift-foundation/blob/559f997199baf4110db7d872adf57735e49bb3ef/Sources/FoundationEssentials/Data/Data.swift#L787) states that

> This is the "one-past-the-end" position, and will always be equal to the `count`.

However, `endIndex` is not always equal to `count`:

```swift
let d1 = Data([0, 1, 2, 3])
let d2 = d1[2 ..< 4]
assert(type(of: d2) == Data.self)
assert(d2.startIndex == 2)
assert(d2.endIndex == 4)
assert(d2.count == 2)
assert(d2.endIndex != d2.count)
```

jmschonfeld:
This is a good catch! Indeed, the `endIndex` is equal to the `count` only when it is not a slice. The documentation for [`Collection.endIndex`](https://developer.apple.com/documentation/swift/collection/endindex) states:

> The collection’s “past the end” position—that is, the position one greater than the last valid subscript argument.

Maybe we should just use similar wording here and remove the reference to `count`? @invalidname do you have thoughts?

invalidname:
> Maybe we should just use similar wording here and remove the reference to `count`? [@invalidname](https://github.com/invalidname) do you have thoughts?

@jmschonfeld  Agreed, this is wrong (and so is the [corresponding apple.com doc](https://developer.apple.com/documentation/foundation/data/endindex)… I'll file a radar for that). After all, if `endIndex` always equals `count`, then why would we need both symbols?

We can definitely lose the reference to `count`, and just using the comment from `Collection` seems like it'd be perfectly valid. TBH, a straightforward conformance to a protocol might not even need its own doc comment, but if there's already something here, probably better to just correct it.

PR #1781, opened 2026-02-25
Correct Data.endIndex doc comment

jmschonfeld:
Removes an incorrect statement from the documentation of `Data.endindex`

### Motivation:

Since `Data` is a self-slicing type, there is no guarantee that `endIndex == count`

Resolves https://github.com/swiftlang/swift-foundation/issues/1764

### Modifications:

Updates the documentation to use phrasing from [`Collection.endIndex`](https://developer.apple.com/documentation/swift/collection/endindex) instead of mentioning the `count`

### Result:

Documentation for `Data.endIndex` is corrected.

### Testing:

This is a documentation-only change so unit testing is not applicable.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data.swift
+++ b/Sources/FoundationEssentials/Data/Data.swift
@@ -948,7 +948,7 @@ public struct Data : RandomAccessCollection, MutableCollection, RangeReplaceable
 
     /// The end `Index` into the data.
     ///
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

invalidname (approved):
LGTM. Thanks for fixing this.

PR #1783, opened 2026-02-26
Fix incorrect Data slice capacity growth behavior

jmschonfeld:
Fixes problematic behavior when `Data` reserves capacity for growth when it is a slice.

### Motivation:

In `Data`s implementation, the various representation types store the slice range and the held `__DataStorage` always owns the entire allocation (which may be larger than the slice). In most cases, when `__DataStorage` needs to grow an allocation, it grows it with the context of the full `__DataStorage` (i.e. `__DataStorage` calls the grow function itself). However, there is a specific instance in which the representation types call the grow function directly: `reserveCapacity(_:)`. In this case, the capacity passed to `reserveCapacity(_:)` is relative to the size of the slice (not including any prefix or suffix not accessible via the slice). However, the representation types simply pass this value through to `__DataStorage` which uses it relative to the size of the full allocation.

This leads to unfortunate behavior: if `reserveCapacity(_:)` is called with a value large enough that it does not fit within the existing slice space + available suffix space but small enough that it is still smaller than the size of the full allocation (including the prefix space, slice space, and suffix space), the capacity does not grow and `reserveCapacity(_:)` is a no-op. We've gotten by so far because functions like `append` and `replaceSubrange` do end up growing with the right context, however it was inefficient as `reserveCapacity` may not pre-allocate as expected.

This becomes problematic with the new `append(addingCapacity:_:)` which relies on `reserveCapacity(_:)` to provide enough uninitialized space and then vends that uninitialized space directly to the user to write into. This can result in situations in which `append(addingCapacity:_:)` can provide an `OutputSpan` that extends beyond the end of an allocation.

### Modifications:

- Removed a quick exit in `Data._Representation.reserveCapacity(_:)` that was incorrect as it does not have enough information to do so here (`__DataStorage` already quick exits elsewhere)
- Fixed `Data.InlineSlice.reserveCapacity(_:)` and `Data.LargeSlice.reserveCapacity(_:)` to provide a `minimumCapacity` relative to the start of the allocation (not the slice) to `__DataStorage`
- Made the two above functions `@_alwaysEmitIntoClient` to ensure that for clients back deploying calls to APIs like `__DataStorage.withUninitializedBytes` that they are guaranteed to also have this fix back deployed
- Added `@usableFromInline` legacy entry points for existing clients using the symbol for these `reserveCapacity` functions that are now AEIC

### Result:

- Calling `reserveCapacity` on a slice now actually ensures there is the specified capacity available after the slice content
- Calling `append(addingCapacity:_:)` on a slice now guarantees the `OutputSpan` provided is valid

### Testing:

New unit tests are provided. Expectations previously failed for direct calls to `reserveCapacity` and calls to `append(addingCapacity:_:)` previously tripped a debug assertion which is now not triggered.

### DIFF
--- a/Tests/FoundationEssentialsTests/DataTests.swift
+++ b/Tests/FoundationEssentialsTests/DataTests.swift
@@ -2861,7 +2861,7 @@ extension DataTests {
         
         let data: Data = Data(bytesNoCopy: bytes.baseAddress!, count: bytes.count, deallocator: .none)
         let copy = data._bridgeToObjectiveC().copy() as! NSData
-        #expect(data.allocationForComparison == copy.allocationForComparison)
+        #expect(data.allocationForComparison != copy.allocationForComparison)
         bytes.deallocate()
     }
 }


Sources/FoundationEssentials/Data/Representations/DataStorage.swift
```diff
@@ -87,24 +87,24 @@ internal final class __DataStorage : @unchecked Sendable {
...
-    //   <─_offset─>
-    //   <─ slice.lowerBound ─>
-    //   <────── slice.upperBound ───────────>
+    //   ┌─ ─ ─ ─ ─ ─┬──────────┬──────────────┬─────────┬───────────────┐
+    //   ┊ prior     │ prefix   ┊ slice        ┊ suffix  ┊ uninitialized │
+    //   ┊ prefix    │          ┊              ┊         ┊ bytes         │
+    //   └─ ─ ─ ─ ─ ─┴──────────┴──────────────┴─────────┴───────────────┘
+    //    <─_offset─>
+    //    <─ slice.lowerBound ─>
+    //    <────── slice.upperBound ───────────>
```
jmschonfeld, line 107: While I was here, I pulled in @glessard 's recommended spacing fix for these comments 🙂

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Tests/FoundationEssentialsTests/DataTests.swift
```diff
@@ -887,6 +917,70 @@ private final class DataTests {
...
+        // Test behavior when the contents should get copied (non-unique)
+        do {
+            let original = Data(count: 80)
+            var slice = original.suffix(1)
+            #expect(slice.count == 1)
+            let startCapacity = capacity(slice)
+            slice.append(Data(repeating: 1, count: 25))
+            #expect(capacity(slice) != startCapacity, "Appending did not trigger a reallocation")
+            #expect(slice.count == 26)
+            _fixLifetime(original) // Ensure original lives beyond the mutations above
```
itingliu, line 930: For my information why did we need this?

jmschonfeld: This guarantees that `original` will be kept alive until at least this line. Otherwise, the compiler would be free to destroy `original` any time after line 924. The key here is that the `slice.append()` call should trigger a copy because the storage is not unique (it is held by both `slice` and `original` at that line). This section confirms that when a copy is made due to a uniqueness check, the correct capacity is allocated. The `do` block below confirms that the behavior still holds even when a copy is not made due to uniqueness (using a `do` block to prevent `original` from living during the `append` call)

jmschonfeld:
@swift-ci please test

PR #1784, opened 2026-02-27
Fast path for initializing Data with NSData

jmschonfeld:
Adds a fast path in `Data.init` for when the provided argument is an `NSData`

### Motivation:

Initializing a `Data` from an `NSData` currently goes down a byte-by-byte copy slow path. We can improve this a lot by dispatching to our bridging code to bridge lazily (when possible) or copy via a `memcpy` otherwise.

### Modifications:

Adds a fast path to `Data.init` when the argument is an `NSData` to dispatch to our bridging code

### Result:

Significant performance improvements when initializing a `Data` with an `NSData`

### Testing:

`NSData` tests are not applicable here


jmschonfeld:
@swift-ci please test

PR #1787, opened 2026-03-02
Update condition for large data test enablement

jmschonfeld:
Ensure that large data tests have sufficient memory to run

### Motivation:

Previously, we used `os` conditionals to determine if we had enough memory to run the large data tests (based on the memory limits of typical devices on those platforms). This validation is fragile and not sustainable when device specifications can change or when allocation sizes change. Instead, we now check whether or not the total amount of physical memory on the device is sufficient to run these tests. In practice, this will disable these tests on devices like watchOS/tvOS but remain enabled on macOS/Linux/Windows when the device has sufficient memory

### Modifications:

Conditionalize large data tests on memory size rather than platform

### Result:

Large data tests are now less likely to run when memory capacity is insufficient.

### Testing:

N/A - this updates unit tests themselves.

### DIFF
--- a/Tests/FoundationEssentialsTests/DataTests.swift
+++ b/Tests/FoundationEssentialsTests/DataTests.swift
@@ -2867,41 +2867,24 @@ extension DataTests {
 }
 #endif
 
-extension Trait where Self == ConditionTrait {
-    static var requiresLargeData: ConditionTrait {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

glessard (approved):

PR #1794, opened 2026-03-04
Follow up for ICUResourceTimeZone

itingliu:
- Remove the inclusive flag sine we're always passing false now
- Use binary search for last index
- Consolidate `isLeapYear` into GregorianCalendar's implementation
- Add tests for failures found in fuzzer. Validate the passed in date is within reasonable range.

170593919

### Testing:

Added tests for edge cases. The existing tests should also continue to pass

### DIFF
--- a/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarRecurrenceRuleTests.swift
@@ -18,6 +18,10 @@ import Testing
 @testable import Foundation
 #endif
 
+#if canImport(TestSupport)
+import TestSupport
+#endif
+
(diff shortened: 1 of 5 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

Sources/FoundationInternationalization/TimeZone/TimeZone_ICUResource.swift
jrflat, line 610: It looks like all callers of `_TimeZoneSingleDSTRule.dstTransition(after:)` also pass `inclusive: false` or use that as the default parameter. Would we want to remove that, too (maybe in a follow-up)?

itingliu: Good catch! Thanks yeah I'll do that as a follow up

PR #1797, opened 2026-03-06
Avoid materializing String in SwiftFTSEnt.name

jmschonfeld:
Uses `Span` instead of a materialized `String` when providing the name of an FTS entry

### Motivation:

`SwiftFTSEnt` is a Swift wrapper around a pointer to an `FTSENT`. It has a `name` property which previously materialized a `String` for the path. However, the only caller of this API just uses the name for a prefix check. Instead of creating a `String` and performing `String` comparison, we can instead vend a `Span<UInt8>` of the path's bytes to the caller. This should be much more performant because we no longer create a `String` on a per-subpath basis if not required, and the prefix comparison no longer performs string normalization.

### Modifications:

`SwiftFTSEnt` is now a lifetime-bound `Span<UInt8>` instead of a `String`. While I was here, I also moved the `fts_skip` call into `SwiftFTSEnt` to avoid escaping the `FTSENT` pointer out of the `SwiftFTSEnt` to decrease the chances of any lifetime issues (in the long term, `SwiftFTSEnt` should be `~Escapable` but we don't have the ability to produce a sequence of `~Escapable` values yet).

### Result:

`SwiftFTSEnt.name` no longer creates a `String`

### Testing:

Existing `FileManager` unit tests cover this code path

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations+Enumeration.swift
@@ -133,20 +133,27 @@ internal import _FoundationCShims
 struct _FTSSequence: Sequence {
     enum Element {
         struct SwiftFTSENT {
-            fileprivate let ptr: UnsafeMutablePointer<FTSENT>
-            
+            private let ptr: UnsafeMutablePointer<FTSENT>
+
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:

PR #1802, opened 2026-03-09
Allow square brackets in Swift CFURL path, query, and fragment

jrflat:
Allow `"["` and `"]"` to exist unencoded in the path, query, and fragment of `CFURL`.

### Motivation:

The old `CFURL` implementation allowed `"["` and `"]"` anywhere in the URL string due to a quirk of how IPv6 literals were implemented (despite both RFC 3986 and the old RFC 1808 only allowing those characters in the host component). We should match this behavior for the path, query, and fragment components for compatibility with both the old implementation and the WHATWG URL spec as a bonus.

### Modifications:

- Allow `"["` and `"]"` unencoded in the path, query, and fragment components
- Update the `addPercentEscapes` function used during string encoding to use the newer, faster `URLComponentAllowedSet` implementation, which I forgot to do earlier. The older `URLComponentAllowedMask` bitmasks used by `URLComponents` and the `CFURL/NSURL` file path initializers still behave the same as before and as expected for those APIs. We can investigate allowing `"["` and `"]"` for those separately.

### Result:

`CFURL` again allows `"["` and `"]"` in path, query, and fragment components and does not return `NULL`.

### Testing:

`CFURL`/`NSURL` unit test with the Swift implementation enabled.

### DIFF
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -1780,4 +1780,26 @@ private struct URLTests {
             #expect(!symlinkURL.hasDirectoryPath)
         }
     }
+
+    @Test func squareBracketsAllowedInPathQueryFragment() {
+        let bracketSpan = "[]".utf8.span
+
+        // Square brackets should be allowed in path, query, and fragment
+        let pathValid = validate(span: bracketSpan, component: .path)
+        let queryValid = validate(span: bracketSpan, component: .query)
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jrflat:
> It looks like this codepath is available everywhere, not just behind `#if FOUNDATION_FRAMEWORK`? If so would it be possible to add a test for this change even though there's no public Swift URL API that exercises this path?

Sure the internal function should be testable, I can add a test

jrflat:
@swift-ci please test

PR #1804, opened 2026-03-09
Add Cache for Built-In CharacterSet for Performance Improvements

chloe-yeo:
Introduce a cache for BMP of built-in CharacterSet options. 

### Motivation:

To improve performance. 

### Modifications:

Introduced a lazy cache for BMPs to create a fast path for frequently-called `getBitmap` method. 

### Result:

No change is expected. 

### Testing:

All pre-existing unit tests pass.

### DIFF
--- a/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
+++ b/Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
@@ -152,10 +152,8 @@ internal struct BuiltInUnicodeScalarSet {
             }
             let set = BuiltInUnicodeScalarSet(type: type)
             let (_, data) = set.bitmap(forPlane: 0, isInverted: isInverted)
-            let bitmap: Data
-            bitmap = data
-            cache[key] = bitmap
-            return bitmap
+            cache[key] = data
+            return data
         }
     }
     


chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
```diff
@@ -129,6 +133,36 @@ internal struct BuiltInUnicodeScalarSet {
...
+    private static func cachedBMP(for type: SetType, isInverted: Bool) -> Data {
+        let key = BMPCacheKey(type: type, isInverted: isInverted)
+        return bmpCache.withLock { cache in
+            if let cached = cache[key] {
+                return cached
+            }
+            let set = BuiltInUnicodeScalarSet(type: type)
+            let (result, data) = set.bitmap(forPlane: 0, isInverted: isInverted)
+            let bitmap: Data
+            switch result {
```
jmschonfeld: This looks like a change in behavior from the previous function below: previously `getBitmap(isInverted:)` always returned the `Data` provided by `bitmap(forPlane:isInverted:)`, but now we potentially cache/return `_CharacterSet.allZeros` or `allOnes` instead. Is that change intentional? If so, should we restructure `bitmap(forPlane:isInverted:)` such that it doesn't create/return a `Data` when the result is going to be discarded anyways?

chloe-yeo: It still needs to return a `Data` when it is called with the argument `bitmap(forPlane: [non-zero], isInverted:)` for cases where we have a `bitmapFilled` case, so we do need to keep it this way.

jmschonfeld: What is the `Data` value that is returned when `result` here is `.bitmapEmpty` or `.bitmapAll`? If it's not meaningful can we remove it for those cases even if it needs to remain for the `.bitmapFilled` case? Or if `data` is also just `_CharacterSet.allZeros`/`_CharacterSet.allOnes` when `result` is `.bitmapEmpty`/`.bitmapAll` respectively, is the `result` enum required at all?

I want to make sure that we're not creating a `Data` and then throwing it away in favor of `allZeros`/`allOnes` if we don't have to

chloe-yeo: As of now, the `Data` that we return defaults to `_CharacterSet.allZeros` when it's either `.bitmapEmpty` or `bitmapAll`. Would it be better then that I return `nil` for cases of `bitmapEmpty` and `bitmapAll`?

jmschonfeld: Should it just return `allZeros` for `bitmapEmpty` and `allOnes` for `bitmapAll`, and then you can remove the result value altogether and just return the `Data`?

chloe-yeo: There are some other helper methods in `CharacterSet` that does need to know if `bitmapEmpty` is the status because it will have to skip the execution in a loop. If I need to keep the status, what would be a good approach to this? Would the performance be equivalent if I use a `bitmap == _CharacterSet.allZeros` to replace `result == .bitmapEmpty`?

jmschonfeld: If the status is needed in other places but isn't really required here, maybe you could use the code that `getBitmap(isInverted:)` had before just so that's clearer? In other words:

```swift
let (_, bitmap) = set.bitmap(forPlane: 0, isInverted: isInverted)
cache[key] = bitmap
return bitmap
```

That way it's a bit simpler and it's clear that the returned `Data` isn't being created + thrown away but rather the `result` just isn't relevant information here?

chloe-yeo: That's a great suggestion, thanks!! :))

chloe-yeo: Fixed!

chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/String/BuiltInUnicodeScalarSet.swift
```diff
@@ -129,6 +133,32 @@ internal struct BuiltInUnicodeScalarSet {
...
+    private static func cachedBMP(for type: SetType, isInverted: Bool) -> Data {
+        let key = BMPCacheKey(type: type, isInverted: isInverted)
+        return bmpCache.withLock { cache in
+            if let cached = cache[key] {
+                return cached
+            }
+            let set = BuiltInUnicodeScalarSet(type: type)
+            let (_, data) = set.bitmap(forPlane: 0, isInverted: isInverted)
+            let bitmap: Data
+            bitmap = data
```
jmschonfeld: minor nit: no need to create a new variable `bitmap` here and assign `data` to it, you can just use `data` on the lines below, right?

chloe-yeo: Yes, thanks for catching this!

chloe-yeo:
@swift-ci please test

PR #1805, opened 2026-03-10
Avoid getattrlist TOCTOU issue for atomic file writing

jmschonfeld:
Use FSOPT_UNIQUE in getattrlist when writing files atomically

### Motivation:

`getattrlist` can provide stale information due to TOCTOU behavior exhibited by the kernel. To avoid this issue, we can use `FSOPT_UNIQUE` to have the kernel fail for us if there are multiple hard link locations rather than checking the returned data ourselves

### Modifications:

Adds `FSOPT_UNIQUE` to the `getattrlist` calls in data atomic writing and remove the `nlink` check of the result

### Result:

We no longer read potentially stale information from the `getattrlist` result

### Testing:

This is a small window for a race condition in the kernel that is not deterministically testable. It is also in `FOUNDATION_FRAMEWORK` code


jmschonfeld:
@swift-ci please test

PR #1808, opened 2026-03-10
NewCodable macros

tevelee:
Added Swift macros for complementing the NewCodable effort

### Motivation:

Several of the existing tests had commented-out `@JSONCodable` annotations sketching out what the macro-driven API should eventually look like. This PR picks up that thread and implements the macros.

### Modifications:

- Added a NewCodableMacros compiler plugin target with the following macros:
    - `@JSONEncodable`: generates the CodingFields enum and JSONEncodable conformance
    - `@JSONDecodable`: generates JSONDecodable conformance with string-based key matching via decodeEachKeyAndValue
    - `@JSONCodable`: combines both, delegating to the individual macro implementations
    - `@CodingKey("custom_name")`: overrides the JSON key for a property
    - `@CodableDefault(value)`: provides a fallback when a key is missing during decoding
    - `@CodableAlias("alt1", "alt2")`: accepts alternative key names during decoding
  - The annotation macros are peer macros that produce no output themselves, they're read as markers by the encoder/decoder macros at expansion time.
  - Replaced the hand-written conformances in CodableRevolutionTests with macro-annotated structs, cutting ~170 lines of manual boilerplate.

### Result:

Structs can now be annotated with `@JSONCodable` to get the same optimized encode/decode code without writing it by hand.

### Testing:

  - Added NewCodableMacrosTests with expansion tests covering basic structs, empty structs, optionals, `@CodingKey`, `@CodableDefault`, `@CodableAlias`, error diagnostics, and the combined `@JSONCodable` macro.
  - Updated existing CodableRevolutionTests to exercise the macros end-to-end against real encoding/decoding.

### DIFF
--- a/Sources/NewCodableMacros/JSONDecodableMacro.swift
+++ b/Sources/NewCodableMacros/JSONDecodableMacro.swift
@@ -275,12 +275,14 @@ extension JSONDecodableMacro: ExtensionMacro {
                 static func decode(from decoder: inout some JSONDecoderProtocol & ~Escapable) throws(CodingError.Decoding) -> \(typeName) {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         \(raw: varDeclarations)
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             \(raw: switchCases)
                             case .unknown: break
                             }
-                            return false
                         }
                         \(raw: guardAndReturn)
                     }
--- a/Tests/NewCodableMacrosTests/JSONCodableMacroTests.swift
+++ b/Tests/NewCodableMacrosTests/JSONCodableMacroTests.swift
@@ -71,13 +71,15 @@ struct JSONCodableMacroTests {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var name: String?
                         var age: Int?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .name: name = try valueDecoder.decode(String.self)
                             case .age: age = try valueDecoder.decode(Int.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let name else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'name'")
@@ -146,13 +148,15 @@ struct JSONCodableMacroTests {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var name: String?
                         var rating: Double?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .name: name = try valueDecoder.decode(String.self)
                             case .rating: rating = try valueDecoder.decode(Double.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let name else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'name'")
@@ -218,13 +222,15 @@ struct JSONCodableMacroTests {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var publishDate: String?
                         var title: String?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .publishDate: publishDate = try valueDecoder.decode(String.self)
                             case .title: title = try valueDecoder.decode(String.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let publishDate else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'date_published'")
@@ -322,13 +328,15 @@ struct JSONCodableMacroTests {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var name: String?
                         var locale: String?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .name: name = try valueDecoder.decode(String.self)
                             case .locale: locale = try valueDecoder.decode(String.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let name else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'name'")
@@ -388,12 +396,14 @@ struct JSONCodableMacroTests {
                 static func decode(from decoder: inout some JSONDecoderProtocol & ~Escapable) throws(CodingError.Decoding) -> User {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var userName: String?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .userName: userName = try valueDecoder.decode(String.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let userName else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'userName'")
@@ -453,12 +463,14 @@ struct JSONCodableMacroTests {
                 static func decode(from decoder: inout some JSONDecoderProtocol & ~Escapable) throws(CodingError.Decoding) -> User {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var userName: String?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
                             case .userName: userName = try valueDecoder.decode(String.self)
                             case .unknown: break
                             }
-                            return false
                         }
                         guard let userName else {
                             throw CodingError.dataCorrupted(debugDescription: "Missing required field 'user_name'")
--- a/Tests/NewCodableMacrosTests/JSONDecodableMacroTests.swift
+++ b/Tests/NewCodableMacrosTests/JSONDecodableMacroTests.swift
@@ -62,13 +62,15 @@ struct JSONDecodableMacroTests {
                     try decoder.decodeStruct { structDecoder throws(CodingError.Decoding) in
                         var name: String?
                         var age: Int?
-                        try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch try CodingFields.field(for: key) {
+                        var _codingField: CodingFields?
+                        try structDecoder.decodeEachField { fieldDecoder throws(CodingError.Decoding) in
+                            _codingField = try fieldDecoder.decode(CodingFields.self)
+                        } andValue: { valueDecoder throws(CodingError.Decoding) in
+                            switch _codingField! {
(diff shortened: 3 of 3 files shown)


codingkey:
这是来自QQ邮箱的假期自动回复邮件。你好，我最近正在休假中，无法亲自回复你的邮件。我将在假期结束后，尽快给你回复，或登录https://www.koudaiui.com，提交您得问题留言，或加微信gpbox7070

Sources/NewCodable/JSON/JSONParserDecoder.swift
```diff
@@ -1749,6 +1749,7 @@ extension JSONParserDecoder.ArrayDecoder: CommonArrayDecoder {
 }
 
 extension JSONParserDecoder.StructDecoder: CommonStructDecoder {
+    @_lifetime(&self)
```
kperryua: These should be `self: copy self`, no?

Sources/NewCodableMacros/JSONDecodableMacro.swift
```diff
@@ -0,0 +1,292 @@
...
+                let guardNames = requiredProperties.map { "let \($0.name)" }.joined(separator: ", ")
+                let args = properties.map { prop -> String in
+                    if let defaultExpr = prop.defaultExpr {
+                        return "\(prop.name): \(prop.name) ?? \(defaultExpr)"
+                    }
+                    return "\(prop.name): \(prop.name)"
+                }.joined(separator: ", ")
+                guardAndReturn = """
+                guard \(guardNames) else {
+                            throw CodingError.dataCorrupted(debugDescription: "Missing required fields")
```
kperryua: It doesn't have to happen right now, but eventually, I think it would be ideal to throw an error that includes the name of the missing field.

Sources/NewCodable/JSON/JSONParserDecoder.swift
kperryua: These should be `self: copy self`, yes?

Sources/NewCodable/Macros.swift
```diff
@@ -0,0 +1,31 @@
...
+public macro JSONCodable() = #externalMacro(module: "NewCodableMacros", type: "JSONCodableMacro")
+
+@attached(peer)
+public macro CodingKey(_ name: String) = #externalMacro(module: "NewCodableMacros", type: "CodingKeyMacro")
+
+@attached(peer)
+public macro CodableDefault<T>(_ value: T) = #externalMacro(module: "NewCodableMacros", type: "CodableDefaultMacro")
+
+@attached(peer)
+public macro CodableAlias(_ names: String...) = #externalMacro(module: "NewCodableMacros", type: "CodableAliasMacro")
```
kperryua: Can we include some documentation herein that clarifies that the form these macros, especially the per-field attribute macros, are provisional and subject to further evolution over time?

Sources/NewCodableMacros/JSONEncodableMacro.swift
```diff
@@ -0,0 +1,244 @@
...
+            var staticString: StaticString {
+                switch self {
+                \(raw: switchCases)
+                }
+            }
+
+            static func field(for key: UTF8Span) throws(CodingError.Decoding) -> CodingFields {
+                switch UTF8SpanComparator(key) {
+                \(raw: fieldForKeyCases)
+                default: throw CodingError.unknownKey(key)
```
kperryua: The behavior of unknown keys will need to be configurable. I think the default macro behavior needs to be permissive, which is more compatible with present-day `Decodable` and makes it easier to evolve a type's encoded representation over time. This is the same behavior as Serde. See https://serde.rs/container-attrs.html#deny_unknown_fields.

I've imagined opting in to `unknownKey` errors will probably look like on option to `@JSONDecodable`, e.g. `@JSONDecodable(.rejectUnknownFields)`.

When unknown keys are allowed, the pattern I've been using is to add an `unknown` case to the decoding field enum. The `switch` statement for this case just `break`s, which causes the StructDecoder to skip the field. There is somewhat of a wart with this design in that `var staticString` must handle the case. I've just been `fatalError`ing in that case, since the request is nonsensical.

Sources/NewCodableMacros/JSONEncodableMacro.swift
```diff
@@ -0,0 +1,244 @@
...
+        let fieldForKeyCases = properties.flatMap { prop -> [String] in
+            var cases = ["case \"\(prop.jsonKey)\": .\(prop.name)"]
+            for alias in prop.aliases {
+                cases.append("case \"\(alias)\": .\(prop.name)")
+            }
+            return cases
+        }.joined(separator: "\n            ")
+
+        let enumDecl: DeclSyntax = """
+        enum CodingFields: Int, JSONOptimizedCodingField {
```
kperryua: Oh, the `Int` `RawValue` type on this enum is a holdover from a previous design that I forgot to remove. I think it's fine to have no `RawRepresentable` conformance at all, but I'm open to thoughts on this.

Tests/NewCodableMacrosTests/JSONEncodableMacroTests.swift
```diff
@@ -0,0 +1,533 @@
...
+
+                    @_transparent
+                    var staticString: StaticString {
+                        switch self {
+                        case .name: "name"
+                        case .age: "age"
+                        }
+                    }
+
+                    static func field(for key: UTF8Span) throws(CodingError.Decoding) -> CodingFields {
```
kperryua: Technically, a type that implements only `JSONEncodable` does not need its fields to conform to `JSONOptimizedDecodingField`.

Sources/NewCodableMacros/JSONEncodableMacro.swift
kperryua: It's not necessary for this PR, but a "sooner than later" requirement will be examining the `CodingKey` overrides for any Unicode which could be represented with byte-for-byte non-equal (but canonically equivalent) UTF-8 normalizations. In those cases, the macro should elide conformance to `JSONOptimizedDecodingField` and use `DecodingFieldUTF8SpanCanonicalEquivalenceComparator` instead.

kperryua: This canonical-equivalence behavior might end up as an explicit option as well. Which behavior will end up as the default is unclear.

Sources/NewCodable/JSON/JSONParserDecoder.swift
tevelee: Right! I'm new to lifetime annotations and the mutating func threw me off guard.

Sources/NewCodableMacros/JSONDecodableMacro.swift
tevelee: It's easy enough to implement. Added separate guard statements (per field)

Sources/NewCodable/Macros.swift
tevelee: Added docs clarifying this

Sources/NewCodableMacros/JSONEncodableMacro.swift
tevelee: Great. I just removed Int RawRepresentable. We can add later if needed

Tests/NewCodableMacrosTests/JSONEncodableMacroTests.swift
tevelee: I separated conformances to JSONOptimizedEncodingField/JSONOptimizedDecodingField

Sources/NewCodableMacros/JSONEncodableMacro.swift
tevelee: I'll leave this up to you then. I only have a shallow understanding of this

Sources/NewCodableMacros/JSONEncodableMacro.swift
tevelee: I'll keep things permissive for now. And we can deal with specialized restrictive features later on

mateusrodriguesxyz:
Maybe `@DecodableAlias` would be more accurate than `@CodableAlias`.

kperryua:
> Maybe `@DecodableAlias` would be more accurate than `@CodableAlias`.

I support that.

I haven't checked whether it is or not, but we also need to make sure that this is compatible with `@CodingKey`.

Sources/NewCodableMacros/JSONDecodableMacro.swift
```diff
@@ -226,17 +226,21 @@ extension JSONDecodableMacro: ExtensionMacro {
...
                 guardAndReturn = "return \(typeName)(\(args))"
             } else {
-                let guardNames = requiredProperties.map { "let \($0.name)" }.joined(separator: ", ")
+                let requiredFieldGuards = requiredProperties.map {
+                    """
+                    guard let \($0.name) else {
+                                throw CodingError.dataCorrupted(debugDescription: "Missing required field '\($0.jsonKey)'")
+                            }
+                    """
+                }.joined(separator: "\n                        ")
```
kperryua, line 260: Forgive my macro ignorance—is there no functionality provided by the macro system for auto-indentation?

Tests/NewCodableMacrosTests/JSONDecodableMacroTests.swift
```diff
@@ -47,8 +47,11 @@ struct JSONDecodableMacroTests {
...
                             return false
                         }
-                        guard let name, let age else {
-                            throw CodingError.dataCorrupted(debugDescription: "Missing required fields")
+                        guard let name else {
+                            throw CodingError.dataCorrupted(debugDescription: "Missing required field 'name'")
+                        }
+                        guard let age else {
+                            throw CodingError.dataCorrupted(debugDescription: "Missing required field 'age'")
                         }
```
kperryua, line 80: Hopefully this doesn't have a huge impact on code size. I expect it should optimize down pretty reasonably. Something we can evaluate later though.

Tests/NewCodableTests/CodableRevolutionTests.swift
```diff
@@ -2139,6 +2144,21 @@ struct JSONDecodableMacroIntegrationTests {
...
+            Issue.record("Expected decoding to fail for a missing required field")
+        } catch let error {
+            guard case .dataCorrupted = error.kind else {
+                Issue.record("Unexpected CodingError.Decoding type: \(error)")
+                return
+            }
+            #expect(error.debugDescription.contains("Missing required field 'date_published'"))
+        }
+    }
+
```
kperryua, line 2159: Can we use `#expect(throws:)`?

Tests/NewCodableMacrosTests/JSONCodableMacroTests.swift
```diff
@@ -315,10 +323,10 @@ struct JSONCodableMacroTests {
                         var name: String?
                         var locale: String?
                         try structDecoder.decodeEachKeyAndValue { key, valueDecoder throws(CodingError.Decoding) in
-                            switch key {
-                            case "name": name = try valueDecoder.decode(String.self)
-                            case "locale": locale = try valueDecoder.decode(String.self)
-                            default: break
+                            switch try CodingFields.field(for: key) {
```
kperryua: Just noticed this with this commit. The macro should definitely generate code that targets `decodeEachField(andValue:)` instead of `decodeEachKeyAndValue()`. The latter always instantiates a String for the key, and then manually calling `CodingFields.field(for: key)` is somewhat wasteful and redundant. The use of `decodeEachKeyAndValue` in the tests was out of pure convenience (and of course, testing that particular function for correctness). With the macro generating the code now, it should bias the most efficient operation.

`decodeEachField` will let us match and create fields from UTF8Spans which will usually let us avoid the temporary String allocations.

tevelee:
Good point, updated it to `@DecodableAlias`. It composes with `@CodingKey` as expected: the `@CodingKey` value remains the canonical serialized key, and `@DecodableAlias` adds additional accepted decode keys.
I also added/updated macro tests for both `@JSONDecodable` and `@JSONCodable` covering `@CodingKey("user_name")` + `@DecodableAlias("username")`.

mateusrodriguesxyz:
Could `@DecodableAlias` support nesting? 

```json
{
  "latitude": 37.332,
  "longitude": -122.011,
  "additionalInfo": {
    "elevation": 30.5
  }
}
```

```swift
@JSONCodable
struct Coordinate {
    var latitude: Double
    var longitude: Double
    @DecodableAlias("additionalInfo.elevation") var elevation: Double
}
```

Sources/NewCodableMacros/JSONDecodableMacro.swift
tevelee: Not automatically in the way this expansion is built right now. Since these members/extensions are being assembled from multiline DeclSyntax string literals, the indentation is basically whatever whitespace I emit there. We could make that cleaner by moving more of it to structured SwiftSyntaxBuilder nodes or running a formatting pass, but I haven't found a macro API hook that auto-indents these raw string-based expansions for us.

Tests/NewCodableMacrosTests/JSONDecodableMacroTests.swift
tevelee: Agreed. I’d expect the separate guard let checks to optimize down reasonably, and I think the clearer per-field diagnostics are worth the slightly larger generated source for now. Once the macro shape settles, we can take a quick look at generated SIL/code size and revisit if it turns out to be measurable.

Tests/NewCodableTests/CodableRevolutionTests.swift
tevelee: Fixed

tevelee:
> Could @DecodableAlias support nesting?

Eventually yes, but I haven’t identified an obvious quick win to tackle it yet.

Tests/NewCodableMacrosTests/JSONCodableMacroTests.swift
tevelee: Updated. The macro now generates `decodeEachField { ... } andValue: { ... }` and decodes `CodingFields` directly from the field decoder

mateusrodriguesxyz:
> > Could @DecodableAlias support nesting?
> 
> Eventually yes, but I haven’t identified an obvious quick win to tackle it yet.

I've brought this up because it's an example from the [documentation](https://developer.apple.com/documentation/Foundation/encoding-and-decoding-custom-types) of having to resort to manually implementing a decoder `init` just to get rid of a simple nesting, which can be a pain.

kperryua:
Agreed that is should be possible. Are you aware of any prior art for this kind of annotation in other macro-based systems? I don't see one in Serde's suite of macros. It could make the implementation rather tricky, introducing more complex cases. The "easier" way out would be collecting parsed JSONPrimitives and then decoding all these values from them after the fact, but it's generally better to avoid that if possible.

One thing to keep in mind is that we should seek maximal compatibility with the `CommonCodable` protocols and forthcoming macro. Not every serialization format is able to provide a "primitive" type (it might not self-describing for instance), so the most macro functionality we can provide without relying on that, the better.

kperryua:
Looks good! The failures are the same ones we started with (that I still need to address). Merging. Thanks again!

kperryua (approved):

mateusrodriguesxyz:
> Are you aware of any prior art for this kind of annotation in other macro-based systems? I don't see one in Serde's suite of macros.

The closest official thing I have found, but not macro related, is pydantic `AliasPath`.

```python
from pydantic import BaseModel, Field, AliasPath

class Coordinate(BaseModel):
    latitude: float
    longitude: float
    elevation: float = Field(validation_alias=AliasPath("additionalInfo", "elevation"))


json_data = """
{
  "latitude": 37.332,
  "longitude": -122.011,
  "additionalInfo": {
    "elevation": 30.5
  }
}
"""

coordinate = Coordinate.model_validate_json(json_data)
```

Serde can custom deserialize a field:

```rust
#[derive(Debug, Deserialize)]
struct Coordinate {
    latitude: f64,
    longitude: f64,

    #[serde(rename = "additionalInfo", deserialize_with = "deserialize_elevation")]
    elevation: f64,
}

fn deserialize_elevation<'de, D>(deserializer: D) -> Result<f64, D::Error>
where
    D: Deserializer<'de>,
{
    let value: Value = Deserialize::deserialize(deserializer)?;
    Ok(value["elevation"].as_f64().unwrap())
}

fn main() {
    let json_data = r#"
    {
      "latitude": 37.332,
      "longitude": -122.011,
      "additionalInfo": {
        "elevation": 30.5
      }
    }
    "#;

    let coordinate: Coordinate = serde_json::from_str(json_data).unwrap();
}
```

There's also this crate: 
https://crates.io/crates/serde_flat_path
https://stackoverflow.com/a/75991416

```rust
#[flat_path]
#[derive(Deserialize)]
struct Foo {
    title: String,
    foo: bool,
    #[flat_path("config.bar.name")]
    bar_name: String,
}
```

PR #1809, opened 2026-03-10
Fixed bug in Currency formatter when `.grouping(.never)` is used with other modifiers

nicked:
Combining `.rounded()` and `.grouping(.never)` on `Currency` format styles (e.g. `IntegerFormatStyle.Currency`) produces broken output: the currency symbol and fraction digits are lost.

### Motivation:

Certain combinations of currency style modifiers loses all formatting:

```swift
  1234.formatted(.currency(code: "USD").rounded().grouping(.never))
  // Expected: "$1234.00"
  // Actual:   "1234"
```

### Modifications:

Add + " " after `group.skeleton` in CurrencyFormatStyleConfiguration.Collection.skeleton, matching the pattern used by all other tokens and by the non-currency equivalent.

### Result:

Currency formatting with `.grouping(.never)` combined with other modifiers works correctly.

### Testing:

Added unit tests.

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Number/NumberFormatStyleConfiguration.swift
+++ b/Sources/FoundationInternationalization/Formatting/Number/NumberFormatStyleConfiguration.swift
@@ -853,7 +853,7 @@ extension CurrencyFormatStyleConfiguration.Collection {
             s += roundingIncrement.skeleton + " "
         }
         if let group = group {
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test Windows

PR #1810, opened 2026-03-10
NSURL dataRepresentation inits should allow some invalid URLs the old parser accepted

jrflat:
Fix Swift `CFURL` parsing to accept some invalid URL strings that the old parser accepted for data representation inits.

### Motivation:

`CFURLCreateWithBytes` and `CFURLCreateAbsoluteURLWithBytes` generally accept any input and percent-encode it until it's valid. However, the new Swift code path for these functions rejected some invalid URL formats. This caused `+[NSURL (absolute)URLWithDataRepresentation:]` to return `nil` for data that previously succeeded, such as IP-literal hosts with invalid characters, unterminated `[` brackets, and relative paths with `:` in the first segment.

### Modifications:

- Gate the strict IP-literal content, unterminated `[`, and colon-in-relative-path rejection on `useModernParsing: true`, so the `CFURL` byte encoding path is as lenient as the old parser.
- When encoding an invalid IP-literal for `CFURL`, preserve the `[` and `]` bracket delimiters and only percent-encode the inner content.
- Use `span.extracting(_:)` more often for safer index validation.

### Result:

Data representation inits accept the same URL strings as the old parser.

### Testing:

Added unit tests for data representation `NSURL`s with invalid IP-literals, unterminated brackets, spaces in hosts, and relative paths with colons.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Validation.swift
+++ b/Sources/FoundationEssentials/URL/URL_Validation.swift
@@ -58,6 +58,10 @@ internal func validateIPLiteral<T: UnsignedInteger & FixedWidthInteger>(
     innerHost: borrowing Span<T>,
     useModernParsing: Bool
 ) -> Bool? {
+    guard useModernParsing else {
+        // For CFURL, allow arbitrary percent-encoding
+        return validate(span: innerHost, component: .hostIPvFuture)
+    }
     var i = 0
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test Windows Platform

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

PR #1812, opened 2026-03-11
Introduce Subprocess 1.0 Proposal

iCharlesHu:
This PR introduces subprocess-1.0 update proposal.

### DIFF
--- a/Proposals/XXXX-subprocess-1.0.md
+++ b/Proposals/XXXX-subprocess-1.0.md
@@ -10,11 +10,30 @@
 
 We introduced Subprocess with SF-0007 and shipped swift-subprocess as a public beta in Spring 2025. Since then, we've received a considerable amount of feedback from the community and updated the Subprocess API to address it. This proposal covers all API changes made since SF-0007 and proposes them for inclusion in the Subprocess 1.0 API.
 
-## Rename `ExecutionResult` and `CollectedResult`
+## Rename `CollectedResult` and Mark `ExecutionResult` Sendable
 
-We propose renaming `ExecutionResult` to `ExecutionOutcome`, to avoid confusion with Swift's `Result` type (since neither are a `Result`). `ExecutionOutcome` indicates that we are presenting the outcome (termination status) of the child process. We also want to add the missing `Sendable` conformances to `ExecutionOutcome`, since the wrapped `value` already has to be `Sendable`.
+We propose renaming `CollectedResult` to `ExecutionRecord`. `ExecutionRecord` indicates that we are presenting the recorded data of the child process, and it fits better with its counterpart `ExecutionResult`.
 
-Similarly, we propose renaming `CollectedResult` to `ExecutionRecord`. `ExecutionRecord` indicates that we are presenting the recorded data of the child process.
+We also want to add the missing `Sendable` conformances to `ExecutionResult`, since the wrapped `value` already has to be `Sendable`.
+
+
+## Adopt `NonisolatedNonsendingByDefault`
+
+We propose adopting the new Swift upcoming feature `NonisolatedNonsendingByDefault`, which allows us to remove all existing `#isolation` parameters:
+
+```diff
+--- a/Sources/Subprocess/API.swift
++++ b/Sources/Subprocess/API.swift
+@@ -132,7 +132,6 @@ public func run<
+     input: Input = .none,
+     output: Output = .discarded,
+     error: Error = .discarded,
+-    isolation: isolated (any Actor)? = #isolation,
+     body: (_ execution: Execution) async throws -> Result
+ ) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void {
+     let configuration = Configuration(
+```
+
 
 ## Remove `.standardOutput` and `.standardError` Properties from `Execution`
 
@@ -62,8 +81,8 @@ The new design eliminates this problem entirely by "promoting" the output `Async
 The full list of closure-based `run()` overloads is listed below:
 
 ```swift
-/// Run an executable with given parameters and a custom closure
-/// to manage the running subprocess’ lifetime.
+/// Runs an executable with given parameters and a custom closure
+/// to manage the running subprocess’s lifetime.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -73,9 +92,12 @@ The full list of closure-based `run()` overloads is listed below:
 ///   - input: The input to send to the executable.
 ///   - output: How to manage executable standard output.
 ///   - error: How to manage executable standard error.
-///   - isolation: the isolation context to run the body closure.
-///   - body: The custom execution body to manually control the running process
-/// - Returns: an `ExecutableResult` type containing the return value of the closure.
+///   - isolation: The isolation context to run the body closure.
+///   - body: A closure to manage the running process.
+///     All arguments passed to this closure are valid only for
+///     the duration of the closure's execution and must not be escaped.
+///     - execution: The running subprocess.
+/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
 public func run<
     Result,
     Input: InputProtocol,
@@ -90,12 +112,11 @@ public func run<
     input: Input = .none,
     output: Output = .discarded,
     error: Error = .discarded,
-    isolation: isolated (any Actor)? = #isolation,
-    body: ((Execution) async throws -> Result)
-) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void
+    body: (_ execution: Execution) async throws -> Result
+) async throws -> ExecutionResult<Result> where Error.OutputType == Void
 
-/// Run an executable with given parameters and a custom closure to manage the
-/// running subprocess' lifetime and stream its standard output.
+/// Runs an executable with given parameters and a custom closure to manage the
+/// running subprocess's lifetime and stream its standard output.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -105,13 +126,17 @@ public func run<
 ///   - input: The input to send to the executable.
 ///   - error: How to manage executable standard error.
 ///   - preferredBufferSize: The preferred size in bytes for the buffer used when reading
-///     from the subprocess's standard error stream. If `nil`, uses the system page size
+///     from the subprocess's standard output stream. If `nil`, uses the system page size
 ///     as the default buffer size. Larger buffer sizes may improve performance for
 ///     subprocesses that produce large amounts of output, while smaller buffer sizes
 ///     may reduce memory usage and improve responsiveness for interactive applications.
-///   - isolation: the isolation context to run the body closure.
-///   - body: The custom execution body to manually control the running process.
-/// - Returns: an `ExecutableResult` type containing the return value of the closure.
+///   - isolation: The isolation context to run the body closure.
+///   - body: A closure to manage the running process.
+///     All arguments passed to this closure are valid only for
+///     the duration of the closure's execution and must not be escaped.
+///     - execution: The running subprocess.
+///     - outputSequence: The standard output as an asynchronous sequence of buffers.
+/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
 public func run<Result, Input: InputProtocol, Error: ErrorOutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -121,12 +146,14 @@ public func run<Result, Input: InputProtocol, Error: ErrorOutputProtocol>(
     input: Input = .none,
     error: Error = .discarded,
     preferredBufferSize: Int? = nil,
-    isolation: isolated (any Actor)? = #isolation,
-    body: ((_ execution: Execution, _ outputSequence: AsyncBufferSequence) async throws -> Result)
-) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void
+    body: (
+        _ execution: Execution,
+        _ outputSequence: AsyncBufferSequence
+    ) async throws -> Result
+) async throws -> ExecutionResult<Result> where Error.OutputType == Void
 
-/// Run an executable with given parameters and a custom closure to manage the
-/// running subprocess' lifetime and stream its standard error.
+/// Runs an executable with given parameters and a custom closure to manage the
+/// running subprocess's lifetime and stream its standard error.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -140,9 +167,13 @@ public func run<Result, Input: InputProtocol, Error: ErrorOutputProtocol>(
 ///     as the default buffer size. Larger buffer sizes may improve performance for
 ///     subprocesses that produce large amounts of output, while smaller buffer sizes
 ///     may reduce memory usage and improve responsiveness for interactive applications.
-///   - isolation: the isolation context to run the body closure.
-///   - body: The custom execution body to manually control the running process
-/// - Returns: an `ExecutableResult` type containing the return value of the closure.
+///   - isolation: The isolation context to run the body closure.
+///   - body: A closure to manage the running process.
+///     All arguments passed to this closure are valid only for
+///     the duration of the closure's execution and must not be escaped.
+///     - execution: The running subprocess.
+///     - errorSequence: The standard error as an asynchronous sequence of buffers.
+/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
 public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -152,12 +183,14 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
     input: Input = .none,
     output: Output,
     preferredBufferSize: Int? = nil,
-    isolation: isolated (any Actor)? = #isolation,
-    body: ((_ execution: Execution, _ errorSequence: AsyncBufferSequence) async throws -> Result)
-) async throws -> ExecutionOutcome<Result> where Output.OutputType == Void
+    body: (
+        _ execution: Execution,
+        _ errorSequence: AsyncBufferSequence
+    ) async throws -> Result
+) async throws -> ExecutionResult<Result> where Output.OutputType == Void
 
-/// Run an executable with given parameters and a custom closure to manage the
-/// running subprocess' lifetime, write to its standard input, and stream its standard output.
+/// Runs an executable with given parameters and a custom closure to manage the
+/// running subprocess's lifetime, write to its standard input, and stream its standard output.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
@@ -170,9 +203,14 @@ public func run<Result, Input: InputProtocol, Output: OutputProtocol>(
 ///     as the default buffer size. Larger buffer sizes may improve performance for
 ///     subprocesses that produce large amounts of output, while smaller buffer sizes
 ///     may reduce memory usage and improve responsiveness for interactive applications.
-///   - isolation: the isolation context to run the body closure.
-///   - body: The custom execution body to manually control the running process
-/// - Returns: An `ExecutableResult` type containing the return value of the closure.
+///   - isolation: The isolation context to run the body closure.
+///   - body: A closure to manage the running process.
+///     All arguments passed to this closure are valid only for
+///     the duration of the closure's execution and must not be escaped.
+///     - execution: The running subprocess.
+///     - inputWriter: A writer for the subprocess's standard input.
+///     - outputSequence: The standard output as an asynchronous sequence of buffers.
+/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
 public func run<Result, Error: ErrorOutputProtocol>(
     _ executable: Executable,
     arguments: Arguments = [],
@@ -181,18 +219,15 @@ public func run<Result, Error: ErrorOutputProtocol>(
     platformOptions: PlatformOptions = PlatformOptions(),
     error: Error = .discarded,
     preferredBufferSize: Int? = nil,
-    isolation: isolated (any Actor)? = #isolation,
     body: (
-        (
-            _ execution: Execution,
-            _ inputWriter: StandardInputWriter,
-            _ outputSequence: AsyncBufferSequence
-        ) async throws -> Result
-    )
-) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void
+        _ execution: Execution,
+        _ inputWriter: StandardInputWriter,
+        _ outputSequence: AsyncBufferSequence
+    ) async throws -> Result
+) async throws -> ExecutionResult<Result> where Error.OutputType == Void
 
-/// Run an executable with given parameters and a custom closure to manage the
-/// running subprocess' lifetime, write to its standard input, and stream its standard error.
+/// Runs an executable with given parameters and a custom closure to manage the
+/// running subprocess's lifetime, write to its standard input, and stream its standard error.
 /// - Parameters:
 ///   - executable: The executable to run.
 ///   - arguments: The arguments to pass to the executable.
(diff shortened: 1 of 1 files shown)


iCharlesHu:
@swift-ci please test

jakepetroules:
Will this be posted on the forums for discussion soon?

This doc looks great to me, although a couple things I'd like to see incorporated:

- I think we should have some discussion on what to do for https://github.com/swiftlang/swift-subprocess/issues/114 (maybe get @compnerd's thoughts), as this seems pretty foundational and hard to change later -- I think it's critical to make a decision on what the final API shape should be before 1.0 goes out.
- https://github.com/swiftlang/swift-subprocess/issues/200 is a small but notable thing that should be in 1.0.
- I've changed my opinion a bit on runDetached... I actually had a use case recently to spawn a daemon process meant to outlive my own process's execution... so maybe we could bring it back, but it specifically returns Void so that callers have no way to control it and therefore not be subject to TOCTOU issues like the previous approach which returned a PID. Not the end of the world if this missed 1.0 as it seems addressable later.

iCharlesHu:
@jakepetroules Thanks for the reminder. I agree number 1 and 2 should be pulled in, and I'm looking at them now. However, I don't think we should include `runDetached` in 1.0. The more I think about it, the more I realized that `runDetached()` as proposed in the original design doesn't provide more functionalities than `NSProcess` itself, since we can't run async file IO or async process exit monitoring. In these situations, I think it's better to just use `NSProcess`.

iCharlesHu:
Addressed @jakepetroules' comments

> Will this be posted on the forums for discussion soon?

Yes that's the plan 😆

iCharlesHu:
@swift-ci please test

jakepetroules:
> However, I don't think we should include runDetached in 1.0. The more I think about it, the more I realized that runDetached() as proposed in the original design doesn't provide more functionalities than NSProcess itself, since we can't run async file IO or async process exit monitoring. In these situations, I think it's better to just use NSProcess.

Yes, agree it's not critical for 1.0. I do think Subprocess should eventually provide a solution here though -- ideally in the long term it could fully replace NSProcess to the point where the latter could be deprecated.

iCharlesHu:
@swift-ci please test

iCharlesHu:
@swift-ci please test

iCharlesHu:
@swift-ci please test

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    ///
+    /// The file descriptor isn't closed afterwards.
+    public static var standardError: Self
+}
+```
+
+
+## Redesign `TerminationStatus` on Windows
+
+The original `TerminationStatus` included two cases: `.exited()` and `.unhandledException()`. While these two cases make sense on Unix systems — where `wait(2)` returns a packed bitfield that distinguishes normal exits from unhandled signals — they do not translate well to Windows. Windows's `GetExitCodeProcess()` returns a single `DWORD` value, making it impossible to reliably distinguish between a normal exit code and an unhandled exception code.
```
jakepetroules, line 862: wording nit: I'm not sure "unhandled exception" is a specific term of art in the context of Windows process termination. I thought that was .NET/CLR specific. @compnerd?

Also, we might be able to somewhat reliably distinguish normal vs abnormal termination using job objects (see JOB_OBJECT_MSG_ABNORMAL_EXIT_PROCESS). Not sure if Subprocess should provide anything there, but could be a post 1.0 thing. I don't think in any case it should involve attempting to reinterpret the numeric exit code.

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+* Status: **Post Review Update**
+* Implementation: https://github.com/swiftlang/swift-subprocess
+
+## Introduction
+
+We introduced Subprocess with SF-0007 and shipped swift-subprocess as a public beta in Spring 2025. Since then, we've received a considerable amount of feedback from the community and updated the Subprocess API to address it. This proposal covers all API changes made since SF-0007 and proposes them for inclusion in the Subprocess 1.0 API.
+
+## Rename `ExecutionResult` and `CollectedResult`
+
+We propose renaming `ExecutionResult` to `ExecutionOutcome`, to avoid confusion with Swift's `Result` type (since neither are a `Result`). `ExecutionOutcome` indicates that we are presenting the outcome (termination status) of the child process. We also want add the missing `Sendable` conformances to `ExecutionOutcome`, since the wrapped `value` already has to be `Sendable`.
```
itingliu:
```suggestion
We propose renaming `ExecutionResult` to `ExecutionOutcome`, to avoid confusion with Swift's `Result` type (since neither are a `Result`). `ExecutionOutcome` indicates that we are presenting the outcome (termination status) of the child process. We also want to add the missing `Sendable` conformances to `ExecutionOutcome`, since the wrapped `value` already has to be `Sendable`.
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+We propose moving `.standardOutput` and `.standardError` properties from `Execution` to the `run()` closure parameter alongside `Execution`:
+
+```swift
+// Before
+public func run<Result>(
+    ...
+    body: (Execution) async throws -> Result
+) async throws -> ExecutionOutcome<Result> { ... }
+
+_ = try await run(...) { execution in
```
itingliu: The fact that you're writing `_ = try await run(...)` in every example in this section makes me wonder if this function should be `@discardableResult`

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+        ) async throws -> Result
+    )
+) async throws -> ExecutionOutcome<Result>
+
+/// Run an executable with given `Configuration` and a custom closure
+/// to manage the running subprocess' lifetime.
+/// - Parameters:
+///   - configuration: The configuration to run.
+///   - input: The input to send to the executable.
+///   - output: How to manager executable standard output.
```
itingliu:
```suggestion
///   - output: How to manage executable standard output.
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+///   - error: How to manager executable standard error.
```
itingliu:
```suggestion
///   - error: How to manage executable standard error.
```

Proposals/XXXX-subprocess-1.0.md
itingliu: And in other `run` functions as well

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    for await item in standardOutput { ... }
+}
+```
+
+This change eliminates the need for `Atomic` and `AtomicBox` within `Execution`, and more importantly, makes these two variables semantically closer to the correct mental model of how output streaming works.
+
+In the original design, since `.standardOutput` and `.standardError` are properties on `Execution`, it creates the illusion that you can repeatedly call these properties and create different output streams:
+
+```swift
+_ = try await run(...) { execution in
```
itingliu: Nit: I find it useful to have a comment when it comes to incorrect code

```suggestion
// Example of undefined behavior 
_ = try await run(...) { execution in
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    for try await item1 in execution.standardOutput { ... }
+    for try await item2 in execution.standardOutput { ... }
+}
+```
+
+However, this is not the case. Once you create an `AsyncBufferSequence` by calling `execution.standardOutput`, the returned sequence effectively "owns" the underlying OS pipe used to read data from. This means calling `execution.standardOutput` multiple times is undefined behavior, and we had to use an internal `Atomic` value to guard against it.
+
+The new design eliminates this problem entirely by "promoting" the output `AsyncBufferSequence`s to be sibling parameters of `Execution`. Since by definition you can't "get" a function parameter multiple times, this eliminates the possibility of the aforementioned undefined behavior. This change also simplified `Execution`'s design by making it non-generic.
```
itingliu: > Since by definition you can't "get" a function parameter multiple times,

I don't quite understand what this means. By "function parameter" did you mean `standardOutput`? Why is that a function parameter? Isn't it a concrete struct `AsyncBufferSequence`?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+///   - workingDirectory: The working directory in which to run the executable.
+///   - platformOptions: The platform-specific options to use when running the executable.
+///   - input: The input to send to the executable.
+///   - output: How to manage executable standard output.
+///   - error: How to manage executable standard error.
+///   - isolation: the isolation context to run the body closure.
+///   - body: The custom execution body to manually control the running process
+/// - Returns: an `ExecutableResult` type containing the return value of the closure.
+public func run<
+    Result,
```
itingliu, line 102: Can this `Result` be `~Copyable`?

@jmschonfeld mentioned that this would help if clients ever want to return a non-copyable value from the closure

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+public func run<Result>(
+    _ configuration: Configuration,
+    preferredBufferSize: Int? = nil,
+    isolation: isolated (any Actor)? = #isolation,
+    body: (
+        (
+            Execution,
+            StandardInputWriter,
+            AsyncBufferSequence,
+            AsyncBufferSequence
```
itingliu: Would it be possible/make sense to assign argument labels to these so we can differentiate between the two `AsyncBufferSequence`? (I believe one is error and the other is output, right?) Something like


```suggestion
            Execution,
            StandardInputWriter,
            _ output: AsyncBufferSequence,
            _ error: AsyncBufferSequence
```

or

```suggestion
            _ execution: Execution,
            _ input: StandardInputWriter,
            _ output: AsyncBufferSequence,
            _ error: AsyncBufferSequence
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+        #if !os(Windows)
+        /// Raw mode — all bytes passed through unmodified. Applies `cfmakeraw()` to the PTY.
+        public static var raw: Self
+
+        /// Custom termios configuration
+        public static func custom(_ info: termios) -> Self
+        #endif
+    }
+
+    /// The initial termianl window size to set to.
```
itingliu:
```suggestion
    /// The initial terminal window size to set to.
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+```swift
+/// An asynchronous sequence of buffers used to stream output from subprocess.
+public struct AsyncBufferSequence: AsyncSequence, Sendable {
+    /// The failure type for the asynchronous sequence.
+    public typealias Failure = any Swift.Error
+    /// The element type for the asynchronous sequence.
+    public typealias Element = Buffer
+
+    /// Iterator for `AsyncBufferSequence`.
+    @_nonSendable
```
itingliu: Why is this _nonSendable?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    ///   line editing, echo, and signal generation. Use when spawning
+    ///   shells or general-purpose command-line tools.
+    /// - `.raw`: Passes all bytes through unmodified. Use when the
+    ///   child process manages its own input handling, or when the
+    ///   parent process implements line editing.
+    ///
+    /// The child process may change this at any time via `tcsetattr`.
+    /// This setting only controls the initial state at spawn time.
+    ///
+    /// For more informationm see `cfmakeraw(3)` and `termios(4)`.
```
itingliu:
```suggestion
    /// For more information see `cfmakeraw(3)` and `termios(4)`.
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+
+        public init(rows: UInt, columns: UInt)
+    }
+    /// The dimensions of this terminal window
+    public let windowSize: WindowSize
+    /// The type of this terminal as defined in `terminfo`
+    public let terminalType: String
+
+    /// Update the dimensions of this terminal window
+    public func update(windowSize: WindowSize) throws(SubprocessError)
```
itingliu: What kind of error does this throw? Would it help to add the type to the comment if it's always one type?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    }
+
+    /// Creates an iterator for this asynchronous sequence.
+    public func makeAsyncIterator() -> Iterator
+}
+```
+
+## Introduce `preferredBufferSize` Parameter
+
+We propose adding a `preferredBufferSize: Int?` parameter to the `run()` overloads whose execution body closure allows standard output and/or error streaming. By default, Subprocess chooses the platform page size as the buffer size when creating `AsyncBufferSequence` for output streaming. This default buffer size might not be suitable for all use cases. In particular, when the child process output is sparse, Subprocess might appear stuck because it's waiting for the child process to write more bytes while the child process might be expecting more input. `preferredBufferSize` allows developers to choose the buffer size most suitable for their particular scenario.
```
itingliu: > This default buffer size might not be suitable for all use cases. In particular, when the child process output is sparse, Subprocess might appear stuck

I can see how this could be interpreted as an argument **against** having a default size because this can turn into a footgun easily. I think you've considered that before. Would you want to add a few sentences about this decision?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    /// the child process are merged into a single output stream. This is equivalent
+    /// to using shell redirection like `2>&1`.
+    ///
+    /// This is useful when you want to capture or redirect both output streams
+    /// together, making it possible to process all subprocess output as a unified
+    /// stream rather than handling standard output and standard error separately
+    ///
+    /// - Returns: A `CombinedErrorOutput` instance that merges standard error
+    ///   with standard output.
+    public static var combineWithOutput: Self
```
itingliu: I wonder if this reads better if it's called `combinedWithOutput` since the type itself is passive voice

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+
+While this design works conceptually, in practice we found that it's impossible to safely vend this API due to PID reuse. Specifically, on Windows a PID does NOT have the concept of `wait()` and reaping — the PID can be reused as soon as the process terminates. This creates a TOCTOU race condition: the PID may not be valid by the time `runDetached()` returns. Rather than designing an elaborate workaround for these race conditions, we elected to simply remove the `runDetached` API since it was never a core part of Subprocess.
+
+
+## Expand Platform-Specific `ProcessIdentifier` on Windows and Linux
+
+To address the potential TOCTOU issue with PIDs described above, we propose exposing platform-specific process file descriptors via `ProcessIdentifier` on Windows and Linux:
+
+```swift
+// For Linux
```
itingliu: This comment contradicts with `#if os(Linux) || os(Android) || os(FreeBSD)` below?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+public struct ProcessIdentifier: Sendable, Hashable {
+    /// The platform specific process identifier value
+    public let value: pid_t
+
+    #if os(Linux) || os(Android) || os(FreeBSD)
+    /// The process file descriptor (pidfd) for the running execution.
```
itingliu: I thought `pidfd` isn't available on FreeBSD? Should this say something like 
```suggestion
    /// The process file descriptor for the running execution. For example, pidfd on Linux 
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+    public let processDescriptor: CInt
+    #endif
+}
+
+// For Windows
+public struct ProcessIdentifier: Sendable, Hashable {
+    /// Windows specific process identifier value
+    public let value: DWORD
+    /// Process handle for current execution.
+    public nonisolated(unsafe) let processDescriptor: HANDLE
```
itingliu, line 818: I don't think it's common to have `public nonisolated(unsafe)` API. Was there any previous discussion that'd be useful to include here for the reader's reference why this needs to be `nonisolated(unsafe)`?

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+        throw MyError()
+        ...
+        throw MyOtherError()
+    }
+} catch let subprocessError as SubprocessError {
+    // Handle errors thrown from within Subprocess itself.
+    // These errors usually indicate some issue with the environment
+    // or a bug within Subprocess itself.
+    switch subprocessError.code {
+    case .spawnFailed:
```
itingliu, line 996: Is this section outdated? `code` is `public struct Code` so you can't `switch` it

Proposals/XXXX-subprocess-1.0.md
jmschonfeld: Agreed, any cases where a function is taking a generic return value of a body closure and just returning it to the caller should permit `~Copyable` return types

Proposals/XXXX-subprocess-1.0.md
jmschonfeld: Also, `@_nonSendable` isn't intended to be used in source code. If the goal is to mark this as explicitly non-Sendable, you'll want to add an unavailable extension. The two don't behave the same - `@_nonSendable` is underscored because it's a byproduct of the clang importer and not for source code

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+extension AsyncBufferSequence {
+    /// Line sequence parses and splits an asynchronous sequence of buffers into lines.
+    ///
+    /// It is the preferred method to convert `Buffer` to `String`
+    public struct LineSequence<Encoding: _UnicodeEncoding>: AsyncSequence, Sendable {
+        /// The element type for the asynchronous sequence.
+        public typealias Element = String
+
+         /// The iterator for line sequence.
+        public struct AsyncIterator: AsyncIteratorProtocol {
```
jmschonfeld, line 568: Looks like this type is missing a `Sendable` annotation

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+## Introduce `AsyncBufferSequence.LineSequence`
+
+The original proposal only included a way to stream a list of `Buffer`s. This makes streaming text difficult since naively converting each `Buffer` to `String` may not always succeed if the `Buffer` happens to break within a grapheme cluster. Since streaming text is one of the most common use cases for `Subprocess`, we propose introducing a new `AsyncBufferSequence.LineSequence` specifically designed to parse and partition an asynchronous sequence of buffers into text lines. Developers can optionally specify a String encoding and a `BufferingPolicy` to control how `LineSequence` handles the exhaustion of a buffer’s capacity.
+
+```swift
+extension AsyncBufferSequence {
+    /// Line sequence parses and splits an asynchronous sequence of buffers into lines.
+    ///
+    /// It is the preferred method to convert `Buffer` to `String`
+    public struct LineSequence<Encoding: _UnicodeEncoding>: AsyncSequence, Sendable {
```
jmschonfeld: It's likely worth documenting what constitutes the end of a line (`\n`, `\r\n`, either of the two, platform dependent, etc.) and whether or not the line will contain the line terminator or if it behaves more like the `split(separator:)` algorithm

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,948 @@
...
+
+We propose two changes to `TerminationStatus`:
+
+1. Remove `.unhandledException()` on Windows, since `TerminationStatus` cannot reliably determine whether the exit code represents a normal exit or an unhandled exception.
+2. Rename `.unhandledException()` to `.signaled()` on Unix systems, since the underlying mechanism is signal delivery, not exception handling.
+
+```swift
+/// An exit status of a subprocess.
+public enum TerminationStatus: Sendable, Hashable {
+    #if canImport(WinSDK)
```
jmschonfeld: The document alternates between `os(Windows)` and `canImport(WinSDK)`. Is there a meaningful difference between them, or should they be made consistent?

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: No this is really just a simple example to illustrate the point. In any real world use case developers would at least want to check and make sure the process exited successfully.

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: Ahh good idea thanks!!

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: By function parameter, I mean that now `standardOutput` is passed to you as a parameter, just like how Execution is passed to your closure as a parameter. Like

```
func doSomething(parameter: AsyncBufferSequence) {
    // I just "get" parameter. I can't call anything like `execution.parameter` again.
}
```

Let me reword this a bit

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: Good idea thanks!

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: The iterator is explicitly not Sendable because this is where we actually perform reads and hold the reference to the unique underlying file descriptor. We don't want reads.next() calls to happen from different threads.

@jmschonfeld ahh thanks. Updated.

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: This is using Swift's switch pattern matching. We are switching on the static values here.

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: I think in practice they are equivalent, but they are still semantically different. I'll stick to `os(Windows)` for behavioral differences and use `canImport` to actually guard the imports.

Proposals/XXXX-subprocess-1.0.md
iCharlesHu: We are moving away from "unhandled exception" anyway. Now it's just `signaled` on Unix.

iCharlesHu:
@swift-ci please test

Proposals/XXXX-subprocess-1.0.md
itingliu: Can we modify this example to include that then?
```suggestion
let result = try await run(...) { execution in
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,863 @@
...
+    for try await item in execution.standardOutput { ... }
+}
+
+// After
+public func run<Result>(
+    ...
+    body: (Execution, StandardInputWriter, AsyncBufferSequence, AsyncBufferSequence) async throws -> Result
+) async throws -> ExecutionOutcome<Result> { ... }
+
+_ = try await run(...) { execution, standardInput, standardOutput, standardError in
```
itingliu:
```suggestion
let result = try await run(...) { execution, standardInput, standardOutput, standardError in
```

Proposals/XXXX-subprocess-1.0.md
```diff
@@ -0,0 +1,863 @@
...
+    for await item in standardOutput { ... }
+}
+```
+
+This change eliminates the need for `Atomic` and `AtomicBox` within `Execution`, and more importantly, makes these two variables semantically closer to the correct mental model of how output streaming works.
+
+In the original design, since `.standardOutput` and `.standardError` are properties on `Execution`, it creates the illusion that you can repeatedly call these properties and create different output streams:
+
+```swift
+_ = try await run(...) { execution in
```
itingliu:
```suggestion
let result = try await run(...) { execution in
```

Proposals/XXXX-subprocess-1.0.md
itingliu: ah yeah ok I see

itingliu (approved):
I left some nitpicky suggestions for the examples.
 
There's still one pending discussion about ~Copyable `Result` but that's not blocking the proposal from going into review.

iCharlesHu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

PR #1813, opened 2026-03-12
fixes a crash when skipping array fields during decoding

t089:
Fixes a crash in `skipValue()` when skipping array/dictionary values during decoding.

### Motivation:

When implementing `JSONDecodable` conformance with `decodeEachKeyAndValue` and a `default: break` to skip unknown keys, decoding crashes with `preconditionFailure("Wrong node type")` if any of the unknown fields values is an array (or a nested object containing arrays). This is because `skipValue()` delegates to `BlackHoleVisitor` via `ArrayDecoder`/`StructDecoder`, but these decoders inherit the parent's coding path node (e.g. `.dictionary`) instead of pushing their own. When `decodeEachElement` calls `incrementArrayIndex()` on the inherited `.dictionary` node, it hits the precondition failure.

### Modifications:

- In `ParserState.skipValue()`, push the appropriate `CodingPathNode` before creating the sub-decoder for both `._openbrace` and `._openbracket` cases.
- Point `currentTopCodingPathNode` at the new node before constructing the `StructDecoder`/`ArrayDecoder`
- After the sub-decoder finishes, unwind back to the parent node (using `withExtendedLifetime` to ensure the node stays alive).

### Result:

Skipping keys in JSON whose values are objects or array no longer crashes.

### Testing:

Validated with a reproducer that no longer crashes after the patch is applied.

### DIFF
--- a/Tests/NewCodableTests/JSONEncodingDecodingTests.swift
+++ b/Tests/NewCodableTests/JSONEncodingDecodingTests.swift
@@ -2166,4 +2166,43 @@ struct JSONEncodingDecodingTests {
         // Optional data should encode the same way.
         _testRoundTrip(of: Optional(data), expectedJSON: expectedJSON)
     }
+
+    /// A simple type that only decodes the "name" field and skips all unknown keys.
+    struct SkipValueOuter: JSONDecodable, Equatable {
+        var name: String
+
+        static func decode(from decoder: inout some JSONDecoderProtocol & ~Escapable) throws(CodingError.Decoding) -> SkipValueOuter {
+            try decoder.decodeStruct { s throws(CodingError.Decoding) in
+                var name: String?
+                try s.decodeEachKeyAndValue {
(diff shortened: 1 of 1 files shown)


kperryua:
@t089 is there a reproducer test case you could add to the suite with this fix?

t089:
Sure, before I check myself, does the Suite Support "exit tests" to observe a crash?

kperryua:
Looks like it's supported on a number of platforms. We'd have to crib a conditional compiler setting from Foundation's `testOnlySwiftSettings` in the Package.swift to enable it safely.

But we're not testing an expected assertion here right? Just a crash fix? The test running to completion should be verification enough. Using the process exit infrastructure might even make it harder for us to get backtrace diagnostics if it were to start crashing again later.

t089:
Ok got it. Yes then I can add a test case for this.

kperryua:
Thanks a ton! You probably want to rebase the branch on top of the latest version that has the various test fixes.

t089:
@kperryua added a couple of tests

PR #1814, opened 2026-03-12
Fix baseline tests

kperryua:
These changes should bring the existing baseline tests to 0 failures.


kperryua:
The remaining test failures are a result of a mistake in the JSONDecodable macro. I'm fixing it with #1815.

PR #1815, opened 2026-03-12
Optional fields should be decoded as optionals to account for literal 'nulls' in the input

kperryua:
This detail evaded my notice in the original macro PR.


kperryua:
This Windows test crash is probably extant, hidden by the other failures. It will need to be diagnosed.

PR #1816, opened 2026-03-13
Add span-based input and output interfaces

kperryua:
The Data interfaces still exist for convenience. They should eventually move to cross-module imports.

### DIFF
--- a/Sources/NewCodable/JSON/NewJSONDecoder.swift
+++ b/Sources/NewCodable/JSON/NewJSONDecoder.swift
@@ -113,12 +113,13 @@ public struct NewJSONDecoder {
         self.options = options
     }
     
+    // TODO: RawSpan-taking functions need to detect Unicode encoding (including BOM) and convert to UTF-8, if necessary.
     @usableFromInline
-    internal func _decode<T: JSONDecodable & ~Copyable>(_ type: T.Type, from data: Data, closure: (inout JSONParserDecoder) throws(CodingError.Decoding) -> T) throws(CodingError.Decoding) -> T {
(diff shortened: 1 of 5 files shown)


Sources/NewCodable/JSON/NewJSONDecoder.swift
```diff
@@ -148,13 +148,27 @@ public struct NewJSONDecoder {
 
     @_disfavoredOverload
     @inlinable
-    public func decode<T: CommonDecodable>(_ type: T.Type, from data: Data) throws(CodingError.Decoding) -> T {
-        try _decode(type, from: data) { parserDecoder throws(CodingError.Decoding) in
+    public func decode<T: CommonDecodable>(_ type: T.Type, from span: Span<UInt8>) throws(CodingError.Decoding) -> T {
```
t089: Should this be `RawSpan`?

kperryua: I'm not entirely clear on that TBH. For UTF-8 input, we very clearly are semantically reading individual bytes, which makes `Span<UInt8>` make more sense to be as input. However, internally we're holding on to it as a `RawSpan`, because we can use SIMD/SWAR techniques to read multiple bytes at a time.

That said, A) it's more straightforward for clients to get a `RawSpan` from a `Span<UInt8>` than to go the other way, and B) eventually I suspect we will need to support other wider Unicode encodings like present-day `JSONDecoder` does, which means we're no longer processing the input byte-by-byte.

kperryua:
Just the known Windows crash on the tests. @t089, any comments on this?

PR #1817, opened 2026-03-13
Use the given CFAllocator for copying CFURL file system path

jrflat:
Honor the `CFAllocator` used to create the `CFURL` when creating its file system path.

### Motivation:

Be more compatible with previous behavior, which used the given `CFAllocator` instead of the default/Swift allocator.

### Modifications:

When copying the POSIX file system path from a canonical file `CFURL`, take in as input the allocator used to create the `CFURL` and use it when creating a `CFString` from the UTF8 bytes instead of creating a Swift String and bridging.

### Result:

No change in swift-foundation behavior. The given `CFAllocator` is used to allocate the resulting path when the Swift implementation is enabled for `CFURL`.

### Testing:

`CFURL` testing with the Swift implementation enabled.


jrflat:
@swift-ci please test

PR #1819, opened 2026-03-16
Add support for `wasm32-unknown-emscripten` platform

MaxDesiatov:
Now that both `swiftc` and `swift-frontend` support Emscripten target triple we can follow up on [the previously pitched Emscripte support](https://forums.swift.org/t/pitch-emscripten-target-support-for-swift/85310) in libraries like `swift-foundation`.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -23,6 +23,8 @@ import CRT
 import WinSDK
 #elseif os(WASI)
 @preconcurrency import WASILibc
+#elseif os(Emscripten)
+@preconcurrency import EmscriptenLibc
 #endif
 
 #if FOUNDATION_FRAMEWORK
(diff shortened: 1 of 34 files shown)


MaxDesiatov:
@swift-ci test

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -532,6 +534,12 @@ enum _FileOperations {
             throw CocoaError.removeFileError(errno, resolve(path: pathStr))
         }
 
+        #if os(Emscripten)
+        // Emscripten doesn't have fts.h; recursive directory removal is not yet supported.
```
parkera: Please avoid manually wrapping comments in Foundation's sources - thanks.

parkera (commented):
Style nitpick, otherwise LGTM.

Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -532,6 +534,12 @@ enum _FileOperations {
...
+        // The rmdir above handles empty directories. For non-empty ones, we need a
+        // recursive implementation that doesn't depend on FTS.
```
MaxDesiatov:
```suggestion
        // Emscripten doesn't have fts.h; recursive directory removal is not yet supported. The `rmdir` above handles empty directories. For non-empty ones, we need a recursive implementation that doesn't depend on FTS.
```

parkera (approved):

MaxDesiatov:
@swift-ci test

kateinoigakukun (approved):
I think we can enable more APIs on Emscripten as it seems like tmpdir is available but this is a good starting point. We can relax restrictions later.

PR #1822, opened 2026-03-16
Fix Data.WritingOptions SetAlgebra semantics violations

jmschonfeld:
Fixes incorrect semantics of `Data.WritingOptions` `SetAlgebra` operations

### Motivation:

`Data.WritingOptions` does not follow the semantic requirements of `SetAlgebra`. Notably, the file protection part is an `enum` value (zero or one may be selected, not multiple) and so the existing bit mask-based arithmetic on these values is not correct. This causes unexpected results when using operations like `.contains`

### Modifications:

Add AEIC versions of each broken `SetAlgebra` function providing better behavior when possible

### Result:

`Data.WritingOptions` now behaves as you'd expect and abides by all `SetAlgebra` semantic constraints. The only slightly problematic behavior remaining is from `formUnion` - it is not possible to form a union between two option sets that both contain file protection parts, so we only keep one of them.

### Testing:

Added new unit tests that run exhaustive conformance validation tests as well as targeted tests of common expected use patterns.

### DIFF
--- a/Sources/FoundationEssentials/Data/CMakeLists.txt
+++ b/Sources/FoundationEssentials/Data/CMakeLists.txt
@@ -30,6 +30,7 @@ target_sources(FoundationEssentials PRIVATE
     Data+Reading.swift
     Data+Searching.swift
     Data+Writing.swift
+    Data+WritingOptions.swift
     DataProtocol.swift
     FileSystemRepresentable.swift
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

kperryua (approved):

Issue #1825, opened 2026-03-17
Find a new solution to the typed-throws conundrum of JSONError/CodingError


kperryua:
NewJSONEncoder/Decoder have a lot of functions that throw CodingErrors but call inner functions that throw *either* JSONError or a CodingError. Because typed throws doesn't support blocks that throw one of a fixed set of errors, we're currently forced to use this construction everywhere:

```
         do {
            ...
        } catch let error as JSONError {
            throw error.at(self.codingPath)
        } catch {
            // TODO: Fix unsavory language workaround
            throw error as! CodingError.Decoding
        }
```

This is extremely undesirable, and not really an option for Embedded Swift either. We need an alternative solution.

kperryua:
I've posited that the compiler *could* sugar this with a local enum that wraps the two types and lets the client do:

```
        } catch let error as JSONError {
            ...
        } catch let error as CodingError.Decoding {
            ...
        }
```

which de-sugars to:

```
        } catch .json(let error) {
            ...
        } catch .coding(let error) {
            ...
        }
```

and is exhaustive. We might need to look into implementing this ourselves and maybe coming up with some non-intrusive function or operator that wraps a function call's error in the desired enum case.

kperryua:
Merged #1830.

PR #1830, opened 2026-03-18
Added error 'lift' operator that more ergonomically handles mixed error types via combined enum

kperryua:
Per https://github.com/swiftlang/swift-foundation/issues/1825, improve the ergonomics of `JSONError` and `CodingError.Decoding` errors being mixed in recursive JSON parsing. This also eliminates the `as!` cast that we currently are using to get by, but which isn't compatible with Embedded Swift.

Wishing for better language support here. This pattern being implicitly sugared into the language would be great. But I'd settle for changes to the typed throws inference / overload disambiguation mechanisms that would allow this bespoke operator to be postfix-unary.

### DIFF
--- a/Sources/NewCodable/JSON/JSONError.swift
+++ b/Sources/NewCodable/JSON/JSONError.swift
@@ -172,3 +172,55 @@ extension JSONError {
         }
     }
 }
+
+// MARK: - Mixed Error Handling
+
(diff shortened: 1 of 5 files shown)


kperryua:
Merging this major infrastructural change to avoid further drifting.

PR #1840, opened 2026-03-20
Replace LockedState with Mutex

jmschonfeld:
Replaces all uses of `LockedState` with `Mutex` and `Atomic`

### Motivation:

`LockedState` is a custom version of `Mutex`, but `Mutex` is strictly better since its concurrency annotations are more correct and it uses the latest language features. Updating to `Mutex` and `Atomic` allows us to remove our custom `LockedState` implementation

### Modifications:

- Removes all uses of `LockedState`
- Removes all conditionals around importing `Synchronization`
- Uses `Mutex` and `Atomic` instead of `LockedState` where appropriate

### Result:

All code still builds successfully using `Mutex`/`Atomic`

### Testing:

All code is tested by existing unit tests


jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/JSON/JSONDecoder.swift
```diff
@@ -167,91 +168,91 @@ open class JSONDecoder {
     /// The strategy to use in decoding dates. Defaults to `.deferredToDate`.
     open var dateDecodingStrategy: DateDecodingStrategy {
         get {
-            optionsLock.lock()
-            defer { optionsLock.unlock() }
+            optionsLock._unsafeLock()
```
itingliu, line 171: Would it work if we write this as
```
            optionsLock.withLock { _ in
                options.dateDecodingStrategy
            }
```
?

Sources/FoundationEssentials/Formatting/FormatterCache.swift
```diff
@@ -43,12 +45,12 @@ package struct FormatterCache<Format : Hashable & Sendable, FormattingType: Send
     }
 
     package subscript(key: Format) -> FormattingType? {
-        _lock.withLock {
+        _lock.withLockExtendingLifetimeOfState {
```
itingliu, line 48: Is this a bug in the existing implementation? Although I guess we have been lucky. It seems that this function isn't actually being used anywhere

Sources/FoundationEssentials/NotificationCenter/ActorQueueManager.swift
```diff
@@ -58,14 +60,22 @@ internal final class _NotificationCenterActorQueueManager: _NotificationCenterAc
...
+
+        init(_ state: consuming Mutex<State>) {
+            self.state = state
+        }
+    }
+
     override init() {
-        state = LockedState(initialState: State())
-        workerTask = Task.detached { [state] in
+        stateRef = StateReference(Mutex(State()))
```
itingliu, line 75: I think I understand why we need to box the mutex inside a `StateReference` , but do you know if this is a general pattern suggested by the language, or is it more akin to a workaround in our particular use case?

Sources/FoundationEssentials/Mutex+ExtendedLifetime.swift
```diff
@@ -0,0 +1,35 @@
...
+//===----------------------------------------------------------------------===//
+package import Synchronization
+
+extension Mutex where Value : Sendable {
+    private struct LockResult<T: ~Copyable, V: ~Copyable>: ~Copyable {
+        let bodyResult: T
+        let extendedValue: V
+    }
+
+    package func withLockExtendingLifetimeOfState<Result: ~Copyable, E>(
```
itingliu, line 19: Do you know if there's a reason why Mutex doesn't provide this function?

Sources/FoundationEssentials/JSON/JSONDecoder.swift
jmschonfeld: I believe so, but the `withLock` strategy does not work for the `_modify` accessor, so I believe I wrote it this way to be consistent across the accessors. At some point we should come back and try to rewrite this and fix the unsafe `_modify` accessor, but for now I left this as close as possible to the original

Sources/FoundationEssentials/NotificationCenter/ActorQueueManager.swift
jmschonfeld: It's a pattern for a workaround until we have more official language features supporting this. This is effectively a borrow that we don't have a way to represent yet. Ideally we'd use the `Borrow` type, but I also think we can't statically guarantee the lifetimes are correct (i.e. statically guarantee the detached task doesn't outlive this object) and so what we'd really need is something like Rust's share type to create a reference to a `~Copyable` type. For now the pattern to do that is wrap it in a class, but language features in the future may be able to wrap it in something better

Sources/FoundationEssentials/Mutex+ExtendedLifetime.swift
jmschonfeld: Good question for @Azoy - it does seem to be a bit higher level than the primitives and not something I'd expect something like a C library's `os_unfair_lock` to offer, but since `Mutex` is made for Swift perhaps it makes sense since I think this does have general utility to prevent `deinit` calls from triggering under the lock?

Sources/FoundationEssentials/Formatting/FormatterCache.swift
jmschonfeld: Yeah I think so - the `removeAllObjects` above already extends the lifetime of state, so I don't think it makes sense to differ the behavior depending on whether one or all objects are removed. I can't remember what exactly prompted me to add this here, but it seems fine to leave in for consistency with the above

PR #1843, opened 2026-03-22
Test updates to support --build-system swiftbuild

owenv:
Similar to the changes in corelibs-foundation in https://github.com/swiftlang/swift-corelibs-foundation/commit/9c46a72b0f96bfd75d372afe135ae79c0652edd1:
- Update test resource lookup to support either --build-system native or --build-system swiftbuild
- Update process name tests to accommodate different test runner naming schemes


owenv:
@swift-ci test

Issue #1190, opened 2025-02-25
Double encoding of query component


lesha1201:
I've noticed that when parsing some URL with reserved characters in `query` component, it does percent-encoding of a whole `query` instead of only invalid characters which is different behavior from Swift 5:

![Image](https://github.com/user-attachments/assets/36431e9d-20de-4e16-b930-4942f29a56d2)
[Playground](https://swiftfiddle.com/?c=H4sIAAAAAAAAA62SUUvDMBSF%2F8rlQqGF2WYbFVcYgzkEwSfFp8WH2mUa6G5CcruX0f9u2m7UBwUH3ryEm%2FOdA4ecsMIC9cEax%2FBgGtqVrA3JcGrF0Lj6hZ2mDw9L2EqCMBI%2Fma0vsmyxyOcuv5kLIdLKv6eltdmKyoNasvIscfJ3YHs1EW2m0d0s2ohonff36bC5LvZfXKLfXSS9dV3ujRu7BE3fiz0NQXoP58ZD1a%2FPT7Hv34tRmlyk3diw4VjivaEgbCpWu44qQMo4EInEZBC3oGqvfkIf6VjWusfgEjbQ57zRQ1KLEzyGr3KbinSG7RfgJF8tOAIAAA%3D%3D)

The problem is that there are a lot of URL that don't encode all reserved characters (e.g. `[`) by RFC3986 and the new behavior might break those URLs when there are already some percent-encoded characters.

I think the old behavior is a more safe approach because it's closer to the parsing of such URL on the web and preserves already encoded parts so it doesn't result in double encoding.

For the context, I'm not a Swift developer and don't know what's the right behavior in this case. Originally, I've encounted that issue when were sharing URL and opening it on iOS. When the URL is `https://9953r5-3000.csb.app/?name=[тест` and you copy it from the browser, it actually copies `https://9953r5-3000.csb.app/?name=[%D1%82%D0%B5%D1%81%D1%82` (notice that `[` symbol isn't encoded). And opening it in many iOS apps actually opens `https://9953r5-3000.csb.app/?name=%5D%25D1%2582%25D0%25B5%25D1%2581%25D1%2582`. Though `https://9953r5-3000.csb.app/?name=[%D1%82%D0%B5%D1%81%D1%82` opens fine on all platfroms (including WebKit browsers) but I guess some iOS apps uses Swift URL to parse the URL first and then opens it in a browser.

adozenlines:
[...]

PR #1844, opened 2026-03-23
Faster `Span`-based `URL` implementation

jrflat:
Implement a new backing class for `URL` focused on performance and gated behind the `FOUNDATION_SWIFT_URL_V2` build flag.

### Motivation:

Many common URL operations such as path manipulation or resolution (currently) involve repeated string allocations, percent-decoding, and re-parsing. By working directly with `Span`(s) of the underlying bytes and replacing just the encoded path during path manipulation (ensuring that nearby ranges and path-related flags are updated correctly), we see 5-8x speedup in use cases like appending or deleting path components.

The new backing type also consolidates and shares code from the Swift `NSURL`/`CFURL` implementation, which improves parsing, validation, encoding, and resolution performance (2-3x speedup), and makes it easier to maintain consistent behavior across all the types.

### Modifications:

- **`URL` (URL.swift):** When `FOUNDATION_SWIFT_URL_V2` is defined, `struct URL` uses the new `_URL` backing class.

- **`_URL` (URL_Impl.swift):** New backing class for `struct URL` when the v2 build flag is enabled. Stores a `_URLInfo` and uses `Span<UInt8>` in its methods where applicable to avoid intermediate `String` allocations.

- **`_URLInfo` (URL_Info.swift):** New storage struct holding the URL string, flags, and component ranges (of UTF8 bytes). Conforms to the existing `_URLParseable` and `_URLHeader` protocols (shared with `__CFSmallURLImpl` and `__CFBigURLImpl`), so it plugs directly into the same parsing and resolution functions that Swift `CFURL` uses. Provides `parse(string:)` and `parse(filePath:)` entry points, file path info construction, and a `replacing(path:)` method for efficient path-only mutations that update flags and shift query/fragment ranges without re-parsing.

- **Span+Path.swift:** `Span<UInt8>` extensions for path and path component operations.

- **URL_File.swift:** Extracted `finalPathLength`, `parseFinalFileSystemRepresentation`, `parsePOSIXPath`, and `parseWindowsPath` from `URL_C+File.swift` into `URL` static methods to share between implementations.

- **URL_Parsing.swift:** Moved the free parsing, validation, and encoding functions into a `URL` extension (indentation change). Removed `#if FOUNDATION_FRAMEWORK` guards to use these for `_URLInfo` on all platforms.

- **URL_Resolution.swift:** Removed `#if FOUNDATION_FRAMEWORK` guards.

- **URLEncoder.swift:** Removed `#if FOUNDATION_FRAMEWORK` guards.

### Result:

<details>
<summary>Ubuntu 24.04 `FOUNDATION_SWIFT_URL_V2` vs. Baseline Performance</summary>
Swift URL v2 vs v1 Performance Linux
</details>

When `FOUNDATION_SWIFT_URL_V2` is enabled, `URL` operations are significantly faster. The flag is not yet enabled, but has been tested. It can be enabled for swift-foundation in a separate PR after further cross-repo testing.

When enabled, `URL`, `NSURL`, and `CFURL` share the same Swift code for parsing, percent-encoding, file path initialization, and resolution.

**Bug fixes/compatibility behaviors:**
  - Prevents double-encoding of already-percent-encoded characters when another invalid character exists in the component (deviating from `URLComponents` encoding to match original `CFURL` byte encoding behavior and the behavior of most other RFC 3986 and WHATWG parsers). Will resolve issues like #1190 when enabled.
  - Allows `[` and `]` unencoded in the path, query, and fragment via #1802, matching WHATWG behavior.
  - `standardized` now preserves leading `..` segments in relative paths so they're not removed until resolution.
  - File path colon-encoding in the first relative path segment is more lenient when the character preceding the colon cannot be part of a scheme (e.g. `%43:` is not mistaken for a scheme).
  - `appendingPathExtension` rejects extensions containing invalid characters that would be problematic when decoded (e.g. spaces or Unicode directional override characters) instead of storing them percent-encoded.

### Testing:

- Added unit tests for path components, last path component, path extensions, appending/deleting path extensions, standardization edge cases, base URL query/fragment inheritance, data representation round-trips, and standardizing file URLs with symlinks.
- Added benchmarks for file and non-file URL path manipulation.
- Tested on Darwin, Linux, and Windows with and without `FOUNDATION_SWIFT_URL_V2` build flag (e.g. `swift test -Xswiftc -DFOUNDATION_SWIFT_URL_V2`).

### DIFF
--- a/Sources/FoundationEssentials/String/Span+Path.swift
+++ b/Sources/FoundationEssentials/String/Span+Path.swift
@@ -10,6 +10,10 @@
 //
 //===----------------------------------------------------------------------===//
 
+#if canImport(Darwin)
+import Darwin
+#endif
+
 @inline(__always)
 private func isSeparator(_ byte: UInt8) -> Bool {
     #if os(Windows)
--- a/Sources/FoundationEssentials/URL/URL_Parsing.swift
+++ b/Sources/FoundationEssentials/URL/URL_Parsing.swift
@@ -18,6 +18,10 @@
 internal import _ForSwiftFoundation
 #endif
 
+#if canImport(Darwin)
+import Darwin
+#endif
+
 // MARK: - Encoding types for URL buffers
 
 // URL types percent-encode invalid characters using their UTF8 representation,


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
I'd also like to investigate changing `final class _URL` to be a `struct` outside of `FOUNDATION_FRAMEWORK` and seeing if there are performance wins, but that requires some decent refactoring around `_URLProtocol` and its conformers to change all inits to static factory methods. Since that would touch un-guarded code and this PR is already fairly large, I'd prefer to do that in a follow-up PR.

PR #1846, opened 2026-03-23
Fix benchmarks build on non-Darwin

jrflat:
Fix `swift package benchmark` build failure on non-Darwin due to `benchmarks` variable being hidden by some `#if` guards.

### Motivation:

Run benchmarks on Linux/Windows et al.

### Modifications:

- Move `#if FOUNDATION_FRAMEWORK` in `BenchmarkCharacterSet.swift` to within the `benchmarks` declaration so the plugin can find the variable, but it has no benchmarks to run outside of `FOUNDATION_FRAMEWORK`.
- Change `#if os(macOS) && USE_PACKAGE` to `#if !FOUNDATION_FRAMEWORK` in `InternationalizationBenchmark.swift`, matching the `#if` guard within the individual benchmark files.

### Result:

`swift package benchmark` succeeds on non-Darwin.

### Testing:

Tested benchmarks on Ubuntu 24.04, `FOUNDATION_FRAMEWORK` macOS, and non-`FOUNDATION_FRAMEWORK` macOS.

### DIFF
--- a/Benchmarks/Benchmarks/CharacterSet/BenchmarkCharacterSet.swift
+++ b/Benchmarks/Benchmarks/CharacterSet/BenchmarkCharacterSet.swift
@@ -10,7 +10,6 @@
 //
 //===----------------------------------------------------------------------===//
 
-#if FOUNDATION_FRAMEWORK
 import Benchmark
 import func Benchmark.blackHole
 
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

PR #1847, opened 2026-03-23
Revert "[Compiletime Performance] Add an overload to Data for concatenation (#1828)"

phausler:
The previous change ended up causing CI failures so this is being reverted to avoid ambiguity causing compatibility test suites to fail.


jmschonfeld:
@swift-ci please test

phausler:
@swift-ci please test

PR #1849, opened 2026-03-24
Field utf8span property

kperryua:
Swap `withUTF8Span` for `utf8Span` with a yielding borrow accessor.

### Motivation:

The `withUTF8Span` API on `CodingField` was an unfortunate accommodation since `StaticString`'s could not guarantee an immortal lifetime for non-pointer-representation strings. This was especially unfortunate given how rare these strings are.

### Modifications:

Use `yielding borrow` accessors to maintain the lifetime of a `UTF8Span` so we don't need the closure.

### Result:

No more unfortunate closures.

### Testing:

The `utf8Span` property implementation is specified by default implementation on `StaticStringCodingField`, so any test/benchmark types currently using that will automatically get this.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -254,6 +254,7 @@ let package = Package(
                 .enableExperimentalFeature("BuiltinModule"),
                 .enableExperimentalFeature("Lifetimes"),
                 .enableExperimentalFeature("SuppressedAssociatedTypes"),
+                .enableExperimentalFeature("CoroutineAccessors"),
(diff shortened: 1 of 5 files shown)


kperryua:
This construction worked on an older toolchain, but simply fails with a compiler error on newer toolchains. Has it been defined as invalid somehow, or is it a bug?

kperryua:
Update: this still fails on 5/4/26 main toolchain.

PR #1850, opened 2026-03-25
Adopt [SE-0503]: SuppressedAssociatedTypesWithDefaults

kperryua:
Newer toolchains are warning that the `SuppressedAssociatedTypes` experimental flag is deprecated, so migrate to the new version.


kperryua:
This CI for this branch evidently doesn't not have the new feature flag supported? I think this will need to wait for a Swift 6.4 based branch for this.

kperryua:
This got handled along with #1947

PR #1851, opened 2026-03-26
Attempt to fix new-codable Windows test crashes

kperryua:


kperryua:
Nope, it doesn't appear to be the calls to strtod/strtof.

kperryua:
A swift-ci windows test indicate the problematic test is probably the `depthTraversal` one.

kperryua:
Woo! First fully green build.

PR #1857, opened 2026-03-26
Enable `Span`-based `URL` implementation

### LINKED ISSUE #1190: Double encoding of query component

jrflat:
Enables the faster `Span`-based `URL` implementation from #1844 by removing the `FOUNDATION_SWIFT_URL_V2` build guard. See more implementation details in #1844

### Motivation:

Many common URL operations such as path manipulation or resolution (currently) involve repeated string allocations, percent-decoding, and re-parsing. By working directly with `Span`(s) of the underlying bytes and replacing just the encoded path during path manipulation (ensuring that nearby ranges and path-related flags are updated correctly), we see 5-8x speedup in use cases like appending or deleting path components.

The new backing type also consolidates and shares code from the Swift `NSURL`/`CFURL` implementation, which improves parsing, validation, encoding, and resolution performance (2-3x speedup), and makes it easier to maintain consistent behavior across all the types.

### Modifications:

Remove `FOUNDATION_SWIFT_URL_V2` guard so that the `_URL` backing class is used.

### Result:

<details>
<summary>Ubuntu 24.04 URL Performance vs. Baseline</summary>
Swift URL v2 vs v1 Performance Linux
</details>

`URL` operations are significantly faster, and `URL`, `NSURL`, and `CFURL` share the same Swift code for parsing, percent-encoding, file path initialization, and resolution.

**Bug fixes/compatibility behaviors:**
  - Prevents double-encoding of already-percent-encoded characters when another invalid character exists in the component (deviating from `URLComponents` encoding to match original `CFURL` byte encoding behavior and the behavior of most other RFC 3986 and WHATWG parsers). Resolves #1190.
  - ~Allows `[` and `]` unencoded in the path, query, and fragment via #1802, matching WHATWG behavior.~ This behavior was too lenient and caused discrepancies between `URL` and `URLComponents` APIs. Mostly reverted in #1875.
  - `standardized` now preserves leading `..` segments in relative paths so they're not removed until resolution.
  - File path colon-encoding in the first relative path segment is more lenient when the character preceding the colon cannot be part of a scheme (e.g. `%43:` is not mistaken for a scheme).
  - `appendingPathExtension` rejects extensions containing invalid characters that would be problematic when decoded (e.g. spaces or Unicode directional override characters) instead of storing them percent-encoded.

### Testing:

- Benchmarks for file and non-file URL path manipulation.
- Unit tests on Darwin, Linux, and Windows.

### DIFF
--- a/Sources/FoundationEssentials/String/CMakeLists.txt
+++ b/Sources/FoundationEssentials/String/CMakeLists.txt
@@ -17,6 +17,7 @@ target_sources(FoundationEssentials PRIVATE
     BuiltInUnicodeScalarSet.swift
     IANACharsetNames.swift
     RegexPatternCache.swift
+    Span+Path.swift
     String+Bridging.swift
     String+Comparison.swift
     String+Encoding.swift
--- a/Sources/FoundationEssentials/URL/CMakeLists.txt
+++ b/Sources/FoundationEssentials/URL/CMakeLists.txt
@@ -15,10 +15,17 @@
 target_sources(FoundationEssentials PRIVATE
     URL.swift
     URL_Bridge.swift
+    URL_File.swift
+    URL_Impl.swift
+    URL_Info.swift
+    URL_Parsing.swift
     URL_Protocol.swift
+    URL_Resolution.swift
     URL_Swift.swift
+    URL_Validation.swift
     URLComponents.swift
     URLComponents_ObjC.swift
+    URLEncoder.swift
     URLParser.swift
     URLTemplate_Expression.swift
     URLTemplate_PercentEncoding.swift
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -620,13 +620,7 @@ internal func foundation_swift_url_v2_enabled() -> Bool {
 }
 #else
 internal func foundation_swift_url_enabled() -> Bool { return true }
-internal func foundation_swift_url_v2_enabled() -> Bool {
-    #if FOUNDATION_SWIFT_URL_V2
-    return true
-    #else
-    return false
-    #endif
-}
+internal func foundation_swift_url_v2_enabled() -> Bool { return true }
 #endif
 
 #if canImport(os)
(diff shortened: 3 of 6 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jmschonfeld (approved):

jrflat:
Squashed and rebased to pick up bug fixes for cross-repo testing

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

PR #1874, opened 2026-04-03
enable DST timezone test

xavgru12:
enable DST timezone test as it was disabled [here](https://github.com/swiftlang/swift-foundation/commit/31466acfc1f39799ac82cc0bdbe46a5941b796ce).
Add a test case for Europe daylight saving.

### Motivation:

Enable and extend test cases for Europe timezone.

### Modifications:

changed to GMT+1

### Result:

Tested both in current time and setting local time to winter(date returns a time in winter).

### Testing:

Test passes again.

### DIFF
--- a/Tests/FoundationInternationalizationTests/TimeZoneTests.swift
+++ b/Tests/FoundationInternationalizationTests/TimeZoneTests.swift
@@ -86,8 +86,7 @@ private struct TimeZoneTests {
 
         test("America/Los_Angeles", "zh_TW", .shortStandard, "PST", "PST")
 
-        // Disabled because it fails when there is DST in Europe
-        //test("Europe/Paris",       "zh_TW", .shortStandard, "GMT+1", "GMT+2")
+        test("Europe/Paris",       "zh_TW", .shortStandard, "GMT+1", "GMT+1")
 
         test("Antarctica/Davis",   "zh_TW", .shortStandard, "GMT+7", "GMT+7")
         test("Asia/Chongqing",      "zh_TW", .shortStandard, "GMT+8", "GMT+8")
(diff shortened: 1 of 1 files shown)


itingliu:
Thanks for the fix! 

While fixing it by changing the test expectation is correct, I don't quite understand the PR description:
> In Summer time, time becomes UTC+2 in Europe. This would mean that this test should have passed for summer time and failed in winter time. It seems that the mechanics under hood must have changed as the opposite was the case.

This sounds like one would expect the function to return UTC+2. Can you clarify what you meant here?

I think the real reason is that `.shortStandard` always formats the date in "standard" time zone, and does not consider DST at all. The other option `.shortDaylightSaving` does consider DST. It might be more clear if we view the expectations side by side:

```
test("Europe/Paris",       "zh_TW", .shortStandard, "GMT+1", "GMT+1") // always in standard time
test("Europe/Paris",       "zh_TW", .shortDaylightSaving, "GMT+1", "GMT+2") // may be GMT+2 if now is in DST
```

xavgru12:
shortDaylightSaving writes it in Taiwanese letters.

I also believe that shortStandard always returns GMT+1 all year. I just wonder how this test could ever pass with GMT+2 and I suspect that internal behavior changed over time. Since winter time is GMT+1, it is safe that this test will pass in winter as well.

xavgru12:
I modified the PR description and added the daylight saving test case. Thanks for your feedback!

itingliu:
> shortDaylightSaving writes it in Taiwanese letters.

Can you clarify which function would return Chinese characters (strictly speaking they're Chinese characters instead of Taiwanese letters)? Doesn't it return "GMT+1" or "GMT+2" like the test showed?

xavgru12:
> > shortDaylightSaving writes it in Taiwanese letters.
> 
> Can you clarify which function would return Chinese characters (strictly speaking they're Chinese characters instead of Taiwanese letters)? Doesn't it return "GMT+1" or "GMT+2" like the test showed?

I tested it again and it returns GMT, not Chinese. I must have run it without "short" before.

itingliu:
@swift-ci please test

itingliu (approved):

PR #1878, opened 2026-04-04
Initializers for joining a sequence of predicates

MatthewTurk247:
This pull request implements the initializers proposed in #1860.

### Testing:

See `PredicateTests/all()` and `PredicateTests/any()`.

### DIFF
--- a/Sources/FoundationEssentials/Predicate/Predicate.swift
+++ b/Sources/FoundationEssentials/Predicate/Predicate.swift
@@ -49,6 +49,52 @@ extension Predicate {
     }
 }
 
+@available(FoundationPreview 6.4, *)
+extension Predicate {
+    public init(all subpredicates: some BidirectionalCollection<Self>) {
+        var iterator = subpredicates.reversed().makeIterator()
+        
+        if let guarded = iterator.next() {
+            self.init({ (input: repeat PredicateExpressions.Variable<each Input>) in
+                var pieces: any StandardPredicateExpression<Bool> = PredicateExpressions.build_evaluate(PredicateExpressions.build_Arg(guarded), repeat each input)
+                
+                while let next = iterator.next() {
+                    pieces = Self.meet(PredicateExpressions.build_evaluate(PredicateExpressions.build_Arg(next), repeat each input), pieces)
+                }
+                
+                return pieces
+            })
+        } else {
+            self.init(value: true)
+        }
+    }
+    
+    public init(any subpredicates: some BidirectionalCollection<Self>) {
+        var iterator = subpredicates.reversed().makeIterator()
+        
+        if let guarded = iterator.next() {
+            self.init({ (input: repeat PredicateExpressions.Variable<each Input>) in
+                var pieces: any StandardPredicateExpression<Bool> = PredicateExpressions.build_evaluate(PredicateExpressions.build_Arg(guarded), repeat each input)
+                
+                while let next = iterator.next() {
+                    pieces = Self.join(PredicateExpressions.build_evaluate(PredicateExpressions.build_Arg(next), repeat each input), pieces)
+                }
+                
+                return pieces
+            })
+        } else {
+            self.init(value: false)
+        }
+    }
+    
+    fileprivate static func meet<T: StandardPredicateExpression<Bool>, U: StandardPredicateExpression<Bool>>(_ lhs: T, _ rhs: U) -> any StandardPredicateExpression<Bool> {
+        PredicateExpressions.build_Conjunction(lhs: lhs, rhs: rhs)
+    }
+    
+    fileprivate static func join<T: StandardPredicateExpression<Bool>, U: StandardPredicateExpression<Bool>>(_ lhs: T, _ rhs: U) -> any StandardPredicateExpression<Bool> {
+        PredicateExpressions.build_Disjunction(lhs: lhs, rhs: rhs)
+    }
+}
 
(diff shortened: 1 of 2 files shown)


Tests/FoundationEssentialsTests/PredicateTests.swift
```diff
@@ -541,4 +541,42 @@ private struct PredicateTests {
             #expect(try expression.evaluate(i) == i + 1)
         }
     }
+    
+    @Test func all() throws {
+        #expect(try Predicate(all: []).evaluate(42))
```
jmschonfeld, line 577: I'm actually a bit surprised this compiles without telling it what the `Input` type of the predicate is here. Does it infer an input of type `Int` from the call to `evaluate` using the type of `42`?

Tests/FoundationEssentialsTests/PredicateTests.swift
```diff
@@ -541,4 +541,42 @@ private struct PredicateTests {
...
+            #Predicate { !$0.f },
+            #Predicate { $0.g.contains(42) },
+            #Predicate { $0.g.min() == 0 }
+        ]
+        #expect(try Predicate(all: predicates).evaluate(input))
+        predicates.append(#Predicate { $0.g.max() == 7 })
+        #expect(try !Predicate(all: predicates).evaluate(input))
+        predicates.append(#Predicate { $0.c < 0.0 })
+        #expect(try !Predicate(all: predicates).evaluate(input))
+    }
```
jmschonfeld, line 595: Should we add tests that validate the short circuiting behavior to validate that predicates later in the list aren't evaluated if a prior expression fails? This should be possible with a custom object and a property (referenced in the predicate via a keypath) that records when the property is accessed.

Tests/FoundationEssentialsTests/PredicateTests.swift
MatthewTurk247: That’s what it seems like. My understanding is that the compiler can propagate the `Int` type backward from the argument list of the instance method and then constrain the array literal to `[Predicate<Int>]`.

jmschonfeld: Neat!

Tests/FoundationEssentialsTests/PredicateTests.swift
MatthewTurk247: Good idea. Just added `PredicateTests/allShortCircuitEvaluation()` and `PredicateTests/anyShortCircuitEvaluation()`.

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
The build failures look to be caused by this change:

```
/__w/swift-foundation/swift-foundation/Sources/FoundationEssentials/Predicate/Predicate.swift:59:90: error: static method 'build_evaluate' requires the types 'PredicateExpressions.PredicateRegex' and 'Expression<repeat each Input, Bool>' be equivalent
```

MatthewTurk247:
Type inference is not working as I was expecting. I’ll add explicit annotations.

Sources/FoundationEssentials/Predicate/Predicate.swift
```diff
@@ -49,6 +49,54 @@ extension Predicate {
...
 }
 
+@available(FoundationPreview 6.4, *)
+extension Predicate {
+    public init(all subpredicates: some BidirectionalCollection<Self>) {
+        var iterator = subpredicates.reversed().makeIterator()
+        
+        if let guarded = iterator.next() {
+            self.init({ (input: repeat PredicateExpressions.Variable<each Input>) in
+                let condition: PredicateExpressions.Value<Self> = PredicateExpressions.build_Arg(guarded)
```
jmschonfeld: The change to pieces makes sense to me, but for condition should we just create it directly via

```swift
let condition = PredicateExpressions.Value(guarded)
```

No need to go through the builder functions here if they're getting in the way, especially `build_Arg` - it's really for the macro which doesn't know what the return type should be, but since we know it we can just create the `Value` directly if it helps the type checker

jmschonfeld:
@swift-ci please test

MatthewTurk247:
Is there an issue with platform availability behind the failed method resolution? I’m starting to suspect that the build routine is not picking up symbols for `PredicateExpressions.PredicateEvaluate`.

jmschonfeld:
> Is there an issue with platform availability behind the failed method resolution? I’m starting to suspect that the build routine is not picking up symbols for `PredicateExpressions.PredicateEvaluate`.

Ah I think this error message has the key detail:

```
/__w/swift-foundation/swift-foundation/Sources/FoundationEssentials/Predicate/Predicate.swift:60:90: error: referencing static method 'build_evaluate' on 'PredicateExpressions.Value' requires that 'Predicate<repeat each Input>' conform to 'Decodable'
```

I had forgotten that `Predicate`'s conformance to `Codable` is gated behind `FOUNDATION_FRAMEWORK` because of dependencies not available in swift-foundation. Since `StandardPredicateExpression` requires a `Codable` conformance that means that the `PredicateEvaluate` operator also isn't available outside of `FOUNDATION_FRAMEWORK`. For now, I think we'll need to put this new API behind `FOUNDATION_FRAMEWORK` and separately we can investigate finding a way to lift that restriction

jmschonfeld (approved):

jmschonfeld:
@swift-ci please test

PR #1888, opened 2026-04-09
Support nesting Decodable types in CommonDecodable types

kperryua:
We need the ability to support embedded `Decodable` types in `CommonDecodable` types, like is currently supported for `JSONDecodable`.

### Modifications:

* Added `decode<Decodable>` requirement to `CommonDecodable`. This requires `decodeAny` support.
* Added `CommonCodablePrimitive` as `DecodedValue` for `decodeAny`.
* Added default implementation that uses `decodeAny` and `AdaptorDecoder`.
* Added `codingPath` requirement to `Common*` encoder/decoders and sub-encoder/decoders.

### Testing:

* Tests that embed `Decodable` and `Encodable` type in `Common*` types.
* Test that `Decodable` types see the correct `codingPath` provided by the `NewJSONDecoder`.

### DIFF
--- a/Sources/NewCodable/CommonCodablePrimitive.swift
+++ b/Sources/NewCodable/CommonCodablePrimitive.swift
@@ -17,8 +17,7 @@ import Foundation
 #endif
 
 /// An internal enum representing all the types that `CommonDecoder` supports.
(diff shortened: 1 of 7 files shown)


kperryua:
Test failures are the expected ones.

PR #1889, opened 2026-04-10
Issue #1818: Serialize access of a rule-based formatter

### LINKED ISSUE #1818: 0.formatted(.byteCount(style: .binary)) crashes because it's not thread-safe
**Describe the bug**
Formatting 0 with ByteCountFormatStyle is not thread-safe, and if you do it enough concurrently it will eventually crash due to corruption of the BreakCache pointer table.

```
💣 Program crashed: Bad pointer dereference at 0x0000000000000000

Platform: x86_64 Linux (Debian GNU/Linux rodete)

Thread 4 crashed:

  0 null
  1 0x00007f824991cbe5 icu::RuleBasedBreakIterator::handleNext() + 276 in lib_FoundationICU.so
  2 0x00007f824992351e icu::RuleBasedBreakIterator::BreakCache::populateFollowing() + 205 in lib_FoundationICU.so
  3 0x00007f824992341e icu::RuleBasedBreakIterator::BreakCache::nextOL() + 13 in lib_FoundationICU.so
  4 0x00007f824991c84d icu::RuleBasedBreakIterator::next() + 28 in lib_FoundationICU.so
  5 0x00007f82499c61bc ustrcase_internalToTitle + 171 in lib_FoundationICU.so
  6 0x00007f82499b0841 icu::UnicodeString::caseMap(int, unsigned int, icu::BreakIterator*, int (*)(int, unsigned int, icu::BreakIterator*, char16_t*, int, char16_t const*, int, icu::Edits*, UErrorCode&)) + 656 in lib_FoundationICU.so
  7 0x00007f82499b18d3 icu::UnicodeString::toTitle(icu::BreakIterator*, icu::Locale const&, unsigned int) + 114 in lib_FoundationICU.so
  8 0x00007f8249af78da icu::RuleBasedNumberFormat::adjustForCapitalizationContext(int, icu::UnicodeString&, UErrorCode&) const + 201 in lib_FoundationICU.so
  9 0x00007f8249af737d icu::RuleBasedNumberFormat::format(long, icu::NFRuleSet*, icu::UnicodeString&, UErrorCode&) const + 268 in lib_FoundationICU.so
 10 0x00007f8249af7258 icu::RuleBasedNumberFormat::format(long, icu::UnicodeString&, icu::FieldPosition&) const + 39 in lib_FoundationICU.so
 11 0x00007f8249b785e1 unum_formatInt64 + 240 in lib_FoundationICU.so
 12 0x00007f824c549ff7 specialized closure #1 in _withResizingUCharBuffer(initialSize:_:) + 102 in libFoundationInternationalization.so
 13 0x00007f824c54cab1 specialized _withResizingUCharBuffer(initialSize:_:) + 144 in libFoundationInternationalization.so
 14 0x00007f824c529300 ByteCountFormatStyle.Attributed._format(_:doubleValue:) + 3327 in libFoundationInternationalization.so
 15 0x00007f824c4e7db2 BinaryInteger.formatted<A>(_:) + 193 in libFoundationInternationalization.so
 16 0x00007f82502798fc
 17 0x00007f82509ccd50 swift::runJobInEstablishedExecutorContext(swift::Job*) + 367 in libswift_Concurrency.so
 18 0x00007f82509cda25 swift_job_run + 148 in libswift_Concurrency.so
... 
```

**To Reproduce**
```swift
import Foundation

func test() async {
  await withTaskGroup(of: Void.self) { group in
    for _ in 0..<100 {
      group.addTask {
        for _ in 0..<100 {
          _ = 0.formatted(.byteCount(style: .binary))
        }
      }
    }
  }
  print("Done")
}

Task {
  await test()
}
RunLoop.main.run(until: Date(timeIntervalSinceNow: 2))
```

**Expected behavior**
It should not crash.

**Configuration:**
Swift version 6.2.4 (swift-6.2.4-RELEASE)
Linux

**Additional context**
Here's my best (relatively quick) guess as to the problem - forgive me if I've jumped to the wrong conclusion on any of the details here, as I e.g. just used the latest source for swift-foundation and ICU, without checking what version is actually in use on my system:

### 1. Swift Foundation's "Zero" Special Case

In `ByteCountFormatStyle.swift` there is a special code path used exclusively for formatting the value **`0`**.  For 0, it wants to "spell out" the answer (e.g., returning `"Zero bytes"` or `"Zero kB"` instead of `"0 kB"`).

To do this, it requests a "descriptive" formatter:

```swift
let configuration = DescriptiveNumberFormatConfiguration.Collection(presentation: .cardinal, capitalizationContext: .beginningOfSentence)
// Fetches from a global, shared cache!
guard let spellOutFormatter = ICULegacyNumberFormatter.formatter(for: .descriptive(configuration), locale: locale) else { ... }

guard let zeroFormatted = spellOutFormatter.format(Int64.zero) else { ... }
```

### 2. The Thread-Safety Assumption in `swift-foundation`

If you look at `ICULegacyNumberFormatter.swift` inside Swift Foundation, it's explicitly marked as thread-safe (`@unchecked Sendable`):

```swift
internal final class ICULegacyNumberFormatter : @unchecked Sendable {
    /// `Sendable` notes: `UNumberFormat` is safe to use from multiple threads after initialization and configuration.
    let uformatter: UnsafeMutablePointer<UNumberFormat?>
```

It looks like Swift caches these formatters globally - presumably to improve performance - on the assumption that calling the C-API `unum_formatInt64` concurrently is safe. 

### 3. ICU's `RuleBasedNumberFormat` is not concurrency-safe

For standard decimal numbers (`DecimalFormat`), ICU might be thread-safe - however, a "spellout" / descriptive configuration forces ICU to use `RuleBasedNumberFormat`.

In ICU's source code (`rbnf.cpp`), `RuleBasedNumberFormat::format` is declared `const` (read-only). But inside this "const" method, it does the following:

```cpp
// From rbnf.cpp
ruleSet->format(number, toAppendTo, toAppendTo.length(), 0, status);
adjustForCapitalizationContext(startPos, toAppendTo, status); // <-- THE CULPRIT
```

To adjust the capitalisation (e.g. capitalising the "Z" in "Zero"), ICU uses a `BreakIterator` (specifically `capitalizationBrkIter`, a class member) to find word and sentence boundaries:

```cpp
currentResult.toTitle(capitalizationBrkIter, locale, U_TITLECASE_NO_LOWERCASE | U_TITLECASE_NO_BREAK_ADJUSTMENT);
```

Here's the fatal flaw: **ICU's `BreakIterator` is stateful.**

When `toTitle` uses the `BreakIterator`, it mutates the iterator's internal traversal state, updating caches and moving positional offsets. 

Because Swift Foundation cached the `UNumberFormat` globally and shared it across all your concurrent `Task`s, multiple threads are simultaneously entering `toTitle` and mutating the exact same `BreakIterator` instance.  This creates a nasty C++ data race that corrupts the internal `BreakCache` pointer table:

```
  0 null  <-- Corrupted pointer / vtable from the data race
  1 0x... icu::RuleBasedBreakIterator::handleNext() 
  2 0x... icu::RuleBasedBreakIterator::BreakCache::populateFollowing() 
```

The CPU jumps to `0x0` and the process crashes.

itingliu:
Resolves #1818 

ICU rule-based number formatter isn't thread safe. Add a lock to serialize access to this formatter if we happen to use one.

This issue is fixed in [upstream ICU](https://github.com/unicode-org/icu/pull/3928), so we will be able to remove this once the fix is available.

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
@@ -15,27 +15,39 @@ import FoundationEssentials
 #endif
 
 internal import _FoundationICU
+internal import Synchronization
 
 @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
 internal final class ICULegacyNumberFormatter : @unchecked Sendable {
 
-    /// `Sendable` notes: `UNumberFormat` is safe to use from multple threads after initialization and configuration.
+    // `Sendable` notes: `UNumberFormat` backed by ICU's `DecimalFormat` is safe to use from multiple threads after initialization and configuration. However, `RuleBasedNumberFormat`is not thread-safe. See https://github.com/unicode-org/icu/pull/3928
     let uformatter: UnsafeMutablePointer<UNumberFormat?>
+    // Optional lock to serializes access to uformatter.
+    // FIXME: Remove the lock once ICU fix is available
+    private let formatLock: Mutex<Void>?
 
-    private init(openedFormatter: UnsafeMutablePointer<UNumberFormat?>) {
+    private init(openedFormatter: UnsafeMutablePointer<UNumberFormat?>, needsLock: Bool) {
         uformatter = openedFormatter
+        formatLock = needsLock ? Mutex(()) : nil
+    }
+
+    private func withLockIfNeeded<T>(_ body: () -> T) -> T {
+        formatLock?.withLock { _ in body() } ?? body()
     }
 
     deinit {
+        // We don't need to lock around since the race in ICU only happens in formatting and parsing.
         unum_close(uformatter)
     }
 
     func parseAsInt(_ string: some StringProtocol) -> Int64? {
-        let arr = Array(string.utf16)
-        var status = U_ZERO_ERROR
-        let parsed = unum_parseInt64(uformatter, arr, Int32(arr.count), nil, &status)
-        guard status.isSuccess else { return nil }
-        return parsed
+        withLockIfNeeded {
+            let arr = Array(string.utf16)
+            var status = U_ZERO_ERROR
+            let parsed = unum_parseInt64(uformatter, arr, Int32(arr.count), nil, &status)
+            guard status.isSuccess else { return nil }
+            return parsed
+        }
     }
 
     // `upperBound`: the utf-16 position in the string where the parse ends
@@ -44,25 +56,27 @@ internal final class ICULegacyNumberFormatter : @unchecked Sendable {
         let arr = Array(string.utf16)
         var status = U_ZERO_ERROR
         var pos = Int32(0) // 0 == start, per ICU docs
-        let parsed = unum_parseInt64(uformatter, arr, Int32(arr.count), &pos, &status)
+        let parsed = withLockIfNeeded { unum_parseInt64(uformatter, arr, Int32(arr.count), &pos, &status) }
         guard status.isSuccess else { return nil }
         upperBound = Int(pos)
         return parsed
     }
 
     func parseAsDouble(_ string: some StringProtocol) -> Double? {
-        let arr = Array(string.utf16)
-        var status = U_ZERO_ERROR
-        let parsed = unum_parseDouble(uformatter, arr, Int32(arr.count), nil, &status)
-        guard status.isSuccess else { return nil }
-        return parsed
+        withLockIfNeeded {
+            let arr = Array(string.utf16)
+            var status = U_ZERO_ERROR
+            let parsed = unum_parseDouble(uformatter, arr, Int32(arr.count), nil, &status)
+            guard status.isSuccess else { return nil }
+            return parsed
+        }
     }
 
     func parseAsDouble(_ string: some StringProtocol, upperBound: inout Int) -> Double? {
         let arr = Array(string.utf16)
         var status = U_ZERO_ERROR
         var pos = Int32(0) // 0 == start, per ICU docs
-        let parsed = unum_parseDouble(uformatter, arr, Int32(arr.count), &pos, &status)
+        let parsed = withLockIfNeeded { unum_parseDouble(uformatter, arr, Int32(arr.count), &pos, &status) }
         guard status.isSuccess else { return nil }
         upperBound = Int(pos)
         return parsed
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

Sources/FoundationInternationalization/Formatting/Number/ICULegacyNumberFormatter.swift
```diff
@@ -15,27 +15,39 @@ import FoundationEssentials
...
 
 internal import _FoundationICU
+internal import Synchronization
 
 @available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
 internal final class ICULegacyNumberFormatter : @unchecked Sendable {
 
-    /// `Sendable` notes: `UNumberFormat` is safe to use from multple threads after initialization and configuration.
+    // `Sendable` notes: `UNumberFormat` backed by ICU's `DecimalFormat` is safe to use from multiple threads after initialization and configuration. However, `RuleBasedNumberFormat`is not thread-safe. See https://github.com/unicode-org/icu/pull/3928
     let uformatter: UnsafeMutablePointer<UNumberFormat?>
```
jmschonfeld, line 24: Part of me wonders if it's worth wrapping this in a sub type that requires you to access it through `withLockIfNeeded` like `Mutex` would. Maybe it's too much ceremony for what it's worth. What do you think?

itingliu: Do you mean a
```
class AlwaysSafeFormatter {
    private let uformatter: UnsafeMutablePointer<UNumberFormat?>
    private let formatLock: Mutex<Void>?
    internal func withLock(...) ... { }
}
```

And we use `AlwaysSafeFormatter.withLock` everywhere?

I can see that, but `ICULegacyNumberFormatter` is already a wrapper. It's also the only (hopefully) place where we optionally lock around an ICU object, so I think for now it's ok to leave it here. But if we ever encounter this pattern again (hopefully not) then I agree we could benefit from this wrapper

jmschonfeld: Yeah something like that, just to avoid accidentally forgetting to use the lock. Sounds good yeah I figured it might be a bit more architecture than it's worth if it's all isolated to a small space

jmschonfeld (approved):

PR #1890, opened 2026-04-10
Fix Observations of ProgressManager's fractionCompleted updates 

chloe-yeo:
This PR adds a `self.withMutation(keyPath: \.completedCount) {}` call in `markChildDirty(at:)` after `markSelfDirty(parents:)` to make sure updates of `fractionCompleted` gets propagated up to top-of-tree. 

### Motivation:

When leaf node has updates, only its direct parent has updates and the root node that is more than one level above the leaf node does not get updates until the very end when `fractionCompleted` is 1.0. 

`markChildDirty` propagates dirty flags up the tree when a child's progress changes, but it never called `withMutation` on intermediate nodes. This means `withObservationTracking`'s `onChange` only fires for the direct parent — ancestors further up the tree never get notified. Async stream patterns like `for await update in Observations { pm.fractionCompleted }` silently miss updates at every level above the direct parent.

### Modifications:

Added a `self.withMutation(keyPath: \.completedCount) {}` call in markChildDirty(at:)`.

### Result:

When leaf node has updates, the root node that is more than one level above the leaf node will also get the updates. 

### Testing:

Added a unit test that reproduced this issue and is resolved with the change.

### DIFF
--- a/Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
+++ b/Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
@@ -384,7 +384,9 @@ internal import _FoundationCollections
             state.markChildDirty(at: position)
         }
         if let parents = parents {
+            // rdar://172950622 (Async stream observation of ProgressManager's fractionCompleted does not receive updates from subsubprogress and beyond.)
             markSelfDirty(parents: parents)
+            self.withMutation(keyPath: \.completedCount) {}
         }
     }
     
--- a/Tests/FoundationEssentialsTests/ProgressManager/ProgressManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/ProgressManager/ProgressManagerTests.swift
@@ -10,6 +10,8 @@
 //
 //===----------------------------------------------------------------------===//
 import Testing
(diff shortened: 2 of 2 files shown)


chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
```diff
@@ -384,7 +384,9 @@ internal import _FoundationCollections
             state.markChildDirty(at: position)
         }
         if let parents = parents {
+            // rdar://172950622 (Async stream observation of ProgressManager's fractionCompleted does not receive updates from subsubprogress and beyond.)
```
itingliu, line 387: For next time I think I'd prefer losing the radar number. This comment doesn't add any value for those who can't access the rdar

itingliu (approved):

Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
chloe-yeo: Thanks, will keep that in mind next time!

PR #1891, opened 2026-04-10
NSString addingPercentEncoding methods should return nil for invalid UTF-16

jrflat:
The new Swift implementation for `-[NSString addingPercentEncoding...]` methods should return `nil` for `NSString`s with invalid (unpaired) UTF-16 surrogates instead of using the `U+FFFD` replacement character.

### Motivation:

Restore previous behavior of returning `nil` when an `NSString` contains an unpaired UTF-16 surrogate.

### Modifications:

The extensions now validate UTF-16 backed `NSString`s via `String(validating: ..., as: UTF16.self)` instead of just bridging to `String` (which replaces invalid UTF-16 with `U+FFFD` when accessing the bytes via `.withUTF8 { ... }`).

### Result:

No behavior change for swift-foundation. On Darwin, these `NSString` methods maintain their old behavior when the Swift implementation is enabled.

### Testing:

`NSString` testing of these methods for strings containing unpaired UTF-16 surrogates.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_C+Encoding.swift
+++ b/Sources/FoundationEssentials/URL/URL_C+Encoding.swift
@@ -228,14 +228,34 @@ extension NSURL {
     }
 }
 
+internal import Foundation_Private.NSString
+
 @objc
 extension NSString {
+    /// Returns `nil` if `self` contains unpaired UTF-16 surrogates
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1893, opened 2026-04-11
Implement Data.range(of:options:in:) for non-Darwin platforms

McNight:
Implement `Data.range(of:options:in:)` for non-Darwin platforms

### Motivation:

`Data.range(of:options:in:)` was only available when building with `FOUNDATION_FRAMEWORK`, leaving non-Darwin platforms without this API. This implements it in pure Swift using the existing `firstRange`/`lastRange` primitives on `DataProtocol`.

### Modifications:

* Provides a pure Swift implementation of `range(of:options:in:)` previously missing outside of FOUNDATION_FRAMEWORK, supporting the `.backwards` and `.anchored` search options.
* Re-enables associated tests previously gated behind FOUNDATION_FRAMEWORK.

### Result:

`Data.range(of:options:in:)` is now available on all platforms. The `SearchOptions` type definition remains platform-conditional (typealias to `NSData.SearchOptions` on Darwin, own `OptionSet` elsewhere), but the function implementation is shared across all platforms.

### Testing:

Re-enabled existing tests in `DataTests` and `PropertyListEncoderTests`  that were previously gated behind `FOUNDATION_FRAMEWORK`.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
+++ b/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
@@ -102,10 +102,8 @@ extension Data {
         into badCharacterShift: inout MutableSpan<Int>
     ) {
         if backwards {
-            var i = needle.count
-            while i > 0 {
-                i -= 1
-                // i is decremented from needle.count before indexing, so it is always in 0..<needle.count.
+            for i in (0..<needle.count).reversed() {
+                // i stays within 0..<needle.count, so safe to use the unchecked subscript.
                 badCharacterShift[Int(needle[unchecked: i])] = i
             }
         } else {
--- a/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
+++ b/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
@@ -1,0 +1,194 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2025 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
+extension Data {
+    func _searchBoyerMoore(_ needle: borrowing Span<UInt8>, in searchRange: Range<Index>, backwards: Bool) -> Range<Index>? {
+        let haystack = span.extracting(_rangeRelativeToStartIndex(searchRange))
+        let needleLength = needle.count
+
+        var badCharacterShift = ContiguousArray(repeating: needleLength, count: Int(UInt8.max) + 1)
+        var goodSubstringShift = ContiguousArray(repeating: needleLength, count: needleLength)
+        var suffixLengths = ContiguousArray(repeating: 0, count: needleLength)
+        var badCharacterShiftSpan = badCharacterShift.mutableSpan
+        var goodSubstringShiftSpan = goodSubstringShift.mutableSpan
+        var suffixLengthsSpan = suffixLengths.mutableSpan
+
+        Self._computeBadCharacterShift(for: needle, backwards: backwards, into: &badCharacterShiftSpan)
+        Self._computeGoodSubstringShift(
+            for: needle,
+            backwards: backwards,
+            shift: &goodSubstringShiftSpan,
+            suffixLengths: &suffixLengthsSpan
+        )
+
+        if backwards {
+            var scanNeedle = 0
+            var scanHaystack = haystack.count - needleLength
+
+            while scanHaystack >= 0, scanNeedle < needleLength {
+                if haystack[scanHaystack] == needle[scanNeedle] {
+                    scanHaystack += 1
+                    scanNeedle += 1
+                } else {
+                    let shift = Swift.max(
+                        badCharacterShift[Int(haystack[scanHaystack])],
+                        goodSubstringShift[scanNeedle]
+                    )
+                    scanHaystack -= shift
+                    scanNeedle = 0
+                }
+            }
+
+            guard scanNeedle == needleLength else {
+                return nil
+            }
+
+            let lowerBound = searchRange.lowerBound + scanHaystack - needleLength
+            return lowerBound..<(lowerBound + needleLength)
+        } else {
+            var scanNeedle = needleLength - 1
+            var scanHaystack = needleLength - 1
+
+            while scanHaystack < haystack.count, scanNeedle >= 0 {
+                if haystack[scanHaystack] == needle[scanNeedle] {
+                    scanHaystack -= 1
+                    scanNeedle -= 1
+                } else {
+                    let shift = Swift.max(
+                        badCharacterShift[Int(haystack[scanHaystack])],
+                        goodSubstringShift[scanNeedle]
+                    )
+                    scanHaystack += shift
+                    scanNeedle = needleLength - 1
+                }
+            }
+
+            guard scanNeedle < 0 else {
+                return nil
+            }
+
+            let lowerBound = searchRange.lowerBound + scanHaystack + 1
+            return lowerBound..<(lowerBound + needleLength)
+        }
+    }
+
+    @_lifetime(badCharacterShift: copy badCharacterShift)
+    private static func _computeBadCharacterShift(
+        for needle: borrowing Span<UInt8>,
+        backwards: Bool,
+        into badCharacterShift: inout MutableSpan<Int>
+    ) {
+        if backwards {
+            for i in (0..<needle.count).reversed() {
+                badCharacterShift[Int(needle[i])] = i
+            }
+        } else {
+            for i in 0..<needle.count {
+                badCharacterShift[Int(needle[i])] = needle.count - i - 1
+            }
+        }
+    }
+
+    @_lifetime(shift: copy shift)
+    @_lifetime(suffixLengths: copy suffixLengths)
+    private static func _computeGoodSubstringShift(
+        for needle: borrowing Span<UInt8>,
+        backwards: Bool,
+        shift: inout MutableSpan<Int>,
+        suffixLengths: inout MutableSpan<Int>
+    ) {
+        if backwards {
+            var reversedNeedle = ContiguousArray(repeating: UInt8.zero, count: needle.count)
+            var reversedNeedleSpan = reversedNeedle.mutableSpan
+            for i in 0..<needle.count {
+                reversedNeedleSpan[i] = needle[needle.count - 1 - i]
+            }
+
+            Self._computeGoodSubstringShift(
+                for: reversedNeedleSpan.span,
+                shift: &shift,
+                suffixLengths: &suffixLengths
+            )
+            Self._reverse(&shift)
+        } else {
+            Self._computeGoodSubstringShift(
+                for: needle,
+                shift: &shift,
+                suffixLengths: &suffixLengths
+            )
+        }
+    }
+
+    @_lifetime(shift: copy shift)
+    @_lifetime(suffixLengths: copy suffixLengths)
+    private static func _computeGoodSubstringShift(
+        for needle: borrowing Span<UInt8>,
+        shift: inout MutableSpan<Int>,
+        suffixLengths: inout MutableSpan<Int>
+    ) {
+        let needleLength = needle.count
+
+        var f = needleLength - 1
+        var g = needleLength - 1
+        suffixLengths[needleLength - 1] = needleLength
+
+        for i in (0..<(needleLength - 1)).reversed() {
+            if i > g, suffixLengths[i + needleLength - 1 - f] < i - g {
+                suffixLengths[i] = suffixLengths[i + needleLength - 1 - f]
+            } else {
+                if i < g {
+                    g = i
+                }
+                f = i
+                while g >= 0, needle[g] == needle[g + needleLength - 1 - f] {
+                    g -= 1
+                }
+                suffixLengths[i] = f - g
+            }
+        }
+
+        var j = 0
+        for i in (0..<needleLength).reversed() {
+            if suffixLengths[i] == i + 1 {
+                while j < needleLength - 1 - i {
+                    if shift[j] == needleLength {
+                        shift[j] = needleLength - 1 - i
+                    }
+                    j += 1
+                }
+            }
+        }
+
+        for i in 0..<(needleLength - 1) {
+            shift[needleLength - 1 - suffixLengths[i]] = needleLength - 1 - i
+        }
+
+        for i in 0..<(needleLength - 1) {
+            shift[i] += needleLength - 1 - i
+        }
+    }
+
+    @_lifetime(span: copy span)
+    private static func _reverse(_ span: inout MutableSpan<Int>) {
+        var lowerBound = 0
+        var upperBound = span.count - 1
+
+        while lowerBound < upperBound {
+            let tmp = span[lowerBound]
+            span[lowerBound] = span[upperBound]
+            span[upperBound] = tmp
+            lowerBound += 1
+            upperBound -= 1
+        }
+    }
+}
--- a/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
+++ b/Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
@@ -12,79 +12,87 @@
 
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 extension Data {
-    func _searchBoyerMoore(_ needle: borrowing Span<UInt8>, in searchRange: Range<Index>, backwards: Bool) -> Range<Index>? {
+    func _searchBoyerMoore(_ needle: Span<UInt8>, in searchRange: Range<Index>, backwards: Bool) -> Range<Index>? {
         let haystack = span.extracting(_rangeRelativeToStartIndex(searchRange))
         let needleLength = needle.count
 
-        var badCharacterShift = ContiguousArray(repeating: needleLength, count: Int(UInt8.max) + 1)
-        var goodSubstringShift = ContiguousArray(repeating: needleLength, count: needleLength)
-        var suffixLengths = ContiguousArray(repeating: 0, count: needleLength)
+        var badCharacterShift: InlineArray<256, Int> = .init(repeating: needleLength)
         var badCharacterShiftSpan = badCharacterShift.mutableSpan
-        var goodSubstringShiftSpan = goodSubstringShift.mutableSpan
-        var suffixLengthsSpan = suffixLengths.mutableSpan
 
         Self._computeBadCharacterShift(for: needle, backwards: backwards, into: &badCharacterShiftSpan)
-        Self._computeGoodSubstringShift(
-            for: needle,
-            backwards: backwards,
-            shift: &goodSubstringShiftSpan,
-            suffixLengths: &suffixLengthsSpan
-        )
 
-        if backwards {
-            var scanNeedle = 0
-            var scanHaystack = haystack.count - needleLength
+        return withUnsafeTemporaryAllocation(of: Int.self, capacity: needleLength) { goodSubstringShift in
+            goodSubstringShift.initialize(repeating: needleLength)
 
-            while scanHaystack >= 0, scanNeedle < needleLength {
-                if haystack[scanHaystack] == needle[scanNeedle] {
-                    scanHaystack += 1
-                    scanNeedle += 1
+            return withUnsafeTemporaryAllocation(of: Int.self, capacity: needleLength) { suffixLengths in
+                suffixLengths.initialize(repeating: 0)
+
+                var goodSubstringShiftSpan = goodSubstringShift.mutableSpan
+                var suffixLengthsSpan = suffixLengths.mutableSpan
+
+                Self._computeGoodSubstringShift(
+                    for: needle,
+                    backwards: backwards,
+                    shift: &goodSubstringShiftSpan,
+                    suffixLengths: &suffixLengthsSpan
+                )
+
+                if backwards {
+                    var scanNeedle = 0
+                    var scanHaystack = haystack.count - needleLength
+
+                    while scanHaystack >= 0, scanNeedle < needleLength {
+                        if haystack[scanHaystack] == needle[scanNeedle] {
+                            scanHaystack += 1
+                            scanNeedle += 1
+                        } else {
+                            let shift = Swift.max(
+                                badCharacterShift[Int(haystack[scanHaystack])],
+                                goodSubstringShift[scanNeedle]
+                            )
+                            scanHaystack -= shift
+                            scanNeedle = 0
+                        }
+                    }
+
+                    guard scanNeedle == needleLength else {
+                        return nil
+                    }
+
+                    let lowerBound = searchRange.lowerBound + scanHaystack - needleLength
+                    return lowerBound..<(lowerBound + needleLength)
                 } else {
-                    let shift = Swift.max(
-                        badCharacterShift[Int(haystack[scanHaystack])],
-                        goodSubstringShift[scanNeedle]
-                    )
-                    scanHaystack -= shift
-                    scanNeedle = 0
+                    var scanNeedle = needleLength - 1
+                    var scanHaystack = needleLength - 1
+
+                    while scanHaystack < haystack.count, scanNeedle >= 0 {
+                        if haystack[scanHaystack] == needle[scanNeedle] {
+                            scanHaystack -= 1
+                            scanNeedle -= 1
+                        } else {
+                            let shift = Swift.max(
+                                badCharacterShift[Int(haystack[scanHaystack])],
+                                goodSubstringShift[scanNeedle]
+                            )
+                            scanHaystack += shift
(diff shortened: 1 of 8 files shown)


parkera:
Ultimately this boils down to `_CFDataFindBytes` (https://github.com/swiftlang/swift-corelibs-foundation/blob/48fc340a66b8cd5605ba764ecaae4b27a1af23b5/Sources/CoreFoundation/CFData.c#L830), so if we have replicated that behavior then we should be good to go.

parkera:
A quick look suggests that we may have a significant algorithmic performance reduction by going with this approach.

jmschonfeld:
Yeah I suspect there are a few performance aspects that we'd want to address here in the Swift implementation to maintain parity with the original C version:

- Early exits: when we can exit early based on analysis of the (constant time) lengths, we don't need to perform a scan at all
- Anchored searches: the current swift implementation performs a full search even when anchored is passed and throws away the result when not anchored. Ideally, we'd only perform a single linear search from the front or back when anchored is specified instead of continuing to look for non-anchored results
- Search algorithm: The C implementation uses the Boyer Moore algorithm, but this Swift implementation uses a simpler repeated linear search. I suspect the Swift implementation will have a higher runtime cost than the Boyer Moore algorithm.
- Collection-based APIs: the current implementation relies on `Data`'s `Collection` conformance which is likely going to be quite slow compared to just accessing the underlying pointer directly. It might be faster to implement it on top of the `Span` or `RawSpan` instead of the collection to avoid accidentally going through `Data`'s complex ABI layer on each access.

@McNight do you think it'd be possible to update your Swift implementation to bring over some of the performance characteristics from the C implementation? Perhaps we could even improve `firstRange`/`lastRange`/similar APIs as a followup if we can implement them in a similar manner but generically over `DataProtocol`

kperryua:
FWIW, the `_StringProcessing` module has a Boyer-Moore-Horspool implementation specific to Strings that it uses in its implementation of `Collection.ranges(of:)`, but IMO it should be relatively simple to generalize this for `Collection<UInt8>`.

https://github.com/swiftlang/swift-experimental-string-processing/blob/e3f7b258f75a42b78624972131a0175cb0b662da/Sources/_StringProcessing/Algorithms/Algorithms/SubstringSearcher.swift#L12

McNight:
@jmschonfeld Definitely, thank you all for the pointers. I'll get back to you asap when I have something!

McNight:
Sorry for the delay!
I changed the implementation and added the Boyer-Moore search path, based closely on the C implementation from `_CFDataFindBytes` and `__CFDataSearchBoyerMoore`.
I also added tests covering the main forward/backward, overlapping, and anchored cases. Test coverage could still be expanded with a few more edge cases, such as empty and single-byte inputs, as well as larger payload scenarios.
Finally, I ran the relevant tests on Linux using the `swiftlang/swift:nightly-main-noble` image and verified they passed there.

Sources/FoundationEssentials/Span+Utils.swift
```diff
@@ -0,0 +1,95 @@
...
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Span<UInt8> {
+    @inline(__always)
```
jmschonfeld: For these types of functions, we don't have a great rule of thumb for when to use `@inline(__always)`, but my personal rule has been to add it if it's very trivial and not inlining would always be a mistake, and otherwise to let the compiler choose. Here, the function is a little more than trivial and the existing `lastIndex(of:)` in `Span+Path.swift` didn't have this attribute, did you find a need to add this here as you benchmarked your change or was this added more preemptively?

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+@available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
+extension Data {
+    func _searchBoyerMoore(_ needle: borrowing Span<UInt8>, in searchRange: Range<Index>, backwards: Bool) -> Range<Index>? {
```
jmschonfeld: nit: no need to use `borrowing` here since `Span<UInt8>` isn't a `~Copyable` type. Same goes for the other `borrowing Span<...>`s below
```suggestion
    func _searchBoyerMoore(_ needle: Span<UInt8>, in searchRange: Range<Index>, backwards: Bool) -> Range<Index>? {
```

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+        let haystack = span.extracting(_rangeRelativeToStartIndex(searchRange))
+        let needleLength = needle.count
+
+        var badCharacterShift = ContiguousArray(repeating: needleLength, count: Int(UInt8.max) + 1)
+        var goodSubstringShift = ContiguousArray(repeating: needleLength, count: needleLength)
+        var suffixLengths = ContiguousArray(repeating: 0, count: needleLength)
```
jmschonfeld: Instead of `ContiguousArray`, it might be great to use a temporary allocation via `withUnsafeTemporaryAllocation` or an `InlineArray<UInt8>` for static sizes here so that it gets put on the stack instead of the heap. Do you think that would be feasible? Let me know if you need any assistance figuring out how to do that.

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+    @_lifetime(shift: copy shift)
+    @_lifetime(suffixLengths: copy suffixLengths)
+    private static func _computeGoodSubstringShift(
+        for needle: borrowing Span<UInt8>,
+        backwards: Bool,
+        shift: inout MutableSpan<Int>,
+        suffixLengths: inout MutableSpan<Int>
+    ) {
+        if backwards {
+            var reversedNeedle = ContiguousArray(repeating: UInt8.zero, count: needle.count)
```
jmschonfeld: This is another good spot for `withUnsafeTemporaryAllocation` so that if needle is small it can go on the stack.

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+            var reversedNeedleSpan = reversedNeedle.mutableSpan
+            for i in 0..<needle.count {
+                reversedNeedleSpan[i] = needle[needle.count - 1 - i]
+            }
```
jmschonfeld: Is it worth using `Self._reversed(&reversedNeedleSpan)` here?

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+                    j += 1
+                }
+            }
+        }
+
+        for i in 0..<(needleLength - 1) {
+            shift[needleLength - 1 - suffixLengths[i]] = needleLength - 1 - i
+        }
+
+        for i in 0..<(needleLength - 1) {
```
jmschonfeld, line 209: Should we carry over the comments from the C implementation here?

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -0,0 +1,194 @@
...
+    @_lifetime(suffixLengths: copy suffixLengths)
+    private static func _computeGoodSubstringShift(
+        for needle: borrowing Span<UInt8>,
+        shift: inout MutableSpan<Int>,
+        suffixLengths: inout MutableSpan<Int>
+    ) {
+        let needleLength = needle.count
+
+        var f = needleLength - 1
+        var g = needleLength - 1
```
jmschonfeld: I know the C implementation uses these variable names so this isn't a requirement, but should we change them to something more meaningful? I'm not certain if the algorithm defines these contents (and these letters are "terms of art" for the algorithm) or if the previous C variable names were just convenient short names in C that we can make more expressive in Swift

Sources/FoundationEssentials/Data/Data+Searching.swift
```diff
@@ -54,4 +29,62 @@ extension Data {
...
+        if isAnchored, searchRangeLength > needleLength {
+            if searchBackwards {
+                searchRange = (searchRange.upperBound - needleLength)..<searchRange.upperBound
+            } else {
+                searchRange = searchRange.lowerBound..<(searchRange.lowerBound + needleLength)
+            }
+        }
+
+        if searchRange.count == needleLength {
+            return span.extracting(_rangeRelativeToStartIndex(searchRange)).elementsEqual(dataToFind.span) ? searchRange : nil
```
jmschonfeld, line 68: Out of curiosity, do you know where this `Span.elementsEqual` function comes from? I didn't think we had that on `Span` (yet)

jmschonfeld (commented):
Thank you so much for putting in the time and effort to rewrite this algorithm in Swift. I reviewed it and validated that it is the same setup/behavior as the C algorithm so I expect it should behave the same. I have a handful of smaller comments, mostly around performance, but overall this is looking really great.

I'll go ahead and start CI on this just so we can validate the tests pass across all platforms, but if any of the comments don't make sense or you feel are beyond the scope of what you can do with this PR let me know and we can either iterate together or find a way to land this as-is and do those as followups.

Out of curiosity, have you done any performance benchmarking of this as compared to the implementation in swift-corelibs-foundation or the implementation from Foundation in the SDK on Darwin today? If not I can help run some of those numbers to see if this matches up (just to make sure we don't regress the performance)

Sources/FoundationEssentials/Data/Data+Searching.swift
jmschonfeld: Oh sorry I just saw that you added this, makes sense!

Sources/FoundationEssentials/Span+Utils.swift
McNight: I think I got a little carried away! Indeed your strategy makes sense, I removed the `@inline(__always)` annotations and kept only the ones that were here before.

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
McNight: Oops, indeed, thx

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
McNight: Done! Tell me what you think

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
McNight: Actually I leveraged the `initializeElement` method on the unsafe mutable buffer pointer

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
McNight: I was thinking about it, I think it makes sense! I brought all the comments, tell me if it's okay or maybe too much...

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
McNight: I renamed a few variables, hope it's clearer

McNight:
I've just done a few tests on my own, but I haven't compared against anything yet... but yes please, if you could show me how to set this up, I'd really appreciate it.

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
jmschonfeld: Looks great, thanks!

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
jmschonfeld: I think it's ok, we can always come back and update it as a followup but carrying over the comments as they are now seems fine to me

jmschonfeld:
Thanks for the updates! The code looks good to me, we'll just want to benchmark it against the existing implementation to make sure the performance is equivalent (or maybe better!). We can do so by adding a benchmark or two to `Benchmarks/Benchmarks/Data/BenchmarkData.swift` in this repo. That `Benchmarks` subdirectory is a separate package that builds benchmarks against various configurations (against a Linux toolchain, against Foundation in the macOS SDK, against the local package, etc.). There's a bit of explanation about this at https://github.com/swiftlang/swift-foundation/blob/main/Foundation_Build_Process.md - I don't remember the right commands to get the configuration you want here offhand, but feel free to take a stab at it or tomorrow I'll see if I can benchmark this on my end. We'll just need to make sure we benchmark your implementation and don't accidentally call the one in Foundation in swift-corelibs-foundation today while testing the new one 🙂

jmschonfeld:
@swift-ci please test

Harry-KNIGHT:
￼￼￼￼

jmschonfeld:
I added the following to the benchmarks:

```swift
    Benchmark("DataRangeOf", closure: { _, box in
        blackHole(box.d1.range(of: box.d2))
    }, setup: { () -> TwoDatasBox in
        var d2 = createSomeData(128, range: 0 ..< 100)
        d2.append(200)
        return TwoDatasBox(d1: createSomeData(1024, range: 0 ..< 100), d2: d2)
    })

    Benchmark("DataRangeOfBackwards", closure: { _, box in
        blackHole(box.d1.range(of: box.d2, options: .backwards))
    }, setup: { () -> TwoDatasBox in
        var d2 = createSomeData(128, range: 0 ..< 100)
        d2.append(200)
        return TwoDatasBox(d1: createSomeData(1024, range: 0 ..< 100), d2: d2)
    })

    Benchmark("DataRangeOfAnchored", closure: { _, box in
        blackHole(box.d1.range(of: box.d2, options: .anchored))
    }, setup: { () -> TwoDatasBox in
        var d2 = createSomeData(128, range: 0 ..< 100)
        d2.append(200)
        return TwoDatasBox(d1: createSomeData(1024, range: 0 ..< 100), d2: d2)
    })
```

and was able to run the benchmarks using:

```
# Record existing baseline against Foundation.framework in the SDK
swift package --build-system native --allow-writing-to-package-directory benchmark baseline update existing --format mark--target DataBenchmarks --filter "DataRangeOf.*"

# Record new performance with local swift-foundation
SWIFTCI_USE_LOCAL_DEPS=/path/to/repos swift package --build-system native --allow-writing-to-package-directory benchmark baseline update new --format mark--target DataBenchmarks --filter "DataRangeOf.*"

# Compare results
SWIFTCI_USE_LOCAL_DEPS=/path/to/repos swift package --build-system native --allow-writing-to-package-directory benchmark baseline compare existing new --format mark--target DataBenchmarks --filter "DataRangeOf.*"
```

the results show that an anchored search is significantly faster (by a few hundred percent) but that a non-anchored forward and reverse search both regressed by about 20%.

<details>
<summary>Full Benchmark Results</summary>

```
DataBenchmarks
============================================================================================================================

----------------------------------------------------------------------------------------------------------------------------
DataRangeOf metrics
----------------------------------------------------------------------------------------------------------------------------

│         Time (wall clock) (ns) *         │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         1 │         1 │         1 │         1 │         1 │         1 │        48 │    928168 │
│                   new                    │         1 │         1 │         1 │         1 │         2 │         2 │       104 │    836298 │
│                    Δ                     │         0 │         0 │         0 │         0 │         1 │         1 │        56 │    -91870 │
│              Improvement %               │         0 │         0 │         0 │         0 │      -100 │      -100 │      -117 │    -91870 │

│         Time (total CPU) (ns) *          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         2 │         2 │         2 │         2 │         2 │         2 │        45 │    928168 │
│                   new                    │         2 │         2 │         2 │         2 │         2 │         3 │        41 │    836298 │
│                    Δ                     │         0 │         0 │         0 │         0 │         0 │         1 │        -4 │    -91870 │
│              Improvement %               │         0 │         0 │         0 │         0 │         0 │       -50 │         9 │    -91870 │

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │      1264 │      1000 │       960 │       889 │       857 │       774 │        21 │    928168 │
│                   new                    │      1044 │       751 │       707 │       686 │       667 │       534 │        10 │    836298 │
│                    Δ                     │      -220 │      -249 │      -253 │      -203 │      -190 │      -240 │       -11 │    -91870 │
│              Improvement %               │       -17 │       -25 │       -26 │       -23 │       -22 │       -31 │       -52 │    -91870 │

----------------------------------------------------------------------------------------------------------------------------
DataRangeOfAnchored metrics
----------------------------------------------------------------------------------------------------------------------------

│         Time (wall clock) (ns) *         │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         0 │         1 │         1 │         1 │         1 │         1 │        32 │   1073756 │
│                   new                    │         0 │         0 │         0 │         0 │         0 │         0 │        15 │   1345122 │
│                    Δ                     │         0 │        -1 │        -1 │        -1 │        -1 │        -1 │       -17 │    271366 │
│              Improvement %               │         0 │         0 │         0 │         0 │         0 │         0 │        53 │    271366 │

│         Time (total CPU) (ns) *          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         1 │         1 │         1 │         2 │         2 │         2 │        33 │   1073756 │
│                   new                    │         1 │         1 │         1 │         1 │         1 │         1 │        43 │   1345122 │
│                    Δ                     │         0 │         0 │         0 │        -1 │        -1 │        -1 │        10 │    271366 │
│              Improvement %               │         0 │         0 │         0 │        50 │        50 │        50 │       -30 │    271366 │

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │      2000 │      1601 │      1601 │      1502 │      1500 │      1334 │        31 │   1073756 │
│                   new                    │     12048 │      8004 │      8004 │      6025 │      5988 │      5988 │        67 │   1345122 │
│                    Δ                     │     10048 │      6403 │      6403 │      4523 │      4488 │      4654 │        36 │    271366 │
│              Improvement %               │       502 │       400 │       400 │       301 │       299 │       349 │       116 │    271366 │

----------------------------------------------------------------------------------------------------------------------------
DataRangeOfBackwards metrics
----------------------------------------------------------------------------------------------------------------------------

│         Time (wall clock) (ns) *         │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         1 │         1 │         1 │         1 │         1 │         1 │        28 │    860935 │
│                   new                    │         2 │         2 │         2 │         2 │         2 │         2 │        21 │    779575 │
│                    Δ                     │         1 │         1 │         1 │         1 │         1 │         1 │        -7 │    -81360 │
│              Improvement %               │      -100 │      -100 │      -100 │      -100 │      -100 │      -100 │        25 │    -81360 │

│         Time (total CPU) (ns) *          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │         2 │         2 │         2 │         2 │         2 │         3 │        20 │    860935 │
│                   new                    │         2 │         2 │         2 │         3 │         3 │         3 │        19 │    779575 │
│                    Δ                     │         0 │         0 │         0 │         1 │         1 │         0 │        -1 │    -81360 │
│              Improvement %               │         0 │         0 │         0 │       -50 │       -50 │         0 │         5 │    -81360 │

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                 existing                 │       858 │       775 │       774 │       774 │       751 │       706 │        35 │    860935 │
│                   new                    │       667 │       632 │       632 │       615 │       615 │       490 │        47 │    779575 │
│                    Δ                     │      -191 │      -143 │      -142 │      -159 │      -136 │      -216 │        12 │    -81360 │
│              Improvement %               │       -22 │       -18 │       -18 │       -21 │       -18 │       -31 │        34 │    -81360 │
```
</details>

So it looks like the new Swift implementation is slightly slower. There might be something worth investigating there, my suspicion is that iterating the spans is slower than the C arrays in C because bounds checking isn't eliminated. I haven't profiled it yet to confirm, but it seems possible that with the use of `.reversed()` on ranges the compiler can't prove that it doesn't need to bounds check

McNight:
Ok so I finally managed to run the benchmarks (my main dev machine is still on Sequoia so I had to use another machine of mine running Tahoe) and indeed I'm seeing similar results. I'm going to see what we can do about the `.reversed()` and Span usage. If you have more insights and things we can try, let me know!

EDIT: I reran the benchmarks and, after using `unchecked` subscripts on `Span`/`MutableSpan`, saw a ~14% speedup in wall/CPU time for `DataRangeOfBackwards` and a ~24% speedup for `DataRangeOf` on an M1 machine. I'll push something asap 👍

jmschonfeld:
> EDIT: I reran the benchmarks and, after using `unchecked` subscripts on `Span`/`MutableSpan`, saw a ~14% speedup in wall/CPU time for `DataRangeOfBackwards` and a ~24% speedup for `DataRangeOf` on an M1 machine. I'll push something asap 👍

Yep, I saw the same thing. I suspect that we'll need to use `unchecked:` subscripts since the compiler indeed can't eliminate bounds checking when iterating in reverse. We can do that, but we should add clear comments to each place we do that explaining why the bounds checking isn't necessary (so that ideally we can remove them in the future when the optimizer becomes smarter). If you want to take a crack at that, that would be great! Thanks!

McNight:
I initially tried replacing most of the hot-path indexing with `unchecked:` subscripts, but that felt broader than what you were suggesting.
This version keeps the unchecked accesses scoped to the reverse/descending indexing paths where the bounds checks seemed most relevant in the benchmarks. Does that match what you had in mind, or would you prefer this to be narrower/broader?
I also added comments documenting the invariants for each unchecked region. Let me know if they add too much noise, we can trim them down.

Sources/FoundationEssentials/Data/Data+Searching+BoyerMoore.swift
```diff
@@ -97,8 +102,11 @@ extension Data {
         into badCharacterShift: inout MutableSpan<Int>
     ) {
         if backwards {
-            for i in (0..<needle.count).reversed() {
-                badCharacterShift[Int(needle[i])] = i
+            var i = needle.count
+            while i > 0 {
+                i -= 1
+                // i is decremented from needle.count before indexing, so it is always in 0..<needle.count.
```
jmschonfeld: If you want I think it's reasonable to keep this with `(0..<needle.count).reversed()` with the comment here and using the `unchecked:` subscript. IMO (with the comment making it explicit) it's clear that reversing a range doesn't change the fact that the index is still within the bounds (even though the compiler doesn't see that)

jmschonfeld:
> This version keeps the unchecked accesses scoped to the reverse/descending indexing paths where the bounds checks seemed most relevant in the benchmarks. Does that match what you had in mind, or would you prefer this to be narrower/broader?

Yep that matches what I was thinking of! I had one comment that you can keep the use of `.reversed()` that you had before, just now with the comment + unchecked subscript if you want - does the new structure read cleaner or to you or do you prefer the `.reversed()` syntax?

McNight:
No actually I preferred the fancy `.reversed()`! It matches with the rest of the function.

jmschonfeld:
> No actually I preferred the fancy `.reversed()`! It matches with the rest of the function.

Cool - it looks like there's actually 2 other cases where you used to use `.reversed()` but now don't, did you want to change those back as well? If so once you do that I'll kick off CI here as well as some other full toolchain tests to make sure that this doesn't break any downstream tests + make sure it's ok to land this now and remove the one from swift-corelibs-foundation afterwards

McNight:
Oops, indeed, sorry for the back and forth...

jmschonfeld:
@swift-ci please test

jmschonfeld:
Full toolchain tests running at https://github.com/swiftlang/swift-corelibs-foundation/pull/5441

jmschonfeld:
@swift-ci please test

jmschonfeld (approved):

jmschonfeld:
@McNight the full toolchain tests passed as did CI on this repo, so now that perf is on-par or better than the C implementation I've merged your implementation! The next step is that we can remove the implementation from swift-corelibs-foundation at https://github.com/swiftlang/swift-corelibs-foundation/blob/3feb2ee8429971ec4499af28b0ef96cd5c596749/Sources/Foundation/NSData.swift#L1265. Did you wanna post a PR there to remove that (or let me know if not and I can help handle that)?

McNight:
Amazing, thanks for your help :) I'll handle it!

McNight:
Here it is: https://github.com/swiftlang/swift-corelibs-foundation/pull/5468

PR #1895, opened 2026-04-13
Do not attempt to use `FICLONE` on Linux+Musl.

grynspan:
The header that defines `FICLONE`, `<linux/fs.h>`, is not available when building the static SDK on Linux with Musl, so don't try to use it there.


grynspan:
@swift-ci test

jmschonfeld (approved):

PR #1898, opened 2026-04-14
Move 0037 to Review

itingliu:
Update status of Proposal 0037


itingliu:
@swift-ci please test

PR #1901, opened 2026-04-14
[6.3] Avoid corruption when downsizing Data via replaceSubrange

### LINKED ISSUE #1900: Data(capacity: ...) causes corruption of data for sizes bigger than 32KB
### Description

When you attempt to use `Data(capacity: ...)` to preallocate at least 32KB of data, you may see data corruption in the payload itself. Using regular `Data()` and appending the same amount of data works as expected.

### Reproduction

Here is an example:

```swift
import Foundation

@main
struct SwiftWASM {
    static func main() {
        print("Allocating data...")
        let capacity = 32*1024
        var data = Data(capacity: capacity)
        for i in 0..<capacity {
            data.append(UInt8(i % 256))
        }

        print("Verifying data...")
        for i in 0..<capacity {
            if data[i] != UInt8(i % 256) {
                fatalError("Data verification failed at index \(i), expected \(UInt8(i % 256)), got \(data[i])")
                
            }
        }

        print("Data verification succeeded")
    }
}
```

You can also use this [link](https://pad.swiftwasm.org/?code=import%20Foundation%0A%0Aprint(%22Allocating%20data...%22)%0Alet%20capacity%20%3D%2032*1024%0Avar%20data%20%3D%20Data(capacity%3A%20capacity)%0Afor%20i%20in%200..%3Ccapacity%20%7B%0A%20%20data.append(UInt8(i%20%25%20256))%0A%7D%0A%0Aprint(%22Verifying%20data...%22)%0Afor%20i%20in%200..%3Ccapacity%20%7B%0A%20%20if%20data%5Bi%5D%20!%3D%20UInt8(i%20%25%20256)%20%7B%0A%20%20%20%20fatalError(%22Data%20verification%20failed%20at%20index%20%5C(i)%2C%20expected%20%5C(UInt8(i%20%25%20256))%2C%20got%20%5C(data%5Bi%5D)%22)%0A%20%20%7D%0A%7D%0Aprint(%22Data%20verification%20succeeded%22)%0A)

### Expected behavior

`Data(capacity: ...)` should not cause memory corruption.

### Environment

swift-6.3-RELEASE_wasm

### Additional information

_No response_

jmschonfeld:
Avoids memory corruption when a `Data` is downsized from a large slice to an inline representation

### Motivation:

When downsizing a `Data` from a large slice to an inline slice, the implementation copies over 3 chunks of bytes:

- The bytes prior to the replacement
- The bytes of the replacement
- The bytes after the replacement

Due to some incorrect indexing math, the third segment above is written over the replaced bytes leading to data corruption.

### Modifications:

Corrected the index arithmetic of these buffer copies

### Result:

Data is copied correctly and no longer corrupted

Resolves #1900 

### Testing:

Additional unit test added

### Future Directions:

In the future, we can investigate updating the `Data` implementation to avoid downsizing entirely (like other collection types) so that manually set capacities are not thrown away by RRC calls.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   tests:
     name: Test (SwiftPM)
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.10
     with:
       linux_swift_versions: '["nightly-6.3"]'
       windows_swift_versions: '["nightly-6.3"]'
@@ -44,7 +44,7 @@ jobs:
 
   soundness:
     name: Soundness
-    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@main
+    uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.10
     with:
       license_header_check_project_name: "Swift.org"
       license_header_check_enabled: false
--- a/Package.swift
+++ b/Package.swift
@@ -69,7 +69,7 @@ if let useLocalDepsEnv = Context.environment["SWIFTCI_USE_LOCAL_DEPS"], !useLoca
         [
             .package(
                 url: "https://github.com/apple/swift-collections",
-                from: "1.1.0"),
+                exact: "1.1.6"),
             .package(
                 url: "https://github.com/apple/swift-foundation-icu",
                 branch: "main"),
--- a/Sources/FoundationEssentials/Data/Data.swift
+++ b/Sources/FoundationEssentials/Data/Data.swift
@@ -1734,8 +1734,8 @@ public struct Data : Equatable, Hashable, RandomAccessCollection, MutableCollect
                             if subrange.lowerBound > 0 {
                                 inlineBuffer.baseAddress?.copyMemory(from: buffer.baseAddress!, byteCount: subrange.lowerBound)
                             }
-                            if subrange.upperBound < resultingUpper {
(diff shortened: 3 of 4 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

justice-adams-apple:
@swift-ci please test linux

.github/workflows/pull_request.yml
```diff
@@ -13,7 +13,7 @@ on:
 jobs:
   tests:
     name: Test (SwiftPM)
-    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@main
```
itingliu, line 16: Is this change intentional as part of this PR?

Package.swift
```diff
@@ -69,7 +69,7 @@ if let useLocalDepsEnv = Context.environment["SWIFTCI_USE_LOCAL_DEPS"], !useLoca
         [
             .package(
                 url: "https://github.com/apple/swift-collections",
-                from: "1.1.0"),
```
itingliu, line 72: same here, though I think we do want to make this change

.github/workflows/pull_request.yml
jmschonfeld: Yes, `main` already has this change and CI on the `release/6.3` release branch fails without this change (due to bugs in the `main` branch of the `swiftlang/github-workflows` repo. I can put this change in a separate PR if you want, but figured since the cadence of 6.3 changes is slow, including it in this PR was simpler

Package.swift
jmschonfeld: Same comment as above, CI fails without this and it's a change we want to make because it matches how the full toolchain builds, so I chose to just include it in this PR for simplicity but can make a separate PR if you prefer

itingliu (approved):

Package.swift
itingliu: Not a problem here. The change looks good to me too

PR #1905, opened 2026-04-14
Mark legacy ABI as obsoleted rather than as spi

glessard:
Addresses rdar://174193354

Using the `@_spi` annotation causes validation issues with the Foundation framework. We can use `obsoleted` annotations instead, at the cost of less straightforward testing. I took advantage of the process of figuring out these tests to round out the legacy ABI tests.


jmschonfeld (approved):

glessard:
@swift-ci please test

itingliu:
@swift-ci please test

PR #1907, opened 2026-04-16
Fix `ProgressManager` Bugs 

chloe-yeo:
This PR fixes two bugs in `ProgressManager`. 

### Motivation:

**Bug A: `ProgressManager` observation consumed by addChild before child `ProgressManager` completes**
`addChild` fired `withMutation(keyPath: \.completedCount)` even though `completedCount` has not changed. This causes the one-shot `withObservationTracking` handlers to be consumed before child `ProgressManager`  reports any progress, so the actual `isFinished` transition, which depends on the change in either `completedCount` or `totalCount`, never got observed. 

Additionally, `markChildDirty` gated `withMutation(keyPath: \.completedCount)` behind the dirty-flag check. This flag is only cleared when `fractionCompleted` or `isFinished` is read. If they are not read between `start` and `complete`, `markChildDirty` skips the `withMutation` call and observation never fires. 

**Bug B: `NSProgress` parent does not receive updates from `ProgressManager` grandchildren** 
When a `Subprogress` is created from `NSProgress` via `Progress.subprogress(assigningCount:)`, and the method consuming this `Subprogress` creates more children which are grandchildren of `NSProgress`, updates from the grandchildren do not propagate up to `NSProgress`.

The interop bridge worked for two-levels, when `complete(count:)` is called on the `ProgressManager` that holds the `SubprogressBridge`, `complete(by:)` notifies the bridge observer via `notifyObservers`, which allows `NSProgress` to receive updates. 
However, when updates are from a child via `markChildDirty`, the code path never called `notifyObservers` on the `SubprogressBridge`, so the `NSProgress` did not receive the updates.

### Modifications:

**Bug A: ProgressManager observation consumed by addChild before child `ProgressManager` completes**
Remove `withMutation` call from `addChild`. Move `withMutation` call in `markChildDirty` outside dirty-flag guard so that it fires unconditionally. 

**Bug B: NSProgress parent does not receive updates from ProgressManager grandchildren** 
Add `notifyInteropObserversOfChildUpdates()` helper method, and call it within `markChildDirty`, which resolves dirty children for interop cases and notifies `SubprogressBridge`. `SubprogressBridge` also sets the `NSProgress` `totalUnitCount` and `completedUnitCount` atomically to avoid a limitation that we hit when `NSProgress` denominator changes from a larger number to a much smaller number.

### Result:

`withObservationTracking` on `isFinished` correctly fires only when values actually change, not on `addChild`.

Updates from descendants of a root `ProgressManager` that are below its direct children are correctly propagated through `SubprogressBridge` to `NSProgress` parents. 

### Testing:

Added unit tests to validate behavior changes.

### DIFF
--- a/Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
+++ b/Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
@@ -387,6 +387,49 @@ internal import _FoundationCollections
             markSelfDirty(parents: parents)
         }
         self.withMutation(keyPath: \.completedCount) {}
+        notifyInteropObserversOfChildUpdate()
+    }
+
+    /// Notifies interop bridge observers so that grandchild progress propagates
+    /// through the SubprogressBridge to the parent NSProgress.
+    private func notifyInteropObserversOfChildUpdate() {
+#if FOUNDATION_FRAMEWORK
+        // Phase 1: Collect bridge manager and dirty children info
+        let info: (bridgeManager: ProgressManager, updates: [PendingChildUpdateInfo]?)? = state.withLock { state in
+            guard case .interopObservation(let observation) = state.interopType,
+                  let bridgeManager = observation.subprogressBridge?.manager else {
+                return nil
+            }
+            return (bridgeManager, state.pendingChildrenUpdates())
+        }
+
+        guard let info else { return }
+
+        // Phase 2: Resolve each dirty child's fraction
+        var childrenUpdates: [PendingChildUpdate] = []
+        if let updates = info.updates {
+            for update in updates {
+                let updatedFraction = update.manager.updatedProgressFraction()
+                childrenUpdates.append(PendingChildUpdate(
+                    index: update.index,
+                    updatedFraction: updatedFraction,
+                    assignedCount: update.assignedCount
+                ))
+            }
+        }
+
+        // Phase 3: Apply updates and compute fraction
+        let observerState = state.withLock { state in
+            if !childrenUpdates.isEmpty {
+                state.updateChildrenProgressFraction(updates: childrenUpdates)
+            }
+            let fraction = state.overallFraction
+            return ObserverState(totalCount: fraction.total ?? 0, completedCount: fraction.completed)
+        }
+
+        // Notify bridge
+        info.bridgeManager.notifyObservers(with: observerState)
+#endif
     }
     
     internal func updatedProgressFraction() -> ProgressFraction {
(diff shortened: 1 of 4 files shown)


chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
```diff
@@ -384,10 +384,50 @@ internal import _FoundationCollections
             state.markChildDirty(at: position)
         }
         if let parents = parents {
-            // rdar://172950622 (Async stream observation of ProgressManager's fractionCompleted does not receive updates from subsubprogress and beyond.)
             markSelfDirty(parents: parents)
-            self.withMutation(keyPath: \.completedCount) {}
         }
+        self.withMutation(keyPath: \.completedCount) {}
+        notifyInteropObserversOfChildUpdate()
```
itingliu, line 391: Can we guard this call with `FOUNDATION_FRAMEWORK` too? This "interop" concept is only relevant to `FOUNDATION_FRAMEWORK`, which you also agreed since the entire body of `notifyInteropObserversOfChildUpdate` is behind `FOUNDATION_FRAMEWORK`. 
```suggestion
#if FOUNDATION_FRAMEWORK
        notifyInteropObserversOfChildUpdate()
#endif
```

chloe-yeo: Yes good catch, fixed, thanks!

Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
```diff
@@ -384,10 +384,52 @@ internal import _FoundationCollections
...
         }
+        self.withMutation(keyPath: \.completedCount) {}
+        #if FOUNDATION_FRAMEWORK
+        notifyInteropObserversOfChildUpdate()
+        #endif
+    }
+
+    /// Notifies interop bridge observers so that grandchild progress propagates
+    /// through the SubprogressBridge to the parent NSProgress.
+    private func notifyInteropObserversOfChildUpdate() {
```
itingliu: Sorry, I forgot to say that this whole function should be behind `#if FOUNDATION_FRAMEWORK` too

chloe-yeo: Oh yeah thanks!

chloe-yeo:
@swift-ci please test

Sources/FoundationEssentials/ProgressManager/ProgressManager.swift
chloe-yeo: Moved this to the file `ProgressManager + Interop` where all the interop methods live!

itingliu (approved):

PR #1912, opened 2026-04-16
new-codable enum macro implementation

kperryua:
Adds *Codable macro implementation for enums

### Motivation:

The initial version of the *Codable macros only worked for structs. Enums have different structure and a different API surface to call into for encoding and decoding them.

### Modifications:

* More refactoring to support generating code for both structs and enums
* Enum generation implemented

### Result:

* Better factored macro code, for potential future *Codable macro library
* Support for enum cases with and without associated values
* Support for labelled and unlabelled associated values
* Support for CodingKey and DecodableAlias for enum case names

### Testing:

* New macro expansion tests
* New integration tests


kperryua:
@t089, I've rebased this onto your access modifier change (#1916) and bundled it into the `CodableExpansion` types as an enum. Please take a look!

t089:
@kperryua Thanks for moving along with this - looks great.

PR #1915, opened 2026-04-16
Data Collection micro optimizations

jmschonfeld:
Adds some micro optimizations to `Data`'s `Collection` APIs to achieve much smaller instruction counts for simple operations and higher level optimizations.

### Motivation:

In general, when accessing multiple bytes from a `Data`, we expect clients to get the span and read from that (which has low overhead). However, as span APIs are relatively new, there's still code that is in use that repeatedly accesses the subscript of a `Data` repeatedly to read multiple bytes. By adding some micro optimizations to `Data`'s basic `Collection` APIs, we can enable some higher level optimizations by the compiler to significantly reduce the overhead of this operation.

### Modifications:

- Remove overflow and optional checking from `__DataStorage.get`/`__DataStorage.set` (index validation is already performed at a higher level
- Remove overflow checking from `Data.count` (via its representations) since `Range` guarantees that the upper bound is greater than the lower bound
- Add `_assumeNonNegative` (already in use by many stdlib types for these APIs) to guarantee that Data's indices and count are always non-negative

### Result:

Many overflow / bounds checking instructions are no longer present in these APIs. Additionally, higher level optimizations are now enabled. For example, this code:

```swift
extension Data {
    func loadNetworkByteOrderUInt32() -> UInt32 {
        precondition(self.count >= MemoryLayout<UInt32>.size)
        return (
            UInt32(self[self.startIndex]) << 24 |
            UInt32(self[self.startIndex &+ 1]) << 16 |
            UInt32(self[self.startIndex &+ 2]) << 8  |
            UInt32(self[self.startIndex &+ 3])
        )
    }
}
```

is currently 45 instructions on Linux (lots of bounds checking, loading the value byte-by-byte and then combining via bit shifts, etc.). With this change in place, the same code above is now only 10 instructions:

```assembly
// Bounds checking:
	subs	x8, x2, x1
	b.ls	.LBB1_3
	cmp	x8, #4
	b.lo	.LBB1_3
// Loading buffer pointer:
	ldp	x8, x9, [x0, #16]
	sub	x8, x1, x8
// Load 32bit integer from buffer:
	ldr	w8, [x9, x8]
	rev	w0, w8
	ret
.LBB1_3:
	brk	#0x1
```

While the same code could already be achieved via a span, it's also a great benefit for the compiler to optimize the instructions when not using a span when feasible.

### Testing:

Code is exercised by existing unit tests

### DIFF
--- a/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
+++ b/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
@@ -119,13 +119,6 @@ internal final class __DataStorage : @unchecked Sendable {
     // The properties below use @exclusivity(unchecked) in release mode to avoid dynamic exclusivity checking. Exclusivity is guaranteed by the copy-on-write implementation of Data itself.
     // Dynamic exclusivity checks are still enabled in debug builds to catch issues with the copy-on-write implementation at-desk and in unit tests.
 
-    /// The size of the prior slice's prefix if the allocation was copied from a slice (to maintain
-    /// stable indexing)
-    #if !DEBUG
-    @exclusivity(unchecked)
-    #endif
-    @usableFromInline var _offset: Int
-    
     /// A pointer to the start of the referenced byte allocation
     #if !DEBUG
     @exclusivity(unchecked)
@@ -143,6 +136,13 @@ internal final class __DataStorage : @unchecked Sendable {
     @exclusivity(unchecked)
     #endif
     @usableFromInline var _capacity: Int
+
+    /// The size of the prior slice's prefix if the allocation was copied from a slice (to maintain
+    /// stable indexing)
+    #if !DEBUG
+    @exclusivity(unchecked)
+    #endif
+    @usableFromInline var _offset: Int
 
     /// The deallocator to use when discarding the byte allocation
     #if !DEBUG
--- a/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
+++ b/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
@@ -168,7 +168,7 @@ internal final class __DataStorage : @unchecked Sendable {
     /// their original allocation, this may point to memory before the actual allocation.
     @inlinable // This is @inlinable as trivially computable.
     var bytes: UnsafeRawPointer? {
-        return UnsafeRawPointer(_bytes)?.advanced(by: -_offset)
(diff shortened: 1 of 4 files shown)


jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/Data/Representations/DataStorage.swift
```diff
@@ -119,6 +119,13 @@ internal final class __DataStorage : @unchecked Sendable {
     // The properties below use @exclusivity(unchecked) in release mode to avoid dynamic exclusivity checking. Exclusivity is guaranteed by the copy-on-write implementation of Data itself.
     // Dynamic exclusivity checks are still enabled in debug builds to catch issues with the copy-on-write implementation at-desk and in unit tests.
 
+    /// The size of the prior slice's prefix if the allocation was copied from a slice (to maintain
+    /// stable indexing)
+    #if !DEBUG
+    @exclusivity(unchecked)
+    #endif
+    @usableFromInline var _offset: Int
```
jmschonfeld: I moved this `_offset` property before `_bytes` instead of after `_capacity`. For platforms where library evolution isn't enabled (Linux/Windows/etc.) it means that the `_offset` and `_bytes` properties can now be loaded together via a single, 2-word load instruction (previously required 2 separate load instructions). `_bytes` is most commonly accessed with `_offset` and `_length` so putting it between the two slightly optimizes these fast paths, and it also makes conceptual sense since the `_offset` is the size of the "fake" space prior to the start of `_bytes`. This has no impact on platforms where library evolution is enabled as properties are accessed separately through getters

jmschonfeld:

jmschonfeld:
It looks like something about this change may be breaking Data-related exit tests on Linux

jmschonfeld:
@swift-ci please test

Sources/FoundationEssentials/Data/Representations/DataStorage.swift
jmschonfeld: This change was removed because it broke the Linux build. I suspect it's because the test runner is using `Data` via the swift-testing in the toolchain and the mismatch in layout is causing issues. We can look at adding this back in the future if we find a way to avoid swift-testing's use of `Data` causing issues

glessard (approved):
This is great!

PR #1916, opened 2026-04-17
[NewCodable] fix macros access modifier

t089:
The `@JSONCodable` / `@JSONEncodable` / `@JSONDecodable` and `@CommonCodable` / `@CommonEncodable` / `@CommonDecodable` macros currently produce an extension whose `encode(to:)` / `decode(from:)` methods have no access modifier. When the attached type is `public` (or `package`), the Swift compiler rejects the generated code:

```
error: method 'encode(to:)' must be declared public because it matches a
requirement in public protocol 'JSONEncodable'
```

Reproducer — this fails to compile today:

```swift
import NewCodable

@JSONCodable
public struct Person {
    public var name: String
    public var age: Int
}
```

## Fix

Inspect the access modifier of the attached declaration and prefix the `encode(to:)` / `decode(from:)` methods accordingly, mirroring Swift's synthesized `Codable` behavior:

- `public` / `open` → `public `
- `package` → `package `
- `internal` / `fileprivate` / `private` / (none) → `` (empty)

Added `accessLevel(of:)` helper in `SharedMacroInfrastructure.swift`; threaded an `accessLevel: String` parameter (defaulting to `""`, preserving existing test expectations for internal types) through `makeEncodableExtension` and `makeDecodableExtension`. All six macro implementations (`JSONCodableMacro`, `JSONEncodableMacro`, `JSONDecodableMacro`, `CommonCodableMacro`, `CommonEncodableMacro`, `CommonDecodableMacro`) now compute the access level once via `accessLevel(of: declaration)` and forward it to the shared generators.

### DIFF
--- a/Sources/NewCodableMacros/CommonCodableMacro.swift
+++ b/Sources/NewCodableMacros/CommonCodableMacro.swift
@@ -36,9 +36,10 @@ extension CommonCodableMacro: ExtensionMacro {
             return []
         }
         
+        let access = accessLevel(of: declaration)
         let codingFields = makeCodingFieldsExtension(for: typeName, from: properties, kind: CommonCodingFieldExpansionKind.both)
-        let encodingImpl = makeEncodableExtension(for: typeName, with: properties, kind: CommonEncodableExpansionKind())
-        let decodingImpl = makeDecodableExtension(for: typeName, with: properties, kind: CommonDecodableExpansionKind())
+        let encodingImpl = makeEncodableExtension(for: typeName, with: properties, kind: CommonEncodableExpansionKind(), accessLevel: access)
+        let decodingImpl = makeDecodableExtension(for: typeName, with: properties, kind: CommonDecodableExpansionKind(), accessLevel: access)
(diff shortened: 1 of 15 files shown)


kperryua (commented):
Nice fix! Do we think there's any value in making `accessLevel` return an `enum` type of some kind instead of a bare string in case we have future need of that information elsewhere?

FYI, I have another in-flight PR (#1912) that adds some even better abstractions to describe the macro expansion. I'm happy to rebase that on top of this though.

kperryua:
> Nice fix! Do we think there's any value in making `accessLevel` return an `enum` type of some kind instead of a bare string in case we have future need of that information elsewhere?
> 
> FYI, I have another in-flight PR (#1912) that adds some even better abstractions to describe the macro expansion. I'm happy to rebase that on top of this though.

To unblock the parallel work, I'll go ahead and merge this, and in my rebasing I'll make the suggested change and let you review @t089.

kperryua (approved):

PR #1917, opened 2026-04-17
Add DocC inline docs for Foundation

parkera:
This is an initial import of DocC documentation for Foundation.

The source is content from the sources that produce the docs at developer.apple.com. Not all documentation exists here. Conceptual documentation, articles, and documentation about ObjC-only types are not included at this time.

For the most part I preferred the documentation phrasing rather than existing comments. In some rare cases, I chose to augment the documentation with content from the source. If there was already docs comments, and it was fairly divergent from the documentation, I kept the comments as actual comments (`//`) and imported the docs for DocC (`///`).

I also added stubs of the high level `.docc` directories for FoundationEssentials and FoundationInternationalization.

### DIFF
--- a/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
+++ b/Sources/FoundationEssentials/AttributedString/AttributeContainer.swift
@@ -10,11 +10,17 @@
 //
 //===----------------------------------------------------------------------===//
 
+/// A container for attribute keys and values.
+///
+/// ``AttributeContainer`` provides a way to store attributes and their values outside of an attributed string. You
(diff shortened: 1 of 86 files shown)


parkera:
@swift-ci test

parkera:
Also - this is intended to be a starting point, not the final form of all the docs. It's a massive enough diff as-is. I expect follow-up PRs to adjust wording, add missing docs, etc.

parkera:
@swift-ci test

jmschonfeld (approved):

parkera:
@swift-ci test

iCharlesHu (approved):

itingliu (approved):

PR #1918, opened 2026-04-17
rdar://162714826 (Remove String(_validating:as:) from Foundation) (#3502)

kperryua:
`Foundation.String(_validating:as:)` is a reimplementation of `Swift.String(validating:as:)` simply to avoid a dependency on a specific version of the Swift stdlib:

```
// TODO: This is part of the stdlib as of 5.11. This is a copy to support building on previous Swift stdlib versions, but should be replaced with the stdlib one as soon as possible.
```

The implementation is deficient compared to the stdlib's because its use of `_tryFromUTF8` redundantly validates the UTF8, and the `isASCII` value it computes is not ever read.

The above TODO is no longer relevant, meaning we can remove it and just use the stdlib implementation.

### DIFF
--- a/Sources/FoundationEssentials/String/String+IO.swift
+++ b/Sources/FoundationEssentials/String/String+IO.swift
@@ -53,7 +53,11 @@ extension String {
         switch encoding {
         case .ascii, .nonLossyASCII:
             func makeString(buffer: UnsafeBufferPointer<UInt8>) -> String? {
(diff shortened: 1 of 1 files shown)


kperryua:
@swift-ci test

jmschonfeld (approved):

parkera (approved):

glessard (approved):

PR #1923, opened 2026-04-22
Expressively initialising calendars

kielgillard:
An implementation for adding a  `Calendar` initializer that accepts a `TimeZone` and optional `Locale` and `firstWeekday` values, allowing developers to expressively and efficiently initialize a fully configured calendar in a single expression.

This is the implementation. The proposal can be found in https://github.com/swiftlang/swift-foundation/pull/1940.

An implementation for the alternative considered in the proposal can be found in https://github.com/swiftlang/swift-foundation/pull/1941.

### Motivation:

Please refer to the Motivation in the proposal text: https://github.com/swiftlang/swift-foundation/pull/1940/changes.

### Modifications:

The new initializer is defined as:

```swift
public init(
    identifier: Identifier,
    timeZone: TimeZone? = nil,
    locale: Locale? = nil,
    firstWeekday: Int? = nil,
    minimumDaysInFirstWeek: Int? = nil
)
```

The initializer is responsible for accepting a value for every publicly settable property of `Calendar` that affects calendrical computations. Presently, it is practically — but not necessarily — a member-wise initializer.

Internally, the initializer sets the values for the properties at once, avoiding the repeated initialization of the underlying calendar storage when setting properties one after the other.

When `timeZone`, `locale`, `firstWeekday`, or `minimumDaysInFirstWeek` are `nil`, the resulting calendar uses the same default values for those corresponding properties as `init(identifier:)`.

### Result:

Consider `CalendarTests.swift`. For example, instead of writing:
```
        var calendar = Calendar(identifier: .gregorian)
        calendar.timeZone = .gmt
        calendar.locale = Locale(identifier: "en_US")
```
…test authors can write:
```
        let calendar = Calendar(identifier: .gregorian, timeZone: .gmt, locale: Locale(identifier: "en_US"))
```

This allows test authors to write a single expression that statically expresses this calendar object does not change after it is defined (that is, the calendar object's value is constant). It also saves approximately 40 lines of code. It is also marginally more efficient.

### Testing:

`CalendarTests.swift` includes a test for the new initializer.


Proposals/NNNN-calendar-timezone-initializer.md
```diff
@@ -0,0 +1,127 @@
...
+
+The new initializer is defined as:
+
+```swift
+public init(
+    identifier: __shared Identifier,
+    timeZone: TimeZone,
+    locale: Locale? = nil,
+    firstWeekday: Int? = nil
+)
```
itingliu: if we're doing this, should this just be a member-wise initializer that includes all settable variables? If we do that we'd want to include `var minimumDaysInFirstWeek: Int` too

kielgillard: Thanks @itingliu for your time and effort, I appreciate it.

In c27294ec7e6e8a70e5564c53614b3bc77894dbc2, I've reframed this initialiser as one for publicly settable properties that affect calendrical computations, rather than as merely an indiscriminate member-wise initialiser.

I do wonder if: 1) that reframing is redundant -- a calendar just is responsible for calendrical computations and; 2) ABI stability prevents the initialiser from changing shape if new properties (with sensible default values) that also affect calendrical computations are added.

For example, let us say that `Calendar` publicly exposed `minimumDaysInFirstWeek` *after* this initialiser was in the ABI. The vast majority of use cases can accept a sensible default value, so the initialiser could take on a new shape that includes `minimumDaysInFirstWeek: Int? = nil`. This couldn't be done in a way that avoids breaking the ABI. So the initialiser can't maintain its responsibility as an initialiser for publicly settable properties that affect calendrical computations. To this I would suggest that such a property is not foreseeable after the mileage `Calendar` and `NSCalendar` has afforded developers all this time. And even if it was foreseeable, I would expect its use case to be so exotic as to make changing the property's default value separate to initialising the `Calendar` a regrettable but tolerable compromise.

itingliu: > For example, let us say that Calendar publicly exposed minimumDaysInFirstWeek after this initialiser was in the ABI. 

I'm not sure what you meant by it. This `minimumDaysInFirstWeek` is already public now. And given that we haven't added new variables on Calendar for a while I think the concerns is fairly small

kielgillard: I was trying to walk through an example of what would happen if a public property was added _after_ the initialiser was in the ABI. I do not think there are any ABI stability concerns impeding this.

I've done some investigation and updated the proposal and implementation again.

Proposals/NNNN-calendar-timezone-initializer.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+## Detailed design
+
+The new initializer is defined as:
+
+```swift
+public init(
+    identifier: __shared Identifier,
```
xwu: Is there a reason to use a superseded underscored ownership modifier here?

kielgillard: 👋 If by "here" you mean the proposal text, then as I understand it, the proposal should contain definition in code. If by "here" you mean in the code, then I am merely following what the Calendar API [uses for its existing initialiser](https://github.com/swiftlang/swift-foundation/blob/b33bbb35330524c21b54b503bb7673a82de59004/Sources/FoundationEssentials/Calendar/Calendar.swift#L468), and I defer to that reason.

Thanks for reading!

kielgillard:
In 5cf8c57430456635c93194f53a5c2fb363816796, explored [tera's suggestion](https://forums.swift.org/t/expressively-initialising-calendars-for-specific-time-zones/86139/9) for a Fluent style configuration initializer. Keen to hear what people think.

Proposals/NNNN-calendar-timezone-initializer.md
xwu: Yeah. I mean both: we don't typically show non-public features in interfaces, and in any case it doesn't seem like the new API requires specifically the superseded ownership modifier.

itingliu: I actually can't find any differences performance-wise with `__shared` or not, so I don't think we strictly need to follow the precedent.

Proposals/NNNN-calendar-timezone-initializer.md
itingliu: I think there would only be an ABI concern if we add another new settable property in the future, and we want to amend the initializer to take that property. But there are a few things we can do that's still ABI safe
1. Don't put the new property in initializers at all. Precedent: `Date.Component`'s initializer has most date component fields, but it does not have `dayOfYear`, which was added later
2. Add another initializer that takes that property. 
3. Add that new property to this initializer with a default argument.

So yeah I don't think this is a deal breaker.

itingliu (commented):
Can you move the proposal file into a separate PR so we can merge that and start the review, if you would also want to do that?

Proposals/NNNN-calendar-timezone-initializer.md
kielgillard: > 3. Add that new property to this initializer with a default argument.

I understood this would in fact be an ABI breaking change because the method signature still changes, even if it has a default value that makes it look "as if" it was the same initialiser at the call site?

Thanks for the tip about DateComponents. Do you think I've adequately captured those three points already in the proposal?

itingliu: Hm yeah you're right about it. We will need a new availability too.

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -348,6 +348,26 @@ public struct Calendar : Hashable, Equatable, Sendable {
             !self.contains(.era) && !self.contains(.year) && !self.contains(.dayOfYear) && !self.contains(.quarter) && !self.contains(.month) && !self.contains(.day) && !self.contains(.weekday) && !self.contains(.weekdayOrdinal) && !self.contains(.weekOfMonth) && !self.contains(.weekOfYear) && !self.contains(.yearForWeekOfYear) && !self.contains(.isLeapMonth) && !self.contains(.isRepeatedDay)
         }
     }
+    
+    /// A description of a `Calendar`'s configuration.
+    public struct Configuration: Hashable, Sendable {
```
parkera: This should be removed now, correct?

Proposals/NNNN-calendar-timezone-initializer.md
kielgillard: Removed in 766b82b.

Sources/FoundationEssentials/Calendar/Calendar.swift
kielgillard: Thanks! I've removed this in 6c36d2e5ebaea97c817f50b87fc44f0e1a588acb. I have a branch for the alternative should people like to consider it in the review (https://github.com/swiftlang/swift-foundation/pull/1941).

kielgillard:
> Can you move the proposal file into a separate PR so we can merge that and start the review, if you would also want to do that?

Done: https://github.com/swiftlang/swift-foundation/pull/1940.

Proposals/NNNN-calendar-timezone-initializer.md
kielgillard: @itingliu Could I please have some guidance for this (what version number, etc)?

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -469,6 +469,21 @@ public struct Calendar : Hashable, Equatable, Sendable {
...
+    /// - parameter timeZone: A `TimeZone` to use, instead of the default.
+    /// - parameter locale: A `Locale` to use, instead of the default.
+    /// - parameter firstWeekday: A first day of the week to use, instead of the default.
+    /// - parameter minimumDaysInFirstWeek: A number of minimum days in the first week to use, instead of the default.
+    public init(identifier: Identifier, timeZone: TimeZone? = nil, locale: Locale? = nil, firstWeekday: Int? = nil, minimumDaysInFirstWeek: Int? = nil) {
+        self.init(identifier: identifier)
+        guard timeZone != nil || locale != nil || firstWeekday != nil || minimumDaysInFirstWeek != nil else {
+            return
+        }
+        _calendar = _calendar.copy(changingLocale: locale, changingTimeZone: timeZone, changingFirstWeekday: firstWeekday, changingMinimumDaysInFirstWeek: minimumDaysInFirstWeek)
```
itingliu: Can you just do this so that we don't create a `_calendar` and recreate one right after it?
```suggestion
        _calendar = CalendarCache.cache.fixed(identifier: identifier, locale: locale, timeZone: timeZone, firstWeekday: firstWeekday, minimumDaysInFirstWeek: minimumDaysInFirstWeek, gregorianStartDate: nil)
```

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -469,6 +469,21 @@ public struct Calendar : Hashable, Equatable, Sendable {
...
     }
 
+    /// Returns a new Calendar using optional, non-default values.
+    ///
+    /// - parameter identifier: The kind of calendar to use.
+    /// - parameter timeZone: A `TimeZone` to use, instead of the default.
+    /// - parameter locale: A `Locale` to use, instead of the default.
+    /// - parameter firstWeekday: A first day of the week to use, instead of the default.
+    /// - parameter minimumDaysInFirstWeek: A number of minimum days in the first week to use, instead of the default.
+    public init(identifier: Identifier, timeZone: TimeZone? = nil, locale: Locale? = nil, firstWeekday: Int? = nil, minimumDaysInFirstWeek: Int? = nil) {
```
itingliu, line 480: This needs an availability annotation. I think it'd be 
```
@available(FoundationPreview 6.5, *)
```

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -201,6 +201,21 @@ private struct CalendarTests {
         #expect(anyHashables[0] != anyHashables[1])
         #expect(anyHashables[1] == anyHashables[2])
     }
+    
+    @Test func longInitializer() {
+        let locale = Locale(identifier: "en_AU")
+        let tz = TimeZone(identifier: "Australia/Sydney")!
```
itingliu: I know we're not super consistent in our tests, but since you're here, can you unwrap with `try` instead of force unwrapping here?

Sources/FoundationEssentials/Calendar/Calendar.swift
kielgillard: This is a great suggestion, thanks. I was not sure if I should use it because I seemed as if it was only intended for use in `NSCoding` and `Codable` conformances. Resolved in fddaee2.

Sources/FoundationEssentials/Calendar/Calendar.swift
kielgillard: I'll wait until https://github.com/swiftlang/swift-foundation/pull/2006 is merged to address this.

Tests/FoundationInternationalizationTests/CalendarTests.swift
kielgillard: Fair! Resolved in b0fc003.

Sources/FoundationEssentials/Calendar/Calendar.swift
kielgillard: Resolved in 95b53e1.

kielgillard:
@itingliu all feedback should be addressed now, happy for any more.

itingliu:
@swift-ci please test

kielgillard:
@itingliu CI seems to be taking a while, any update on this?

itingliu (approved):
Thank you!!

PR #1924, opened 2026-04-22
Remove Swift 4 Obsoleted APIs

jmschonfeld:
Removes all APIs obsoleted in Swift 4

### Motivation:

APIs obsoleted in Swift 4 cannot be called by any code today built with the latest toolchain because the latest toolchain does not support Swift 3 mode. Therefore, this code is dead and cannot be called and therefore can be removed.

### Modifications:

Removes all APIs obsoleted in Swift 4

### Result:

APIs obsoleted in Swift 4 are no longer present

### Testing:

N/A, this is removing code that is by definition uncallable


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

PR #1925, opened 2026-04-23
Restore backslash decoding behavior for Windows file URL paths

jrflat:
Restore Windows `URL.path` behavior of decoding `%5C` to `\`.

### Motivation:

`URL`'s POSIX path behavior is to decode all percent-escapes sequences except `%2F -> /` and `%00 -> \0`. The implementation in #1857 extended this behavior for Windows paths such that `%5C` was no longer decoded to `\`. A test failure in sourcekit-lsp revealed this is problematic because clients can pass a `file://` URL string containing backslashes to `URL(string:)`, which will encode all `\` to `%5C`. These clients then expect to receive the decoded `\` characters from the file path APIs.

### Modifications:

Replace usages of the `.windowsPath` pecent-decoding exclusion mask with `.posixPath` to restore previous behavior of only excluding `%2F` and `%00`.

### Result:

`URL(string: #"file:///C:\hello\world"#)?.path` will now return `#"/C:\hello\world"#` as it did previously instead of `"/C:%5Chello%5Cworld"`. Note that `URL(filePath:)` is the recommended initializer to use in this scenario as it avoids leading `/` issues by canonicalizing on initialization.

### Testing:

Added unit test to verify prior behavior.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Impl.swift
+++ b/Sources/FoundationEssentials/URL/URL_Impl.swift
@@ -641,11 +641,14 @@ extension _URL {
                     fromSpan: urlPath.extracting(inputStart...)
                 )
             } else {
-                // Percent-decode, excluding "%00", "%2F", and "%5C"
+                // Percent-decode, excluding "%00", "%2F"
                 guard let decodedLength = URLEncoder.percentDecodeUnchecked(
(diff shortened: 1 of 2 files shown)


jrflat:
@swift-ci please test

itingliu (approved):

PR #1934, opened 2026-04-30
Disable the test while I investigate a fix for the format change caused by ICU

itingliu:
Darwin ICU recently changed how "J" and "j" symbols resolve into date formats, and caused these test failures. Disabling the tests while figuring out the fix for now.

175861506


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1937, opened 2026-05-01
Cleanup Platform.swift import conditionals

jmschonfeld:
A minor cleanup of a few conditionals in `Platform.swift`

### Motivation:

This is a small adjustment to the conditionals in `Platform.swift` to make sure it fully encompasses all of the various modules Foundation may depend on and that it is calling the right functions when supported

### Modifications:

Updated a few `canImport`-style checks in `Platform.swift`

### Result:

Platform.swift correctly builds across all platforms

### Testing:

Validated that `Platform.swift` still builds on all platforms

### DIFF
--- a/Sources/FoundationEssentials/Platform.swift
+++ b/Sources/FoundationEssentials/Platform.swift
@@ -9,7 +9,7 @@
 //
 //===----------------------------------------------------------------------===//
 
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

kperryua (approved):

PR #1938, opened 2026-05-01
Remove _specializingCast from Foundation

jmschonfeld:
Removes `_specializingCast` in favor of `_specialize`

### Motivation:

We added `_specializingCast` to Foundation, but now the stdlib offers `_specialize` so we can just use that instead and remove our implementation

### Modifications:

Removes `_specializingCast` and replaces all uses with `_specialize`

### Result:

All code uses `_specialize` from the stdlib instead of our own implementation

### Testing:

All code still builds and tests still pass


jmschonfeld:
@swift-ci please test

parkera (approved):

kperryua (approved):

PR #1939, opened 2026-05-02
Update CMakeLists.txt

tkrajacic:
Adds a file to CMakeLists.txt that was missing

### Motivation:

I tried to use HTTPFormatStyle on Linux and mysteriously the compiler didn't find it.

### Modifications:

Add the implementation file to CMakeLists.txt

### Result:

It works

Credit goes to @jmschonfeld since he immediately discovered why the code wasn't available on Linux even though the Swift file was there.


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #1940, opened 2026-05-04
Proposal: Calendar initializer for non-default values.

kielgillard:
A proposal for adding a  `Calendar` initializer that accepts a `TimeZone` and optional `Locale` and `firstWeekday` values, allowing developers to expressively and efficiently initialize a fully configured calendar in a single expression.

This is the proposal. The implementation can be found in https://github.com/swiftlang/swift-foundation/pull/1923.

An implementation for the alternative considered can be found in https://github.com/swiftlang/swift-foundation/pull/1941.


itingliu (approved):

itingliu:
@swift-ci please test

PR #1941, opened 2026-05-04
Configuration based calendar initializer

kielgillard:
Do not merge. 

This is an alternative considered for the proposal in https://github.com/swiftlang/swift-foundation/pull/1940. That proposal is implemented in https://github.com/swiftlang/swift-foundation/pull/1923.

I will close this upon the end of the review.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -349,6 +349,26 @@ public struct Calendar : Hashable, Equatable, Sendable {
         }
     }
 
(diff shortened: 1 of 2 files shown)


kielgillard:
Review has ended. Thanks to all for the [discussion in the review thread](https://forums.swift.org/t/review-sf-0040-calendar-initializer-with-time-zone-locale-first-weekday-and-minimum-days-in-first-week/86559). An adjacent discussion around "Fluent" style value construction can be found [here](https://forums.swift.org/t/implicit-fluent-chaining/86660).

PR #1942, opened 2026-05-04
Correct `String.removingDotSegments` for relative paths

jrflat:
Correct `String.removingDotSegments` behavior for relative paths and align `_SwiftURL.standardized` to use the consolidated dot segment implementation for easier comparison testing.

### Motivation:

The internal `String.removingDotSegments` extension followed the RFC 3986 algorithm, which collapses a relative path like `a/../b` to `/b`, returning an unexpected root path. It should instead preserve leading dot segments (like RFC 1808) until base path resolution is performed.

`_URL` already contains this bug fix, but this PR also aligns `_SwiftURL` for easier API comparison testing.

### Modifications:

- Consolidates the internal `String.removingDotSegments` extension to call into the `resolveDotSegmentsInPlace(buffer:useRFC1808:)` function used by `CFURL`, `NSURL`, and the v2 `_URL` backing class.
- Updates the behavior of `_SwiftURL.standardized` to match the behavior of `_URL`.
- Updates the behavior of `String.removingDotSegments` to not accidentally resolve a relative path like `a/../b` to an absolute path `/b`.

### Result:

- `String.removingDotSegments` returns `""` instead of `"/"` for various relative path inputs that shouldn't become absolute, aligned with other URL APIs.
- `_SwiftURL.standardized` matches other URL APIs' behavior of preserving leading dot segments (RFC 1808).

### Testing:

- Updated unit tests for relative paths returning `""` instead of a root path.
- Removed v2 guards from `URL.standardized` unit tests that now pass on either backing class implementation.


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #1943, opened 2026-05-04
Remove JSON-related stringshims

jmschonfeld:
Removes a few string shims used by JSON encoding/decoding

### Motivation:

These functions are currently called via a shim in `string_shims.h`, but we can instead call them directly from Swift. The only thing we actually need a shim for is the Android SDK version check (which has no equivalent in Swift code) and the `LC_ALL_MASK` check (which can't be imported into Swift)

### Modifications:

- Remove some shims and replace with direct calls in Swift code

### Result:

Swift code directly calls C library functions and C code is removed from `_FoundationCShims`

### Testing:

Covered by existing JSON unit tests


jmschonfeld:
@swift-ci please test

kperryua (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

PR #1945, opened 2026-05-05
CodableBy macro, CodingStrategy protocols, and CodingStrategySyntax

kperryua:
Adds the `@CodableBy` macros and associated APIs that allow transformation of an encoded value into arbitrary runtime values.

### Motivation:

Sometimes an encoded value needs to be decoded into a different type at runtime. Or sometimes a client wants to use a different encoding/decoding algorithm than the default. A very common example of the former case is dates. The native runtime type in Swift for dates is, of course, `Date`, but most serialization formats don't natively handle dates. Instead they're often encoded as strings and then reinterpreted as dates at runtime. With the exception of the `Date`- or `Data`-specific "global strategies" that exist on `JSONEncoder` & `JSONDecoder`, this kind of customization used to require manual implementations or clever propertyWrappers.

The `@CodableBy` macro allows clients to define arbitrary transformations of encoded values to decoded types, and vice versa. These attributes instruct a `@XYZCodable` macro to call the specified "strategy" implementation instead of the runtime type's default implementation. These strategies can be format-agnostic (`CommonCodableStrategy`), and can also be format-specific (`JSONCodableStrategy`) to take full advantage of the format's own capabilities.

### Modifications:

* Adds `CodableBy`, `EncodableBy`, and `DecodableBy` macros.
* Adds `CodableStrategy` macros for both "Common" and "JSON"
* Adds `CodingStrategySyntax` to allow more concise and expressive `CodableBy` attribute declarations.
* Implements the `passthrough`, `losslessStringConversion`, `.format()`, `.dateFormat()`, and `.base64` strategies (the last 3 destined for Foundation)

### Testing:

Adds both macro expansion and integration tests to verify macro functionality.


kperryua:
I'm seriously reconsidering the use of `FormatStyle` and `ParseStrategy` here. Obviously these will be Foundation-specific additions, but so many of these formats are inherently localized, which is not a good candidate for serialization formats.

I'm considering whether we should just resort to ad hoc strategies (e.g. ISO8601 specific) or perhaps generalize over a similar, but specifically locale-unaffected protocol.

Sources/NewCodable/CodingStrategies.swift
```diff
@@ -0,0 +1,190 @@
...
+
+// MARK: Format Style Convenience
+
+extension CommonCodingStrategy {
+    /// Encode/decode a value as a string using a `ParseableFormatStyle`.
+    public static func format<F: ParseableFormatStyle>(_ style: F) -> StringFormattedCodingStrategy<F> where Self == StringFormattedCodingStrategy<F>, F.FormatOutput == String {
+        .init(style)
+    }
+    
+    public static func dateFormat<F: ParseableFormatStyle>(_ style: F) -> StringFormattedCodingStrategy<F> where Self == StringFormattedCodingStrategy<F>, F.FormatInput == Date, F.FormatOutput == String {
```
itingliu: Sorry if I'm missing some context, but why do you want another override particularly for `Date`?

kperryua: I actually would rather not have to do this! But because macros don't (yet?) have a way to bind generic parameter types to the attachee's types (in this case the `FormatStyle.FormatOutput == Date`), there's no way to disambiguate between `Date`'s and `DateComponent`'s overloads of `.iso8601`.

An alternative to `.dateFormat()` is an overload of `.format()` that takes a metatype value to manually bind the type where necessary, e.g. `.format(.iso8601, type: Date.self)`.

itingliu:
> I'm considering whether we should just resort to ad hoc strategies (e.g. ISO8601 specific) or perhaps generalize over a similar, but specifically locale-unaffected protocol.

As we talked offline, I agree with this. I think it would help avoid exposing a footgun

Sources/NewCodable/CodingStrategies.swift
kperryua: Alternatively, I could wish for the macro system to ignore the ambiguity since its main job here is to guarantee that there's well-formed AST, which there is. The code this would expand to already has the `Date` type available to bind to, so the complaining here about ambiguity is unhelpful.

I don't really see that happening anytime soon so here are the alternative `@CodableBy()` parameter syntaxes I can come up with:
1. `.dateFormat(.iso8601)`
2. `.format(.iso8061, type: Date.self)`
3. just `.iso8601`
4. `.format(Date.ISO8601FormatStyle())`

(1) implies we'd likely also need an overload for `.dateComponents()` as well, most likely.
(2) is fully featured, but requires the replication of the attachee property's type. Formats with unambiguous static members could use the `.format()` overload without the `type:` parameter just fine.
(3) Because this would be declared on `CodingStrategySyntax`, there *is* no bound type—just AST that is identical to what we'd make the macro generate in actual code where the type *can* be bought. But this style makes customizing the resulting format style more difficult (because it's not actually resolving to a an `ISO8601FormatStyle`!) and it has to be repeated for other formats (e.g. `.http`).
(4) is just requiring clients to fully specify the name of the `FormatStyle` type without static member syntax. This will always work, but it's still repeating "Date" and seems ergonomically worse than (1) or (2).

We also have the option of doing *both* (1) and (2). `.dateFormat()` is a little prettier than `type: Date.self`, but likely infinitely more common than `dateComponentsFormat()`, for which I think `type: DateComponents.self` is acceptable.

kperryua:
Latest commit adds a proposed protocol replacement for `FormatStyle`, et al. and adopts it for `Date` and `DateComponents` (soon `URL`). Also adds `.format(_:type:)`.

kperryua:
Another direction I could take `CodingFormatStyle` is to eschew `FormatInput`/`String` usage entirely and either take an encoder/decoder directly, or maybe just a `Span` / `UTF8Span`. The latter would let us avoid having to split the protocol into format-specialized mirrors, but limits our ability to access other input types besides strings.

kperryua:
Merged changes that removes NewCodable -> FoundationEssentials dependency.
This brought the `.iso8601` extension declarations into FoundationEssentials module, which unfortunately necessitated `@_disfavoredOverload` attributes on all of them. Pretty obnoxious, but it seems like that's the common pattern for these repeated static member declarations.

kperryua:
RE: All the discussion about FormatStyle—I realized that with the design of this API, it's actually not a NewCodable concern at all. FormatStyle is owned by Foundation, so its adoption of CodingStrategies and CodingStrategySyntax will be too. I'll keep what's there for now as a provisional proof-of-concept, but it will not end up as part of the final NewCodable design.

PR #1950, opened 2026-05-06
Avoid crashing when test resources are not found

jmschonfeld:
Throw errors instead of crashing when test resources are not found during test execution

### Motivation:

Currently, unit tests crash when test resources are not found. On platforms like Windows where there are no backtraces, this results in hard to debug CI logs and the remaining tests stop running.

### Modifications:

Update the `testData` function to `throws` instead of returning an `Optional` that is force unwrapped

### Result:

When a test resource is not found, the test is marked as a failure instead of a crash preventing further testing.

### Testing:

Validated that all tests pass and that when removing a test file locally an error is thrown.


jmschonfeld:
@swift-ci please test

itingliu (approved):

jmschonfeld:
@swift-ci please test

glessard (commented):
Yes! Force unwrapping is a scourge in tests!

PR #1953, opened 2026-05-07
Add Swift Hebrew calendar implementation 

dra8an:
Add pure Swift Hebrew calendar implementation 

### Motivation:

The Hebrew calendar is currently implemented via ICU/C bridging, which incurs overhead and requires translating between ICU's calendar APIs and Foundation's _CalendarProtocol contract. 

This PR provides a Swift arithmetic implementation based on the Reingold & Dershowitz algorithms (Calendrical Calculations), enabling the Hebrew calendar to conform directly to _CalendarProtocol without the indirection of adapting ICU's interface. 

It also introduces an optional nextDate(after:matching:direction:) fast path protocol method that calendar implementations can override to answer common date matching queries in O(1) instead of going through the generic enumeration framework.

### Modifications:

Calendar_Hebrew.swift (new): Complete _CalendarProtocol conformance for .hebrew including date components, date arithmetic, dateInterval(of:for:), ordinality, and a nextDate fast path covering year, month, weekday, and weekdayOrdinal patterns. Includes the HebrewArithmetic engine with a single slot YearData cache. The cache stores precomputed year level metadata (new year RD, year length, leap status) so that downstream calls within the same year reuse it without redundantly recomputing the Hebrew new year and elapsed days arithmetic.

Calendar_Protocol.swift: Added optional nextDate(after:matching:direction:) protocol method with a default nil implementation (no op for calendars that don't override it).

Calendar.swift: Wired up the fast path in nextDate(after:matching:) and enumerateDates(startingAfter:matching:) for default policy calls.

Calendar_Cache.swift: Registered .hebrew in the calendar cache.

New test files covering unit tests, DST policy parity tests, and performance benchmarks.

### Result:

Calendar(identifier: .hebrew) is now backed by a Swift implementation. Date computation, component extraction, and date enumeration all work without ICU. The fast path yields significant speedups for nextDate/enumerateDates on supported patterns.

### Testing:

All 1527 tests pass across 3 test targets (FoundationEssentials, FoundationInternationalization, FoundationMacros). Hebrew specific suites include unit tests,  DST policy parity tests, and performance benchmarks. Built and tested with Swift 6.4

### Follow up:

This PR is already fairly large, so a couple of items will be addressed in separate PRs: additional fast paths for recurrence rules, and a shared utility layer for APIs currently based on ICU that are common across calendars (such as weekNumber(), isDateInWeekend(), weekday() and similar).


itingliu:
@swift-ci please test

Tests/FoundationInternationalizationTests/HebrewCalendarPerformanceTests.swift
```diff
@@ -0,0 +1,201 @@
+//===----------------------------------------------------------------------===//
```
itingliu: Can you rewrite these performance tests as a benchmark test? See https://github.com/swiftlang/swift-foundation/blob/main/Benchmarks/Benchmarks/Internationalization/BenchmarkCalendar.swift

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
```diff
@@ -0,0 +1,208 @@
...
+                        divergences.append(
+                            "[\(label)] (rep=\(repeated), skip=\(skipped)): hebrew=\(hebrewDate) greg=\(gregDate) diff=\(hebrewDate.timeIntervalSince(gregDate))s"
+                        )
+                    }
+                }
+            }
+        }
+
+        if !divergences.isEmpty {
+            print("\nDST policy parity divergences (\(divergences.count)):")
```
itingliu: Can we replace `print` with something more discoverable, like a test assert? `print` statements tend to be buried in the log.

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
```diff
@@ -0,0 +1,208 @@
...
+            firstWeekday: nil, minimumDaysInFirstWeek: nil, gregorianStartDate: nil
+        )
+        let hebrewCal = Calendar(inner: hebrewInner)
+        var gregCal = Calendar(identifier: .gregorian)
+        gregCal.timeZone = la
+
+        // For each Greg wall-clock at a DST boundary, build the same wall-clock
+        // via the Hebrew calendar (using its civil-month decomposition of the
+        // corresponding Greg date) and compare resulting Dates.
+        let probes: [(String, Int, Int, Int, Int, Int)] = [
```
itingliu: Can these probes be passed in as the test arguments?

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
```diff
@@ -0,0 +1,208 @@
...
+        )
+
+        // Compute the RD (proleptic Gregorian fixed-day number) of midnight UTC for a given Greg date.
+        // Independent of any calendar's internals — uses Foundation's Date arithmetic.
+        func rdAtMidnightUTC(_ y: Int, _ m: Int, _ d: Int) -> Int64 {
+            var cal = Calendar(identifier: .gregorian)
+            cal.timeZone = .gmt
+            var dc = DateComponents()
+            dc.year = y; dc.month = m; dc.day = d; dc.timeZone = .gmt
+            let date = cal.date(from: dc)!
```
itingliu: Can we avoid force unwrap from tests?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1788 @@
...
+@preconcurrency import Bionic
+#elseif canImport(Glibc)
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif canImport(CRT)
+import CRT
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
```
itingliu, line 25: Why are these needed?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1788 @@
...
+        Logger(subsystem: "com.apple.foundation", category: "hebrew_calendar")
+    }()
+#endif
+
+    let kSecondsInWeek = 604_800
+    let kSecondsInDay = 86400
+    let kSecondsInHour = 3600
+    let kSecondsInMinute = 60
+
+    let inf_ti: TimeInterval = 4398046511104.0
```
itingliu, line 49: As we mentioned previously, these constants overlap with those in GregorianCalendar. We should extra these out so they can be shared

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1788 @@
...
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(identifier)
+        hasher.combine(timeZone)
+        hasher.combine(firstWeekday)
+        hasher.combine(minimumDaysInFirstWeek)
+        hasher.combine(localeIdentifier)
+        hasher.combine(preferredFirstWeekday)
+        hasher.combine(preferredMinimumDaysInFirstweek)
+    }
```
itingliu, line 152: if I read it correctly, this file is a duplicate of Gregorian Calendar up at least until this point. Can we factor out common logic?

Tests/FoundationInternationalizationTests/HebrewCalendarPerformanceTests.swift
dra8an: Rewritten

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
dra8an: Replaced with except

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
dra8an: Done

Tests/FoundationInternationalizationTests/HebrewDSTPolicyParityTests.swift
dra8an: Done

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Removed unused imports, leftover for some of my parity testing

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: I was planning to do this in a separate PR, but please let me know if you think I should do it in this one.

itingliu: ah thanks that works too

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1778 @@
...
+        case .nanosecond:
+            return DateInterval(start: date, duration: 1e-9)
+        case .isLeapMonth, .isRepeatedDay, .calendar, .timeZone:
+            return nil
+        }
+    }
+
+    // MARK: - Weekend queries
+
+    // Matches _CalendarGregorian.isDateInWeekend exactly. Uses locale's weekendRange
```
itingliu: Can we add a // TODO for these future refactoring opportunities?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1778 @@
...
+
+    /// Convert a local-seconds-since-reference value to (RD, seconds-in-day).
+    private static func rataDieAndSecondsInDay(localSeconds: Double) -> (rd: Int64, secondsInDay: Double) {
+        let totalDays = (localSeconds / 86400).rounded(.down)
+        let rd = Int64(totalDays) &+ rataDieAtDateReference
+        let secondsInDay = localSeconds - totalDays * 86400
+        return (rd, secondsInDay)
+    }
+
+    /// Given a fixed-day RD + local-seconds-within-day, build a UTC Date after
```
itingliu: these comments are so verbose

itingliu (commented):
Can we add a feature flag?

Benchmarks/Benchmarks/Internationalization/BenchmarkCalendar.swift
```diff
@@ -538,5 +538,62 @@ func calendarBenchmarks() {
...
+                stop = true
+            }
+        }
+    }
+
+    Benchmark("HebrewCalendar-allocationsForFixedCalendar", configuration: allocationsConfiguration) { benchmark in
+        for _ in benchmark.scaledIterations {
+            let cal = Calendar(identifier: .hebrew)
+            let date = cal.date(byAdding: .day, value: 1, to: hebrewStart)
+            blackHole(date)
```
itingliu, line 562: It's not clear to me what we're measuring here. Do we want to measure the effects of `cal.date`, or do we also want to measure initializing `let cal = Calendar(identifier: .hebrew)`?

If the former I suggest that we reuse `hebrewCal` on line 544 so this benchmark closure does not count the initialization

Benchmarks/Benchmarks/Internationalization/BenchmarkCalendar.swift
```diff
@@ -538,5 +538,62 @@ func calendarBenchmarks() {
...
+        }
+    }
+
+    Benchmark("HebrewCalendar-copyOnWritePerformance", configuration: allocationsConfiguration) { benchmark in
+        var cal = Calendar(identifier: .hebrew)
```
itingliu, line 567: similarly I also recommend moving this out to outside of the benchmark scope, or use benchmark's "start measuring" functionality

itingliu:
@swift-ci please test

Benchmarks/Benchmarks/Internationalization/BenchmarkCalendar.swift
dra8an: For allocationsForFixedCalendar we intentionally measure both — creating a fresh Calendar(identifier: .hebrew) and calling date(byAdding:) on it. This matches the existing  Gregorian allocationsForFixedCalendars benchmark, which does the same thing. The goal is to verify that fetching a cached calendar instance + doing one operation stays cheap.

For the other benchmarks (nextThousandHanukkahs, dateComponents, roundTrip), the calendar is initialized once outside the closure so we're only measuring the computation itself.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -0,0 +1,1706 @@
...
+            fixedTime = floor(fixedTime / 3600.0) * 3600.0
+            fixedTime = fixedTime - ti
+            return DateInterval(start: Date(timeIntervalSinceReferenceDate: fixedTime), duration: 3600.0)
+        case .minute:
+            // Minute and second don't depend on TZ — float floor on UTC seconds is fine.
+            let time = date.timeIntervalSinceReferenceDate
+            return DateInterval(start: Date(timeIntervalSinceReferenceDate: floor(time / 60.0) * 60.0), duration: 60.0)
+        case .second:
+            let time = date.timeIntervalSinceReferenceDate
+            return DateInterval(start: Date(timeIntervalSinceReferenceDate: floor(time)), duration: 1.0)
```
itingliu, line 559: This caused the linux build failure: 

```
Calendar_Hebrew.swift:549:77: [1;31merror: [1;39mcannot find 'floor' in scope[0;0m
```

You probably need to put back the header import that you removed

```
#if canImport(os)
internal import os
#elseif canImport(Bionic)
@preconcurrency import Bionic
#elseif canImport(Glibc)
@preconcurrency import Glibc
#elseif canImport(Musl)
@preconcurrency import Musl
#elseif canImport(CRT)
import CRT
#elseif os(WASI)
@preconcurrency import WASILibc
```
 or change to use `rounded` or something else

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

itingliu (approved):

PR #1954, opened 2026-05-07
Guard legacy ContiguousBytes ABI with DATA_LEGACY_ABI

jmschonfeld:
Guard the legacy `ContiguousBytes` ABI entry points behind the `DATA_LEGACY_ABI` flag

### Motivation:

These ABI entry points are only necessary on platforms where we need to preserve legacy ABI

### Modifications:

Guard the legacy `ContiguousBytes` ABI entry points behind the `DATA_LEGACY_ABI` flag

### Result:

ABI entry points are not exposed when not required

### Testing:

N/A because we are removing ABI entry points that are unnecessary, existing tests validate the ABI entry points are present when required


jmschonfeld:
@swift-ci please test

itingliu (approved):

PR #1955, opened 2026-05-07
Remove module specifier on FormatStyle within Date

jmschonfeld:
Remove extra unnecessary module qualification in `Date.FormatStyle`-related code

### Motivation:

This code currently `#if`'s around whether to reference `FormatStyle` as `Foundation.FormatStyle` or `FoundationEssentials.FormatStyle` resulting in duplicated code. However, when `protocol FormatStyle` is in the same module as the code, name lookup prefers the top level protocol over the sibling `struct Date.FormatStyle`. This means we can remove the extra qualification inside of FoundationEssentials.

Note that this lookup precedence rule only applies within `FoundationEssentials`. `Decimal+FormatStyle` has a similar workaround in `FoundationInternationalization`, but we cannot make this change there because the sibling `struct Decimal.FormatStyle` is preferred over the imported `protocol FormatStyle` from `FoundationEssentials`.

### Modifications:

Remove extra module qualification in `Date+FormatStyle.swift`

### Result:

Project builds without duplicated code

### Testing:

Verified that project still builds.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Date+FormatStyle.swift
+++ b/Sources/FoundationEssentials/Calendar/Date+FormatStyle.swift
@@ -45,15 +45,9 @@ extension Date {
     ///
     /// - Parameter format: The format for formatting `self`.
     /// - Returns: A representation of `self` using the given `format`. The type of the representation is specified by `FormatStyle.FormatOutput`.
-#if FOUNDATION_FRAMEWORK
-    public func formatted<F: Foundation.FormatStyle>(_ format: F) -> F.FormatOutput where F.FormatInput == Date {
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

jmschonfeld:
Unfortunately my analysis was incorrect. The only reason why this works here is because `Date.FormatStyle` is defined in `FoundationInternationalization`, thus it is not visible to name lookup when building `Date.formatted`. However, when building all source files together in one Foundation module for Foundation.framework, it is visible and fails to compile there, so we unfortunately still need the `#if`s here for now

Issue #1960, opened 2026-05-08
Get new-codable optionally building in embedded mode


kperryua:
Now that the `NewCodable` target is entirely independent of Foundation, we can more easily pursue getting it building cleanly in Embedded Swift—which is one of the main goals of this project.

tishin:
I believe the set of PRs above combined should make it possible to build `NewCodable` using Embedded Swift. At least as of 6.3.1, I checked it on `swift-6.3.1-RELEASE_wasm-embedded` with `.enableExperimentalFeature("SuppressedAssociatedTypes")` in the target's swiftSettings.

kperryua:
Awesome!

All of these are merged.

@tishin Can you share your specific setup for building this in Embedded mode locally? I've attempted to set something up myself, but I'm quite new to this mode, and I'm probably doing things incorrectly. 😅

The Embedded platforms we may target are quite varied, so I wonder if we need some platform adaptation layer. My local attempts are complaining about the lack of `memcmp`, `strtod`, and `strtof` from libc. I'd definitely prefer to turn these into calls into the stdlib and let it handle that, but we still don't have Span-taking versions of `String(_:)` initializers for numbers, and I'm still rather distrustful in the ability of the optimizer to optimize various equality calls into `memcmp` equivalents. (Plus, there's `-Onone` performance to care about there as well.)

I'd love to have some kind of CI automation to verify that we don't regress this, but I'm not sure the infrastructure is established for that yet.

tishin:
> [@tishin](https://github.com/tishin) Can you share your specific setup for building this in Embedded mode locally?

I just followed [Getting started with WASM guide](https://www.swift.org/documentation/articles/wasm-getting-started.html). No extra setup was needed. Then I just built NewCodable with 
`swift build --swift-sdk swift-6.3.1-RELEASE_wasm-embedded --target NewCodable`

I haven't checked any Embedded platforms other than WASM.

kperryua:
#1984 is available for review to officially start building Embedded mode in CI.

kperryua:
Merged #1984.

PR #1964, opened 2026-05-09
new-codable: Specify error types for the methods directly used in NewCodable implementation

tishin:
Specifies error types for NewCodable throws.

### Motivation:

Untyped throws won't compile on Embedded Swift. This PR only covers functionality used in the NewCodable implementation, not covering traditional Codable adapters. Adapters will be the subject of another PR, as they have a different kind of issue with Embedded. It does touch the "Traditional Codable support" block of `JSONPrimitive`, because `decode(UInt8.self)` from it is also used in the new implementation.

### Modifications:

Specified error types.

### Result:

Resolves some of the build errors on Embedded Swift (#1960).


kperryua:
Thanks for this!

`JSONError` is *probably* the right type to throw for `JSONPrimitive`, but it was never really well designed from an API point of view. We'll need to eventually circle back to evaluating that type in detail if we're going to formally make it part of the API surface.

PR #1965, opened 2026-05-09
new-codable: Address traditional Codable unavailability on Embedded Swift

tishin:
Disables `Codable` adapters for Embedded Swift builds.

### Motivation:

Traditional `Codable` is unavailable on Embedded Swift, and using any of its interfaces (`CodingKey`, `Encodable`, `Decodable`, etc.) will result in compilation errors.

### Modifications:

Added conditional compilation checks for everything related to traditional Codable adapters.

### Result:

Resolves some of the build errors on Embedded Swift (#1960).

### DIFF
--- a/Sources/NewCodable/Utilities.swift
+++ b/Sources/NewCodable/Utilities.swift
@@ -15,6 +15,7 @@
 // Shared Key Type
 //===----------------------------------------------------------------------===//
 
+#if !hasFeature(Embedded)
 @usableFromInline
 internal enum _CodingKey : CodingKey {
     case string(String)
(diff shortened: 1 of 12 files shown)


Sources/NewCodable/Utilities.swift
```diff
@@ -16,7 +16,7 @@
 //===----------------------------------------------------------------------===//
 
 @usableFromInline
-internal enum _CodingKey : CodingKey {
+internal enum _CodingKey: Sendable {
```
kperryua: Are we actually using `_CodingKey` somewhere in Embedded-native code? That would be surprising to me. I'd expect that this entire type is just unavailable in Embedded.

tishin: You're right, it's only used in adapters, so it can also be disabled for Embedded.

kperryua (approved):

PR #1970, opened 2026-05-11
[CMAKE] compare CMAKE_SYSTEM_NAME against string literals

justice-adams-apple:
Change FoundationEssentials to check CMAKE_SYSTEM_NAME against string literals

### Motivation:

As part of the bring up of debain13, the platform is using a newer cmake 3.31. cmake 3.31 sets WASI=1 so:

```cmake
    if(CMAKE_SYSTEM_NAME STREQUAL WASI)
```
evaluates as "WASI" STREQUAL "1" on cmake 3.31 (Which Is false), silently skipping the -D_GNU_SOURCE compile option for WASI targets. This results in 
```
swift-foundation/Sources/FoundationEssentials/Platform.swift:421:16: error: cannot find 'strtod_l' in scope
```
during toolchain builds.

Debain 12 has Cmake 3.25 which has no WASI variable, so WASI fell through to a literal string and the condition evaluated correctly by comparing against the string literal

### Modifications:
Change all platform checks against CMAKE_SYSTEM_NAME to use the string literal

### Result:

The toolchain will build on debain13

### Testing:

Will test on CI


justice-adams-apple:
@swift-ci smoke test

jmschonfeld:
`Sources/FoundationMacros/CMakeLists.txt` also has `if(CMAKE_SYSTEM_NAME STREQUAL Darwin)`, should we change that as well so this repo is consistently using strings everywhere (and I suspect this change is probably needed somewhere for swift-corelibs-foundation and swift-foundation-icu as well)

justice-adams-apple:
@jmschonfeld probably? should I include that in this pr?

justice-adams-apple:
@swift-ci test

jmschonfeld:
@justice-adams-apple including it seems like a good idea to me unless you think there's something different we should do there, just so that the repo is self-consistent and we change everything at once if we know the non-quoted approach is incorrect. But if this piece is urgent to unblock something and the FoundationMacros change isn't, we could do that in a separate PR.

justice-adams-apple:
@swift-ci test

jmschonfeld (approved):

PR #1974, opened 2026-05-12
Update `.hour(.defaultDigits(amPM: .omitted))` test expectation

itingliu:
There's a recent ICU change that intentionally strips extra text when the skeleton is just "J" (hour only, amPM omitted), so `.hour(.defaultDigits(amPM: .omitted))` now returns "3" instead of "3時" for zh_TW. Update the test expectations behind a flag for now in case CI hasn't all caught up with the new behavior.


itingliu:
@swift-ci please test

itingliu:
@swift-ci please test

jmschonfeld (approved):

iCharlesHu (approved):

PR #1978, opened 2026-05-14
Decimal crash

wadetregaskis:
Fixed a buffer over-run crash in `stringViewContainsDecimalSeparator` when the given string is a strict subset of the separator.

### Motivation:

Crashing is bad? 😆

### Modifications:

Now both strings are walked in parallel until _either_ of them ends (previously it was only checking for the end of the separator - in a way that would have worked fine for a C-string, but doesn't for Swift `String`s - well, in this case a `UTF8View` of one - that are not null-terminated).

### Result:

It'll now return false in this case, rather than crashing, because the separator is not found _in its entirety_ within the given string.

All other cases should behave as before.

### Testing:

I added a new unit test for this particular case.


Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -281,6 +281,20 @@ private struct DecimalTests {
         #expect(notDecimal == nil)
     }
 
+    @Test func decimalParseTruncatedMultiByteSeparator() {
+        // Regression: Decimal._decimal walked past `endIndex` when the input ended on a
```
itingliu: sorry another nit pick about comments... These comments will become irrelevant after the fix is merged (because the bug is fixed!) Can we leave them in the PR description or commit message, but leave them out of the code?

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -281,6 +281,20 @@ private struct DecimalTests {
...
+        // proper prefix of a multi-byte decimal separator. The check that the separator
+        // matches at the current index only tested that the first byte was in range, then
+        // unconditionally indexed forward by `decimalSeparator.count - 1` more bytes.
+        // With separator ",," and input "1,", parsing crashed on `utf8View[endIndex]`.
+        let result = Decimal._decimal(from: "1,".utf8, decimalSeparator: ",,".utf8, matchEntireString: false)
```
itingliu, line 285: This is calling an internal method on `Decimal`. Is there a test case that we can add for calling the public `Decimal` API that would trigger this crash?

wadetregaskis: I think it's ultimately a hypothetical bug, more than a real one [today].  I found it not by triggering it organically but by static analysis.

I can make the test use the `decimal(from:decimalSeparator:matchEntireString:)` that's _technically_ public (though marked @_spi for SwiftCorelibsFoundation)…

…but beyond that, I'm not sure if there's a way to trigger it today.  I can't find any existing locales with multi-character separators, and there's no public API for customising the decimal separator.

It looks like maybe I could create a custom one using `localeAsIfCurrent` (package internal), and then use the public APIs to create a `Decimal` and trigger the problem… still using private API, but mimicking [more realistically] a hypothetical future locale addition or change that introduced a multi-character separator.

'course, how likely that is to happen, I'm not sure… I assume the existing locales are pretty comprehensive, and that new locales don't come along often.

itingliu:
@swift-ci please test

iCharlesHu (approved):

PR #1981, opened 2026-05-15
Benchmarks: default to local swift-foundation branch

itingliu:
With no environment variable set, benchmarks now build against the local swift-foundation branch with other package dependencies fetched from remote. Previously the default was the system Foundation / toolchain, which is rarely useful for everyday work at desk.

USE_TOOLCHAIN=1 opts back in to the old default (Foundation.framework on macOS, toolchain Foundation on Linux).

Adds Benchmarks/README.md documenting prerequisites, run commands, and all four configurations (local default, SWIFTCI_USE_LOCAL_DEPS, USE_PACKAGE, USE_TOOLCHAIN). Condenses the Benchmarks section in Foundation_Build_Process.md to a pointer to that README.

### DIFF
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -6,9 +6,11 @@ enum UsePackage {
     /// Use a known package hash downloaded from GitHub. Set with env var `USE_PACKAGE`
     case useGitHubPackage
 
-    /// Use a local package, rooted at `String`. For example ".." to back up one directory from the Benchmark package. This is useful when testing local changes.
-    /// Set with env var `SWIFTCI_USE_LOCAL_DEPS=<path to root>`, for example `SWIFTCI_USE_LOCAL_DEPS=..`
+    /// Use a local package, rooted at `String`. `SWIFTCI_USE_LOCAL_DEPS` is also visible to `../Package.swift`, which resolves other package dependencies from local sibling checkouts as well.
     case useLocalPackage(String)
+
+    /// Use a local swift-foundation checkout, rooted at `String`, but fetch other package dependencies from remote.
+    case useLocalPackageRemoteDeps(String)
 
     /// Use Foundation.framework (Darwin) or the toolchain (Linux)
     case useToolchain
@@ -17,7 +19,7 @@ enum UsePackage {
         switch self {
             case .useGitHubPackage:
                 return "Using GitHub package"
-            case .useLocalPackage(let root):
+            case .useLocalPackage(let root), .useLocalPackageRemoteDeps(let root):
                 #if os(macOS)
                     return "Using local package checkout at \(root)/swift-foundation"
                 #else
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

Benchmarks/Package.swift
```diff
@@ -61,7 +67,7 @@ var i18nTargetDependencies : [Target.Dependency] = []
 var swiftSettings : [SwiftSetting] = [.unsafeFlags(["-Rmodule-loading"]), .enableUpcomingFeature("MemberImportVisibility")]
 
 switch usePackage {
-    case .useLocalPackage(let root):
+    case .useLocalPackage(let root), .useLocalPackageRemoteDeps(let root):
         #if os(macOS)
             packageDependency.append(.package(name: "foundation-local", path: "\(root)/swift-foundation"))
             targetDependency.append(.product(name: "FoundationEssentials", package: "foundation-local"))
```
itingliu, line 73: This `foundation-local` is set at the main Package.swift. Its value depends on whether `SWIFTCI_USE_LOCAL_DEPS` is set. When it is set it's using local check, and remote one if unset.

jrflat:
There are 2 obstacles to how `swift-foundation` benchmarks are currently built and run through `swift-corelibs-foundation`:
1. `swift-corelibs-foundation` requires the `"SWIFTCI_USE_LOCAL_DEPS"` env var to be set in order to use the local copy of `swift-foundation`, so we'll need to update _that_ `Package.swift`, too (or find a way to define an env var automatically so that it knows to use the local copy of `swift-foundation`). Otherwise, we depend on the local `swift-corelibs-foundation` by default, but SCL-F still pulls in the remote `swift-foundation` dependency and doesn't pick up local changes.
2. The default `swift package benchmark` still requires a local checkout of `swift-corelibs-foundation` in the same directory as `swift-foundation`. I think this is OK but should be documented.

To solve both, maybe we can figure out a way to avoid re-exporting the packages through `swift-corelibs-foundation`, at least for running benchmarks?

itingliu:
@jrflat I'm not sure I follow why `swift-corelibs-foundation` matters here. It's not a dependency of swift-foundation. Where is it needed?

jrflat:
On non-macOS the current setup routes through a local `swift-corelibs-foundation` because that produces the `Foundation` that re-exports FoundationEssentials/I18n

```
    case .useLocalPackage(let root), .useLocalPackageRemoteDeps(let root):
        #if os(macOS)
            packageDependency.append(.package(name: "foundation-local", path: "\(root)/swift-foundation"))
            targetDependency.append(.product(name: "FoundationEssentials", package: "foundation-local"))
            targetDependency.append(.product(name: "FoundationInternationalization", package: "foundation-local"))
        #else
            packageDependency.append(.package(name: "foundation-local", path: "\(root)/swift-corelibs-foundation"))
            // Foundation re-exports FoundationEssentials and FoundationInternationalization
            targetDependency.append(.product(name: "Foundation", package: "foundation-local"))
        #endif
```

PR #1985, opened 2026-05-19
Work around ContiguousBytes cast optimization regression

jmschonfeld:
Resolves a regression in `ContiguousBytes` casting caused by `~Copyable`/`~Escapable` adoption

### Motivation:

`Data` uses `as? ContiguousBytes` casts in some of its `RangeReplaceableCollection` functions to provide fast-paths when provides `Sequence`s are known to be `ContiguousBytes` (after checking `withContiguousStorageIfAvailable`). When calling these APIs with non-generic sequences, the cast is typically folded at compile time based on whether the sequence conforms. However, when `ContigiousBytes` was updated to refine `~Copyable` and `~Escapable`, this optimization is no longer taking place resulting in expensive dynamic casts. This is particularly important because some APIs like `append(UInt8)` relied on the default implementation which call to the RRC funnel point with `CollectionOfOne` or `EmptyCollection` which do not implement `withContiguousStorageIfAvailable`.

### Modifications:

Add manual implementations of all `Data` inherited APIs that would funnel to these `as? ContiguousBytes` casts with `EmptyCollection` or `CollectionOfOne`

### Result:

By using `UnsafeRawBufferPointer` instead, the dynamic casts are eliminated because we can statically take the `withContiguousStorageIfAvailable` fast path.

### Testing:

Validated that clients calling these APIs have proper codegen.

Note: I will be following up separately to both ensure that this cast is optimized in the general case as well as implementing `withContiguousStorageIfAvailable` on `CollectionOfOne` and `EmptyCollection`

### DIFF
--- a/Sources/FoundationEssentials/Data/Data.swift
+++ b/Sources/FoundationEssentials/Data/Data.swift
@@ -695,6 +695,13 @@ public struct Data : RandomAccessCollection, MutableCollection, RangeReplaceable
         _append(buffer)
     }
 
+    @_alwaysEmitIntoClient
+    public mutating func append(_ byte: UInt8) {
+        Swift.withUnsafeBytes(of: byte) { buffer in
+            _representation.append(contentsOf: buffer)
+        }
+    }
+
     #if FOUNDATION_FRAMEWORK
     @available(macOS, obsoleted: 1.0)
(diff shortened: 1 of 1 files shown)


jmschonfeld:
@swift-ci please test

glessard (approved):

PR #1989, opened 2026-05-20
Add File List Validation to CMake CI

jmschonfeld:
Validates that the list of files built by CMake is the same as the list built by SwiftPM

### Motivation:

In order to prevent issues like https://github.com/swiftlang/swift-foundation/pull/1939 from happening in the future, we can validate that the list of files built by CMake is the same as the list built by SwiftPM

### Modifications:

Update the CMake CI to use `swift package describe` and the CMake API to produce a list of files for FoundationEssentials and FoundationInternationalization, and then diff the list of files

### Testing:

Validated via CI

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -35,10 +35,24 @@ jobs:
         uses: actions/checkout@v6
       - name: Install Dependencies
         shell: bash
-        run: apt update && apt install -y cmake ninja-build
+        run: apt update && apt install -y cmake ninja-build jq
       - name: Configure Project
         shell: bash
-        run: cmake -G 'Ninja' -B build -S . -DCMAKE_C_COMPILER=clang -DCMAKE_Swift_COMPILER=swiftc -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=YES
+        run: |
+          mkdir -p build/.cmake/api/v1/query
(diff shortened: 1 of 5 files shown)


FranzBusch (commented):
I did propose such a reusable workflow here: https://github.com/swiftlang/github-workflows/pull/229/changes#diff-8e8fa3873fa4aa830f0548d1e6280951a03da36c37f703f67624cc72433bbcc4. Would be interested to hear if that would solve your need here.

jmschonfeld:
> I did propose such a reusable workflow here: https://github.com/swiftlang/github-workflows/pull/229/changes#diff-8e8fa3873fa4aa830f0548d1e6280951a03da36c37f703f67624cc72433bbcc4. Would be interested to hear if that would solve your need here.

I don't think it would quite work as-is with our setup since we list sources across multiple `CMakeLists.txt` files rather than a single list in `add_library` (which is why this uses the CMake API to produce the combined list of sources after configuring the build) but if we were to update our CMake build a bit we might be able to make it compatible.

jmschonfeld:
@swift-ci please test

jmschonfeld:
I've added the missing files to the CMake build and now the new check passes. For context, this is an older build where it failed so you can see what a failure looks like: https://github.com/swiftlang/swift-foundation/actions/runs/26180590263/job/77022260799?pr=1989

parkera (approved):
Awesome, thank you!

PR #1991, opened 2026-05-20
add native ListFormatStyle implementation

cjchapman:
**Updated — switched the data shape to a packed C header in the style of the Swift stdlib's `ScalarPropsData.h`, in response to the earlier feedback.** The data is no longer stored as Swift collection literals.

---

Add native `ListFormatStyle` implementation behind a `FOUNDATION_LIST_FORMAT_NATIVE` build flag.

### Motivation:

`ListFormatStyle` currently bridges to ICU at runtime via `ulistfmt_*`, which makes the type unavailable on platforms without ICU and ties output to whatever ICU version the host links. The goal of this change is to ship CLDR list-pattern data inside swift-foundation as static data, alongside a Swift formatter. The new path is opt-in for now so it can land without disturbing existing consumers.

### Modifications:

The data pipeline is two-stage:

- **Stage 1 (offline, maintainer-run):** the generator (`utils/update-list-format-data` + `-impl.swift`) reads CLDR XML from a `CLDR_PATH` checkout and emits a committed JSON intermediate at `Sources/FoundationInternationalization/Formatting/ListFormatData.json`. The resolution algorithm mirrors ICU's interleaved alias + locale fallback (e.g. `es or-narrow → or-short → or` redirects, with locale fallback restarted at each step so `es::or` is found before `root::or`). The emission is sparse — only entries whose resolved row differs from the locale's parent's row are kept — and the parent map is greedily simplified by trying to drop entries that don't change any lookup. The generator self-checks correctness by simulating the runtime walk against the fully resolved table.

- **Stage 2 (build-time):** `utils/build-list-format-data` reads the JSON plus `--locales=...` and `--fallback=...` arguments and emits a packed C header at `Sources/_FoundationCShims/include/_ListFormatData.h`. Slot and parent entries reference locale identifiers indirectly through a shared `_ListFormatLocales[]` string pool, indexed by `uint16_t`, so each entry is 4 bytes rather than two pointers. The pattern pool and row table are transitively pruned to only what's reachable from the requested locale set, so subsetted builds compile in only the data they actually need. The configured fallback locale (default `root`) is auto-included in the kept set.

- Added `ListFormatPatterns.swift`: the `ListPatterns` struct, a `ListPatternCondition` enum for the contextual rules (Spanish y→e, Spanish o→u, Hebrew vav prefix, Thai joiner), the locale-walking lookup, and predicate functions translated from `i18n/listformatter.cpp`. The lookup walks the parent chain and binary-searches the C arrays via `strcmp` on the pooled locale strings, avoiding per-step Swift `String` allocation. When the walk exhausts without a match, runtime consults the fallback locale's row; direct lookups for `"root"` still return root's data. Conditions aren't stored in the data — they're computed at format time from `(locale.language, type, default-pattern)` so the data tables stay deduped against actual patterns and the rule definitions stay in one place.

- Added `NativeListFormatter.swift`: the empty/single/pair/3+ formatting algorithm, plus FSI/PDI bidi-isolate wrapping for items whose direction conflicts with the list's overall direction. Direction detection reuses `BuiltInUnicodeScalarSet`'s strong-RTL bitmap via a new `package` predicate `Unicode.Scalar._isStrongRTL`.

- Wired the build flag into `ListFormatStyle.format(_:)` (single conditional at the call site). The runtime sources, the generated C header, and its `#include` line in `_FoundationCShims.h` are all wrapped in `#if FOUNDATION_LIST_FORMAT_NATIVE` so they add no code or data to the binary when the flag is off.

### Result:

- With the flag off (default), behavior is unchanged — `ICUListFormatter` continues to serve all `ListFormatStyle` calls.

- With `FOUNDATION_LIST_FORMAT_NATIVE` defined, `ListFormatStyle` formats lists without any ICU dependency, sourcing patterns from the packed C header.

- For local testing: pass both `-Xswiftc -DFOUNDATION_LIST_FORMAT_NATIVE` and `-Xcc -DFOUNDATION_LIST_FORMAT_NATIVE` to `swift build` / `swift test`. Both are required because the C-side `-D` is what gates the data inclusion in `_FoundationCShims.h`.

### Testing:

- Ported the bulk of ICU's `listformattertest.cpp` into `Tests/FoundationInternationalizationTests/Formatting/ListFormatStyleTests.swift`: locale coverage (root, en, en_US, en_GB, nn, zh_Hant_HK, ru, ml, zu), the pt/en type×width matrix, all four contextual rules (Spanish y→e, Spanish o→u, Hebrew vav prefix, Thai joiner), literal-brace and empty-string edge cases, and FSI/PDI bidi wrapping. The full suite passes in both flag states (32/32). The Malayalam test is conditioned on the flag because upstream CLDR/ICU has switched to commas-only output, but `swift-foundation-icu` still has the older patterns.

- The stage-1 generator runs a sparse-lookup self-check at emit time: every `(locale, slot)` resolved value must be reachable through the runtime walk against the emitted sparse data. Both stage outputs are byte-deterministic across runs (sorted JSON keys; sorted slot tables and locale pool).


parkera:
You would perhaps expect that the generated Swift code produces const data, but unfortunately we do not have support for that in Swift yet. (https://github.com/swiftlang/swift-evolution/blob/main/proposals/0359-build-time-constant-values.md).

I also see a lot of duplicated data in it. For example: `ListPatterns(start: "{0}, {1}", middle: "{0}, {1}", ...` appears 783 times. That could probably be deduplicated far more efficiently in a binary format of some kind. Even binary property list will deduplicate strings (although not short strings, iirc).

One more optimization idea -- I assume that locale identifiers will be very common in this kind of data. Should we define them as integers as well, so that `ab` is always 0, `af` is always 1, etc -- then deduplicate those as well, across different entries?

Or maybe if we relied upon a common compression format for storing all of this we could get some of the de-duping behavior for free.

parkera:
The list format parent locales is a good example of where I think we'd benefit from some more thinking about compression, given that the vast majority have the same parent value of `en_001`.

cjchapman:
I'm happy to pack the data into a more space-efficient binary format.

Is a C header file the right place to store the data? Or some other way? Any examples you can point me to?

parkera:
I don't think we have any examples in this project, yet.

We'll have to brainstorm about the right way to do it, and some requirements that we should put in place to allow clients to pick and choose which data they want to include.

cjchapman:
The commit I just pushed show the full (all locales) data file in `_ListFormatData.h`.

Here's an example of the data subsetted for just English (`en`) & Spanish (`es`) with `en_001` designated as the fallback if the user requests a locale that doesn't fall back to either of those. (I had to change the extension from `.h` to `.c` for GitHub to let me attach it.):
[_ListFormatData_subset.c](https://github.com/user-attachments/files/28158677/_ListFormatData_subset.c)

cjchapman:
The commit I just pushed adds some benchmarks and some performance improvements. Here's where things stand now:

| Benchmark | ICU (ns) | Native (ns) | Speedup |
|---|---:|---:|---:|
| en-empty | 331 | 259 | 1.28× |
| en-single | 625 | 405 | 1.54× |
| en-pair | 826 | 520 | 1.59× |
| en-three | 1,083 | 673 | 1.61× |
| en-long (20 items) | 6,136 | 2,329 | 2.63× |
| es-contextual | 1,242 | 762 | 1.63× |
| he-contextual | 924 | 627 | 1.47× |
| th-contextual | 1,046 | 698 | 1.50× |
| bidi-mixed | 1,128 | 720 | 1.57× |
| multi-locale | 11,000 | 7,000 | 1.57× |
| fresh-style | 1,149 | 704 | 1.63× |

p50 wall-clock per call.
Captured with `BENCHMARK_DISABLE_JEMALLOC=1`.
Speedup is `icu_time / native_time`.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,400 @@
...
+            self.canBuildLinearly = true
+        } else {
+            self.canBuildLinearly = false
+        }
+    }
+
+    func format(strings: [String]) -> String {
+        switch strings.count {
+        case 0: return ""
+        case 1: return wrapBidi(strings[0])
```
itingliu, line 269: Would we be able to add a fast path if we know the language doesn't need BIDI at all?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,400 @@
...
+    }
+
+    func format(strings: [String]) -> String {
+        switch strings.count {
+        case 0: return ""
+        case 1: return wrapBidi(strings[0])
+        case 2:
+            let a = wrapBidi(strings[0]), b = wrapBidi(strings[1])
+            if isThaiAnd, let parsed = pairParsed {
+                return applyThai(parsed: parsed, isPair: true, before: a, after: b)
```
itingliu: I guess I have a similar question here as the BIDI comment above: should the formatter have the concept of a language code? say instead of `var isThaiAnd: Bool` would it make more sense to store it as a language code? We can then use the language code to check if it's Thai or whether we need to do BIDI path

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,400 @@
...
+        result += wrapBidi(items[0])
+        result += start.connector
+        result += wrapBidi(items[1])
+        for i in 2..<(items.count - 1) {
+            result += middle.connector
+            result += wrapBidi(items[i])
+        }
+        result += end.connector
+        result += wrapBidi(items[items.count - 1])
+        return result
```
itingliu: I think we can get some nice performance wins if we use `Span` instead of appending `String`s after one another. You can take a look at URL_Impl.swift for example, or the `format` function in `DateComponents+ISO8601FormatStyle.swift`

(The latter uses an in-house version of Span called `OutputBuffer`, but the idea is the same)

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: No, this code is meant to handle situations like an English-locale list containing an Arabic word, or an Arabic-locale list containing an English word, so we have to examine the items.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: See my earlier comment about the bidi part. Regarding the Thai part, we do have (and use) a language code:
```swift
self.isThaiAnd = (language == "th") && (signature.listType == 0)
```

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: I see. Thanks. But if we somehow know that the client doesn't support Arabic in their app, say, we get this information at build time (assuming that's when we will be excluding Arabic list format templates), would it be possible to do something about that?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Yes, that helped. Thanks!

cjchapman:
Here's the performance with the commit I just pushed which follows the examples of `URL_Impl.swift` and `DateComponents+ISO8601FormatStyle.swift` as Tina suggested (thanks Tina!):

| Benchmark              |  ICU (ns) | Native (ns) | Speedup |
|------------------------|----------:|------------:|--------:|
| en-empty               |       331 |         246 |   1.35× |
| en-single              |       625 |         380 |   1.64× |
| en-pair                |       826 |         471 |   1.75× |
| en-three               |     1,083 |         587 |   1.84× |
| en-long (20 items)     |     6,136 |       1,835 |   3.34× |
| es-contextual          |     1,242 |         657 |   1.89× |
| he-contextual          |       924 |         567 |   1.63× |
| th-contextual          |     1,046 |         624 |   1.68× |
| bidi-mixed             |     1,128 |         620 |   1.82× |
| multi-locale           |    11,000 |       6,230 |   1.77× |
| fresh-style            |     1,149 |         608 |   1.89× |

p50 wall-clock, jemalloc disabled

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: You could have an app that only supports a language like English and still have a loan word from another language (such as Arabic or Hebrew) in an otherwise English list. This code is meant to handle such situations. The bidi check already has an ASCII fast path (see `direction(of item: String)` in `NativeListFormatter.swift`), and this new code is now more performant than the code it's meant to replace, so to me it doesn't seem worthwhile to take on the risk (and complexity) of disabling it when the list of supported locales doesn't include languages with RTL scripts.

cjchapman:
My earlier commit didn't take Tina's 'Span' suggestion literally, but the one I just pushed does (using `UTF8Span` and `OutputSpan` specifically).

Here's the latest performance:

| Benchmark              |  ICU (ns) | Native (ns) | Speedup |
|------------------------|----------:|------------:|--------:|
| en-empty               |       331 |         255 |   1.30× |
| en-single              |       625 |         385 |   1.62× |
| en-pair                |       826 |         476 |   1.74× |
| en-three               |     1,083 |         606 |   1.79× |
| en-long (20 items)     |     6,136 |       1,937 |   3.17× |
| es-contextual          |     1,242 |         677 |   1.83× |
| he-contextual          |       924 |         577 |   1.60× |
| th-contextual          |     1,046 |         636 |   1.64× |
| bidi-mixed             |     1,128 |         639 |   1.77× |
| multi-locale           |    11,000 |       6,713 |   1.64× |
| fresh-style            |     1,149 |         652 |   1.76× |

p50 wall-clock, jemalloc disabled

marcprux:
I'm late to the party here, but I am keenly interested in this due to ICU app size being a major barrier to Swift adoption on Android (see https://forums.swift.org/t/android-app-size-and-lib-foundationicu-so/78399).

Is there a broader context for this change? I understand the goal is to be able to implement `ListFormatStyle` without needing ICU's `ulistfmt_*`. Are there other efforts along these lines, such as for number/date formatting/parsing or string collation? Is the goal to just be able to pull `ListFormatStyle` into `FoundationEssentials` and enable its use without the burden of `FoundationInternationalization`, or is this part of some larger project to divvy up i18n-related functionality to be closer to the APIs that utilize it?

parkera:
@cjchapman is doing some exploration of approaches that may allow us to reduce the size of the ICU dependency. It's all early prototyping at this stage (thus the draft PR).

cjchapman:
That force-push (about 45 minutes ago) was a rebase and refactoring based on: https://github.com/swiftlang/swift-foundation/pull/1999

utils/update-list-format-data-impl.swift
```diff
@@ -0,0 +1,564 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/*
+
+ Swift script that reads CLDR XML and emits a Swift source file containing
```
richgillam: It doesn't emit a Swift source file, does it?  I think it emits a JSON file.  (I think there are several other references to this code emitting Swift source code in here...)

Also, do you need to check in the JSON file?  It seems like it could be treated as an intermediate build product.

utils/build-list-format-data-impl.swift
```diff
@@ -0,0 +1,366 @@
...
+    ? nil
+    : Set(localesArg.split(separator: ",").map(String.init))
+
+// MARK: - Load JSON
+
+let bytes: Data
+do {
+    bytes = try Data(contentsOf: URL(fileURLWithPath: inputPath))
+} catch {
+    FileHandle.standardError.write("error: failed to read \(inputPath): \(error)\n".data(using: .utf8)!)
```
richgillam: Maybe I'm betraying my Swift ignorance, but all of these lines referring to `FileHandle.standardError` felt a little overcomplicated to me.  It seems like you ought to be able to do this kind of thing with `print()`.  I did a little bit of research and `print()` takes a `TextOutputStream`, and there's apparently no built-in `TextOutputStream` that writes to stderr.  A thread in the Swift forum (https://forums.swift.org/t/how-to-print-to-stderr-swift-6-1/79863) suggested writing a custom `TextOutputStream` to do this-- it's only seven lines of code-- and it might make all these writes look a little more idiomatic (you wouldn't need the `.data(using: .utf8)` stuff, for example).

utils/update-list-format-data-impl.swift
```diff
@@ -0,0 +1,564 @@
...
+
+    var allLocales: [String] = []
+    var data: [String: LocaleData] = [:]
+    for file in xmlFiles {
+        let locale = file.deletingPathExtension().lastPathComponent
+        allLocales.append(locale)
+        // Cheap pre-filter: most files have no <listPatterns> block.
+        // Parsing every CLDR file via XMLDocument is slow; skip when possible.
+        guard let bytes = try? Data(contentsOf: file),
+              let s = String(data: bytes, encoding: .utf8),
```
richgillam: Isn't there a `String(contentsOf:)` you could use here instead of doing it in two steps?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+        let listType: Int
+        let width: Int
+    }
+
+    /// A list pattern of the form `<prefix>{0}<connector>{1}<suffix>` parsed
+    /// once at formatter-construction time. The Thai fast path uses this to
+    /// avoid re-scanning the pattern on every format call. Slices are
+    /// materialized into concrete `String`s so format-time concatenation can
+    /// take the fast byte-copy path rather than scalar-by-scalar append.
+    internal struct ParsedPattern: Equatable {
```
richgillam: I think ICU has something called `SimpleFormat` that basically does this, and as we expand this to cover more areas of functionality, I think we're going to want a similar class that does this.  I think you should think about factoring the code that takes a pattern like `"{0} and {1}"` and plugs other strings into the `{0}` and `{1}` into a general utility class that everyone can use.

richgillam (commented):
A few first comments.  I may have more tomorrow.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,505 @@
...
+            for scalar in item.unicodeScalars {
+                // ASCII is never strong-RTL and never an Arabic number, so skip
+                // the bitmap/range checks for it (the common case) — calling
+                // `_isStrongRTL` per ASCII scalar would pay a full bitmap lookup.
+                if scalar.value < 0x80 { continue }
+                if scalar._isStrongRTL || isArabicNumber(scalar) { return true }
+            }
+        }
+        return false
+    }
```
richgillam, line 574: All of the `wrapBidi()` stuff seems like it could be factored out too.  Maybe this functionality could be included in the `SimpleFormat` class I was suggesting earlier (or into another class that either uses or is used by `SimpleFormat`?).  I suspect we're going to need _that_ in a lot more places as well.

(I think I asked you offline if this already existed in Foundation and the answer was that it's in CoreFoundation but not in `swift-foundation`, so maybe this is a first step toward porting that code into Swift from CF.)

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,505 @@
...
+        }
+        if scalar._isStrongRTL { return false }
+        switch scalar.properties.generalCategory {
+        case .uppercaseLetter, .lowercaseLetter, .titlecaseLetter,
+             .modifierLetter, .otherLetter:
+            return true
+        default:
+            return false
+        }
+    }
```
richgillam: It'd be nice if we didn't have to do this.  If the stdlib functions aren't fast or simple enough to support this use case without this kind of extra work, it kind of feels like we should be trying to fix it (or get somebody else to fix it) there.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,505 @@
...
+             0x0660...0x0669,  // Arabic-Indic digits
+             0x066B...0x066C,  // Arabic decimal / thousands separators
+             0x06DD,           // Arabic end of ayah
+             0x0890...0x0891,  // Arabic pound / piastre marks
+             0x08E2:           // Arabic disputed end of ayah
+            return true
+        default:
+            return false
+        }
+    }
```
richgillam: This functionality should be in the stdlib, along with all the other Unicode character property queries.  I don't know if we need to file a Radar or something, but at the least, I think I'd put a comment here indicating that, along with a TODO.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+        default:
+            return false
+        }
+    }
+
+    private static func language(of locale: String) -> String {
+        if let u = locale.firstIndex(of: "_") { return String(locale[..<u]) }
+        if let u = locale.firstIndex(of: "-") { return String(locale[..<u]) }
+        return locale
+    }
```
richgillam, line 601: At some point, we should be using the `Locale` object in here instead of using strings to denote the locale, and I think (hope?) there's already API on `Locale` to do this.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+    private let listDirection: Locale.LanguageDirection
+    /// Whether this formatter's locale + type combination triggers the Thai
+    /// contextual joiner rule. Cached to avoid re-deciding per format call.
+    private let isThaiAnd: Bool
+    /// Cached condition decisions per slot so format calls don't re-classify
+    /// the rule per call. `nil` means no contextual rule applies for this slot.
+    private let startCondition: ListPatternCondition?
+    private let middleCondition: ListPatternCondition?
+    private let endCondition: ListPatternCondition?
+    private let pairCondition: ListPatternCondition?
```
richgillam: All of the code for handling contextual substitutions seems really complicated and obscures the logic flow for somebody reading this code.  I wonder if it's possible to improve on what we have here?  (I'll think about this more and see if I can come up with a more concrete suggestion.)

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// MARK: - Binary search over the C tables
+
+/// Look up `locale` in the slot table for `(type, width)`. Returns the row
+/// index if present.
+private func _searchSlot(locale: String, type: Int, width: Int) -> UInt16? {
+    switch (type, width) {
+    case (0, 0):
+        return withUnsafePointer(to: _ListFormatSlot_AndWide) { ptr in
```
richgillam: I wonder if there's a way to get rid of all this `withUnsafePointer` stuff?  Maybe not, since we're looking at a C source file containing static data.  And you're containing it here where nobody else has to worry about it, so maybe it's okay this way.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#endif
+
+internal import _FoundationInternationalizationData
+
+// MARK: - Types
+
+/// Resolved list-formatting patterns for one (type, width) slot, parameterized by locale.
+internal struct ListPatterns: Hashable, Sendable {
```
richgillam: The class is called `ListPatterns`, but the file is `ListFormatPatterns`.  Do the names need to be different?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+/// extensions (`ThaiHandler`'s contextual joiner and FSI/PDI bidi isolate
+/// wrapping, which isolates any item containing a character of the opposite
+/// direction to the list, matching `ListFormatter::needsBidiIsolates`).
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+internal final class NativeListFormatter: @unchecked Sendable {
+
+    struct Signature: Hashable {
+        let localeIdentifier: String
+        let listType: Int
+        let width: Int
```
richgillam: It'd make the code more readable (both here and in `ListFormatPatterns`) if `listType` and `width` were enums, or at least had constants for the values.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+    switch (type, width) {
+    case (0, 0):
+        return withUnsafePointer(to: _ListFormatSlot_AndWide) { ptr in
+            _bsearchSlot(target: locale, base: UnsafeRawPointer(ptr).assumingMemoryBound(to: _ListFormatSlotEntry.self),
+                         count: Int(_ListFormatSlot_AndWide_Count))
+        }
+    case (0, 1):
+        return withUnsafePointer(to: _ListFormatSlot_AndShort) { ptr in
+            _bsearchSlot(target: locale, base: UnsafeRawPointer(ptr).assumingMemoryBound(to: _ListFormatSlotEntry.self),
+                         count: Int(_ListFormatSlot_AndShort_Count))
```
richgillam: All of the cases in this function look the same, except for the range of memory you're pointing to.  It'd probably read better if the switch just selected the pointer and count and the common code was down below.  It'd also read better if `type` and `width` were enums or at least had symbolic constants for their values.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+                    let entry = base[mid]
+                    let cmp = strcmp(cChild, pool[Int(entry.child)]!)
+                    if cmp == 0 { return String(cString: pool[Int(entry.parent)]!) }
+                    if cmp > 0 { lo = mid &+ 1 } else { hi = mid &- 1 }
+                }
+                return nil
+            }
+        }
+    }
+}
```
richgillam: Some functions in this class, such as `_parentLookup` and `_bsearchSlot`, seem like general functions that could apply to a lot of other data types.  Have you given any thought to how this approach to handling the data would generalize out to handle other pieces of data in the CLDR database, and which pieces of this logic could be factored out into a class all the internationalization components could share?

richgillam (commented):
More comments.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+                let pool = UnsafeRawPointer(poolPtr).assumingMemoryBound(to: UnsafePointer<CChar>?.self)
+                var lo = 0
+                var hi = Int(_ListFormatParentCount) - 1
+                while lo <= hi {
+                    let mid = (lo &+ hi) / 2
+                    let entry = base[mid]
+                    let cmp = strcmp(cChild, pool[Int(entry.child)]!)
+                    if cmp == 0 { return String(cString: pool[Int(entry.parent)]!) }
+                    if cmp > 0 { lo = mid &+ 1 } else { hi = mid &- 1 }
+                }
```
richgillam: Too bad we have to code up our own binary search here.  Is there any utility function (maybe even on the C side, since we're searching a C data structure) that'd save us this?

Sources/FoundationInternationalization/Formatting/ListFormatStyle.swift
```diff
@@ -153,8 +153,12 @@ public struct ListFormatStyle<Style: FormatStyle, Base: Sequence>: FormatStyle w
...
     /// - Returns: A string representation of the provided sequence.
     public func format(_ value: Base) -> String {
-        let formatter = ICUListFormatter.formatter(for: self)
-        return formatter.format(strings: value.map(memberStyle.format(_:)))
+        let strings = value.map(memberStyle.format(_:))
+        #if FOUNDATION_LIST_FORMAT_NATIVE
+        return NativeListFormatter.formatter(for: self).format(strings: strings)
+        #else
+        return ICUListFormatter.formatter(for: self).format(strings: strings)
+        #endif
```
richgillam, line 161: Do we have a process for determining when we can get rid of `FOUNDATION_LIST_FORMAT_NATIVE` and `ICUListFormatter` and just use the new class (hopefully renamed) all the time?

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+    if language == "th", type == 0 {
+        return .thaiContextual
+    }
+    return nil
+}
+
+// MARK: - Predicate implementations
+
+internal func _spanishYToEFires(on text: String) -> Bool {
+    var iter = text.unicodeScalars.makeIterator()
```
richgillam: With this and the functions under it, this seems like a complicated way to get what we're after, and is hard to read.  First, it doesn't seem to make sense to have to go down to the `unicodeScalars` to do these tests.  Couldn't you just do this at the `String` level?  It seems like you could just do something like

```
let lc = text.lowercased
if lc.starts(with:"i") || (lc.starts(with:"hi") && !lc.starts(with:"hie") && !lc.starts(with:"hia")) {
    return true
}
return false
```

What you have looks like a direct translation of the C++ code and is probably more performance-optimized, but how performance-optimized do we have to be here, and is there a way to do it that's more readable?

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+/// A locale-specific contextual rule that may modify the `end`/`pair` pattern
+/// based on the surrounding text. Application happens in
+/// `NativeListFormatter.apply`; the data tables don't carry the tag because
+/// it's determined by `(locale.language, type, default-pattern)` at format time.
+internal enum ListPatternCondition: Sendable, Hashable {
+    case spanishYToE
+    case spanishOToU
+    case hebrewNonHebrewPrefix
+    case thaiContextual
+}
```
richgillam: I'd like to suggest an alternative approach to the contextual patterns that I think is simpler.

First, all the contextual patterns are "2" and "end" patterns, and they're all in languages where the "2" and "end" patterns are the same.  So you don't need to do anything with the "start" and "middle" patterns, and the logic for the "2" and "end" patterns can be the same.

Have `NativeListFormatter` have two members for contextual changes: `alternativeCondition` and `alternativePattern`.  Both of these are set up at construction time: The `alternativeCondition` is the function returned by `_listPatternCondition()` for the language and formatter type.   The `alternativePattern` is the list format pattern you want to use if the condition fires (for Spanish "and", it's `{0} e {1}`).  Both are optional.  If you're about to use the "2" and "end" pattern, you check `alternativeCondition` (if it exists) and if it returns true, you use `alternativePattern` to format the string instead of the regular "2" and "end" pattern (which I'm assuming will be the same).  You only need two members, you don't need the enum, and everything's ready to go.

This works better, I think, if you factor out the actual formatting code into a separate `SimpleFormat` object that takes the pattern string, as I suggested earlier, but I think it'll leave you with something simpler and easier to understand and maintain.

I'd also recommend moving `_listPatternCondition()` and the predicate functions out of `ListFormatPatterns` and into either `NativeListFormat` or a new class-- I think it'd be better to keep `ListFormatPatterns` focused on loading the regular resource data.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+            var lo = 0
+            var hi = count - 1
+            while lo <= hi {
+                let mid = (lo &+ hi) / 2
+                let entry = base[mid]
+                let cmp = strcmp(cTarget, pool[Int(entry.locale)]!)
+                if cmp == 0 { return entry.row }
+                if cmp > 0 { lo = mid &+ 1 } else { hi = mid &- 1 }
+            }
+            return nil
```
richgillam: Same question about existing binary-search functions here.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
```diff
@@ -0,0 +1,280 @@
...
+        let base = UnsafeRawPointer(ptr).assumingMemoryBound(to: UnsafePointer<CChar>?.self)
+        return String(cString: base[Int(index)]!)
+    }
+}
+
+// MARK: - Contextual conditions
+
+/// Returns the contextual rule that applies to a given `(language, type,
+/// default-pattern)`, or `nil` if no rule applies.
+internal func _listPatternCondition(language: String, type: Int, defaultPattern: String) -> ListPatternCondition? {
```
richgillam: You really want the type to be an enum or named constant.

richgillam (commented):
More comments...

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,505 @@
...
+    /// Pre-classifies bidi conflicts in a sizing pass (using a stack-allocated
+    /// `Bool` buffer via `withUnsafeTemporaryAllocation`), then writes UTF-8
+    /// bytes directly into a `String(unsafeUninitializedCapacity:)` buffer
+    /// wrapped in an `OutputSpan`. Skips the per-`+=` UTF-8 validity checks
+    /// and copy-on-write uniqueness checks `String +=` paid on every append.
+    private func buildLinear(items: [String],
+                             start: ParsedPattern,
+                             middle: ParsedPattern,
+                             end: ParsedPattern) -> String {
+        return withUnsafeTemporaryAllocation(of: Bool.self, capacity: items.count) { conflicts in
```
itingliu, line 371: Can we use `withTemporaryAllocation`?

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,505 @@
...
+            for i in items.indices {
+                let wrap = shouldWrapBidi(items[i])
+                conflicts.initializeElement(at: i, to: wrap)
+                total += items[i].utf8.count + (wrap ? 6 : 0)
+            }
+            total += start.connector.utf8.count
+                + middle.connector.utf8.count * (items.count - 3)
+                + end.connector.utf8.count
+
+            return String(unsafeUninitializedCapacity: total) { buffer in
```
itingliu, line 383: Is there a safe alternative?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+
+    /// Apply a contextual rule (if any) to `pattern`, then interpolate.
+    /// `before` substitutes for `{0}`; `after` for `{1}`. Top-level fast paths
+    /// in `format(strings:)` cover the no-condition case for simple patterns
+    /// and the Thai case; this path handles Spanish/Hebrew, plus the
+    /// no-condition case for non-simple patterns (those with a non-empty
+    /// prefix/suffix), plus a fallback for malformed data.
+    private func apply(_ pattern: String, condition: ListPatternCondition?,
+                       parsed: ParsedPattern?, isPair: Bool,
+                       before: String, after: String) -> String {
```
itingliu: Experiment with changing into `before: inout String` instead of returning a String

And/or see if we can introduce an internal "StringBuilder" helper

Tests/FoundationInternationalizationTests/Formatting/ListFormatStyleTests.swift
```diff
@@ -61,25 +61,425 @@ private struct ListFormatStyleTests {
...
+        let arabicIndic = "\u{0661}\u{0662}\u{0663}"   // ١٢٣ — Bidi_Class AN
+        let persian = "\u{06F1}\u{06F2}\u{06F3}"       // ۱۲۳ — Bidi_Class EN
+        #expect([arabicIndic].formatted(s) == isolate(arabicIndic))
+        #expect([arabicIndic, arabicIndic].formatted(s) == "\(isolate(arabicIndic)) and \(isolate(arabicIndic))")
+        #expect([persian, persian].formatted(s) == "\(persian) and \(persian)")
+    }
+
+    // MARK: - Internal pattern parser
+
+    #if FOUNDATION_LIST_FORMAT_NATIVE
```
itingliu: Can we move these tests into a separate file?

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+///
+/// Reimplements Apple-ICU's `i18n/listformatter.cpp`
+/// (`ListFormatter::formatStringsToValue` and its `PatternHandler` family):
+/// the CLDR "List Patterns" empty/single/pair/3+ composition, the contextual
+/// rules (Spanish y→e / o→u and the Hebrew non-Hebrew prefix), and the Apple
+/// extensions (`ThaiHandler`'s contextual joiner and FSI/PDI bidi isolate
+/// wrapping, which isolates any item containing a character of the opposite
+/// direction to the list, matching `ListFormatter::needsBidiIsolates`).
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
+internal final class NativeListFormatter: @unchecked Sendable {
```
itingliu: This should just be Sendable, no unchecked

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+    /// take the fast byte-copy path rather than scalar-by-scalar append.
+    internal struct ParsedPattern: Equatable {
+        let prefix: String
+        let connector: String
+        let suffix: String
+        let connectorStartsWithSpace: Bool
+        let connectorEndsWithSpace: Bool
+    }
+
+    internal static let cache = FormatterCache<Signature, NativeListFormatter>()
```
itingliu: We prefer to use `Mutex` instead of `FormatterCache`. The latter was introduced as a workaround before Mutex was available

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
itingliu: Can we consolidate these four variables into one check if the language doesn't need special conditions? If I read it correctly, for most use cases these would all be nil

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
```diff
@@ -0,0 +1,505 @@
...
+            }
+            return apply(patterns.pair, condition: pairCondition, parsed: pairParsed,
+                         isPair: true, before: a, after: b)
+        default:
+            // Fast path: simple patterns + no contextual rule. Build the
+            // result in one buffer in a single pass (O(N) total work) rather
+            // than producing N intermediate strings. `wrapBidi` is inlined
+            // so we don't allocate an intermediate `[String]` for the wrapped
+            // items.
+            if canBuildLinearly,
```
itingliu: Since this `canBuildLinearly` is used only once here, can we just discard this variable and check directly with the following before using it? 
```
if s = startParsed, let m = middleParsed, let e = endParsed,
           s.prefix.isEmpty, s.suffix.isEmpty,
           m.prefix.isEmpty, m.suffix.isEmpty,
           e.prefix.isEmpty, e.suffix.isEmpty, startCondition == nil, middleCondition == nil, endCondition == nil,
               let startP = startParsed, let middleP = middleParsed, let endP = endParsed {
    return buildLinear(items: strings, start: startP, middle: middleP, end: endP)
}
```

Sources/FoundationInternationalization/Formatting/ListFormatStyle.swift
cjchapman: As of commit c54f4adbce966a24c75e9760250993e55c22d587, the Swift implementation now builds and runs by default, and the ICU path is opt-in behind FOUNDATION_LIST_FORMAT_ICU (renamed from FOUNDATION_LIST_FORMAT_NATIVE).

I'm planning to do some more renaming later, after I've worked my way through the current feedback.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Fixed in commit 30b0c35f13642424c5f7a9d3e398e26e16584c20.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
itingliu: We don't have one. As far as I know we already have two custom implementations of binary search in swift-foundation. 

This does look pretty similar to the one above so maybe at lease we can see if we want to consolidate them

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Done in commit 6ac6da49186d00629d581a9961dcb6e9f4451431.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Done in commit 6ac6da49186d00629d581a9961dcb6e9f4451431.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Done in commit 6ac6da49186d00629d581a9961dcb6e9f4451431.

utils/update-list-format-data-impl.swift
cjchapman: Fixed in commit c2a4af42de240444c4a68f7c57fd0fb251136242.

Yes, we need to check in the JSON file. It's a compact representation of all the list formatting data we need from CLDR since we won't have access to the full CLDR data at the time we subset the data to the locales needed.

utils/build-list-format-data-impl.swift
cjchapman: Done in commit c2a4af42de240444c4a68f7c57fd0fb251136242.

utils/update-list-format-data-impl.swift
cjchapman: Done in commit c2a4af42de240444c4a68f7c57fd0fb251136242.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Fixed in commit 411be3927c0e8740d848df6ec277816bd5b9b097.

utils/update-list-format-data-impl.swift
richgillam: >Yes, we need to check in the JSON file.

We've talked offline, but I'll say here that I'm still skeptical of this claim.  I'm concerned that the JSON files aren't that much smaller than the original XML files and that they're a whole lot less human-readable, and I'm wondering whether we can either do subsetting straight from a full copy of the binary data, or just publish the CLDR XML data in another repo or something just so that it doesn't clog up this one.  (Maybe we could use the regular Unicode CLDR repo, along with some collection of Apple-specific overlay files?

I'm fine with what you have right now for now (i.e., I don't mind if you merge this PR as is), but I think this part of the issue deserves more discussion.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Done in commit 6421c0b934357e02a6a4c04e73c6ab39c5f5a81d.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Done in commit 6421c0b934357e02a6a4c04e73c6ab39c5f5a81d.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Done in commit 59c1e54f5da00d6b66e42d13edb0ee19d8709bb2.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Introduced a `SimpleFormatter` helper in commit 59c1e54f5da00d6b66e42d13edb0ee19d8709bb2. It formats by appending into a caller-supplied buffer, and the 3+ item path now builds into a single buffer instead of returning/re-allocating an intermediate string per join.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Added a "TODO" comment in commit 78542f497c5fc19a79154b87d74b98149a868afa to remind us to use `withTemporaryAllocation` once `swift-tools-version` is 6.4 or higher.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Done in commit 9ae4b8928ad2c26c9206ef3692e37bab6f38e218.

Sources/FoundationInternationalization/Formatting/NativeListFormatter.swift
cjchapman: Done in commit 2064b4527584ead48264078499b0a7328869747e.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,582 @@
...
+
+/// A compiled `<prefix>{0}<connector>{1}<suffix>` pattern — a minimal Swift
+/// analogue of ICU's `SimpleFormatter`. The positional placeholders are parsed
+/// once (`init?`), then formatting appends into a caller-supplied buffer (ICU's
+/// `appendTo` model) so a whole list can be built in a single allocation.
+///
+/// Internal for now, but shaped to graduate into a shared formatter utility:
+/// `Duration+UnitsFormatStyle` open-codes the same `{0}`/`{1}` substitution, and
+/// other CLDR-driven formatters will want it too.
+internal struct SimpleFormatter: Equatable {
```
itingliu, line 89: If I read this description correctly, this is not only useful for List formatter, but also a base for other pattern-based formatter. Would it make sense to give it a descriptive name for that role? Like, PatternBasedFormatter or something like that

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Struct renamed to `ListFormatPatterns` in commit 787defb437a32b47e72168d86e44fe6766f19b89.

Tests/FoundationInternationalizationTests/Formatting/ListFormatStyleTests.swift
cjchapman: Moved to `SimpleFormatterTests.swift‎` in commit 5eb84c572afd2ea7d90b4f78ccee5d6ccc9c429b.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+    /// or Arabic number (weak RTL, `AN`); in an RTL list, if it contains any
+    /// strong-LTR character (`L`). Scanning the whole item (rather than only its
+    /// leading direction) catches a same-leading item with a trailing opposite
+    /// run, e.g. "David \u{05DB}\u{05D4}\u{05DF}" in an English list. Factored out so
+    /// `buildLinear` can pre-classify items (sizing the output buffer exactly)
+    /// without materializing the wrapped strings.
+    @inline(__always)
+    private func shouldWrapBidi(_ item: String) -> Bool {
+        if listDirection == .rightToLeft {
+            for scalar in item.unicodeScalars where isStrongLTR(scalar) {
```
itingliu: Can we add a fast path for all -ascii Strings? You can check so with `item.utf8Span.isKnownASCII`?

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+    @inline(__always)
+    private func isStrongLTR(_ scalar: Unicode.Scalar) -> Bool {
+        let v = scalar.value
+        // ASCII fast path: ASCII letters are strong LTR; nothing else in the
+        // ASCII range is strong-directional.
+        if v < 0x80 {
+            return (v >= 0x41 && v <= 0x5A) || (v >= 0x61 && v <= 0x7A)
+        }
+        if scalar._isStrongRTL { return false }
+        switch scalar.properties.generalCategory {
```
itingliu: Can we add a TODO for BIDI_Category if/when it becomes available in stdlib? If it's not available or in the work, we probably want to follow up with stdlib for this ER.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: Generally I think we'd want to avoid iterating through unicodeScalars if we can

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+    /// behave as ordinary LTR and don't disturb the separators. This matches
+    /// Apple-ICU's `needsBidiIsolates` (which triggers on `AN` but not `EN`).
+    ///
+    /// TODO: Replace these hand-coded ranges with a real Bidi_Class lookup once
+    /// swift-foundation exposes one without depending on ICU. The bundled
+    /// CFUniChar bitmaps cover the strong-RTL set used by `_isStrongRTL` but
+    /// carry no Arabic-number / bidi-class data, so the ranges are maintained by
+    /// hand for now (and don't cover the SMP `AN` ranges, e.g. Rumi numerals).
+    @inline(__always)
+    private func isArabicNumber(_ scalar: Unicode.Scalar) -> Bool {
```
cjchapman: Note to self: add a comment here to using bidiCategory instead when it becomes available.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: Meanwhile we can consolidate these helper functions into a utility file like this 
```
extension Unicode.Scalar {
     var isStrongLTR: Bool { ... }
}
```

Ditto for the other helper methods here

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
itingliu: At least we can consolidate this and the `_parentLookup` below

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: Maybe it's FINE

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Per discussion with Tina, it's alright to use this approach for now.

Benchmarks/Benchmarks/Formatting/BenchmarkListFormat.swift
```diff
@@ -0,0 +1,158 @@
...
+
+    Benchmark("list-format-th-contextual", configuration: scaling) { benchmark in
+        for _ in benchmark.scaledIterations {
+            blackHole(thaiItems.formatted(thStyle))
+        }
+    }
+
+    // MARK: FSI/PDI bidi wrapping
+
+    Benchmark("list-format-bidi-mixed", configuration: scaling) { benchmark in
```
garywade, line 130: I'd like to see this allow for a statistical directional mode rather than relying on the first strong character of an item.  This doesn't need an implementation, but providing some hook that can make it easy to add at some point.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+        default:
+            // Fast path: simple patterns, not Thai. Build the result in one
+            // buffer in a single pass (O(N) total work) rather than producing N
+            // intermediate strings. `wrapBidi` is inlined so we don't allocate an
+            // intermediate `[String]` for the wrapped items. A contextual rule
+            // (Spanish/Hebrew) only changes which connector the final join uses,
+            // so it stays on this path.
+            if canBuildLinearly, !isThai,
+               let startF = startFormatter, let middleF = middleFormatter, let endF = endFormatter {
+                let lastJoin = firedAlternative(forNextItem: strings[strings.count - 1]) ?? endF
```
itingliu: This indentation seems off. 
- [ ] For me: find a linter to help with this

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+        guard oneDigit < scalars.endIndex, scalars[oneDigit] == "1" else { return nil }
+        let oneClose = scalars.index(after: oneDigit)
+        guard oneClose < scalars.endIndex, scalars[oneClose] == "}" else { return nil }
+        let suffixStart = scalars.index(after: oneClose)
+
+        let connector = String(scalars[connectorStart..<oneOpen])
+        self.prefix = String(scalars[..<zeroOpen])
+        self.connector = connector
+        self.suffix = String(scalars[suffixStart...])
+        self.connectorStartsWithSpace = connector.unicodeScalars.first == " "
```
itingliu: For follow up: for all these iterations through `unicodeScalars` we should see if we can use string.utf8 instead, and / or add a fast path for ASCII in other cases

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -37,6 +37,75 @@ extension OutputSpan where Element == UInt8 {
...
+///
+/// Internal for now, but shaped to graduate into a shared formatter utility:
+/// `Duration+UnitsFormatStyle` open-codes the same `{0}`/`{1}` substitution, and
+/// other CLDR-driven formatters will want it too.
+internal struct SimpleFormatter: Equatable {
+    let prefix: String
+    let connector: String
+    let suffix: String
+    let connectorStartsWithSpace: Bool
+    let connectorEndsWithSpace: Bool
```
richgillam, line 94: Let me start by saying I'm not saying you should do this _now_.  I'm fine if you check this in as-is and we (or maybe _I_) make additional tweaks and improvements later.

That said, I think we're going to need `SimpleFormatter` (or whatever we call it) to be more general.  I don't know that it matters for list formatting, but for some of the other uses, we're going to need a variable number of substitutions, and we're going to need to be able to vary the order (i.e., you're going to need to be able to do both "{0}, {1}" and "{1}, {0}").

You're probably also going to want to move this class into its own source file at some point, but that can happen later too.

Other than that, this new class looks good!

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
richgillam: I suspect Tina's right here.  Since the special syntax characters are all ASCII, it might be simpler and faster to operate in the UTF8 domain for this whole function.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -135,67 +199,88 @@ internal final class NativeListFormatter: Sendable {
...
-            // Thai or contextual or non-simple-pattern locale: per-call apply.
-            if isThaiAnd,
-               let startP = startParsed,
-               let middleP = middleParsed,
-               let endP = endParsed {
-                var result = applyThai(parsed: startP, isPair: false,
+            if isThai,
+               let startF = startFormatter,
+               let middleF = middleFormatter,
+               let endF = endFormatter {
```
richgillam: You don't have to rename something in an `if let`.  If all you're doing is converting from a non-nil optional to a non-optional, you can just say `if let startFormatter`.  I don't know for sure, but I'm guessing this'll work for the comma-delimited list of conditions here too.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -135,67 +199,88 @@ internal final class NativeListFormatter: Sendable {
...
+
+    /// The formatter for the end/pair join when this formatter's contextual
+    /// substitution rule fires for `nextItem` (the item that follows the
+    /// connector); `nil` if there's no rule or it doesn't fire.
+    private func firedAlternative(forNextItem nextItem: String) -> SimpleFormatter? {
+        guard let condition = alternativeCondition, let formatter = alternativeFormatter,
+              condition(nextItem) else {
+            return nil
         }
+        return formatter
```
richgillam, line 346: The code calling this function might be simpler if, instead of returning `nil`, this function returned a `SimpleFormatter` all the time-- the alternative formatter when appropriate and the regular end formatter when not.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -135,67 +199,88 @@ internal final class NativeListFormatter: Sendable {
...
-                             start: ParsedPattern,
-                             middle: ParsedPattern,
-                             end: ParsedPattern) -> String {
+                             start: SimpleFormatter,
+                             middle: SimpleFormatter,
+                             end: SimpleFormatter) -> String {
+        // TODO: use withTemporaryAllocation once swift-tools-version is 6.4 or higher
         return withUnsafeTemporaryAllocation(of: Bool.self, capacity: items.count) { conflicts in
             var total = 0
             for i in items.indices {
```
richgillam, line 373: If I'm reading this right, building the list linearly means that the list formatter has to know about the innards of the `SimpleFormatter`, which seems questionable from an object-oriented-design point of view.  I think it's possible to solve this while still making `SimpleFormatter` fully general, but I haven't thought it all the way through yet.

Again, I don't want to hold up this PR over that, but I think this is something we should find a way to remind ourselves to go back and think about some more later.  Maybe add a TODO or file a Radar or something?

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -245,132 +330,45 @@ internal final class NativeListFormatter: Sendable {
...
-            var output = OutputSpan(buffer: buffer, initializedCount: 0)
-            output.append(copying: p.prefix.utf8Span.span)
-            output.append(copying: s0.utf8Span.span)
-            output.append(copying: p.connector.utf8Span.span)
-            output.append(copying: s1.utf8Span.span)
-            output.append(copying: p.suffix.utf8Span.span)
-            return output.finalize(for: buffer)
+    /// Format with the compiled formatter, or fall back to scanning the raw
+    /// pattern when parsing failed (malformed data).
+    private func interpolate(_ formatter: SimpleFormatter?, _ pattern: String,
```
richgillam: I don't understand the comment, and don't really understand what this function does.  Can you explain a little more?

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -245,132 +330,45 @@ internal final class NativeListFormatter: Sendable {
...
+                             _ s0: String, _ s1: String) -> String {
+        if let formatter {
+            return formatter.format(s0, s1)
         }
+        return interpolate(pattern, s0, s1)
     }
 
     /// Single-pass Thai formatter: emit the final string directly using the
     /// pre-parsed pattern decomposition, with any contextually-required spaces
     /// adjacent to the connector inserted.
```
richgillam, line 440: Did you explain somewhere else in here just what the Thai contextual change is?  I don't remember seeing it.

Again, I don't want to hold up this PR over it, but is it possible to generalize this so that we don't have to have special Thai-specific logic at the top level?  I gather that this is fundamentally different from the Spanish- and Hebrew-specific logic, but I'm not grokking exactly how and I'm wondering if the language- and context-specific logic can be generalized to the point where you don't need special stuff just for Thai.

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
richgillam: +1

richgillam (approved):
Looking SO much better!

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+/// `FOUNDATION_LIST_FORMAT_ICU` build flag is set.
+///
+/// Reimplements Apple-ICU's `i18n/listformatter.cpp`
+/// (`ListFormatter::formatStringsToValue` and its `PatternHandler` family):
+/// the CLDR "List Patterns" empty/single/pair/3+ composition, the contextual
+/// rules (Spanish y→e / o→u and the Hebrew non-Hebrew prefix), and the Apple
+/// extensions (`ThaiHandler`'s contextual joiner and FSI/PDI bidi isolate
+/// wrapping, which isolates any item containing a character of the opposite
+/// direction to the list, matching `ListFormatter::needsBidiIsolates`).
+@available(macOS 12.0, iOS 15.0, tvOS 15.0, watchOS 8.0, *)
```
itingliu, line 183: I don't think you need this for internal types

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: correction: `string.withUTF8 { ... `

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    /// UTF-8 byte count of formatting two arguments of the given byte lengths.
+    func utf8Count(_ arg0Bytes: Int, _ arg1Bytes: Int) -> Int {
+        prefix.utf8.count + arg0Bytes + connector.utf8.count + arg1Bytes + suffix.utf8.count
+    }
+
+    /// Append `prefix + s0 + connector + s1 + suffix` into `output`. The caller
+    /// is responsible for having reserved enough capacity (see `utf8Count`).
+    @inline(__always)
```
itingliu: Did you find any actual impact with ` @inline(__always)`? This would increase code size, so beware before sticking it everywhere

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
itingliu: Totally agreed that that it'd be great if we don't sprinkle language specific behaviors here and there

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -0,0 +1,577 @@
...
+    }
+
+    /// Substitute `{0}` and `{1}` in `pattern`. Other braces are emitted
+    /// literally (CLDR list patterns use exactly the two placeholders, but
+    /// items themselves may contain literal braces — see the
+    /// `literalBracePlaceholders` test).
+    private func interpolate(_ pattern: String, _ s0: String, _ s1: String) -> String {
+        var result = ""
+        result.reserveCapacity(pattern.count + s0.count + s1.count)
+        let scalars = pattern.unicodeScalars
```
itingliu: same comment about using utf8 if applicable

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Done in commit 2a67f51660711bf7163c129ea2dc46768600e529 titled "Scan patterns and items as UTF-8 in the list formatter".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Done in commit 2a67f51660711bf7163c129ea2dc46768600e529 titled "Scan patterns and items as UTF-8 in the list formatter".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Done in commit 2a67f51660711bf7163c129ea2dc46768600e529 titled "Scan patterns and items as UTF-8 in the list formatter".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: I did some more benchmarking and found that this and two other instances of `@inline(__always)` did not significantly affect benchmarks, so I've removed them in commit 6f4c5569bcbc91235b2edd23f25a88a5b40e7848 titled "Drop @inline(__always) from list-format helpers that don't benefit". The remaining three instances (combined) result in a ~10% performance improvement in the `en-long` benchmark. If that's not enough of a benefit to warrant keeping them, please let me know and I'll remove them too.

Sources/FoundationInternationalization/Formatting/ListFormatPatterns.swift
cjchapman: Done in commit 4278b2b1c975a2b773c054f7af77077b702982c4 titled "Consolidate the list-format slot and parent binary searches".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Done in commit b79d2c7cfe7185e402590db2404f9124d95612ae titled "Apply swift-format to the new source files, with hand-tuned alignment".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: Done in commit a520d917a08a0c8ec0f7418d87bf27e8f17493e5 titled "Use if-let shorthand instead of renaming the formatter bindings".

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
```diff
@@ -417,15 +481,33 @@ internal final class ListFormatterImpl: Sendable {
...
+        // numbers, and (in an RTL list) anything but a Latin letter — is either
+        // outside ASCII or trivially classified by byte, so branch on the cheap
+        // `isKnownASCII` bit first and only decode scalars when we must.
+        let isASCII = item.utf8Span.isKnownASCII
         if listDirection == .rightToLeft {
+            // Only a strong-LTR character (`L`) disturbs an RTL list. Within
+            // ASCII that means a Latin letter, so scan the UTF-8 bytes directly
+            // and skip scalar decoding entirely.
+            if isASCII {
+                for byte in item.utf8 where (0x41...0x5A).contains(byte) || (0x61...0x7A).contains(byte) {
```
richgillam, line 537: I had to read this a couple times to get it.  I take it `for x in y where c { doSomething }` is equivalent to `for x in y { if c { doSomething } }`?  So the `return false` below only executes after you've iterated through the wrong string and not found any letters?

richgillam (approved):
The changes you've made since the last time I looked at this all look good, but some of the bigger issues I raised aren't addressed.  I'm fine with you checking this in as-is, but I'm wondering what your plan is for
- The bidi class.  You were going to change to the existing function that calls into ICU4C for this, weren't you?  And then maybe try to get a pure-Swift version into the stdlib later?  Is changing to the ICU4C part happening as part of this PR or later?
- The data files.  We've been talking about trying to convert that from C into Swift, weren't we?  Is that going to happen in a later PR?
- A more general version of SimpleFormatter.  A later PR?  Maybe somebody else does it?
- Moving the parent-lookup logic into `Locale`.  A later PR?  Maybe somebody else does it?

Sources/FoundationInternationalization/Formatting/ListFormatterImpl.swift
cjchapman: I've changed the code to get bidi class from ICU for now, on a new `Unicode.Scalar._bidiClass` property, with a comment explaining the plan to switch to getting it from stdlib once it becomes available. This supersedes the old per-scalar helpers (`isStrongLTR`, `isArabicNumber`, `_isStrongRTL`), which are now gone. This is in commit 1377219e080d9ca9a9cd8e109a883b5e3878268a "Get Bidi_Class from ICU for list-format bidi wrapping".

richgillam (approved):
The new bidi changes look great.

cjchapman:
That force push just now was to rebase this off #2137.

utils/update-list-format-data-impl.swift
```diff
@@ -63,6 +63,95 @@ let slots: [(cldrType: String, type: String, width: String)] = [
...
+//
+// To refresh at a CLDR/ICU upgrade: re-extract the %%ALIAS values from a matching
+// icu4c/source/data/locales/*.txt checkout (each aliased locale's .txt file is a
+// single line `xx{ "%%ALIAS"{"yy"} }`), or re-run ICU's LdmlConverter over the
+// config's <localeIds>. Only the sources reachable through the public
+// ListFormatStyle surface (list patterns) need to appear here.
+//
+// This is the full curated record. The generator emits only the subset that the
+// current CLDR data doesn't already resolve correctly by plain truncation (see
+// trimRedundantAliases), so the shipped JSON is smaller.
```
richgillam: This is very long and clearly AI-generated, and I didn't completely follow what it says it's doing.  It seems to be saying that it's generating the list of locale-ID aliases from the `%%ALIAS` tags in individual ICU resource files.  I don't think that's something we want to depend on long-term, and I'm wondering if ICU is getting that list from somewhere else (it sounds like maybe it's hard-coded somewhere inside the `cldr2icu` tool?).  I'm wondering if we just want to have that alias list hard-coded in our own stuff (and hand-maintain it), or lobby to get it into `supplementalData.xml` in CLDR, or something, so we're not relying on the current ICU source code.

utils/update-list-format-data-impl.swift
```diff
@@ -489,9 +581,54 @@ func simplifyParentMap(
     _ = emissions // silenced; emit() recomputes from the simplified parentMap
 }
 
+// Drop aliases that the plain fallback walk (truncation only, no alias redirect)
+// already resolves to the same patterns the alias would. Such an alias is
+// redundant *for list formatting*: e.g. zh_TW → zh_Hant_TW changes nothing,
```
richgillam, line 562: The "*for list formatting*" thing here makes me nervous.  We're eventually going to want to generalize these locale-manipulation processes, and I'm not sure what that means I'm going to have to do, or how it'd affect list formatting.  I'm wondering if we need to talk...

(Or are you just merging list-formatting data here, and not locale-fallback data?  Maybe I misunderstood...)

utils/update-list-format-data-impl.swift
cjchapman: The list below this comment was derived from ICU's `%%ALIAS` tags which derived from some hard-coded stuff in ICU's `cldr-to-icu/config.xml` (not CLDR), but it's now effectively an "alias list hard-coded in our own stuff" .

utils/update-list-format-data-impl.swift
cjchapman: This is for minimizing list-formatting-related data specifically.

utils/update-list-format-data-impl.swift
cjchapman: I just pushed a revised version of this comment.

Sources/FoundationInternationalization/BidiClass+ICU.swift
itingliu, line 1: For consistency can we name this class BIDIClass+ICU.swift?

cjchapman: Done.

cjchapman:
That force push just now was due to a rebase to pick up `utf8SpanMakingContiguous`.

itingliu (approved):
Looking great thank you!

PR #1999, opened 2026-05-26
Create _FoundationInternationalizationData library

jmschonfeld:
Creates a library named `_FoundationInternationalizationData` to hold CLDR data that Foundation uses directly for its native implementations of ICU functionality

### Motivation:

This is a precursor to https://github.com/swiftlang/swift-foundation/pull/1991 which will introduce the first set of data to this library and begin using it.

### Modifications:

- Create a new `_FoundationInternationalizationData` static library
- Statically link it into the `FoundationInternationalization` dynamic library for dynamic builds, or install the library for static builds (mirroring the behavior of `_FoundationCShims`)

### Result:

`_FoundationInternationalizationData` library is created

### Testing:

Will test via this repo's CI as well as full toolchain testing alongside the modification to the windows installer script to include the new header

### DIFF
--- a/Sources/_FoundationInternationalizationData/CMakeLists.txt
+++ b/Sources/_FoundationInternationalizationData/CMakeLists.txt
@@ -1,0 +1,50 @@
+##===----------------------------------------------------------------------===##
+##
+## This source file is part of the Swift open source project
+##
+## Copyright (c) 2026 Apple Inc. and the Swift project authors
+## Licensed under Apache License v2.0
+##
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.md for the list of Swift project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+add_library(_FoundationInternationalizationData STATIC
+    ListFormatData.c)
+
+target_include_directories(_FoundationInternationalizationData PUBLIC include)
+
+target_compile_options(_FoundationInternationalizationData INTERFACE
+  "$<BUILD_INTERFACE:$<$<COMPILE_LANGUAGE:Swift>:SHELL:-Xcc -fmodule-map-file=${CMAKE_CURRENT_SOURCE_DIR}/include/module.modulemap>>")
+
+set_property(GLOBAL APPEND PROPERTY SWIFT_FOUNDATION_EXPORTS _FoundationInternationalizationData)
+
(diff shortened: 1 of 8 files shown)


jmschonfeld:
@swift-ci please test

Sources/_FoundationInternationalizationData/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+set_property(GLOBAL APPEND PROPERTY SWIFT_FOUNDATION_EXPORTS _FoundationInternationalizationData)
+
+if(BUILD_SHARED_LIBS)
+    set(install_directory swift)
+else()
+    set(install_directory swift_static)
+endif()
+
+# Copy Headers to known directory for direct client (XCTest) test builds
```
parkera, line 31: Not sure I should question your build-script-fu here, but do we need to do this copy for this library?

jmschonfeld: Yes, unfortunately. During toolchain builds, XCTest builds before Foundation gets installed into the toolchain, so instead of using default include paths, it manually looks for Foundation build products. Even though XCTest doesn't directly use this functionality, there's still a visible module dependency chain from `XCTest` --> `Foundation` --> `FoundationInternationalization` --> `_FoundationInternationalizationData` that must be present otherwise the build will fail due to a missing module dependency. In an ideal world, Foundation would install before XCTest builds so XCTest doesn't need to look for the build products manually, but sadly the build script is not ideal.

parkera (approved):

jmschonfeld:
@swift-ci please test

jmschonfeld:
@swift-ci please test

Sources/_FoundationInternationalizationData/CMakeLists.txt
finagolfin: > In an ideal world, Foundation would install before XCTest builds so XCTest doesn't need to look for the build products manually, but sadly the build script is not ideal.

If you turn Foundation from [a bash `build-script-impl` product](https://github.com/swiftlang/swift/blob/306b4396f455f16ea1cc4000cab17ce10096ff7f/utils/build-script-impl#L2397) into a Python `build-script` product, as done for cmark in swiftlang/swift#37102, I believe it will be installed right after building. It would take translating about 130 lines.

jmschonfeld:
@swift-ci please test

PR #2003, opened 2026-05-27
[Windows] Atomic write: clear READONLY on destination and retry rename

clintonpi:
### Motivation:

`SetFileInformationByHandle` with `FileRenameInfoEx` and `FILE_RENAME_FLAG_POSIX_SEMANTICS | FILE_RENAME_FLAG_REPLACE_IF_EXISTS` can fail with `ERROR_ACCESS_DENIED` (mapped from NTSTATUS `STATUS_CANNOT_DELETE`) when the destination has `FILE_ATTRIBUTE_READONLY`. POSIX `rename(2)` is not affected if the destination file itself is read-only. This is the final piece for #1485.

### Modifications:

- When the rename fails with `ERROR_ACCESS_DENIED` and the destination has `FILE_ATTRIBUTE_READONLY` set, clear that attribute and retry the rename once before falling through to the existing `MoveFileExW` fallback.

### Result:

Atomic writes on Windows now replace a read-only destination, aligning with POSIX semantics.

### Testing:

Added a cross-platform test to verify the modification and its consistency.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -447,17 +447,51 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable 
                     wcscpy_s($0, cchLength + 1, pwszPath)
                 }
 
-                if !SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize) {
-                    let dwError = GetLastError()
+                var renameOk = SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize)
+
+                if !renameOk {
+                    var dwError = GetLastError()
+
+                    // FileRenameInfoEx with POSIX_SEMANTICS + REPLACE_IF_EXISTS
+                    // returns ERROR_ACCESS_DENIED (mapped from NTSTATUS STATUS_CANNOT_DELETE)
+                    // when the destination has FILE_ATTRIBUTE_READONLY.
+                    // Clear it on the destination (in line with POSIX semantics)
+                    // and retry once before falling through.
+                    if dwError == ERROR_ACCESS_DENIED {
+                        let dwAttributes = GetFileAttributesW(pwszPath)
+
+                        if dwAttributes != INVALID_FILE_ATTRIBUTES
+                            && dwAttributes & FILE_ATTRIBUTE_READONLY != 0
+                        {
+                            // TOCTOU is possible here between GetFileAttributesW and SetFileAttributesW.
+                            // Only relevant though in the atypical case when SetFileInformationByHandle
+                            // returns false, where the thread is already on an error path.
+                            // Hence, skip expensive mitigation and defer to caller.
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -447,18 +447,52 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable
...
 
-                if !SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize) {
-                    let dwError = GetLastError()
+                var renameOk = SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize)
+
+                if !renameOk {
+                    var dwError = GetLastError()
+
+                    // FileRenameInfoEx with POSIX_SEMANTICS + REPLACE_IF_EXISTS
+                    // returns ERROR_ACCESS_DENIED (mapped from NTSTATUS STATUS_CANNOT_DELETE)
```
parkera: Just as an FYI, we don't manually wrap comments in Foundation sources.

parkera (approved):
Looks good to me. I'd like one more set of eyes before merging

Tests/FoundationEssentialsTests/DataIOTests.swift
```diff
@@ -211,6 +215,30 @@ private final class DataIOTests {
         let readData = try Data(contentsOf: url, options: [])
         #expect(readData == data)
     }
+
+#if os(Windows)
```
jmschonfeld: Since you mentioned that this matches POSIX behavior, would we be able to structure the unit test in such a way that it tests this behavior across all platforms to validate the consistency? It looks like if you use `FileManager.setAttributes` with a read-only mode that would allow us to produce a readonly file on any platform (we map `_S_IWRITE` to `FILE_ATTRIBUTE_READONLY` on Windows and use the mode to call `chmod` on Darwin + Linux), then theoretically we could validate the `Data.write(to:)` behavior for all platforms to ensure it's consistent in handling of readonly files, right?

jmschonfeld (commented):
Change itself looks good to me based on your explanation, but one comment about expanding the unit test to validate the cross-platform consistency

Tests/FoundationEssentialsTests/DataIOTests.swift
clintonpi: Yes, great idea!

jmschonfeld:
@swift-ci please test

jmschonfeld:
@parkera should we consider taking this for 6.4? I know at one point this was a source of problems for swift-driver, I think that was worked around but I don't know if there are still lingering issues there

jmschonfeld:
Discussed with Tony and we'd like to take this for 6.4 - could you update/rebase your PR against `release/6.4.x`? It will merge forward to `main` automatically from there

clintonpi:
Closing to re-open against `release/6.4.x` branch.

clintonpi:
Re-opened: https://github.com/swiftlang/swift-foundation/pull/2007

PR #2007, opened 2026-05-29
[Windows] Atomic write: clear READONLY on destination and retry rename

clintonpi:
### Motivation:

`SetFileInformationByHandle` with `FileRenameInfoEx` and `FILE_RENAME_FLAG_POSIX_SEMANTICS | FILE_RENAME_FLAG_REPLACE_IF_EXISTS` can fail with `ERROR_ACCESS_DENIED` (mapped from NTSTATUS `STATUS_CANNOT_DELETE`) when the destination has `FILE_ATTRIBUTE_READONLY`. POSIX `rename(2)` is not affected if the destination file itself is read-only. This is the final piece for #1485.

_Fixes #1507_

### Modifications:

- When the rename fails with `ERROR_ACCESS_DENIED` and the destination has `FILE_ATTRIBUTE_READONLY` set, clear that attribute and retry the rename once before falling through to the existing `MoveFileExW` fallback.

### Result:

Atomic writes on Windows now replace a read-only destination, aligning with POSIX semantics.

### Testing:

Added a cross-platform test to verify the modification and its consistency.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -447,17 +447,44 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable 
                     wcscpy_s($0, cchLength + 1, pwszPath)
                 }
 
-                if !SetFileInformationByHandle(hFile, FileRenameInfoEx, pInfo, dwSize) {
-                    let dwError = GetLastError()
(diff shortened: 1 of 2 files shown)


clintonpi:
@swift-ci please test

jmschonfeld (approved):

clintonpi:
@swift-ci please test

clintonpi:
@swift-ci please test

jmschonfeld (approved):

Issue #874, opened 2024-08-18
Sorting with multiple SortComparator uses a Sequence instead of Collection 


jamiejamesjk:
I see there is a function to sort a `Sequence` of elements using any `Sequence` of `SortComparator` here:

https://github.com/apple/swift-foundation/blob/main/Sources/FoundationEssentials/SortComparator.swift#L234

However, shouldn't the sequence of sort comparators being passed to `sorted` be of type `Collection` instead? Because a `Sequence` doesn't mandate the ability to iterate non-destructively, but the `sorted` function iterates through the provided `Sequence` multiple times.

PR #2009, opened 2026-05-31
SortComparator: materialise comparator `Sequence` before reusing it for sort (#874)

wadetregaskis:
`SortComparator` shouldn't assume it can iterate its `Sequence` of `SortComparator`s multiple times.

### Motivation:

The public `Sequence.sorted(using: S)` overload that takes a `Sequence` of `SortComparator`s forwards each pair of elements through `comparators.compare(_:_:)`, which itself iterates the entire sequence.  That works for `Array` (or any multi-pass collection) but `Sequence` does not promise non-destructive iteration: a single-use sequence (e.g. `AnySequence` wrapping an `AnyIterator`) is exhausted after the first comparison, so every subsequent call to `compare` walks an empty iterator and returns `.orderedSame` — silently leaving the bulk of the input unsorted.

I don't know how common it is to do this - I imagine the concrete type used is just an `Array` 99% of the time, which happens to work just fine - but this is easy to fix and the kind of bug that's quite hard to find if you stumble into it unwittingly, as a user of this API.

### Modifications:

This patch fixes that by snapshotting the comparator sequence into an `Array` before sorting so it can be replayed safely.

### Result:

Sorting works as the caller intended, irrespective of what type of `Sequence` they use.

### Testing:

New unit test added.

### DIFF
--- a/Sources/FoundationEssentials/SortComparator.swift
+++ b/Sources/FoundationEssentials/SortComparator.swift
@@ -258,6 +258,12 @@ extension Sequence {
         S.Element == Comparator,
         Element == Comparator.Compared
     {
+        // Materialize the comparator sequence into an array: `compare(_:_:)` is
+        // invoked once per pair of elements, and `Sequence` does not promise
+        // non-destructive multi-pass iteration. A single-use sequence (e.g. one
+        // backed by `AnyIterator`) would otherwise be exhausted after the first
+        // comparison and silently produce an unsorted result.
+        let comparators = Array(comparators)
         return self.sorted {
             comparators.compare($0, $1) == .orderedAscending
         }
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/SortComparator.swift
```diff
@@ -258,6 +258,12 @@ extension Sequence {
         S.Element == Comparator,
         Element == Comparator.Compared
     {
+        // Materialize the comparator sequence into an array: `compare(_:_:)` is
+        // invoked once per pair of elements, and `Sequence` does not promise
+        // non-destructive multi-pass iteration. A single-use sequence (e.g. one
+        // backed by `AnyIterator`) would otherwise be exhausted after the first
+        // comparison and silently produce an unsorted result.
+        let comparators = Array(comparators)
```
jmschonfeld, line 266: Have you measured the performance impact of requiring the comparators to be copied into a heap allocation for every call site, even those that could have been iterated twice?

We also don't typically wrap comments in this repo, nor do I think we need a long comment here - a short comment would suffice.

PR #2010, opened 2026-05-31
URL: don't append a trailing '?' when the queryItems list is empty (#1548)

wadetregaskis:
### Motivation:

`URL.appending(queryItems:)` blindly assigned `c.queryItems = newItems` even when `newItems` was empty. `URLComponents` distinguishes between `queryItems == nil` (no query, no '?') and `queryItems == []` (empty query, trailing '?'), so appending an empty array to a URL that had no query introduced a stray '?':

```swift
URL(string: "https://www.example.com/path")!
    .appending(queryItems: [])
// -> "https://www.example.com/path?"
````

### Modifications:

This patch makes it short-circuit on an empty input list and return the receiver unchanged.

### Result:

Appending an empty list won't change the URL.

### Testing:

Added a new unit test which checks that the errant '?' is no longer appended, but also that the '?' remains if already there, and that appending a non-empty set of items still works (technically redundant with other tests - IMO makes sense locally since it encompasses the full set of '?'-related cases, but I can remove that if you prefer).

### DIFF
--- a/Sources/FoundationEssentials/URL/URL.swift
+++ b/Sources/FoundationEssentials/URL/URL.swift
@@ -1713,6 +1713,9 @@ extension URL {
     /// - Parameter queryItems: A list of `URLQueryItem` to append to the receiver.
     @available(macOS 13.0, iOS 16.0, tvOS 16.0, watchOS 9.0, *)
     public func appending(queryItems: [URLQueryItem]) -> URL {
+        // Appending nothing should not alter the URL — in particular, must not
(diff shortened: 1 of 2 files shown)


jrflat:
Issue #1548 is already closed for the reasons discussed there and at length in PR #1650. Your PR description here also mentions a major reason for this: the APIs and the RFC distinguish specifically between an empty query (a lone `?`) and no query (no `?` at all).

If you have new arguments or a use-case this is blocking, feel free to re-open #1548 and add to the discussion there. Thanks!

Issue #1098, opened 2024-12-16
`Data.write(to:options:)` crashes if `[.atomic, .withoutOverwriting]` are passed together


weissi:
https://github.com/swiftlang/swift-foundation/blob/c0532f8c99491949d146fbd9b9f2bac5384e6dbd/Sources/FoundationEssentials/Data/Data.swift#L2445-L2447

Severe, source breaking bug here. In 5.10 and below, `[.atomic, .withoutOverwriting]` worked just fine (possibly not achieving the desired behaviour). We've got to back out this change because

1. It's source breaking
2. The [docs](https://swiftinit.org/docs/swift/foundationessentials/data/writingoptions) don't mention that `.atomic` and `.withoutOverwriting` are mutually exclusive
3. They shouldn't be mutually exclusive
  3.1. On Linux, the [man page for open](https://www.man7.org/linux/man-pages/man2/open.2.html)'s example for `O_TMPFILE` contains how to implement `[.atomic, .withoutOverwriting]`
  3.2. On systems that lack `O_TMPFILE`, this can still be approximated with write to `.tmp_XXXXXX`, then rename tmp file (and remove if that fails). That still prevents half-written files

beltradini:
Hi! My idea is to remove the `fatalError` and make it safe to use `.atomic` and `.withoutOverwriting` together by:

- Writing the data to a temporary file first (using `.atomic`)
- If the destination file already exists, delete the temp file and throw a standard Cocoa error (`NSFileWriteFileExistsError`)
- If it doesn't exist, move the temp file to the destination path

This should match the behavior we had in Swift 5.10 and keep things safe without overwriting files unexpectedly.
Let me know if this direction works — I’ll send in a draft PR soon so we can review it together.

Thanks!

PR #2011, opened 2026-05-31
Data.write: adds support for the combination of `.atomic` & `.withoutOverwriting`, instead of crashing (#1098)

wadetregaskis:
### Motivation:

Prior to this patch, `Data.write(to:options:)` called `fatalError` when both `.atomic` and `.withoutOverwriting` were passed together.  That combination worked in Swift 5.10 and is documented as supported (Apple's Foundation maps it to "write to aux file with O_EXCL, then exchange"), so trapping is not just innately bad but also a source-breaking regression.

### Modifications:

  - POSIX: after writing the auxiliary file, `link` is used to map it under the final name too.  `link` fails with `EEXIST` if the destination already exists, which is exactly the contract of `.withoutOverwriting`.  And whether successful or not, the original name is unlinked (which leaves only the intended final name referring to the file, or deletes the file if the `link` failed).
  - Windows: in `SetFileInformationByHandle`, `FILE_RENAME_FLAG_REPLACE_IF_EXISTS` / `MOVEFILE_REPLACE_EXISTING` are dropped if `.withoutOverwriting` is specified, and the `ERROR_FILE_EXISTS` & `ERROR_ALREADY_EXISTS` errors are mapped to Cocoa's `fileWriteFileExists`.

### Result:

Writing atomically without overwriting now works [again].

### Testing:

New unit test added.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -438,8 +438,10 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable 
             let dwSize = DWORD(MemoryLayout<FILE_RENAME_INFO>.size + cbSize + MemoryLayout<WCHAR>.size)
             try withUnsafeTemporaryAllocation(byteCount: Int(dwSize),
                                               alignment: MemoryLayout<FILE_RENAME_INFO>.alignment) { pBuffer in
+                let withoutOverwriting = options.contains(.withoutOverwriting)
+
                 var pInfo = pBuffer.baseAddress?.bindMemory(to: FILE_RENAME_INFO.self, capacity: 1)
-                pInfo?.pointee.Flags = FILE_RENAME_FLAG_POSIX_SEMANTICS | FILE_RENAME_FLAG_REPLACE_IF_EXISTS
+                pInfo?.pointee.Flags = FILE_RENAME_FLAG_POSIX_SEMANTICS | (withoutOverwriting ? 0 : FILE_RENAME_FLAG_REPLACE_IF_EXISTS)
                 pInfo?.pointee.RootDirectory = nil
                 pInfo?.pointee.FileNameLength = DWORD(cbSize)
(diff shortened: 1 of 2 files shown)


jmschonfeld:
> That combination worked in Swift 5.10 and is documented as supported (Apple's Foundation maps it to "write to aux file with O_EXCL, then exchange"), so trapping is not just innately bad but also a source-breaking regression.

Can you please elaborate on this statement? It seems mostly incorrect to me. This combination of options is explicitly documented as unsupported:

> You can’t combine this constant with [atomic](https://developer.apple.com/documentation/foundation/nsdata/writingoptions/atomic) because atomic allows the system to overwrite the original file.

https://developer.apple.com/documentation/foundation/nsdata/writingoptions/withoutoverwriting

Additionally, this source code is shared with `Foundation.framework` on Apple platforms and it terminates the same way on Apple platforms as it does on Linux currently. I did confirm it did work with 5.10 before the swift-foundation recoring, but it looks like Apple platforms have always performed this way and were documented as such. Can you please share what platform + OS version you tested on when validating this?

kperryua:
The semantics of `WithoutOverwriting` referring that of the auxiliary file was never intended by the original API design. The option is specifically documented to affect the *destination* file only.

When originally implemented, we had no silver bullet syscall that could support an atomic rename() that had the same semantics as `O_EXCL`. We have that now on Darwin *for APFS* in the form of `RENAME_EXCL`. It's also available for Linux in terms of `RENAME_NOREPLACE`. And per your description, sounds like we have it for Windows. In order to adopt however, we'd have to decide whether we're willing to start dynamically throwing "feature unsupported" errors on file systems where the behavior is not supported.

Furthermore, without any kind of availability declaration, we'd probably have to predicate this behavior on deployment target so that it doesn't surprisingly behave differently on current OS's and then crash immediately when running on older OS's.

RE: `link()`. The solution is clever, but there is known friction with Sandbox on Darwin when using `link()` on a aux file that exists in a temporary directory and certain destination directories. I'd rather avoid that headache if we can.

PR #2016, opened 2026-06-02
DNM: test cmake build

itingliu:


itingliu:
@swift-ci please test

PR #2020, opened 2026-06-03
(rdar://173406489) Plist parsing should reject \U sequences without any valid hex digits (#4168)

kperryua:
The OpenStep plist format supports a variable 1-4 hex characters following a \U escape. The current implementation basically tries this algorithm four times:

* Read current character
* If it's valid hex, accumulate it and advance to the next character

Then:

* Return the accumulated value as the character.

The issue is that \U followed by ZERO valid hex characters is treated as a NUL character instead of an error.


kperryua:
@swift-ci please test

jmschonfeld (approved):

PR #2021, opened 2026-06-03
`CharacterSet`: Memory-related Performance Improvements

chloe-yeo:
This PR introduces a method that fills a buffer for a specified plane of a `BuiltInUnicodeScalarSet` so that mutating methods in `CharacterSet` do not instantiate 8KB of `Data` that are immediately thrown away after that `Data` has been used for mutating another buffer via mutating methods such as union and intersection. The method is equivalent in functionality to the previous `bitmap(forPlane:isInverted:)` method. 

### Motivation:

This is a change that helps to reduce the memory usage of `CharacterSet` when trying to fill `Annex` planes of built-in `CharacterSet`s. 

### Modifications:

Added a method `bitmap(forPlane:isInverted:into:apply:)` method and refactored `bitmap(forPlane:isInverted:)` to call into it to avoid duplicate logic. 

### Result:

There should be no behavioral change. 

### Testing:

All the existing unit tests should still pass.

### DIFF
--- a/Sources/FoundationEssentials/String/String+Bridging.swift
+++ b/Sources/FoundationEssentials/String/String+Bridging.swift
@@ -17,6 +17,7 @@
 internal import CoreFoundation_Private.CFString
 internal import ObjectiveC_Private.objc_internal
 internal import CoreFoundation_Private.ForFoundationOnly
+internal import Foundation_Private.NSString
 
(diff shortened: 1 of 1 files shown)


chloe-yeo:
@swift-ci please test

PR #2022, opened 2026-06-03
`Decimal`: use `UInt128` significand to speed up operations

xwu:
This PR introduces a new internal computed property (called `_significand` to distinguish itself from `_mantissa`) of type `UInt128`, which allows us to perform arithmetic operations bypassing `VariableLengthInteger`.

Although conceptually low-hanging fruit, fully threading the changes through the implementation represents an overhaul of some scale but with performance gains to match. The resulting implementations (written by hand) are fortunately imminently readable. Latent bugs are addressed along the way, substantially improving the precision of mathematical operations on `Decimal`.

### Motivation:

#1754 demonstrated that making `VariableLengthInteger` non-allocating (and not really variable) dramatically improves performance. While improved, however, `Decimal` operations are still by no means optimized for performance. Sadly, this state of affairs encourages the erroneous impression that decimal floating-point is intrinsically much less performant than it could be as compared to alternative numeric representations.

The prior PR was a fantastic and inspiring first move. However, absent context about other advances in Swift, LLM-driven efforts overlook that performing arithmetic limb-by-limb (which is what `VariableLengthInteger` encapsulates) is no longer necessary for implementing basic operations, as the 128-bit mantissa can be bitwise copied into a `UInt128` so that we can leverage more performant compiler primitives.

### Modifications:

This PR replaces `VariableLengthInteger` operations with `UInt128` operations, rewriting comparison, addition (and subtraction), multiplication, and division. Normalization is also rewritten to remove the last consumer of `VariableLengthInteger`, but it is also now only called by the `NSDecimalNormalize` shim.

Along the way, latent bugs are either annotated or fixed altogether--[see added tests](https://github.com/swiftlang/swift-foundation/pull/2022/changes/09bd889db9fa209b09437bc46e892d8708e54411). For example:

- The existing implementation truncates the 'refitted' mantissa in the case of arithmetic overflow during addition, which is not correct for the documented default `.plain` rounding mode (it also makes no attempt to behave correctly for other rounding modes). The revised implementation now respects rounding mode.

- The existing implementation exhibits unexpected behavior when multiplying two values with small exponents that should lead to an underflow result. (Reading the code suggests there should be a runtime trap, but in the REPL there's just a very large arbitrary result.) The revised implementation now correctly throws `underflow`.

- The existing implementation always rounds _towards zero_ (i.e., truncates) for division. The revised implementation now respects rounding mode (crucially, the documented default rounding mode, `.plain`).

- The existing implementation normalizes dividend and divisor by an arbitrary criterion chosen in 1999, which has been associated with bugs; code comments reference <rdar://problem/5197585> and <rdar://problem/2354750>. The revised implementation now scales the dividend's significand appropriately to fill 128 bits.

- The existing implementation produces a NaN value during normalization if the smaller of the two inputs has a finite, negative value that truncates to zero. The revised implementation now respects rounding mode and, if rounding up such a negative value, produces zero rather than spurious NaN.

- In the existing implementation, legacy `NSDecimal*` functions other than `Add` never signal loss of precision, as such information was neither consistently computed nor plumbed through. The revised implementation now indicates loss of precision whenever an inexact result is returned.

### Result:

Using benchmarks added in #1754, this PR results in a <strike>~350%</strike> ~500% boost in addition performance, a <strike>~750%</strike> ~950% boost in multiplication performance, and a ~7000% boost in division performance as measured by throughput.

And, as described above, arithmetic operations now have improved precision and latent bugs have been fixed. `VariableLengthInteger` is removed entirely.

```
----------------------------------------------------------------------------------------------------------------------------
Decimal add metrics
----------------------------------------------------------------------------------------------------------------------------

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                   main                   │        13 │        13 │        13 │        13 │        13 │        13 │        13 │        26 │
│               Current_run                │        80 │        78 │        78 │        77 │        76 │        71 │        69 │       154 │
│                    Δ                     │        67 │        65 │        65 │        64 │        63 │        58 │        56 │       128 │
│              Improvement %               │       515 │       500 │       500 │       492 │       485 │       446 │       431 │       128 │


----------------------------------------------------------------------------------------------------------------------------
Decimal divide metrics
----------------------------------------------------------------------------------------------------------------------------

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                   main                   │         1 │         1 │         1 │         1 │         1 │         1 │         1 │         2 │
│               Current_run                │        74 │        73 │        73 │        72 │        72 │        71 │        71 │       145 │
│                    Δ                     │        73 │        72 │        72 │        71 │        71 │        70 │        70 │       143 │
│              Improvement %               │      7300 │      7200 │      7200 │      7100 │      7100 │      7000 │      7000 │       143 │


----------------------------------------------------------------------------------------------------------------------------
Decimal multiply metrics
----------------------------------------------------------------------------------------------------------------------------

│          Throughput (# / s) (M)          │        p0 │       p25 │       p50 │       p75 │       p90 │       p99 │      p100 │   Samples │
│                   main                   │         8 │         8 │         8 │         8 │         8 │         8 │         8 │        17 │
│               Current_run                │        85 │        84 │        84 │        83 │        82 │        80 │        80 │       166 │
│                    Δ                     │        77 │        76 │        76 │        75 │        74 │        72 │        72 │       149 │
│              Improvement %               │       962 │       950 │       950 │       938 │       925 │       900 │       900 │       149 │
```

### Testing:

All 33 existing unit tests for `Decimal` pass (with modifications to account for now-corrected rounding with division and improved precision--see comments below). Additional unit tests are added for corrected behavior.
All 5 existing benchmark tests show improved performance compared to the current baseline as described above.


Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -319,25 +319,18 @@ private struct DecimalTests {
...
+        #expect(bNormalized._mantissa.0 == 18026)
+        #expect(bNormalized._mantissa.1 == 7848)
+        #expect(bNormalized._mantissa.2 == 21680)
+        #expect(bNormalized._mantissa.3 == 24817)
+        #expect(bNormalized._mantissa.4 == 48933)
+        #expect(bNormalized._mantissa.5 == 3665)
+        #expect(bNormalized._mantissa.6 == 52911)
+        #expect(bNormalized._mantissa.7 == 162)
+        #expect(bNormalized._length == 8)
         #expect(b != bNormalized)   // b had a loss Of Precision when normalising
```
xwu: The result now loses only one digit's precision while matching exponents of `a` and `b`.

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -519,26 +512,26 @@ private struct DecimalTests {
...
-        expected._mantissa.7 = 10941
+        expected._mantissa.0 = 43837
+        expected._mantissa.1 = 8
+        expected._mantissa.2 = 0
+        expected._mantissa.3 = 0
+        expected._mantissa.4 = 0
+        expected._mantissa.5 = 0
+        expected._mantissa.6 = 0
+        expected._mantissa.7 = 0
         #expect(Decimal._compare(lhs: expected, rhs: result) == .orderedSame)
```
xwu, line 563: The result is now exact, so `expected` is updated to match.

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -551,20 +544,20 @@ private struct DecimalTests {
...
-                11611,
-                5147
+                56474,
+                1226,
+                57626,
+                58413,
+                21451,
+                2282,
+                50579,
+                51471
```
xwu, line 588: The result now gains additional precision.

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -894,11 +887,11 @@ private struct DecimalTests {
             let aDivD = a / d
             let caDivD = c * aDivD
             #expect(try ab == #require(Decimal(string: "498.7509045")))
-            #expect(try aDivD == #require(Decimal(string: "0.46424522863476857959755328492004843907")))
-            #expect(try caDivD == #require(Decimal(string: "8.453441368210501065891847765109162027")))
+            #expect(try aDivD == #require(Decimal(string: "0.46424522863476857959755328492004843908")))
```
xwu, line 939: The result is now correctly rounded (using default `.plain`) instead of truncated.

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -894,11 +887,11 @@ private struct DecimalTests {
...
+            #expect(try caDivD == #require(Decimal(string: "8.4534413682105010658918477651091620272")))
 
             let result = (a * b) + (c * (a / d))
-            #expect(try result == #require(Decimal(string: "507.2043458682105010658918477651091")))
+            #expect(try result == #require(Decimal(string: "507.20434586821050106589184776510916203")))
```
xwu, line 943: The result now gains additional precision.

xwu:
@swift-ci test macOS

xwu:
cc @stephentyrone :)

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
```diff
@@ -74,7 +74,7 @@ extension Decimal /* : FloatingPoint */ {
 
     /// The decimal value that represents the smallest possible non-zero value for the underlying representation.
     public static let leastNonzeroMagnitude = Decimal(
-        _exponent: -127,
+        _exponent: -128,
```
stephentyrone: Probably makes sense to pull this out as a separate PR, since it's effectively a policy change rather than just the perf fix.

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -24,49 +24,99 @@ import CRT
...
-    /*^30*/ [0x0000, 0x4000, 0xedea, 0x4674, 0x9cd0, 0x9f2c, 0x000c],
-    /*^31*/ [0x0000, 0x8000, 0x4b26, 0xc091, 0x2022, 0x37be, 0x007e],
-    /*^32*/ [0x0000, 0x0000, 0xef81, 0x85ac, 0x415b, 0x2d6d, 0x04ee],
-    /*^33*/ [0x0000, 0x0000, 0x5b0a, 0x38c1, 0x8d93, 0xc644, 0x314d],
-    /*^34*/ [0x0000, 0x0000, 0x8e64, 0x378d, 0x87c0, 0xbead, 0xed09, 0x0001],
-    /*^35*/ [0x0000, 0x0000, 0x8fe8, 0x2b87, 0x4d82, 0x72c7, 0x4261, 0x0013],
-    /*^36*/ [0x0000, 0x0000, 0x9f10, 0xb34b, 0x0715, 0x7bc9, 0x97ce, 0x00c0],
-    /*^37*/ [0x0000, 0x0000, 0x36a0, 0x00f4, 0x46d9, 0xd5da, 0xee10, 0x0785],
-    /*^38*/ [0x0000, 0x0000, 0x2240, 0x098a, 0xc47a, 0x5a86, 0x4ca8, 0x4b3b],
+private let _pow10: [39 of UInt128] = [
```
stephentyrone, line 27: 😃

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -24,49 +24,99 @@ import CRT
...
+    }
+
+    @inline(__always)
+    func _quotientAndRemainderDividingBy10000() -> (quotient: Self, remainder: Self) {
+        let m = 0xD1B71758E219652BD3C36113404EA4A9 as UInt128
+        let q = self.multipliedFullWidth(by: m).high &>> 13
+        let r = self &- q &* 10000
+        return (q, r)
+    }
+
```
stephentyrone, line 103: This operation is weird enough that a comment would be welcome.

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
xwu: Yeah, I debated about it. The practical issue is, now that the arithmetic operations correctly use the full range of the type, the _results_ of ordinary operations (and not just carefully constructed inputs) can more easily be smaller than `leastNonzeroMagnitude` as defined here--an overt violation of the documented guarantee for this API.

The other bit is that the existing implementation (originally by me in the overlay, IIRC) was a straight-up copypasta error and not a policy decision: [existing (longstanding) documentation](https://developer.apple.com/documentation/foundation/nsdecimalnumber) is that the type allows exponents of `-128`:

> `NSDecimalNumber`, an immutable subclass of `NSNumber`, provides an object-oriented wrapper for doing base-10 arithmetic. An instance can represent any number that can be expressed as `mantissa x 10^exponent` where mantissa is a decimal integer up to 38 digits long, and **exponent is an integer from –128 through 127**.
>
> Important: The Swift overlay to the Foundation framework provides the `Decimal` structure, which bridges to the `NSDecimalNumber` class.

Based on these, I concluded that it was a bugfix actually quite intimately related with the other changes here. But happy to split out and discuss separately if this rationale isn't convincing.

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
xwu: Agree, and I'm realizing on re-reading that I should probably annotate the addition and division implementation with additional comments (it did not occur to me in the moment as it's not my first go-round with these, but some of it is non-obvious on first read for the general reader).

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -183,6 +202,7 @@ extension Decimal {
                 r = divisor - r
             }
         } else {
+            // Opposite sign, |b| > |a|.
```
xwu:
```suggestion
            // Opposite sign, |b| >= |a|.
```

xwu:
@swift-ci test macOS

Sources/FoundationEssentials/Decimal/Decimal+Math.swift
```diff
@@ -24,49 +24,116 @@ import CRT
...
+    @inline(__always)
+    func _quotientAndRemainderDividingBy10000() -> (quotient: Self, remainder: Self) {
+        let m = 0xD1B71758E219652BD3C36113404EA4A9 as UInt128
+        let q = self.multipliedFullWidth(by: m).high &>> 13
+        let r = self &- q &* 10000
+        return (q, r)
+    }
+
+    // Full-width division of `(high * 2**128 + low)` by a constant divisor,
+    // using a single step of schoolbook division in base `2**128` (cf. Knuth Algorithm D).
```
xwu:
```suggestion
    // using a single step of schoolbook short division in base `2**128` (cf. Knuth exercise 4.3.1-16).
```

xwu:
@swift-ci test macOS

xwu:
@swift-ci test macOS

xwu:
@swift-ci test macOS

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
xwu: Hmm, while I think `leastNonzeroMagnitude` has an obvious correct answer worth fixing as a bug, there _are_ policy-level changes regarding other constants nearby that (eventually) I'd love to have considered. The churn here wouldn't be great so I'll back this out.

xwu:
@swift-ci test macOS

Sources/FoundationEssentials/Decimal/Decimal+Conformances.swift
```diff
@@ -753,6 +753,8 @@ extension Decimal : SignedNumeric {
         do {
             let result = try lhs._multiply(by: rhs, roundingMode: .plain)
             lhs = result
+        } catch _CalculationError.underflow {
+            lhs = .zero
```
xwu, line 757: Note that this _may_ be formally tantamount to a policy change as compared to `NSDecimal*` guarantees, but is also probably the more (only?) reasonable behavior.

The prior implementation never threw `.underflow`, so there is no actual precedent for this specific operation. In practice, that implementation also had sufficient issues with correctness, precision, and not respecting rounding mode that I'm not sure users could rely upon it to produce zero or NaN (or sometimes a totally unspecified arbitrarily large result—see above).

It is already the behavior in existing code with respect to at least _some_ operations to underflow to zero:
https://github.com/swiftlang/swift-foundation/blob/5da00f0dc72b182f50dd292421b0436b55ddc869/Sources/FoundationEssentials/Decimal/Decimal%2BConformances.swift#L288-L289

xwu:
@swift-ci test macOS

xwu:
Ping @swiftlang/swift-foundation-codeowners  

This is the first tranche of performance (and precision) improvements which, I think, will be welcomed by those who use this type. Hopefully you'll find the approach here to be acceptable, maintaining the documented behavior of all APIs and to the greatest extent possible also maintaining observed behavior (modulo clear bugs as compared to documentation, discussed above). Changes which touch on policy decisions have been deliberately deferred.

Each commit is separately compilable and, if desired, separately reviewable. However, since this work was done by hand, there's some iterative refinement or doubling back in subsequent commits, and the first commit has some undesired whitespace changes (which were subsequently zapped but not easy to retroactively fix up).

I have a follow-up set of changes which, stacked on this one, yield cumulative improvements as follows:

| Benchmark | Time (wall clock) (ns, baseline &rarr; revised, p50) | Throughput (improvement %, p50) |
| ---- | ---: | ---: |
| Add | 111 &rarr; 10 | 1033 |
| Compare | 32 &rarr; 7 | 350 |
| Divide | 1928 &rarr; 10 | 10100 |
| Hash | 44 &rarr; 37 | 17 |
| Init from `Double(1)` | 825 &rarr; 14 | 7200 |
| Init from `Double.pi` | 109 &rarr; 9 | 1100 |
| Multiply | 169 &rarr; 13 | 1167 |

stephentyrone:
These changes are definitely welcome. I've been chatting with Tony about how Foundation wants to get these landed; I'll give them a careful review this weekend and we'll go from there.

xwu:
Awesome!

For context (in case it helps so that you're not expending time to suggest revisions for portions that may be superseded in short order), the next tranche of changes is essentially complete and [viewable](https://github.com/xwu/swift-foundation/compare/decimal-performance...xwu:swift-foundation:decimal-more-performance).

The major _changes_ there that iterate on work in _this_ PR (as opposed to new work) will be elimination of (most) iterative scaling by factors of 10 or 1e4 for [at-most-two-shot scaling by estimation](https://github.com/xwu/swift-foundation/commit/4485bdada8c2fd58cd53247299e603945919c218) (and, consequently, generalization of some `UInt128` helpers to all powers of 10); `compact()` will be optimized with a different approach for [divisibility by 10 without computing remainder](https://github.com/xwu/swift-foundation/commit/40200dc83d7451cc3b573f9ba61cb96969637d29#diff-8f9aeaa7782df20ada5c6b7282f9ea823c0b538563fcb41ad00076a34d65fd10R148).

[The major _new work_ in the next tranche will be an implementation of Russ Cox's unrounded scaling method for binary floating point conversion, a modern reworking of string conversion (preserving quirks but Unicode correct! and fixes JSON decoding bugs!), restoration of correct rounding to a given scaling factor for arithmetic operations (only internal for now), and proper hashing.]

xwu:
Friendly ping :)

stephentyrone (approved):
There's some room for further improvements here, but let's take it as-is and work on the next batch.

xwu:
@swift-ci test macOS

jmschonfeld:
The macOS GitHub actions test was recently added and isn't triggered via `@swift-ci` you may need to push a new commit and/or rebase in order to trigger it

xwu:
As before, lacking merge permissions for this repo, could someone please kindly push the button? Thanks!

parkera:
hey @xwu - can you apply this patch to your branch? It will allow your fix to apply to the framework build, where the definition of `Decimal` comes from `NSDecimal.h` in C rather than our Swift struct.

```patch
diff --git a/Sources/FoundationEssentials/Decimal/Decimal.swift b/Sources/FoundationEssentials/Decimal/Decimal.swift
index 056b801c..b7decaee 100644
--- a/Sources/FoundationEssentials/Decimal/Decimal.swift
+++ b/Sources/FoundationEssentials/Decimal/Decimal.swift
@@ -23,9 +23,6 @@ import ucrt
 /// A structure representing a base-10 number.
 public struct Decimal: Sendable {
-    @_spi(SwiftCorelibsFoundation)
-    public typealias Mantissa = (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)
-
     internal struct Storage: Sendable {
         var exponent: Int8
         // Layout:
@@ -113,18 +110,6 @@ public struct Decimal: Sendable {
         }
     }
 
-    internal var _significand: UInt128 {
-        // Note that per project policy we no longer consider big-endian architectures.
-        @inline(__always) get {
-            return unsafeBitCast(storage.mantissa, to: UInt128.self)
-        }
-        // Note that if `_significand` is set to `0` while `_isNegative == 1`, setting `_length` results in NaN.
-        @inline(__always) set {
-            self.storage.mantissa = unsafeBitCast(newValue, to: Mantissa.self)
-            self._length = UInt32((128 &- newValue.leadingZeroBitCount &+ 15) / 16)
-        }
-    }
-
     internal var _lengthFlagsAndReserved: UInt8 {
         get {
             return self.storage.lengthFlagsAndReserved
@@ -197,6 +182,36 @@ extension Decimal {
 
 #endif // !FOUNDATION_FRAMEWORK
 
+extension Decimal {
+#if FOUNDATION_FRAMEWORK
+    internal typealias Mantissa = (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)
+#else
+    @_spi(SwiftCorelibsFoundation)
+    public typealias Mantissa = (UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16, UInt16)
+#endif
+    
+    internal var _significand: UInt128 {
+        // Note that per project policy we no longer consider big-endian architectures.
+        @inline(__always) get {
+#if FOUNDATION_FRAMEWORK
+            return unsafeBitCast(_mantissa, to: UInt128.self)
+#else
+            return unsafeBitCast(storage.mantissa, to: UInt128.self)
+#endif
+        }
+        // Note that if `_significand` is set to `0` while `_isNegative == 1`, setting `_length` results in NaN.
+        @inline(__always) set {
+#if FOUNDATION_FRAMEWORK
+            _mantissa = unsafeBitCast(newValue, to: Mantissa.self)
+            _length = UInt32((128 &- newValue.leadingZeroBitCount &+ 15) / 16)
+#else
+            self.storage.mantissa = unsafeBitCast(newValue, to: Mantissa.self)
+            self._length = UInt32((128 &- newValue.leadingZeroBitCount &+ 15) / 16)
+#endif
+        }
+    }
+}
+
 // MARK: - String
 extension Decimal {
 #if FOUNDATION_FRAMEWORK
```

parkera:
It moves `Mantissa` and `_significand` to an extension so it applies to both types.

xwu:
Got it: I'll push that change presently.

Please let me know if there's anything I can usefully do locally going forward to test the `FOUNDATION_FRAMEWORK` build configuration (I'm guessing not).

parkera:
Normally there aren't many conflicts like this. `Decimal` is just weird because of its heritage. I don't mind taking care of that side.

xwu:
Thanks @parkera--if it looks good to you from that end, then I think this PR is ready to merge (I don't have permissions for that in this repo).

xwu:
@parkera I forgot to mention—I don't have access, but if these bugs are still open, it might be fun to check if they can be closed once these changes land in the framework build:

rdar://problem/5197585  
rdar://problem/2354750

:)

parkera:
Thanks @xwu - will check

PR #2023, opened 2026-06-03
Post review update for SF-0037

iCharlesHu:
This PR introduces post review updates for SF-0037 to address feedback from SwiftSubprocess 1.0 [Review](https://forums.swift.org/t/review-sf-0037-subprocess-1-0/86004/4) and [Pitch](https://forums.swift.org/t/pitch-subprocess-1-0/85589/32) threads.

### Modifications:

Updated `SF-0037`.

### Result:

N/A

### Testing:

N/A

### DIFF
--- a/Proposals/0037-subprocess-1.0.md
+++ b/Proposals/0037-subprocess-1.0.md
@@ -10,96 +10,87 @@
 
 We introduced Subprocess with SF-0007 and shipped swift-subprocess as a public beta in Spring 2025. Since then, we've received a considerable amount of feedback from the community and updated the Subprocess API to address it. This proposal covers all API changes made since SF-0007 and proposes them for inclusion in the Subprocess 1.0 API.
 
-## Rename `CollectedResult` and Mark `ExecutionResult` Sendable
 
-We propose renaming `CollectedResult` to `ExecutionRecord`. `ExecutionRecord` indicates that we are presenting the recorded data of the child process, and it fits better with its counterpart `ExecutionResult`.
+## Collapse the `run()` Overloads Behind a Generic `Execution` Type
 
-We also want to add the missing `Sendable` conformances to `ExecutionResult`, since the wrapped `value` already has to be `Sendable`.
+In SF-0007, `Execution` is generic over its `Output` and `Error` types, exposing `standardOutput` and `standardError` as conditional properties available only when the corresponding stream is redirected to a sequence. Writing to the child's standard input, however, is handled differently: it requires a *separate* family of `run()` overloads whose body closure receives an extra `StandardInputWriter` argument. This means standard input is handled out of step with the other two streams, and every combination of "writes to standard input or not" needs its own closure-based overload. This design created a combinatory explosion of `run()` overloads.
 
-
-## Adopt `NonisolatedNonsendingByDefault`
-
-We propose adopting the new Swift upcoming feature `NonisolatedNonsendingByDefault`, which allows us to remove all existing `#isolation` parameters:
-
-```diff
---- a/Sources/Subprocess/API.swift
-+++ b/Sources/Subprocess/API.swift
-@@ -132,7 +132,6 @@ public func run<
-     input: Input = .none,
-     output: Output = .discarded,
-     error: Error = .discarded,
--    isolation: isolated (any Actor)? = #isolation,
-     body: (_ execution: Execution) async throws -> Result
- ) async throws -> ExecutionOutcome<Result> where Error.OutputType == Void {
-     let configuration = Configuration(
-```
-
-
-## Remove `.standardOutput` and `.standardError` Properties from `Execution`
-
-We propose moving `.standardOutput` and `.standardError` properties from `Execution` to the `run()` closure parameter alongside `Execution`:
+We propose unifying all three standard streams by making `Execution` generic over its `Input` type as well. The body closure now receives a single `Execution<Input, Output, Error>` value, and type-conditional extensions expose `standardInputWriter`, `standardOutput`, and `standardError` only for the matching stream types:
 
 ```swift
-// Before
-public func run<Result>(
-    ...
-    body: (Execution) async throws -> Result
-) async throws -> ExecutionOutcome<Result> { ... }
-
-let result = try await run(...) { execution in
-    for try await item in execution.standardOutput { ... }
+/// A running subprocess.
+///
+/// The three generic parameters determine which streaming properties are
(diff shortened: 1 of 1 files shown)


Proposals/0037-subprocess-1.0.md
```diff
@@ -3,503 +3,279 @@
...
+    /// buffers.
+    public var standardError: SubprocessOutputSequence { get }
 }
 ```
 
-However, this is not the case. Once you create an `AsyncBufferSequence` by calling `execution.standardOutput`, the returned sequence effectively "owns" the underlying OS pipe used to read data from. This means calling `execution.standardOutput` multiple times is undefined behavior, and we had to use an internal `Atomic` value to guard against it.
+To make this work, the previously internal `CustomWriteInput` and `SequenceOutput` types are now public, along with the `.inputWriter` and `.sequence` factory methods. Callers opt in to each stream independently:
 
-The new design eliminates this problem entirely by "promoting" the output `AsyncBufferSequence`s to be sibling parameters of `Execution`. Since by definition you can't "get" a parameter passed to a closure multiple times, this eliminates the possibility of the aforementioned undefined behavior. This change also simplified `Execution`'s design by making it non-generic.
+- `input: .inputWriter` unlocks `execution.standardInputWriter`;
```
itingliu: sorry... but what does `unlock` mean exactly?

Proposals/0037-subprocess-1.0.md
```diff
@@ -3,503 +3,279 @@
...
 
+## Introduction
 
-## Adopt `NonisolatedNonsendingByDefault`
+We introduced Subprocess with SF-0007 and shipped swift-subprocess as a public beta in Spring 2025. Since then, we've received a considerable amount of feedback from the community and updated the Subprocess API to address it. This proposal covers all API changes made since SF-0007 and proposes them for inclusion in the Subprocess 1.0 API.
 
-We propose adopting the new Swift upcoming feature `NonisolatedNonsendingByDefault`, which allows us to remove all existing `#isolation` parameters:
+The single largest change is a redesign of how the `run()` family is structured: instead of more than a dozen overloads, `Subprocess` now exposes a small set of `run()` functions built on top of a generic `Execution` type. Most of the other changes such as the unified `ExecutionResult`, the public `SubprocessOutputSequence`, and the per-stream opt-in factories follow from this redesign.
```
itingliu: If I read it correctly, this redesign doesn't really change the call-site much, right? If that's the case, can we specifically call it out? Or if not, can we have a before-after example for what the usage looks like?

Proposals/0037-subprocess-1.0.md
```diff
@@ -3,503 +3,279 @@
...
-///   - body: A closure to manage the running process.
-///     All arguments passed to this closure are valid only for
-///     the duration of the closure's execution and must not be escaped.
-///     - execution: The running subprocess.
-///     - outputSequence: The standard output as an asynchronous sequence of buffers.
-/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
+    input: borrowing Span<InputElement>,
+    output: Output,
+    error: Error = .discarded
+) async throws -> ExecutionResult<Void, Output, Error>
```
itingliu, line 157: Can I suggest that we put this `Span`-based input as the second method in the `MARK: - Configuration based` section? That's the order you have above in the `Executable`-based section:
1. input: InputProtocol
2. input: borrowing Span
3. closure-based

This way it's easier to compare them one to one

Proposals/0037-subprocess-1.0.md
```diff
@@ -3,503 +3,279 @@
...
-///   - body: A closure to manage the running process.
-///     All arguments passed to this closure are valid only for
-///     the duration of the closure's execution and must not be escaped.
-///     - execution: The running subprocess.
-/// - Returns: An ``ExecutionResult`` that contains the closure's return value.
+    input: Input,
+    output: Output,
+    error: Error,
+    body: (Execution<Input, Output, Error>) async throws -> Result
+) async throws -> ExecutionResult<Result, Output, Error>
```
itingliu, line 142: With this redesign are we still not able to make `Result: ~Copyable`?

iCharlesHu: I just did an experiment and unfortunately no. It boils down to we need to return this result from `withThrowingTaskGroup` and it requires its `GroupResult` to be `Copyable`.

itingliu: Just to avoid this question coming up and back, can you add this discussion to future direction mentioning the blocking issue?

iCharlesHu: Actually I lied: https://github.com/swiftlang/swift-subprocess/pull/305/changes

itingliu: Awesome! @jmschonfeld will also be glad to know

itingliu:
@swift-ci please test

Proposals/0037-subprocess-1.0.md
```diff
@@ -779,11 +656,117 @@ let result = try await run(
...
+
+```swift
+extension OutputProtocol where Self == FileDescriptorOutput {
+    /// Create a Subprocess output that writes output to the standard output of
+    /// current process.
+    ///
+    /// The file descriptor isn't closed afterwards.
+    public static var currentStandardOutput: Self
+
+    /// Create a Subprocess output that write output to the standard error of
```
itingliu:
```suggestion
    /// Create a Subprocess output that writes output to the standard error of
```

Proposals/0037-subprocess-1.0.md
```diff
@@ -779,11 +656,117 @@ let result = try await run(
 `result.standardOutput` will print `Hello Stdout;\nHello Stderr`.
```
itingliu: Would the result actually be "Hello Stdout\nHello Stderr"? I thought ";" is the shell command separating syntax

Proposals/0037-subprocess-1.0.md
itingliu:
```suggestion
    /// making `nonisolated(unsafe)` safe here.
```

itingliu (approved):

itingliu:
@swift-ci please test

Proposals/0037-subprocess-1.0.md
iCharlesHu: Ahh good catch!

iCharlesHu:
@swift-ci please test

PR #2025, opened 2026-06-08
Redo bplist string bridging to preserve embedded NULs

itingliu:
Should also save absolutely comical amounts of memory in some cases since it no longer autoreleases the string contents

Resolves 178206456


jmschonfeld (approved):

itingliu:
@swift-ci please test

Catfish-Man (approved):
Thank you!

PR #2017, opened 2026-06-02
Reapply hebrew calendar feature

itingliu:
- **Reapply "Add Swift Hebrew calendar implementation  (#1953)" (#2015)**
- **Add Hebrew Calendar file to cmakelist**

PR #2028, opened 2026-06-09
Hebrew calendar fast paths + shared calendar helpers

dra8an:
## Hebrew calendar fast paths + shared calendar helpers

Follow up to PR #1953 (Hebrew calendar port, merged as #2017).

### Summary

**Commit 1**: Add Hebrew calendar fast paths + shared protocol method

Adds `_CalendarProtocol.nextDate(after:matching:direction:)` as an optional fast path hook (default returns nil, existing paths unchanged for all other calendars). `_CalendarHebrew` implements it for:
- `{month, day}` (annual recurrence, e.g. Hanukkah)
- `{month, weekday, weekdayOrdinal}` (Nth weekday of month)
- `{month, weekday, weekOfMonth}` (weekday in Nth week of month)
- Time only `{hour, minute, second}`

`Calendar_Enumerate.swift` probes the fast path at init; `Calendar_Recurrence.swift` adds single combination, multi combination cartesian, and negative ordinal translation short circuits for `RecurrenceRule`.

**Commit 2**: Extract shared calendar helpers into `_CalendarConstants` + `_CalendarUtility`

Addresses PR #1953 review feedback (comments #6 + #7). Introduces shared static helpers for time unit constants, `hash(into:)`, `firstWeekday`, `minimumDaysInFirstWeek`, `copy()`, and `isDateInWeekend`. Hebrew adopts them; Gregorian adoption deferred to a follow up PR.

### Performance

Measured on arm64 (M3 Max), release build, `swift-benchmark`. Zero heap allocations in all cases.

| Benchmark | ICU (C++) | Native Swift | Speedup |
|---|---|---|---|
| `nextThousandHanukkahs` (enumerate {month,day}) | 55 μs | 167 ns | **325×** |
| `dateComponents` {year,month,day} (10K dates) | 3 ns/date | 1 ns/date | **2–3×** |
| `roundTripDateComponents` (10K dates) | 6 ns/date | 2 ns/date | **2–3×** |
| Calendar instantiation + `date(byAdding:)` | 514 ns | 122 ns | **4.2×** |
| Copy on write mutation | 1,736 ns | 31 ns | **54×** |

The 325× enumeration speedup comes from O(1) direct computation fast paths for common date matching patterns, bypassing the generic iterate and test framework entirely.

### Testing

- `HebrewRecurrenceRuleParityProbe.swift`: 13 tests, 392 rule shapes × 2,088 date comparisons, 0 divergences vs Foundation's ICU backed Hebrew calendar.
- All existing tests pass (1100 tests, 59 suites).

### Cross calendar safety

- Non implementing calendars are unchanged: the protocol default returns nil, leaving all existing paths intact.
- The `RecurrenceRule` short circuits probe with a sentinel `_calendarNextDate` call first; non Hebrew calendars bail before any expensive work.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1299,12 +1299,13 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - parameter direction: Which direction in time to search. The default value is `.forward`, which means later in time.
     /// - parameter block: A closure that is called with search results.
     @available(iOS 8.0, *)
+    /// Proxy to the calendar implementation's fast-path `nextDate`.
+    internal func _calendarNextDate(after date: Date, matching components: DateComponents, direction: SearchDirection) -> Date? {
+        _calendar.nextDate(after: date, matching: components, direction: direction)
+    }
+
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
-        // Fast-path: if the calendar implementation has a direct answer for the
-        // first match AND default policies are in effect, drive the loop with
-        // repeated nextDate(after:matching:) calls instead of the generic
-        // enumerate framework. The fast path opts out (returns nil) for any
-        // pattern it can't handle, so this is safe.
+        // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
         if matchingPolicy == .nextTime && repeatedTimePolicy == .first,
            let _ = _calendar.nextDate(after: start, matching: components, direction: direction) {
             var current = start
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1302,19 +1302,17 @@ public struct Calendar : Hashable, Equatable, Sendable {
         _calendar.nextDate(after: date, matching: components, direction: direction)
     }
 
-    /// Whether the calendar implementation supports the `nextDate` fast path.
-    internal var _supportsNextDateFastPath: Bool {
-        _calendar.supportsNextDateFastPath
+    /// Whether the calendar implementation supports the `nextDate` fast path for the given pattern.
+    internal func _supportsNextDateFastPath(for components: DateComponents) -> Bool {
+        _calendar.supportsNextDateFastPath(for: components)
     }
 
     @available(iOS 8.0, *)
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
-        if matchingPolicy == .nextTime && repeatedTimePolicy == .first, _supportsNextDateFastPath,
-           let firstMatch = _calendar.nextDate(after: start, matching: components, direction: direction) {
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first, _supportsNextDateFastPath(for: components) {
+            var current = start
             var stop = false
-            block(firstMatch, true, &stop)
-            var current = firstMatch
             while !stop {
                 guard let next = _calendar.nextDate(after: current, matching: components, direction: direction) else {
                     block(nil, false, &stop)
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1312,9 +1312,11 @@ public struct Calendar : Hashable, Equatable, Sendable {
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
         if matchingPolicy == .nextTime && repeatedTimePolicy == .first,
-           _supportsNextDateFastPath {
-            var current = start
+           _supportsNextDateFastPath,
+           let firstMatch = _calendar.nextDate(after: start, matching: components, direction: direction) {
             var stop = false
+            block(firstMatch, true, &stop)
+            var current = firstMatch
             while !stop {
                 guard let next = _calendar.nextDate(after: current, matching: components, direction: direction) else {
                     block(nil, false, &stop)
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1285,7 +1285,7 @@ public struct Calendar : Hashable, Equatable, Sendable {
         _calendar.nextDate(after: date, matching: components, direction: direction)
     }
 
-    internal func _supportsNextDateFastPath(for components: DateComponents) -> Bool {
+    internal func _supportsNextDateFastPath(for components: ComponentSet) -> Bool {
         _calendar.supportsNextDateFastPath(for: components)
     }
 
@@ -1309,7 +1309,7 @@ public struct Calendar : Hashable, Equatable, Sendable {
     @available(iOS 8.0, *)
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
-        if matchingPolicy == .nextTime && repeatedTimePolicy == .first, _supportsNextDateFastPath(for: components) {
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first, _supportsNextDateFastPath(for: components._populatedComponentSet) {
             var current = start
             var stop = false
             while !stop {
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1298,8 +1298,6 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - parameter repeatedTimePolicy: Determines the behavior of the search algorithm when the input produces a time that occurs twice on a particular day.
     /// - parameter direction: Which direction in time to search. The default value is `.forward`, which means later in time.
     /// - parameter block: A closure that is called with search results.
-    @available(iOS 8.0, *)
-    /// Proxy to the calendar implementation's fast-path `nextDate`.
     internal func _calendarNextDate(after date: Date, matching components: DateComponents, direction: SearchDirection) -> Date? {
         _calendar.nextDate(after: date, matching: components, direction: direction)
     }
@@ -1309,10 +1307,10 @@ public struct Calendar : Hashable, Equatable, Sendable {
         _calendar.supportsNextDateFastPath
     }
 
+    @available(iOS 8.0, *)
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
-        if matchingPolicy == .nextTime && repeatedTimePolicy == .first,
-           _supportsNextDateFastPath,
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first, _supportsNextDateFastPath,
            let firstMatch = _calendar.nextDate(after: start, matching: components, direction: direction) {
             var stop = false
             block(firstMatch, true, &stop)
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1281,6 +1281,14 @@ public struct Calendar : Hashable, Equatable, Sendable {
         }
     }
 
+    internal func _calendarNextDate(after date: Date, matching components: DateComponents, direction: SearchDirection) -> Date? {
+        _calendar.nextDate(after: date, matching: components, direction: direction)
+    }
+
+    internal func _supportsNextDateFastPath(for components: DateComponents) -> Bool {
+        _calendar.supportsNextDateFastPath(for: components)
+    }
+
     /// Computes the dates which match (or most closely match) a given set of components, and calls the closure once for each of them, until the enumeration is stopped.
     ///
     /// There will be at least one intervening date which does not match all the components (or the given date itself must not match) between the given date and any result.
@@ -1298,15 +1306,6 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - parameter repeatedTimePolicy: Determines the behavior of the search algorithm when the input produces a time that occurs twice on a particular day.
     /// - parameter direction: Which direction in time to search. The default value is `.forward`, which means later in time.
     /// - parameter block: A closure that is called with search results.
-    internal func _calendarNextDate(after date: Date, matching components: DateComponents, direction: SearchDirection) -> Date? {
-        _calendar.nextDate(after: date, matching: components, direction: direction)
-    }
-
-    /// Whether the calendar implementation supports the `nextDate` fast path for the given pattern.
-    internal func _supportsNextDateFastPath(for components: DateComponents) -> Bool {
-        _calendar.supportsNextDateFastPath(for: components)
-    }
-
     @available(iOS 8.0, *)
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
--- a/Sources/FoundationEssentials/Calendar/Calendar.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar.swift
@@ -1304,10 +1304,15 @@ public struct Calendar : Hashable, Equatable, Sendable {
         _calendar.nextDate(after: date, matching: components, direction: direction)
     }
 
+    /// Whether the calendar implementation supports the `nextDate` fast path.
+    internal var _supportsNextDateFastPath: Bool {
+        _calendar.supportsNextDateFastPath
+    }
+
     public func enumerateDates(startingAfter start: Date, matching components: DateComponents, matchingPolicy: MatchingPolicy, repeatedTimePolicy: RepeatedTimePolicy = .first, direction: SearchDirection = .forward, using block: (_ result: Date?, _ exactMatch: Bool, _ stop: inout Bool) -> Void) {
         // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
         if matchingPolicy == .nextTime && repeatedTimePolicy == .first,
-           let _ = _calendar.nextDate(after: start, matching: components, direction: direction) {
+           _supportsNextDateFastPath {
             var current = start
             var stop = false
             while !stop {
--- a/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
+++ b/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
@@ -1,0 +1,22 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Time-unit constants shared across calendar implementations.
+internal enum _CalendarConstants {
+    static let kSecondsInWeek = 604_800
+    static let kSecondsInDay = 86400
+    static let kSecondsInHour = 3600
+    static let kSecondsInMinute = 60
+
+    /// Sentinel used by unbounded-range loops in date arithmetic.
+    static let inf_ti: TimeInterval = 4398046511104.0
+}
--- a/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
+++ b/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
@@ -10,13 +10,13 @@
 //
 //===----------------------------------------------------------------------===//
 
-/// Time-unit constants shared across calendar implementations.
-internal enum _CalendarConstants {
-    static let kSecondsInWeek = 604_800
-    static let kSecondsInDay = 86400
-    static let kSecondsInHour = 3600
-    static let kSecondsInMinute = 60
+extension Calendar {
+    /// Time unit constants shared across calendar implementations.
+    static let _kSecondsInWeek = 604_800
+    static let _kSecondsInDay = 86400
+    static let _kSecondsInHour = 3600
+    static let _kSecondsInMinute = 60
 
-    /// Sentinel used by unbounded-range loops in date arithmetic.
-    static let inf_ti: TimeInterval = 4398046511104.0
+    /// Sentinel used by unbounded range loops in date arithmetic.
+    static let _inf_ti: TimeInterval = 4398046511104.0
 }
--- a/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
+++ b/Sources/FoundationEssentials/Calendar/CalendarConstants.swift
@@ -12,11 +12,11 @@
 
 extension Calendar {
     /// Time unit constants shared across calendar implementations.
-    static let _kSecondsInWeek = 604_800
-    static let _kSecondsInDay = 86400
-    static let _kSecondsInHour = 3600
-    static let _kSecondsInMinute = 60
+    static let _secondsInWeek = 604_800
+    static let _secondsInDay = 86400
+    static let _secondsInHour = 3600
+    static let _secondsInMinute = 60
 
-    /// Sentinel used by unbounded range loops in date arithmetic.
-    static let _inf_ti: TimeInterval = 4398046511104.0
+    /// Upper bound for date interval durations in unbounded range loops.
+    static let _maxDateIntervalDuration: TimeInterval = 4398046511104.0
 }
--- a/Sources/FoundationEssentials/Calendar/CalendarUtility.swift
+++ b/Sources/FoundationEssentials/Calendar/CalendarUtility.swift
@@ -71,8 +71,7 @@ internal enum _CalendarUtility {
     // MARK: - isDateInWeekend
 
     /// Whether `weekday` + `timeInDay` falls within `weekendRange`.
-    static func isDateInWeekend(weekday: Int, timeInDay: TimeInterval,
-                                 weekendRange: WeekendRange) -> Bool {
+    static func isDateInWeekend(weekday: Int, timeInDay: TimeInterval, weekendRange: WeekendRange) -> Bool {
         if weekendRange.start == weekendRange.end && weekday != weekendRange.start {
             return false
         } else if weekendRange.start < weekendRange.end && (weekday < weekendRange.start || weekday > weekendRange.end) {
--- a/Sources/FoundationEssentials/Calendar/CalendarUtility.swift
+++ b/Sources/FoundationEssentials/Calendar/CalendarUtility.swift
@@ -1,0 +1,101 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Static helpers shared by `_CalendarProtocol` conformers.
+internal enum _CalendarUtility {
+
+    // MARK: - firstWeekday
+
+    /// Validates a `firstWeekday` value (must be 1...7).
+    static func validatedFirstWeekday(_ value: Int) -> Int {
+        precondition(value >= 1 && value <= 7, "Weekday should be in the range of 1...7")
+        return value
+    }
+
+    /// Resolves `firstWeekday`: stored value → locale preference → 1 (Sunday).
+    static func resolveFirstWeekday(stored: Int?, locale: Locale?) -> Int {
+        if let stored {
+            return stored
+        } else if let locale {
+            return locale.firstDayOfWeek.icuIndex
+        } else {
+            return 1
+        }
+    }
+
+    // MARK: - minimumDaysInFirstWeek
+
+    /// Clamps `minimumDaysInFirstWeek` to 1...7.
+    static func clampedMinimumDaysInFirstWeek(_ value: Int) -> Int {
+        if value < 1 { return 1 }
+        if value > 7 { return 7 }
+        return value
+    }
+
+    /// Resolves `minimumDaysInFirstWeek`: stored value → locale preference → 1.
+    static func resolveMinimumDaysInFirstWeek(stored: Int?, locale: Locale?) -> Int {
+        if let stored {
+            return stored
+        } else if let locale {
+            return locale.minimumDaysInFirstWeek
+        } else {
+            return 1
+        }
+    }
+
+    // MARK: - copy()
+
+    /// Resolves `copy(...)` arguments: override if supplied, else current value.
+    static func resolvedCopyArgs(
+        currentTimeZone: TimeZone, changingTimeZone: TimeZone?,
+        currentLocale: Locale?, changingLocale: Locale?,
+        currentFirstWeekday: Int?, changingFirstWeekday: Int?,
+        currentMinimumDaysInFirstWeek: Int?, changingMinimumDaysInFirstWeek: Int?
+    ) -> (timeZone: TimeZone, locale: Locale?, firstWeekday: Int?, minimumDaysInFirstWeek: Int?) {
+        let newTimeZone = changingTimeZone ?? currentTimeZone
+        let newLocale = changingLocale ?? currentLocale
+        let newFirstWeekday: Int? = changingFirstWeekday ?? currentFirstWeekday
+        let newMinDays: Int? = changingMinimumDaysInFirstWeek ?? currentMinimumDaysInFirstWeek
+        return (newTimeZone, newLocale, newFirstWeekday, newMinDays)
+    }
+
+    // MARK: - isDateInWeekend
+
+    /// Whether `weekday` + `timeInDay` falls within `weekendRange`.
+    static func isDateInWeekend(weekday: Int, timeInDay: TimeInterval,
+                                 weekendRange: WeekendRange) -> Bool {
+        if weekendRange.start == weekendRange.end && weekday != weekendRange.start {
+            return false
+        } else if weekendRange.start < weekendRange.end && (weekday < weekendRange.start || weekday > weekendRange.end) {
+            return false
+        } else if weekendRange.start > weekendRange.end && (weekday > weekendRange.end && weekday < weekendRange.start) {
+            return false
+        }
+
+        if weekday == weekendRange.start {
+            guard let onsetTime = weekendRange.onsetTime, onsetTime != 0 else {
+                return true
+            }
+            return timeInDay >= onsetTime
+        } else if weekday == weekendRange.end {
+            guard let ceaseTime = weekendRange.ceaseTime, ceaseTime < 86400 else {
+                return true
+            }
+            return timeInDay < ceaseTime
+        } else {
+            return true
+        }
+    }
+
+    /// Default weekend range (region 001): Sat–Sun, full day.
+    static let defaultWeekendRange = WeekendRange(onsetTime: 0, ceaseTime: 86400, start: 7, end: 1)
+}
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -331,8 +331,7 @@ extension Calendar {
                 let validates = matchingComponents._validate(for: calendar)
                 finished = !validates
 
-                self.usesFastPath = validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first
-                    && calendar._supportsNextDateFastPath(for: matchingComponents)
+                self.usesFastPath = validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first && calendar._supportsNextDateFastPath(for: matchingComponents)
             }
             
             mutating func next() -> Element? {
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -332,8 +332,7 @@ extension Calendar {
                 finished = !validates
 
                 self.usesFastPath = validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first
-                    && calendar._supportsNextDateFastPath
-                    && calendar._calendarNextDate(after: start, matching: matchingComponents, direction: direction) != nil
+                    && calendar._supportsNextDateFastPath(for: matchingComponents)
             }
             
             mutating func next() -> Element? {
@@ -539,7 +538,7 @@ extension Calendar {
                                          previouslyReturnedMatchDate: Date?) throws -> SearchStepResult {
 
         // Fast-path: ask the calendar directly. Returns nil for unrecognized patterns.
-        if _supportsNextDateFastPath && matchingPolicy == .nextTime && repeatedTimePolicy == .first {
+        if _supportsNextDateFastPath(for: matchingComponents) && matchingPolicy == .nextTime && repeatedTimePolicy == .first {
             if let fast = _calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction) {
                 return SearchStepResult(result: (fast, true), newSearchDate: fast)
             }
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -310,6 +310,9 @@ extension Calendar {
             // Calculated at init, checked on `next`
             var finished: Bool
             
+            // Fast-path: skip `_enumerateDatesStep` when the calendar can answer directly.
+            let usesFastPath: Bool
+
             internal init(_ calendar: Calendar, start: Date, range: Range<Date>?, matching matchingComponents: DateComponents, matchingPolicy: Calendar.MatchingPolicy, repeatedTimePolicy: Calendar.RepeatedTimePolicy, direction: Calendar.SearchDirection) {
                 self.calendar = calendar
                 
@@ -325,12 +328,35 @@ extension Calendar {
                 searchingDate = start
                 
                 // If this fails we'll short circuit the next `next`
-                finished = !matchingComponents._validate(for: calendar)
+                let validates = matchingComponents._validate(for: calendar)
+                finished = !validates
+
+                // Probe: commit to the fast loop if the calendar recognizes this pattern.
+                if validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first,
+                   calendar._calendarNextDate(after: start, matching: matchingComponents, direction: direction) != nil {
+                    self.usesFastPath = true
+                } else {
+                    self.usesFastPath = false
+                }
             }
             
             mutating func next() -> Element? {
                 guard !finished else { return nil }
                 
+                if usesFastPath {
+                    guard let next = calendar._calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction) else {
+                        finished = true
+                        return nil
+                    }
+                    if let range, !range.contains(next) {
+                        finished = true
+                        return nil
+                    }
+                    searchingDate = next
+                    previouslyReturnedMatchDate = next
+                    return next
+                }
+
                 repeat {
                     iterations += 1
                     do {
@@ -515,6 +541,13 @@ extension Calendar {
                                          direction: SearchDirection,
                                          inSearchingDate searchingDate: Date,
                                          previouslyReturnedMatchDate: Date?) throws -> SearchStepResult {
+
+        // Fast-path: ask the calendar directly. Returns nil for unrecognized patterns.
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first {
+            if let fast = _calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction) {
+                return SearchStepResult(result: (fast, true), newSearchDate: fast)
+            }
+        }
 
         // Step A: Call helper method that does the searching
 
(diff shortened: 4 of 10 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -325,12 +328,35 @@ extension Calendar {
                 searchingDate = start
                 
                 // If this fails we'll short circuit the next `next`
-                finished = !matchingComponents._validate(for: calendar)
+                let validates = matchingComponents._validate(for: calendar)
+                finished = !validates
+
+                // Probe: commit to the fast loop if the calendar recognizes this pattern.
+                if validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first,
+                   calendar._calendarNextDate(after: start, matching: matchingComponents, direction: direction) != nil {
```
itingliu: It looks like we don't really need to call this function. We're only calling it to see if the specific calendar implements this function. 

If my understanding of this is correct, this is probably not the best way to do it because
1. It's possible for the protocol witness to have a fast path, but returns `nil` with this given date. In that case we'd miss out the opportunity for the fast path
2. it's possible this function is actually expensive. So it seems wasteful to call it even when we don't really need the result

I can think of two ways to do it
1. Separate this fast path into a separate conformance
2. Add another variable for the witness to implement to signify if they opt into fast path or not

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -516,6 +542,13 @@ extension Calendar {
                                          inSearchingDate searchingDate: Date,
                                          previouslyReturnedMatchDate: Date?) throws -> SearchStepResult {
 
+        // Fast-path: ask the calendar directly. Returns nil for unrecognized patterns.
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first {
+            if let fast = _calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction) {
```
itingliu, line 541: Another issue with using `_calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction)` to check if it's a fast path is that we will be doing repeated work again if this function returns `nil`. We don't know if it's `nil` because there's no fast path, or if this function genuinely has to return `nil`

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
```diff
@@ -0,0 +1,22 @@
...
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Time-unit constants shared across calendar implementations.
+internal enum _CalendarConstants {
```
itingliu: Why do we need a new scope instead of just using `struct Calendar`?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,218 @@ extension Calendar {
...
+    /// Whether the combinations contain any `.nth(N, day)` with `N < 0`.
+    fileprivate func _unadjustedDatesHasNegativeOrdinal(_ c: _DateComponentCombinations) -> Bool {
+        guard let weekdays = c.weekdays else { return false }
+        for w in weekdays {
+            if case .nth(let n, _) = w, n < 0 { return true }
+        }
+        return false
+    }
+
+    /// Expand `_DateComponentCombinations` into a flat array of single-valued
```
itingliu: Like mentioned in https://github.com/swiftlang/swift-foundation/blob/main/CONTRIBUTION_GUIDELINE.md we don't wrap comments

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,218 @@ extension Calendar {
...
+                                for miIdx in 0..<minutesCount {
+                                    for sIdx in 0..<secondsCount {
+                                        guard let dc = make(
+                                            monthIdx: mIdx, weekdayIdx: wIdx,
+                                            daysOfMonthIdx: domIdx,
+                                            daysOfYearIdx: doyIdx,
+                                            weeksOfYearIdx: woyIdx,
+                                            hoursIdx: hIdx, minutesIdx: miIdx,
+                                            secondsIdx: sIdx) else { return nil }
+                                        result.append(dc)
```
parkera: This looks extremely inefficient to me... for one thing, `DateComponents` is a pretty expensive struct to create in the first place because it has storage for all the fields. Also we're in a gigantic nested loop here. There has to be a better way to do this, right?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Good point. Added supportsNextDateFastPath boolean property to _CalendarProtocol (default false). _CalendarHebrew returns true. The probe call is gone, check the boolean instead of calling nextDate speculatively.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Fixed by the same change as #1. The boolean gates all fast path entry points, so _calendarNextDate is only called when we know the calendar implements it. A nil return now unambiguously means "no match in this direction."

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
dra8an: Moved to extension Calendar instead.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Replaced the deep nested loop with an iterative expansion: build a single base DateComponents with all single valued axes pre-filled, then only iterate over axes that actually have multiple values, cloning and patching the base. For example, a rule with one month, one weekday, and two hours creates only two DateComponents  (the base is cloned once per hour value) instead of running through all nesting levels and rebuilding from scratch each time

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -1299,16 +1299,24 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - parameter direction: Which direction in time to search. The default value is `.forward`, which means later in time.
     /// - parameter block: A closure that is called with search results.
     @available(iOS 8.0, *)
```
itingliu, line 1309: This availability and all of the header docs are positioned in the wrong spot. They need to follow `public func enumerateDates`

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -1299,16 +1299,24 @@ public struct Calendar : Hashable, Equatable, Sendable {
...
-        // Fast-path: if the calendar implementation has a direct answer for the
-        // first match AND default policies are in effect, drive the loop with
-        // repeated nextDate(after:matching:) calls instead of the generic
-        // enumerate framework. The fast path opts out (returns nil) for any
-        // pattern it can't handle, so this is safe.
+        // Fast-path: drive the loop with direct nextDate calls when default policies are in effect.
         if matchingPolicy == .nextTime && repeatedTimePolicy == .first,
-           let _ = _calendar.nextDate(after: start, matching: components, direction: direction) {
-            var current = start
+           _supportsNextDateFastPath,
```
itingliu: Nit: Can we please not have an orphan check at its own line? i.e. can we move this to line 1314

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -1299,16 +1299,24 @@ public struct Calendar : Hashable, Equatable, Sendable {
...
+           let firstMatch = _calendar.nextDate(after: start, matching: components, direction: direction) {
```
itingliu: Is there any reason for checking both `_supportsNextDateFastPath` and `firstMatch` instead of just checking only `_supportsNextDateFastPath` ? 

It looks like we'll call `nextDate(after:)` below anyways inside the `while !stop {` loop. So it looks like it'd be fine if we remove this check since we'd still call `nextDate(after:)` at least once. And if we do that we can eliminate this extra check, which is also duplicated code

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -325,12 +328,31 @@ extension Calendar {
                 searchingDate = start
                 
                 // If this fails we'll short circuit the next `next`
-                finished = !matchingComponents._validate(for: calendar)
+                let validates = matchingComponents._validate(for: calendar)
+                finished = !validates
+
+                self.usesFastPath = validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first
+                    && calendar._supportsNextDateFastPath
+                    && calendar._calendarNextDate(after: start, matching: matchingComponents, direction: direction) != nil
```
itingliu: Same question: Why does `usesFastPath` depend on both `_supportsNextDateFastPath` AND `_calendarNextDate(after: start, matching: matchingComponents, direction: direction)` being nil? This doesn't really separate the concept of supporting-fastpath-or-not from the result of the enumeration, which basically brings us back to the previous version, right? Am I missing something?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -1630,34 +1659,46 @@ extension Calendar {
         var result = startDate
         var dateMonth = component(.month, from: result)
         if month != dateMonth {
-            repeat {
-                let lastResult = result
-                
-                guard let foundRange = dateInterval(of: .month, for: result) else {
-                    throw CalendarEnumerationError.dateOutOfRange(.month, result)
+            // Fast-path: advance to target month directly.
+            if !isLeapMonthDesired || !strictMatching {
```
itingliu, line 1662: Can you elaborate why this check?

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
```diff
@@ -50,21 +50,24 @@ package protocol _CalendarProtocol: AnyObject, Sendable, CustomDebugStringConver
     func date(byAdding components: DateComponents, to date: Date, wrappingComponents: Bool) -> Date?
     func dateComponents(_ components: Calendar.ComponentSet, from start: Date, to end: Date) -> DateComponents
 
-    /// Optional fast path for `Calendar.nextDate(after:matching:)`. Default
-    /// returns nil, falling through to the generic enumerate framework. Calendar
-    /// implementations may override for patterns they can answer in O(1).
+    /// Optional fast path for `Calendar.nextDate(after:matching:)`. Returns nil to fall through to the generic enumerate framework.
```
itingliu: Reading this comment again, I wonder if I've misunderstood what this function is supposed to mean

Do we really want to always fall through the generic enumerate function if this function returns nil? As I mentioned last time, isn't it also valid for a calendar to implement a fast-path, but also needs to return nil if we don't find anything? If we really exhausted the search and know for sure there's no matching result, it seems that we should just bail there instead of falling through the generic enumeration.

So I'm asking should we update this comment, or am I missing something about this design?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+        let monthsCount = c.months?.count ?? 1
+        let weekdaysCount = c.weekdays?.count ?? 1
+        let daysOfMonthCount = c.daysOfMonth?.count ?? 1
+        let daysOfYearCount = c.daysOfYear?.count ?? 1
+        let weeksOfYearCount = c.weeksOfYear?.count ?? 1
+        let hoursCount = c.hours?.count ?? 1
+        let minutesCount = c.minutes?.count ?? 1
+        let secondsCount = c.seconds?.count ?? 1
+
+        let total = monthsCount * weekdaysCount * daysOfMonthCount
```
itingliu: Is it ever possible for this to overflow?

Sources/FoundationEssentials/Calendar/Calendar.swift
dra8an: The boolean is calendar-level: "this calendar implements fast paths." The probe is pattern-level: "this specific pattern is supported." Hebrew opts in via the boolean but only handles some patterns ({month, day}, {month, weekday, weekdayOrdinal}, etc.). For patterns it doesn't handle, nextDate returns nil. Without the probe, the while loop would call `nextDate, get nil, call block(nil, false, &stop)`, and terminate the sequence — the user gets nothing instead of falling through to the generic path that would have found results.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Same reasoning as above. The boolean avoids calling nextDate at all for non-opt-in calendars (zero cost). The probe confirms the specific pattern is handled. If the probe returns nil, usesFastPath stays false and the iterator uses _enumerateDatesStep for every call which are correct results via the slow path.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: The fast path advances to the target month directly via nextDate. But when isLeapMonthDesired && strictMatching, the logic below needs to distinguish between the leap and non-leap variant of the month (e.g. Adar vs Adar I in Hebrew). The fast path doesn't carry that distinction, so we skip it and let the existing month-matching logic handle it.

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
dra8an: nextDate returning nil has two meanings depending on context. At the entry gate (the probe), nil means "I don't handle this pattern" so fall through. Inside the committed fast loop (after the probe confirmed the pattern), nil genuinely means "no more matches" so the sequence is done. The probe separates these two cases: it runs once to confirm support, and after that, nil in the loop is unambiguous.

Sources/FoundationEssentials/Calendar/Calendar.swift
dra8an: Fixed. Moved @available back to its correct position above enumerateDates.

Sources/FoundationEssentials/Calendar/Calendar.swift
dra8an: Fixed. Moved _supportsNextDateFastPath to the same line.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+        func translateWeekday(_ entry: RecurrenceRule.Weekday, into dc: inout DateComponents) -> Bool {
+            guard case .nth(let n, let dayOfWeek) = entry else { return false }
+            let wdIdx = dayOfWeek.icuIndex
+            if n > 0 {
+                dc.weekdayOrdinal = n
+                dc.weekday = wdIdx
+                return true
+            }
+            // Negative ordinal → {month, weekday, weekOfMonth}.
+            guard let info = monthInfo else { return false }
```
itingliu: I find it confusing that this function uses a `monthInfo` that is populated outside of the scope. It makes it very difficult to trace when this call is needed.... 

Is there a better way to structure this? For example, pass `monthInfo` as an argument to `translateWeekday` instead of capturing it? Or calculate `monthInfo` directly inside this function?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+    fileprivate func _unadjustedDatesHasNegativeOrdinal(_ c: _DateComponentCombinations) -> Bool {
+        guard let weekdays = c.weekdays else { return false }
+        for w in weekdays {
+            if case .nth(let n, _) = w, n < 0 { return true }
+        }
+        return false
+    }
+
+    /// Expand `_DateComponentCombinations` into a flat array of single-valued `DateComponents`. Negative ordinals are translated to `{month, weekday, weekOfMonth}` using `anchor`'s month structure. Returns nil if the pattern can't be expanded.
+    fileprivate func _expandedDateComponents(
```
itingliu: It looks like this is only called when we're taking the fast path. Any chance we can make it obvious?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+
+            monthInfo = MonthInfo(
+                month: targetMonth,
+                isLeapMonth: targetIsLeap,
+                day1Weekday: day1Weekday,
+                daysInMonth: daysInMonth,
+                firstWeekday: self.firstWeekday,
+                minDays: self.minimumDaysInFirstWeek
+            )
+        }
```
itingliu, line 736: Did I read it right that everything we're doing here is to create a `MonthInfo`? If so, can we add an initializer in `MonthInfo`, and move this block into its initializer? Something like

```
    init?(calendar: Calendar, combinations: _DateComponentCombinations, anchor: Date) {
        // current lines 725-760
    }
```

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+                    return nil
+                case .nth(let n, _):
+                    if n == 0 { return nil }
+                    if n < 0 {
+                        guard anchor != nil else { return nil }
+                        hasNegativeOrdinal = true
+                    }
+                }
+            }
+        }
```
itingliu, line 683: This block is almost the same as `_unadjustedDatesHasNegativeOrdinal`. 

Also, when we call it at line 1092, we first call `_unadjustedDatesHasNegativeOrdinal`, and then `_expandedDateComponents` right after we know there's already a negative ordinal. 

Can we consolidate the chec? Or maybe just remove `_unadjustedDatesHasNegativeOrdinal` if it's always called inside `_expandedDateComponents`?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -828,6 +1061,53 @@ extension Calendar {
...
+                    guard let fast = _calendarNextDate(after: startDate, matching: dc, direction: .forward) else {
+                        allFastPathed = false
+                        break
+                    }
+                    results.append((fast, dc))
+                }
+                if allFastPathed && !results.isEmpty {
+                    results.sort { $0.0 < $1.0 }
+                    return results
+                }
```
itingliu: This is exactly the same as line 1075 above. Anything we can do to avoid duplicated code?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,239 @@ extension Calendar {
...
+            dc.month = info.month
+            dc.isLeapMonth = info.isLeapMonth
+            dc.weekday = wdIdx
+            dc.weekOfMonth = weekOfMonth
+            return true
+        }
+
+        // Build a base DateComponents with single valued axes, then vary only the multi valued ones.
+        var base = DateComponents()
+        var axes: [(WritableKeyPath<DateComponents, Int?>, [Int])] = []
```
itingliu: Keypath isn't the most performant way to do this because it involves runtime dispatch. I'd encourage trying profiling it with plain accessors to get an idea of its cost

Sources/FoundationEssentials/Calendar/Calendar.swift
```diff
@@ -1298,15 +1298,19 @@ public struct Calendar : Hashable, Equatable, Sendable {
     /// - parameter repeatedTimePolicy: Determines the behavior of the search algorithm when the input produces a time that occurs twice on a particular day.
     /// - parameter direction: Which direction in time to search. The default value is `.forward`, which means later in time.
     /// - parameter block: A closure that is called with search results.
+    internal func _calendarNextDate(after date: Date, matching components: DateComponents, direction: SearchDirection) -> Date? {
```
parkera: This new function broke the connection of `enumerateDates` with the public documentation above it.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -325,12 +328,30 @@ extension Calendar {
                 searchingDate = start
                 
                 // If this fails we'll short circuit the next `next`
-                finished = !matchingComponents._validate(for: calendar)
+                let validates = matchingComponents._validate(for: calendar)
+                finished = !validates
+
+                self.usesFastPath = validates && matchingPolicy == .nextTime && repeatedTimePolicy == .first
+                    && calendar._supportsNextDateFastPath(for: matchingComponents)
```
parkera: Please go ahead and keep this on one line, don't split at 80 chars.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -425,10 +392,10 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
         case .era:
             // Hebrew has a single AM era spanning from epoch to effectively forever.
             // Matches ICU's reported start (Hebrew epoch, -181,778,083,200 s before
-            // Date reference = year 1 AM, Tishrei 1, midnight UTC) and duration (inf_ti).
+            // Date reference = year 1 AM, Tishrei 1, midnight UTC) and duration (Calendar._inf_ti).
             return DateInterval(
                 start: Date(timeIntervalSinceReferenceDate: -181_778_083_200.0),
-                duration: inf_ti
+                duration: Calendar._inf_ti
```
parkera: Isn't this a constant anyway...?

parkera (dismissed):

Sources/FoundationEssentials/Calendar/Calendar.swift
dra8an: Fixed. Moved proxy functions before the doc block

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Calendar._inf_ti is already a static let. Did you mean something else here?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Replaced single multiplication with overflow-safe loop using multipliedReportingOverflow. Bails on overflow or exceeding maxCombinations

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Renamed to _fastPathDateComponents

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: _translateWeekday is now a static func taking monthInfo as an explicit parameter

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Removed _unadjustedDatesHasNegativeOrdinal. The check now lives only inside _fastPathDateComponents.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Extracted _probeAllFastPath helper; merged paths (2) and (3) into a single _fastPathDateComponents call that handles both positive and negative ordinals

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Replaced with direct property assignment. No more KeyPath dispatch.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Created _NegativeOrdinalMonthInfo struct with init?(calendar:combinations:anchor:)

Sources/FoundationEssentials/Calendar/CalendarUtility.swift
```diff
@@ -0,0 +1,101 @@
...
+        let newFirstWeekday: Int? = changingFirstWeekday ?? currentFirstWeekday
+        let newMinDays: Int? = changingMinimumDaysInFirstWeek ?? currentMinimumDaysInFirstWeek
+        return (newTimeZone, newLocale, newFirstWeekday, newMinDays)
+    }
+
+    // MARK: - isDateInWeekend
+
+    /// Whether `weekday` + `timeInDay` falls within `weekendRange`.
+    static func isDateInWeekend(weekday: Int, timeInDay: TimeInterval,
+                                 weekendRange: WeekendRange) -> Bool {
```
itingliu: Nit: can we not wrap functions like this? Either align all arguments or not at all, but I don't understand this style

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -516,6 +536,13 @@ extension Calendar {
                                          inSearchingDate searchingDate: Date,
                                          previouslyReturnedMatchDate: Date?) throws -> SearchStepResult {
 
+        // Fast-path: ask the calendar directly. Returns nil for unrecognized patterns.
+        if _supportsNextDateFastPath(for: matchingComponents) && matchingPolicy == .nextTime && repeatedTimePolicy == .first {
+            if let fast = _calendarNextDate(after: searchingDate, matching: matchingComponents, direction: direction) {
```
itingliu, line 541: If `_calendarNextDate` returns nil, doesn't it mean that we truly cannot find a matching date? In this case can we short path this to return `SearchStepResult(result: nil...)`?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -516,6 +536,13 @@ extension Calendar {
                                          inSearchingDate searchingDate: Date,
                                          previouslyReturnedMatchDate: Date?) throws -> SearchStepResult {
 
+        // Fast-path: ask the calendar directly. Returns nil for unrecognized patterns.
```
itingliu: What does this comment "Returns nil for unrecognized patterns." mean?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
itingliu: > The fast path doesn't carry that distinction, so we skip it and let the existing month-matching logic handle it.

How would the callsites know what the fast path supports and what not? Is this something we can fold into `supportsNextDateFastPath()`? 

Also, it looks like elsewhere we always only use the fast path when this condition holds:
```
_supportsNextDateFastPath(for: matchingComponents) && matchingPolicy == .nextTime && repeatedTimePolicy == .first 
```

Would it make sense to modify `_supportsNextDateFastPath` to 
```
_supportsNextDateFastPath(for:matchingPolicy:repeatedTimePolicy:)
```
so we consolidate that check?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -1718,6 +1757,15 @@ extension Calendar {
...
         }
 
+        // Fast-path: enrich with current month and ask the calendar directly.
+        if components.month == nil, components.weekday != nil {
+            var enriched = components
+            enriched.month = component(.month, from: startingAt)
+            if let fast = _calendarNextDate(after: startingAt, matching: enriched, direction: direction) {
+                return fast
+            }
+        }
```
itingliu, line 1770: I thought we switched to use `_supportsNextDateFastPath` to check if fast-path is supported, and use `_calendarNextDate` to only get the result? So, similar question about whether we can truly return `nil` here instead of falling through to the generic path below. 


i.e. can we do 
```
if _supportsNextDateFastPath(...) {
    return _calendarNextDate(after: startingAt, matching: enriched, direction: direction)
}
```

Also my question applies to all uses of `if let fast = _calendarNextDate` in this file

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -1807,6 +1855,11 @@ extension Calendar {
             return nil
         }
 
+        // Fast-path: ask the calendar directly.
+        if let fast = _calendarNextDate(after: startingAt, matching: components, direction: direction) {
+            return fast
+        }
```
itingliu, line 1864: ditto

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -1913,6 +1966,13 @@ extension Calendar {
             return nil
         }
 
+        // Fast-path: use minimal {weekday}-only components.
+        var minimal = DateComponents()
+        minimal.weekday = weekday
+        if let fast = _calendarNextDate(after: startingAt, matching: minimal, direction: direction) {
+            return fast
+        }
+
```
itingliu, line 1978: ditto

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -828,6 +1029,23 @@ extension Calendar {
...
                                       repeatedTimePolicy: RepeatedTimePolicy) throws -> [(Date, DateComponents)]? {
 
+        // Fast-path short-circuits. Only fires when the calendar opts in.
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first {
+
+            // (1) Single-combination: one value per field.
+            if let dc = _singleCombinationDateComponents(combinationComponents),
+               _supportsNextDateFastPath(for: dc),
+               let fast = _calendarNextDate(after: startDate, matching: dc, direction: .forward) {
+                return [(fast, dc)]
```
itingliu: Nit: Please pay attention to the indentation

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -828,6 +1029,23 @@ extension Calendar {
...
+            }
+
+            // (2) Multi-combination: cartesian product (handles both positive and negative ordinals).
+            if let allDCs = _fastPathDateComponents(combinationComponents, anchor: startDate),
+               let results = _probeAllFastPath(allDCs, after: startDate) {
+                return results
```
itingliu, line 983: Nit: I think this indentation is also off

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
```diff
@@ -78,5 +81,16 @@ extension _CalendarProtocol {
...
+    /// Default `hash(into:)` for the standard calendar identity tuple. Calendars with non-standard hashing (e.g. `_CalendarAutoupdating`, `_CalendarICU`, `_CalendarBridged`) override this.
+    package func hash(into hasher: inout Hasher) {
+        hasher.combine(identifier)
+        hasher.combine(timeZone)
+        hasher.combine(firstWeekday)
+        hasher.combine(minimumDaysInFirstWeek)
+        hasher.combine(localeIdentifier)
+        hasher.combine(preferredFirstWeekday)
+        hasher.combine(preferredMinimumDaysInFirstweek)
+    }
```
itingliu: I'm not comfortable providing a default hash implementation. I don't think there's such a thing as "non-standard" or "standard" hashing when we currently already have three different implementations in the code. Providing a default implementation makes it prone to oversight.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,207 @@ extension Calendar {
...
+        for dc in allDCs {
+            guard let fast = _calendarNextDate(after: startDate, matching: dc, direction: .forward) else { return nil }
+            results.append((fast, dc))
+        }
+        results.sort { $0.0 < $1.0 }
+        return results
+    }
+
+    /// Single-valued `DateComponents` from combinations, or nil if expansion is needed.
+    fileprivate func _singleCombinationDateComponents(
```
itingliu: Do we really need this function? it reads to me that `_single​Combination​Date​Components` is just the degenerate case of `_fastPathDateComponents`. Can we not use that one by folding fast paths into that one? Or alternatively is there any common logic we can extract from this function and `_fastPathDateComponents` since they look very similar?

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
```diff
@@ -50,21 +50,24 @@ package protocol _CalendarProtocol: AnyObject, Sendable, CustomDebugStringConver
...
     func dateComponents(_ components: Calendar.ComponentSet, from start: Date, to end: Date) -> DateComponents
 
-    /// Optional fast path for `Calendar.nextDate(after:matching:)`. Default
-    /// returns nil, falling through to the generic enumerate framework. Calendar
-    /// implementations may override for patterns they can answer in O(1).
+    /// Optional fast path for `Calendar.nextDate(after:matching:)`. Nil means no more matches (not "unsupported pattern" — use `supportsNextDateFastPath(for:)` to check that).
     func nextDate(after date: Date, matching components: DateComponents, direction: Calendar.SearchDirection) -> Date?
 
+    /// Whether this calendar can fast path the given pattern. Default returns false; calendar implementations override for patterns they handle.
+    func supportsNextDateFastPath(for components: DateComponents) -> Bool
```
itingliu: I wonder if it's better not to use date component as the argument for validation. Instead, we can pass in all of the fields separately
```
_supportsNextDateFastPath(era:year:month:day:...)
```

The reason is that `DateComponents` aren't cheap memory and performance wise. 

It makes sense if we already have a `DateComponent` such as when we enter from the Calendar enumeration API. But in, say, Calendar_RecurrenceRule call site, we construct `DateComponents` solely for passing into this function, so that we can get the value out of the `DateComponents`. This roundtrip feels unnecessary.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,207 @@ extension Calendar {
...
+            self.isLeapMonth = targetIsLeap
+            self.day1Weekday = calendar.component(.weekday, from: monthStart)
+            self.daysInMonth = dayRange.upperBound - 1
+            self.firstWeekday = calendar.firstWeekday
+            self.minDays = calendar.minimumDaysInFirstWeek
+        }
+    }
+
+    /// Translate a `.nth(N, day)` weekday entry into DateComponents fields. Returns false if out of range.
+    private static func _translateWeekday(_ entry: RecurrenceRule.Weekday, into dc: inout DateComponents, monthInfo: _NegativeOrdinalMonthInfo?) -> Bool {
```
itingliu: I don't understand why we need to use an intermediary `_NegativeOrdinalMonthInfo` instead of just writing into `DateComponents` directly from callsite

Sources/FoundationEssentials/Calendar/CalendarUtility.swift
dra8an: Fixed. Put all parameters on one line.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Now uses if _supportsNextDateFastPath(for:) { return _calendarNextDate(...) } so nil unambiguously means "no results."

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
dra8an: Removed the default. Hebrew now has an explicit hash(into:).

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Removed _singleCombinationDateComponents entirely. _fastPathDateComponents now handles total == 1. _unadjustedDates has a single call to _fastPathDateComponents

Sources/FoundationEssentials/Calendar/Calendar_Protocol.swift
dra8an: Changed to supportsNextDateFastPath(for components: Calendar.ComponentSet). ComponentSet is a UInt bitmask, zero allocation. Call sites that have a DateComponents use ._populatedComponentSet helper that know their literals like [.month] or [.weekday] directly.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Removed. Values are now local variables computed inline and passed directly to _translateWeekday as explicit parameters.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: With supportsNextDateFastPath guarding entry, nil now always means "no results."

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Now returns SearchStepResult(result: nil, newSearchDate: searchingDate) when _calendarNextDate returns nil inside the fast path guard.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Fixed.

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: Fixed.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
dra8an: `supportsNextDateFastPath(for:)` already tells callsites whether a given pattern is supported. For the policy check, the fast path only works with `.nextTime` + `.first` because it does exact matching (no approximation). 

Folding that into `supportsNextDateFastPath(for:matchingPolicy:repeatedTimePolicy` is reasonable but would mean the helper methods need access to those parameters (they're currently not passed in). 

For now the policy check lives at the top level entry points (enumerateDates, DatesByMatching.init, _enumerateDatesStep, _unadjustedDates) which are the only places policies are available. The helpers below are only reached after that check already passed.

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -663,6 +663,207 @@ extension Calendar {
...
+            let (product, overflow) = total.multipliedReportingOverflow(by: count)
+            if overflow || product > maxCombinations { return nil }
+            total = product
+        }
+        if total <= 1 { return nil }
+
+        var monthInfo: _NegativeOrdinalMonthInfo? = nil
+        if hasNegativeOrdinal, let anchor = anchor {
+            guard let info = _NegativeOrdinalMonthInfo(calendar: self, combinations: c, anchor: anchor) else { return nil }
+            monthInfo = info
```
itingliu: My point is, instead of creating a `monthInfo` here then later translating it into a `DateComponents` at line 721 and 728 below, can't we just modify `base` directly? That not only removes the need of a new `_NegativeOrdinalMonthInfo` type but also avoids one extra intermediate allocation?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
itingliu: See my comment above of consolidating the two phase (1) write into a monthInfo and (2) translate that into a `DateComponent` into one thing, with the intention that we  minimize code duplication and unnecessary back-and-forth

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
itingliu: Am I looking at the wrong commit? It doesn't look like it's fixed. I still see a one-space indentation here

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
itingliu: not sure what you meant.. This function is still here?

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
itingliu: Sorry I'm looking at an outdated version. Thanks!

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
```diff
@@ -828,6 +976,14 @@ extension Calendar {
                                       matchingPolicy: MatchingPolicy,
                                       repeatedTimePolicy: RepeatedTimePolicy) throws -> [(Date, DateComponents)]? {
 
+        // Fast-path short-circuits. Only fires when the calendar opts in.
+        if matchingPolicy == .nextTime && repeatedTimePolicy == .first {
```
itingliu, line 980: // Only fires when the calendar opts in.

Which part is this? It looks like this code is called without checking for fast path?

If this is stale can you go over the existing comments to make sure they're updated as you update the implementation?

itingliu (approved):
Just one more comment

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: I checked the latest commit https://github.com/dra8an/swift-foundation/commit/76b9730 and it is showed as removed

Sources/FoundationEssentials/Calendar/Calendar_Recurrence.swift
dra8an: Indeed. The opt-in check (`supportsNextDateFastPath`) happens inside `_probeAllFastPath` per pattern, not at this level. Updated the comment to reflect that. Also did a pass over other comments to make sure they're current

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -566,47 +541,16 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
-        } else if dayOfWeek == weekendRange.end {
-            guard let ceaseTime = weekendRange.ceaseTime, ceaseTime < 86400 else {
-                return true
-            }
-            return timeInDay < ceaseTime
-        } else {
-            return true
-        }
+        let timeInDay = TimeInterval(
+            (comps.hour ?? 0) * Calendar._kSecondsInHour
```
parkera: Seems inconsistent that we define a constant for seconds in hour but not for minutes in second... maybe it should just be 3600 here. I'm ok with limited magic numbers. Not blocking, I guess.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -1038,28 +982,14 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
-        // weekdayOrdinal is only fast-pathed for {month, weekday, weekdayOrdinal}
-        // (e.g. "4th Thursday of November" / "last Friday of Adar"). Any other shape
-        // falls through to the generic enumerate framework.
-        if hasWdOrd && !(hasMonth && hasWeekday && !hasDay) { return nil }
-        // {weekday, day} and {weekday, month} (without wdOrd) aren't fast-pathed.
-        if hasWeekday && hasDay { return nil }
-        if hasWeekday && hasMonth && !hasWdOrd { return nil }
-        if !hasMonth && !hasDay && !hasWeekday { return nil }
+        let hasWeekOfMonth = components.weekOfMonth != nil
+        let timeOnly = !hasMonth && !hasDay && !hasWeekday
```
parkera, line 992: Do other fields like era or dayOfYear matter here? Maybe ignoring them is specific to the Hebrew calendar implementation...

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -1070,15 +1000,39 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
+            return nextMonthWeekdayWeekOfMonthMatch(
                 rd: rd, currentSecsInDay: currentSecsInDay, currentYear: year,
                 targetMonth: components.month!,
                 targetWeekday: components.weekday!,
-                targetWdOrd: components.weekdayOrdinal!,
+                targetWeekOfMonth: components.weekOfMonth!,
                 targetSecsInDay: secsInDay, forward: forward, tz: tz)
         }
 
+        if hasWdOrd {
```
parkera, line 1018: There's really no need to make these abbreviations for argument labels.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -1191,6 +1145,20 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
         return nil
     }
 
+    /// Fast path for `{h, mi, s, ns?}` — next date with the requested time-of-day.
+    private func nextTimeOfDayMatch(rd: Int64, currentSecsInDay: Double,
```
parkera: Same thing here about argument labels - why `rd`?  I'm actually not sure what it means in this context.

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
```diff
@@ -0,0 +1,22 @@
...
+
+extension Calendar {
+    /// Time unit constants shared across calendar implementations.
+    static let _kSecondsInWeek = 604_800
+    static let _kSecondsInDay = 86400
+    static let _kSecondsInHour = 3600
+    static let _kSecondsInMinute = 60
+
+    /// Sentinel used by unbounded range loops in date arithmetic.
+    static let _inf_ti: TimeInterval = 4398046511104.0
```
parkera: Let's give this a sensible name while we're at it.

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+extension Calendar {
+    /// Time unit constants shared across calendar implementations.
+    static let _kSecondsInWeek = 604_800
```
parkera: This `_k` prefix is very "C". It can just be `secondsInWeek`.

parkera:
I removed my CR and just left some really minor style nitpicks for now. Otherwise looks good to me with @itingliu 's approval.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Fixed. Now uses Calendar._secondsInMinute instead of the magic number 60.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: They're already rejected by supportsNextDateFastPath which returns false for era, dayOfYear, weekOfYear, yearForWeekOfYear. By the time nextDate is called, those are guaranteed absent.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Renamed to rataDie (Rata Die is the fixed day number convention from Calendrical Calculations). Internal parameter stays rd for brevity in the arithmetic

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
dra8an: Renamed to _maxDateIntervalDuration

Sources/FoundationEssentials/Calendar/CalendarConstants.swift
dra8an: Renamed to lose k

itingliu:
@swift-ci please test

PR #2029, opened 2026-06-09
Prepare Data(capacity:_:) and Data(rawCapacity:_:) for renaming

jmschonfeld:
Marks `Data`'s relatively new initializers as non-public while its naming is under review

### Motivation:

I'm preparing a pitch for new APIs across the `Data` type that align with new APIs proposed for `UniqueArray` and `RigidArray`. I'd like to reconsider the naming of these initializers to make them consistent with new naming proposed for other to-be-added APIs. Since we haven't shipped them yet, we can avoid a deprecate-and-replace by hiding the new APIs for now until the naming is confirmed 

### Modifications:

Marks the two initializers as `internal` or `@_spi` as needed to hide them without breaking clients.

### Result:

These APIs are not widely accessible until the naming is finalized

### Testing:

Existing unit tests still pass


jmschonfeld:
@swift-ci please test

glessard (approved):

PR #2039, opened 2026-06-15
PropertyListEncoder can create XML documents containing null bytes

kperryua:
The XML spec explicitly forbids NUL bytes. We should prevent clients from being able to create XML property lists containing them.

Since this is a restriction of just XML, this prohibition should not apply to binary plists. As annoying as embedded NUL strings are when used without care, I think we're obligated to preserve them in places where they are not forbidden.

Also, in the spirit of "be liberal in what you accept, but strict in what you produce"—and to reduce the potential risk of this change—I'm not proposing any change to XML plist deserialization, which currently DOES accept NUL bytes.


kperryua:
@swift-ci Please test

parkera (approved):

PR #2041, opened 2026-06-16
6.4: compare CMAKE_SYSTEM_NAME against string literals

MaxDesiatov:
**Explanation**: As part of the bring up of debian13, the platform is using a newer cmake 3.31. cmake 3.31 sets `WASI=1` so:

```cmake
    if(CMAKE_SYSTEM_NAME STREQUAL WASI)
```
evaluates as `"WASI" STREQUAL "1"` on cmake 3.31 (which Is false), silently skipping the `-D_GNU_SOURCE` compile option for WASI targets. This results in 
```
swift-foundation/Sources/FoundationEssentials/Platform.swift:421:16: error: cannot find 'strtod_l' in scope
```
during toolchain builds.

Debian 12 has Cmake 3.25 which has no WASI variable, so WASI fell through to a literal string and the condition evaluated correctly by comparing against the string literal
**Scope**: limited to WASI builds;
**Risk**: low due to limited scope;
**Testing**: tested in local builds and Debian 13 CI.
**Issue**: rdar://179825652
**Reviewer**: @jmschonfeld


MaxDesiatov:
@swift-ci test

jmschonfeld (approved):

PR #2042, opened 2026-06-16
Fixed buffer overrun in error message generation

kperryua:
In the previous fix, I coalesced two scenarios in the error generation—getting a non-hex character right after \U, and getting EOF. In the attempt to helpfully print out the character in the former, I introduced a buffer overrun in the latter.


kperryua:
@swift-ci please test

jmschonfeld (approved):

PR #2044, opened 2026-06-17
Add an alternative considered about using an enum for the version field

parkera:
Update UUIDv7 proposal with an alternative considered about using an enum.


itingliu (approved):

itingliu:
@swift-ci please test

PR #2046, opened 2026-06-17
Benchmarks: add two `Decimal` benchmarks

xwu:
This PR adds two `Decimal` benchmarks (for comparison and hashing).

### Motivation:

These benchmarks are useful to lay the groundwork for subsequent performance optimization.

### Modifications:

A small fixture of 'typical' values is added, with each benchmark using a simple cursor-based method to iterate through these values with minimal in-loop overhead (rather than measuring multiple operations per closure invocation).

### Result:

Two additional benchmarks will be available for `Decimal`.

### Testing:

Benchmarks have been run against `main` (and against subsequent work) for appropriateness of scaling, etc.


parkera (approved):

xwu:
@swift-ci test

xwu:
Looks like I don't have merge permissions in this repo... @parkera would you mind?

PR #2049, opened 2026-06-18
Atomic write: preserve file mode on non-framework platforms too

clintonpi:
Motivation:

`FOUNDATION_FRAMEWORK` preserves the file mode for atomic writes (best-effort). Replicate this behaviour on other platforms.

Modifications:

- Store the file mode before rename operation and try to reapply after.
  - On non-framework platforms (excluding Windows), `mode` was already initialized but calls to `fchmod(mode)` was unreachable since mode stayed `nil`.
  - On Windows, only FILE_ATTRIBUTE_READONLY is preserved.

Result:

Improved cross-platform consistency.

Testing:

Adjusted the `DataIOTests/atomicWriteReplacesReadOnlyDestination` to expect this behaviour on non-framework platforms.


clintonpi:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -433,6 +433,14 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable
         }
 
         try inPath.path.withNTPathRepresentation { pwszPath in
+            let preRenameAttributes: DWORD = {
+                let dwAttributes = GetFileAttributesW(pwszPath)
+                guard dwAttributes != INVALID_FILE_ATTRIBUTES else { return 0 }
+
+                // Hand-pick the necessary attributes. Also, not all attributes can be set with SetFileAttributesW.
+                return dwAttributes & (FILE_ATTRIBUTE_READONLY | FILE_ATTRIBUTE_HIDDEN | FILE_ATTRIBUTE_SYSTEM | FILE_ATTRIBUTE_ARCHIVE)
```
kperryua: We need to be pretty careful in which attributes we carry over. The Darwin implementation does *just* the permissions part of the `mode_t` for now. Sometimes these APis are used to atomically write a file that is  intentionally replacing the original file with one that has a new "identity". In macOS parlance, this could be like a "Save As" operation for a new document over an existing file that the user is choosing to replace. We still do it atomically so that the original file isn't lost, but it's not the "same" file being updated.

But it's also used sometimes used for in-place updates… However, we have the `FileManager.replaceItem` API that is more targeted for that case.

We've discussed the possibility of the API gaining options that preserve more information from the original file when other semantics are desired. The choice of attributes for other platforms should be considered under that light. I'm not sufficiently familiar with Windows to make a judgement call here. Naïvely, I could see `READONLY` being preserved as it's permissions-like, but I'm unsure of the others.

clintonpi:
The Linux Platform check failed because the CI container user running the test was root and bypassed permissions.

clintonpi:
@swift-ci please test

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -525,6 +535,13 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable
         }
 #else
         let newPath = inPath.path
+        var preRenameStat = stat()
+
+        if stat(inPathFileSystemRep, &preRenameStat) == 0 {
+            mode = preRenameStat.st_mode & 0o7777
```
kperryua: Should we mask by `~S_IFMT` instead?

Sources/FoundationEssentials/Data/Data+Writing.swift
clintonpi: Ah, this makes sense.

kperryua:
@clintonpi

The non-Windows portions of this PR are supplanted by my PR in #2077

Once that's merged, I expect you'll want to rebase on top of those changes and reapply the Windows-specific bits.

PR #2050, opened 2026-06-18
[6.4] Add an alternative considered about using an enum for the version field

itingliu:
Cherry-pick of #2044 onto release/6.4.x since the proposal is meant for 6.4

Update UUIDv7 proposal with an alternative considered about using an enum.


itingliu:
@swift-ci please test

PR #2051, opened 2026-06-18
URL v2 file path performance and bug fixes

### LINKED ISSUE #1931: URL appending behavior change?
**Describe the bug**
It looks like there was a behavior change in the appending method around the slash character `/`

**To Reproduce**
`test.swift`
```
import Foundation

let baseURL = try URL(string: "https://example.com")!
let result1 = baseURL.appending(component: "folder/", directoryHint: .inferFromPath)
if result1.absoluteString == "https://example.com/folder%2F" {
  print("pass")   // Previous behavior
} else {
  print("fail")   // New behavior
  exit(1)
}
```

Command line
```
>swiftc test.swift
>.\test.exe
fail
```

**Expected behavior**

At commit b935c8fbdcde72965143affbef3d7690a7f6b2e3 after two PRs/commits
- https://github.com/swiftlang/swift-foundation/pull/1857 649fd827e30bd7be7971d49a86560fdb411906b1
- https://github.com/swiftlang/swift-foundation/pull/1921 b935c8fbdcde72965143affbef3d7690a7f6b2e3

**Regression information:**
It didn't occur at commit cab05a5e9e04218348079c8a8b6235e94bcc398f before the above two commits

jrflat:
Improve correctness, compatibility, and performance of file-path parsing and manipulation in the Swift 6.4 `URL` v2 implementation, with ~2,000 new lines of unit tests to lock down the behavior and achieve near-100% code coverage.

### Motivation:

The path-manipulation methods should not be able to introduce or change a `URL`'s scheme or host, and the stored string should always be unambiguous. Prior `CF`/`NS`/`URL` implementations have ambiguities here because they don't follow RFC 3986 Section 4.2 in many cases. This change closes those ambiguities, aligns behavior with the previous Swift `URL` (v1) for compatibility in some edge cases, and improves path manipulation performance with minor code size wins along the way.

### Modifications:

**Prevent path manipulation from introducing a new scheme or host**

- Following RFC 3986 Section 4.2, path manipulation methods now prepend `./` to a relative path that has no scheme or host and whose first segment contains `:`. For example, `URL("?query").appending(path: "fake:scheme")` becomes `./fake:scheme?query`, so it can't re-parse as having a scheme.
- A path with no host is no longer allowed to begin with `//` (the leading slashes are compressed), which would otherwise re-parse as having a host. Note that this does not affect the common case of canonical `file://` URLs with empty host, e.g. it does not affect `URL(filePath: #"\\server\share"#)` on Windows.
- A new debug `assert` in `_URLInfo.replacingPath` requires that the result is identical to re-parsing the whole string, a strong invariant for the wide range of test and fuzzer inputs to these APIs.

**Align with previous Swift `URL` (v1) in some edge cases**

- Appending `""` to e.g. `http://example.com/path` appends a trailing slash regardless of `isDirectory == false`. Some apps rely on this as a way to append a trailing slash.
- `URL(filePath:)` encodes _every_ `:` in the first segment of a relative path, even when preceding characters aren't scheme-legal.
- `URL(filePath:)`, `appending(path:)`, and `appendingPathExtension(_:)` strip trailing null bytes for file URLs.
- `URL(filePath:)` fixes and logs errors for API misuse when given a full URL string instead of a file path (gated on linked-on-or-after).
- `.standardized` expands `file:/path` to `file:///path` and `scheme://` to `scheme:///`.
- `.appending(component: "folder/", directoryHint: .inferFromPath)` no longer both encodes and appends a trailing slash (resolves #1931).

**Re-organized the appending code to highlight a clear set of rules:**

- `path/` + `/component` drops one slash to avoid a double slash.
- `path` + `component` inserts a single slash.
- `isDirectory: false` strips trailing slashes from the component, but not from the current path.
- `isDirectory: true` adds a trailing slash if one isn't already present.
- An empty relative path with no scheme or host is treated as `.`, so `.` + `fake:scheme` becomes `./fake:scheme` (unambiguous).
- Appending to a scheme-only URL appends directly (`mailto:` + `user@example.com` becomes `mailto:user@example.com`), compressing leading `//` so it isn't read as an authority.

**Performance / code size**

- `_URLInfo.replacingPath` writes the new path directly into the destination `String` buffer (one allocation) instead of building it in a temporary allocation and copying.
- Removed `@inline(__always)` annotations that were either already inlined or bloated code size without a perf benefit. In particular, the `_URLInfo.substring(_:)` helper (~179 lines of `String.utf8` index arithmetic) was pushing the simple component accessors over the inlining size threshold and preventing them from being inlined upward.
- Avoid a `String` copy when the appended `StringProtocol` is already a `String`, via `_specialize`.

### Result:

- Path-manipulation methods can no longer introduce or change a scheme or host, and always produce an unambiguous stored string that round-trips through the parser.
- Behavior matches the previous Swift `URL` (v1) in some edge cases.
- Appending edge cases now follow a clearer, more organized set of rules.
- `URL` path manipulation is 10-20% faster than the prior v2 implementation.

### Testing:

- ~2,000 new lines of unit tests lock down this behavior and reach near-100% coverage of `URL_Impl.swift`.
- Running these URL APIs with fuzzer inputs and the `assert` in `_URLInfo.replacingPath` enabled shows no ambiguities.

### DIFF
--- a/Sources/FoundationEssentials/String/Span+Path.swift
+++ b/Sources/FoundationEssentials/String/Span+Path.swift
@@ -228,4 +228,43 @@ extension Span<UInt8> {
             memcmp(buffer.baseAddress.unsafelyUnwrapped, prefix.utf8Start, prefixLength) == 0
         }
     }
+
+    /// Returns a `String` by compressing consecutive forward slashes.
+    func slashCompressedString() -> String {
+        String(unsafeUninitializedCapacity: count) { buffer in
+            _compressSlashes(into: buffer)
+        }
+    }
+
+    /// - Precondition: `buffer.count >= self.count`
+    private func _compressSlashes(into buffer: UnsafeMutableBufferPointer<UInt8>) -> Int {
+        precondition(buffer.count >= count)
+        var writeIndex = 0, prev = UInt8(0), i = 0
+        while i < count {
+            let v = self[i]
+            buffer[writeIndex] = v
+            // Branchless skip of duplicate slashes
+            writeIndex &+= (v == ._slash && prev == ._slash) ? 0 : 1
+            prev = v
+            i &+= 1
+        }
+        return writeIndex
+    }
+}
+
+extension UnsafeBufferPointer<UInt8> {
+    func count(trailingNullsRemoved: Bool) -> Int {
+        var length = self.count
+        if trailingNullsRemoved {
+            while length > 0 && self[length - 1] == 0 {
+                length -= 1
+            }
+        }
+        return length
+    }
+
+    @_lifetime(borrow self)
+    func span(trailingNullsRemoved: Bool) -> Span<UInt8> {
+        self.span.extracting(first: count(trailingNullsRemoved: trailingNullsRemoved))
+    }
 }
--- a/Sources/FoundationEssentials/URL/URL_Impl.swift
+++ b/Sources/FoundationEssentials/URL/URL_Impl.swift
@@ -26,8 +26,9 @@ import WinSDK
 
 #if FOUNDATION_FRAMEWORK
 internal import _ForSwiftFoundation
-#if canImport(Synchronization)
 internal import Synchronization
+#if canImport(os)
+internal import os
 #endif
 #endif
 
@@ -45,11 +46,7 @@ final class _URL: Sendable {
     #if FOUNDATION_FRAMEWORK
     // Note: We use a lock instead of a lazy var to ensure that we always
     // bridge to the same NSURL even if the URL was copied across threads.
-    #if canImport(Synchronization)
     private let _nsurlLock = Mutex<NSURL?>(nil)
-    #else
-    private let _nsurlLock = LockedState<NSURL?>(initialState: nil)
-    #endif
     private var _nsurl: NSURL {
         return _nsurlLock.withLock {
             if let nsurl = $0 { return nsurl }
@@ -60,7 +57,6 @@ final class _URL: Sendable {
     }
     #endif
 
-    @inline(__always)
     private init(_info: _URLInfo, _baseURL: URL._Impl?) {
         // We shouldn't be passing a base URL if we have a scheme
         assert(_baseURL == nil || !_info.hasScheme)
@@ -73,7 +69,6 @@ final class _URL: Sendable {
 extension _URL: _URLProtocol {}
 #endif
 
-@inline(__always)
 private func hasTrailingSeparator(_ path: String) -> Bool {
     #if os(Windows)
     path.utf8.last == ._backslash || path.utf8.last == ._slash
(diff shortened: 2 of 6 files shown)


jrflat:
@swift-ci please test

jrflat:
@swift-ci please test

jrflat:
Cross-repo smoke testing and toolchain builds succeeded on Linux and macOS. Windows hit infra errors that I'm working to resolve so we can have full confidence.

jrflat:
The Windows toolchain build hit another unrelated error, but Windows PR testing passed, so we should have all the cross-repo bases covered now.

jrflat:
@swift-ci test macOS Platform

Sources/FoundationEssentials/String/Span+Path.swift
```diff
@@ -228,4 +228,43 @@ extension Span<UInt8> {
...
+
+    /// Returns a `String` by compressing consecutive forward slashes.
+    func slashCompressedString() -> String {
+        String(unsafeUninitializedCapacity: count) { buffer in
+            _compressSlashes(into: buffer)
+        }
+    }
+
+    /// - Precondition: `buffer.count >= self.count`
+    private func _compressSlashes(into buffer: UnsafeMutableBufferPointer<UInt8>) -> Int {
```
itingliu, line 240: As a follow up we should convert these `UnsafeMutableBufferPointer` to `OutputSpan`

Sources/FoundationEssentials/URL/URL_Impl.swift
```diff
@@ -915,170 +950,215 @@ extension _URL {
...
-                return replacing(
-                    path: buffer.span.extracting(first: pathLength),
-                    isDirectory: true,
-                    encodingState: encodingState
+            // E.g. "?query" or "#fragment". A few cases of ambiguity to avoid:
+            // - Appending "//path" shouldn't interpret "//" as the authority
+            // - Appending "/path" shouldn't change a relative path to absolute
+            // - Appending "foo:bar" shouldn't interpret "foo" as the scheme
+            // To achieve this, we always treat the current empty path as "."
+            return ("." as StaticString).withUTF8Buffer { dotBuffer in
```
itingliu, line 1060: Also for future: If we find ourselves having to do this StaticString dance more than once, we can also consider adding a specialized version of `appending` that takes a single `UInt8`?

itingliu (approved):
Other than the general comment about avoiding adding more code using unsafe API, the change itself looks good

Sources/FoundationEssentials/String/Span+Path.swift
jrflat: Will follow this up by adding relevant internal helpers for `OutputSpan` (such as bulk appending) and making that transition!

Sources/FoundationEssentials/URL/URL_Impl.swift
jrflat: Yeah, we might be able to use something like `CollectionOfOne(UInt8(ascii: ".")).span` here but would need to be careful about the lifetime. I'll do some investigation in a follow-up! In general, it would be useful if there was an API to get a span of UTF-8 bytes from a `StaticString`, too.

PR #2052, opened 2026-06-18
Add a swift-format configuration file and apply the format

itingliu:
The configuration was generated in a way that minimizes the diff if we apply it

There isn't currently a way to join comments. Tracked in swift-format issue https://github.com/swiftlang/swift-format/issues/1235

The multi-line statement break is also wonky. See the stale discussions below

### DIFF
--- a/.swift-format
+++ b/.swift-format
@@ -1,0 +1,80 @@
+{
+  "fileScopedDeclarationPrivacy" : {
+    "accessLevel" : "private"
+  },
+  "indentBlankLines" : false,
+  "indentConditionalCompilationBlocks" : false,
+  "indentSwitchCaseLabels" : false,
+  "indentation" : {
+    "spaces" : 4
+  },
+  "lineBreakAroundMultilineExpressionChainComponents" : false,
+  "lineBreakBeforeControlFlowKeywords" : false,
+  "lineBreakBeforeEachArgument" : false,
+  "lineBreakBeforeEachGenericRequirement" : false,
+  "lineBreakBetweenDeclarationAttributes" : false,
+  "lineLength" : 250,
+  "maximumBlankLines" : 2,
+  "multiElementCollectionTrailingCommas" : true,
+  "noAssignmentInExpressions" : {
+    "allowedFunctions" : [
+      "XCTAssertNoThrow"
+    ]
+  },
+  "orderedImports" : {
+    "includeConditionalImports" : false,
+    "shouldGroupImports" : false
+  },
+  "prioritizeKeepingFunctionOutputTogether" : false,
+  "reflowMultilineStringLiterals" : "always",
+  "respectsExistingLineBreaks" : true,
+  "rules" : {
+    "AllPublicDeclarationsHaveDocumentation" : false,
+    "AlwaysUseLiteralForEmptyCollectionInit" : false,
+    "AlwaysUseLowerCamelCase" : true,
+    "AmbiguousTrailingClosureOverload" : true,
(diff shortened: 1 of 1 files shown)


Benchmarks/Benchmarks/Data/BenchmarkData.swift
```diff
@@ -77,41 +77,51 @@ let benchmarks: @Sendable () -> Void = {
...
+        })
 
-    Benchmark("DataInitSequence", configuration: .init(tags: ["kind": "smallSlice"]), closure: { benchmark in
-        for _ in benchmark.scaledIterations {
-            blackHole(Data([1, 3, 5, 7, 9, 11, 13, 15, 17, 19, 21, 23, 25, 27, 29, 31, 33]))
-        }
-    })
+    Benchmark(
+        "DataInitSequence", configuration: .init(tags: ["kind": "smallSlice"]),
+        closure: { benchmark in
```
jmschonfeld: The rules here seem inconsistent. I'm not sure when it chooses to wrap the parameters onto its own line (here one parameter is on its own line, 2 are on the same line, and there are still lines longer than 80 chars so I don't think it's column-based wrapping)

Sources/FoundationEssentials/AttributedString/AttributedString+UnicodeScalarView.swift
```diff
@@ -194,13 +194,16 @@ extension AttributedString.UnicodeScalarView: BidirectionalCollection {
     ) -> AttributedString.Index? {
         precondition(i >= startIndex && i <= endIndex, "AttributedString index out of bounds")
         precondition(limit >= startIndex && limit <= endIndex, "AttributedString index out of bounds")
-        guard let j = _guts.string.unicodeScalars.index(
-            i._value, offsetBy: distance, limitedBy: limit._value
-        ) else {
+        guard
+            let j = _guts.string.unicodeScalars.index(
```
jmschonfeld: Another place where the line wrapping/breaking looks off to me

Benchmarks/Benchmarks/Data/BenchmarkData.swift
itingliu: Apparently this has nothing to do with line wrapping length. It's internal to swift-format's handling of multi-line arguments... basically if an argument value spans across multiple lines, the argument would be moved to the next line.

We can workaround this format rule by switching to the trailing closure by removing `closure: `, but it won't work when we have multiple closures like the other benchmark below:
```
           closure: { _, box in
                blackHole(box.d1 == box.d2)
            },
            setup: { () -> TwoDatasBox in
                TwoDatasBox(d1: data, d2: data)
            })
```

I found a few similar issues in swift-format, though I'm not sure if this is exactly the same as those

Sources/FoundationEssentials/AttributedString/AttributedString+UnicodeScalarView.swift
itingliu: This is similar to the other closure case actually. You manually broke after `index`, and thanks to a "respect manual break: true" setting in the config, that line break is preserved. Because of that line break, this statement is treated as a multi-line statement. swift-format's implementation insists on moving argument whose value spans multiple lines into its own line 

I haven't tried but I believe a workaround of this is to keep this entire statement in one line:
```
guard let j = _guts.string.unicodeScalars.index(i._value, offsetBy: distance, limitedBy: limit._value) else {
    return nil
}
```

I think there's an issue opened on swift-format that says this doesn't follow the preferred SwiftUI style

PR #2057, opened 2026-06-23
Read POSIX locale environment on non-Darwin

Kyle-Ye:
## Summary

Read the POSIX locale environment in the non-Darwin `LocaleCache` fallback path before falling back to `en_001`.

The current non-Darwin implementation always reports `en_001`, which makes APIs such as `Locale.current.region?.identifier` return `001` even when the process was launched with a concrete locale such as `en_US.UTF-8` or `zh_CN.UTF-8`.

## Changes

- Check `LC_ALL`, `LC_MESSAGES`, then `LANG` on Linux/Android/BSD/musl builds.
- Normalize POSIX locale values by removing encoding/modifier suffixes and converting hyphen separators to underscores for locale identifiers.
- Preserve the existing `en_001` fallback when the environment is unset or explicitly `C`/`POSIX`.
- Add coverage for `LANG`, `LC_MESSAGES`, `LC_ALL`, `LC_ALL=C`, and hyphenated locale values.

## Links

- Forum Discussion: https://forums.swift.org/t/foundation-locale-current-is-incorrect/77231
- Related PR: https://github.com/swiftlang/swift-foundation/pull/1308

### DIFF
--- a/Sources/FoundationEssentials/Locale/Locale_Cache.swift
+++ b/Sources/FoundationEssentials/Locale/Locale_Cache.swift
@@ -291,9 +291,22 @@ struct LocaleCache : Sendable, ~Copyable {
     }
 #else
     private static let fallbackLocaleIdentifier = "en_001"
+    private static let maximumPOSIXLocaleIdentifierUTF8Count = 156 // Source: ICU's ULOC_FULLNAME_CAPACITY - 1.
+    private static let allowedPOSIXLocaleIdentifierScalars = Set("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789_-.@".unicodeScalars)
 
+    // Source: ICU's POSIX default-locale lookup uses LC_MESSAGES, indirectly
+    // following POSIX setlocale(3)'s category order: LC_ALL, the category
+    // variable, then LANG. POSIX locale names commonly use
+    // language[_territory][.codeset][@modifier]; Foundation's locale identifier
+    // for this fallback only needs the base language/territory/variant portion.
     private static func _localeIdentifier(fromPOSIXLocaleIdentifier identifier: String) -> String? {
         guard !identifier.isEmpty && identifier != "C" && identifier != "POSIX" else {
+            return nil
+        }
+        guard identifier.utf8.count <= Self.maximumPOSIXLocaleIdentifierUTF8Count else {
+            return nil
+        }
+        guard identifier.unicodeScalars.allSatisfy(Self.allowedPOSIXLocaleIdentifierScalars.contains) else {
             return nil
         }
 
--- a/Sources/FoundationEssentials/Locale/Locale_Cache.swift
+++ b/Sources/FoundationEssentials/Locale/Locale_Cache.swift
@@ -290,19 +290,57 @@ struct LocaleCache : Sendable, ~Copyable {
         return preferredLocaleID
     }
 #else
+    private static let fallbackLocaleIdentifier = "en_001"
+
+    private static func _localeIdentifier(fromPOSIXLocaleIdentifier identifier: String) -> String? {
+        guard !identifier.isEmpty && identifier != "C" && identifier != "POSIX" else {
+            return nil
+        }
+
+        let endIndex = identifier.firstIndex { character in
+            character == "." || character == "@"
+        } ?? identifier.endIndex
+        let baseIdentifier = String(identifier[..<endIndex])
+        guard !baseIdentifier.isEmpty && baseIdentifier != "C" && baseIdentifier != "POSIX" else {
+            return nil
+        }
+
+        return baseIdentifier.replacing("-", with: "_")
+    }
+
+    private static func _preferredLanguageIdentifier(fromLocaleIdentifier identifier: String) -> String {
+        let languageIdentifier = String(identifier.split(separator: "@", maxSplits: 1)[0]).replacing("_", with: "-")
+        return Locale.canonicalLanguageIdentifier(from: languageIdentifier)
+    }
+
+    private static func _localeIdentifierFromEnvironment() -> String? {
+        #if !NO_PROCESS && (os(Linux) || os(Android) || os(FreeBSD) || os(OpenBSD) || canImport(Musl))
+        for environmentVariable in ["LC_ALL", "LC_MESSAGES", "LANG"] {
+            guard let value = Platform.getEnvSecure(environmentVariable), !value.isEmpty else {
+                continue
+            }
+            return _localeIdentifier(fromPOSIXLocaleIdentifier: value)
+        }
+        #endif
+
+        return nil
+    }
+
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Locale/Locale_Cache.swift
```diff
@@ -290,19 +290,57 @@ struct LocaleCache : Sendable, ~Copyable {
...
+        let endIndex = identifier.firstIndex { character in
+            character == "." || character == "@"
+        } ?? identifier.endIndex
+        let baseIdentifier = String(identifier[..<endIndex])
+        guard !baseIdentifier.isEmpty && baseIdentifier != "C" && baseIdentifier != "POSIX" else {
+            return nil
+        }
+
+        return baseIdentifier.replacing("-", with: "_")
+    }
```
parkera, line 322: I'm curious about the provenance of this algorithm (and the one below). Is it based on some other code we have in Foundation, or some external source?

Sources/FoundationEssentials/Locale/Locale_Cache.swift
```diff
@@ -290,19 +290,57 @@ struct LocaleCache : Sendable, ~Copyable {
         return preferredLocaleID
     }
 #else
+    private static let fallbackLocaleIdentifier = "en_001"
+
+    private static func _localeIdentifier(fromPOSIXLocaleIdentifier identifier: String) -> String? {
+        guard !identifier.isEmpty && identifier != "C" && identifier != "POSIX" else {
```
parkera, line 303: I think we may want to add a few more checks here, considering the untrusted source of this string. Perhaps a length limit, maybe verifying all ASCII.

parkera:
Thanks @Kyle-Ye, I think this is a good approach.

Sources/FoundationEssentials/Locale/Locale_Cache.swift
Kyle-Ye: Done. I added defensive validation before using the environment value: a maximum UTF-8 length, plus an ASCII-only allowed scalar set for the POSIX locale characters we accept.

Invalid values now fall back to `en_001`, including non-ASCII input, unexpected punctuation such as `/`, and overly long strings. I also added tests covering those cases.

Sources/FoundationEssentials/Locale/Locale_Cache.swift
Kyle-Ye: Good point. This line is only normalizing to Foundation/ICU-style locale identifiers. POSIX locale names usually already use `_`, but this also accepts BCP 47-shaped input such as `pt-BR.UTF-8` and turns it into the identifier shape used elsewhere by Foundation (`pt_BR`). This matches the existing CLDR conversion in `Locale.identifier(.cldr)`, which changes BCP 47 `-` separators to `_`, and `Locale.Components.icuIdentifier` also uses ICU-style identifiers like `de_DE`.

Based on this https://github.com/swiftlang/swift-foundation/blob/b58f9fedfcc466c179f8a16d75f1a9c342f21ee5/Sources/FoundationInternationalization/Locale/Locale_ICU.swift#L403

Kyle-Ye:
@swift-ci please test

Sources/FoundationEssentials/Locale/Locale_Cache.swift
```diff
@@ -290,19 +290,70 @@ struct LocaleCache : Sendable, ~Copyable {
...
-        prefs.languages = ["en-001"]
+        let localeIdentifier = Self._localeIdentifierFromEnvironment() ?? Self.fallbackLocaleIdentifier
+        prefs.locale = localeIdentifier
+        prefs.languages = [Self._preferredLanguageIdentifier(fromLocaleIdentifier: localeIdentifier)]
         return (prefs, true)
     }
 
     func preferredLanguages(forCurrentUser: Bool) -> [String] {
-        [Locale.canonicalLanguageIdentifier(from: "en-001")]
+        let localeIdentifier = Self._localeIdentifierFromEnvironment() ?? Self.fallbackLocaleIdentifier
```
itingliu, line 351: Why don't we just call `preferredLocale()` so we don't need to duplicate this logic?

Sources/FoundationEssentials/Locale/Locale_Cache.swift
```diff
@@ -290,19 +290,70 @@ struct LocaleCache : Sendable, ~Copyable {
...
+        let baseIdentifier = String(identifier[..<endIndex])
+        guard !baseIdentifier.isEmpty && baseIdentifier != "C" && baseIdentifier != "POSIX" else {
+            return nil
+        }
+
+        return baseIdentifier.replacing("-", with: "_")
+    }
+
+    private static func _preferredLanguageIdentifier(fromLocaleIdentifier identifier: String) -> String {
+        let languageIdentifier = String(identifier.split(separator: "@", maxSplits: 1)[0]).replacing("_", with: "-")
```
itingliu, line 325: If I read it correctly, currently all of the callers of this function already sanitize the identifier so it would not actually contain "`@`". Does that preprocess at callsites allow us to remove this extra work?

Sources/FoundationEssentials/Locale/Locale_Cache.swift
```diff
@@ -290,19 +290,70 @@ struct LocaleCache : Sendable, ~Copyable {
...
+    // language[_territory][.codeset][@modifier]; Foundation's locale identifier
+    // for this fallback only needs the base language/territory/variant portion.
+    private static func _localeIdentifier(fromPOSIXLocaleIdentifier identifier: String) -> String? {
+        guard !identifier.isEmpty && identifier != "C" && identifier != "POSIX" else {
+            return nil
+        }
+        guard identifier.utf8.count <= Self.maximumPOSIXLocaleIdentifierUTF8Count else {
+            return nil
+        }
+        guard identifier.unicodeScalars.allSatisfy(Self.allowedPOSIXLocaleIdentifierScalars.contains) else {
```
itingliu, line 309: I think iterating through UTF8 and check if the respective valid range contains the utf8 code unit would be faster than iterating through the unicode scalar, like this
```
func isAllowed(_ byte: UInt8) -> Bool {
    switch byte {
    case UInt8(ascii: "A")...UInt8(ascii: "Z"),
         UInt8(ascii: "a")...UInt8(ascii: "z"),
         UInt8(ascii: "0")...UInt8(ascii: "9"),
         UInt8(ascii: "_"), UInt8(ascii: "-"), UInt8(ascii: "."), UInt8(ascii: "@"):
        return true
    default:
        return false
    }
}

// to call it
identifier.utf8.allSatisfy(isAllowedByte)
```
I think this still works even if the string contains characters that are multiple-utf8-bytes long

itingliu: Although, currently we're iterating the identifier multiple times in this function. First time here for validation, second time below to find "." and '@", and third time for replacing "-" with "_". Would it make sense to collapse the traversal into one loop?

PR #2058, opened 2026-06-23
Avoid crashing when removing empty subrange from empty Data

### LINKED ISSUE #2054: Data.removeSubrange crashes (after allocating unnecessary memory)
**Describe the bug**
`Data` `removeSubrange` crashes if `Data` is empty and the given range has no length.  And it wastes time creating an unnecessary, duplicate empty `Data` instance beforehand.

**To Reproduce**
```swift
import Foundation
var d = Data()
d.removeSubrange(0..<0)
```

**Expected behaviour**
Nothing (`d` remains empty, no runtime failure, and no additional memory allocations or retain traffic).

**Configuration**
Swift version 6.4-dev (LLVM 0aa1511f1bca4a6, Swift 80702ec6ad4159b)
Linux (Debian-based).

**Regression information**
The same code does not crash in Swift 6.2.4 (though I didn't check if it wastes memory too).

**Additional context**

#### Detailed Control Flow Trace

1. **Initialization of Empty `Data`**: 
   * `var d = Data()` creates a `Data` instance that [initially] points to the shared empty singleton representation in `Data+Representation.swift`:

     ```swift
     @usableFromInline static let empty = __DataStorage.init(bytes: nil, length: 0)
     ```

     This singleton has `_capacity = 0`, `_length = 0`, and `_bytes = nil`.

2. **Calling `removeSubrange(0..<0)`**:
   * `removeSubrange` basically just does `replaceSubrange(0..<0, with: nil, count: 0)`, inside Data+Representation.swift:

     ```swift
     mutating func replaceSubrange(_ subrange: Range<Index>, with bytes: UnsafeRawPointer?, count cnt: Int) {
         …
         ensureUniqueReference()
         let upper = _slice.upperBound  // 0
         let nsRange = (
                location: subrange.lowerBound,  // 0
                length: subrange.upperBound - subrange.lowerBound)  // 0 - 0 = 0
         _storage.replaceBytes(in: nsRange, with: bytes, length: cnt)  // Arguments: (0, 0), nil, 0
         ...
     }
     ```

3. **`ensureUniqueReference` & Copy-On-Write**:
   * `ensureUniqueReference()` checks `isKnownUniquelyReferenced(&_storage)`. Because the singleton is shared, it evaluates to `false`.
   * It makes a copy via `_storage = _storage.mutableCopy(_slice)`.  This produces a new `__DataStorage` instance that is now uniquely-referenced but still has a `_capacity` of `0`, `_length` of 0, and a `nil` `_bytes` pointer.

     **That's the first bug - `removeSubrange` shouldn't even call `ensureUniqueReference` when it's not going to have an effect, as that wastes CPU time and RAM with an unnecessary, duplicate `__DataStorage` instance.**

4. **Replacing Bytes (`replaceBytes`)**:
   * The execution in `replaceSubrange(_:with:count:)` moves to `_storage.replaceBytes(in: (0, 0), with: nil, length: 0)` inside `DataStorage.swift`.
   * `replaceBytes` checks: `if resultingLength > currentLength`. Since `0 > 0` is false, it executes the `else` branch:
     ```swift
     } else {
         ensureUniqueBufferReference()
     }
     ```

5. **Erroneous Force-Unwrap in `ensureUniqueBufferReference`**:
   * Inside `ensureUniqueBufferReference(...)`:
     * The guard fails [at its job]; it short-circuits the key check that would avoid the crash.
        ```swift
        guard isExternallyOwned || newLength > _capacity else { return }
        ```

        `isExternallyOwned` is **`true`** because `_capacity == 0`.  So the length check never runs.
     * The `if newLength == 0` block runs:
       ```swift
       if newLength == 0 {
           if isExternallyOwned {
               let newCapacity = malloc_good_size(_length) // _length is 0, yields newCapacity = 0
               let newBytes = __DataStorage.allocate(newCapacity, false)
               __DataStorage.move(newBytes!, _bytes!, _length)
               …
       ```

       **That's the second bug - it assumes `_bytes` is not nil, even though _nothing_ in the control flow has tried to ensure that.**

       Standard Swift best-practice - avoiding force-unwrapping - would have prevented the crash.

jmschonfeld:
Fixes a crash introduced by the new Data ABI

### Motivation:

Removing the empty subrange `0 ..< 0` from an empty `Data` is a valid operation (although it has no effect). It should not crash and in an ideal world should also not trigger a copy.

### Modifications:

- Fixed a discrepancy in the handling of `nil` pointers in `__DataStorage` to ensure we only copy from non-null pointers
- Added an early-exit in `Data._Representation.replaceSubrange()` to avoid triggering CoW for empty replacements of an empty subrange

### Result:

Removing an empty subrange of `0 ..< 0` from an empty `Data` no longer crashes or triggers CoW

Resolves #2054 

### Testing:

Added an additional test that crashed prior to this change but passes now

### DIFF
--- a/Sources/FoundationEssentials/Data/Representations/Data+Representation.swift
+++ b/Sources/FoundationEssentials/Data/Representations/Data+Representation.swift
@@ -193,6 +193,8 @@ extension Data {
             precondition(startIndex <= subrange.lowerBound, "index \(subrange.lowerBound) is out of bounds of \(startIndex)..<\(endIndex)")
             precondition(subrange.upperBound <= endIndex, "index \(subrange.upperBound) is out of bounds of \(startIndex)..<\(endIndex)")
 
+            if subrange.isEmpty && cnt == 0 { return }
+
             ensureUniqueReference()
             let upper = _slice.upperBound
             let nsRange = (
--- a/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
+++ b/Sources/FoundationEssentials/Data/Representations/DataStorage.swift
@@ -297,7 +297,9 @@ internal final class __DataStorage : @unchecked Sendable {
             if isExternallyOwned {
                 let newCapacity = malloc_good_size(_length)
                 let newBytes = __DataStorage.allocate(newCapacity, false)
-                __DataStorage.move(newBytes!, _bytes!, _length)
+                if let bytes = _bytes {
+                    __DataStorage.move(newBytes!, bytes, _length)
+                }
                 _freeBytes()
                 _bytes = newBytes
                 _capacity = newCapacity
--- a/Tests/FoundationEssentialsTests/DataTests.swift
+++ b/Tests/FoundationEssentialsTests/DataTests.swift
@@ -2992,7 +2992,13 @@ extension DataTests {
         }
         #expect(mutateMe == expected)
     }
-    
+
+    @Test func emptyDataRemoveSubrange() {
+        var empty = Data()
+        empty.removeSubrange(0 ..< 0)
+        #expect(empty.count == 0)
+    }
+
     @Test func rangeOfSlice() throws {
         let data = try #require("FooBar".data(using: .ascii))
         let slice = data[3...] // Bar


jmschonfeld:
@swift-ci please test

iCharlesHu (approved):

itingliu (approved):

PR #2060, opened 2026-06-25
Return CFString from certain Swift CFURL methods to avoid bridging

jrflat:
Return/use `CFString` directly from the Swift implementation of some `CFURL` functions that can be called before Foundation/`NSString` bridging classes are initialized.

### Motivation:

`String` to `NSString` bridging and `NSString` ObjC dispatch require the respective classes to be initialized in Foundation. If certain `CFURL` functions (implemented in Swift) return `String` and are called from CoreFoundation before this initialization occurs, it fails to find the concrete bridging class.

### Modifications:

Return a true `CFString` from `__copySwiftEncodedFSRPath`, and store a true `CFString` in `CFURL`s initialized from a file path/representation.

### Result:

N/A for swift-foundation. These CF APIs work as expected regardless of Foundation initialization status.

### Testing:

Test suites for `NSURL` and `CFURL`.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_C+Encoding.swift
+++ b/Sources/FoundationEssentials/URL/URL_C+Encoding.swift
@@ -109,13 +109,13 @@ extension NSURL {
         return URLEncoder.percentEncode(string: path, encoding: encoding, component: component, skipAlreadyEncoded: false)
     }
 
-    static func __copySwiftEncodedFSRPath(_ path: String) -> String? {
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

jrflat:
@swift-ci test macOS Platform

itingliu (approved):

PR #2062, opened 2026-06-25
Fix two-digit era year parsing

### LINKED ISSUE #2061: Date.ParseStrategy parses two-digit Japanese and ROC era years in the wrong century
### Describe the bug

`Date.ParseStrategy`, `Date.VerbatimFormatStyle.parseStrategy`, and `Date.FormatStyle.parse(_:)` can parse two-digit year-of-era values in Japanese and ROC calendars as dates 100 years too far in the future.

For example, `平成10年` is parsed as 2098 instead of 1998, and `民國01年` is parsed as 2012 instead of 1912.

### To Reproduce

```swift
import Foundation

let timeZone = TimeZone(secondsFromGMT: 0)!
var gregorian = Calendar(identifier: .gregorian)
gregorian.timeZone = timeZone

func ymd(_ date: Date) -> String {
    let c = gregorian.dateComponents([.year, .month, .day], from: date)
    return String(format: "%04d-%02d-%02d", c.year!, c.month!, c.day!)
}

let japaneseLocale = Locale(identifier: "ja_JP@calendar=japanese")
var japaneseCalendar = Calendar(identifier: .japanese)
japaneseCalendar.locale = japaneseLocale
japaneseCalendar.timeZone = timeZone

let japaneseFormat: Date.FormatString =
    "\(era: .wide)\(year: .defaultDigits)年\(month: .defaultDigits)月\(day: .defaultDigits)日"
let japaneseStrategy = Date.ParseStrategy(
    format: japaneseFormat,
    locale: japaneseLocale,
    timeZone: timeZone,
    calendar: japaneseCalendar,
    isLenient: false
)

print("平成09:", ymd(try japaneseStrategy.parse("平成09年1月1日")))
print("平成10:", ymd(try japaneseStrategy.parse("平成10年1月1日")))
print("令和01:", ymd(try japaneseStrategy.parse("令和01年5月1日")))

let rocLocale = Locale(identifier: "zh_TW@calendar=roc")
var rocCalendar = Calendar(identifier: .republicOfChina)
rocCalendar.locale = rocLocale
rocCalendar.timeZone = timeZone

let rocFormat: Date.FormatString =
    "\(era: .wide)\(year: .defaultDigits)年\(month: .defaultDigits)月\(day: .defaultDigits)日"
let rocStrategy = Date.ParseStrategy(
    format: rocFormat,
    locale: rocLocale,
    timeZone: timeZone,
    calendar: rocCalendar,
    isLenient: false
)

print("民國01:", ymd(try rocStrategy.parse("民國01年1月1日")))
print("民國58:", ymd(try rocStrategy.parse("民國58年1月1日")))
print("民國59:", ymd(try rocStrategy.parse("民國59年1月1日")))
```

Actual output:

```text
平成09: 2097-01-01
平成10: 2098-01-01
令和01: 2119-05-01
民國01: 2012-01-01
民國58: 2069-01-01
民國59: 1970-01-01
```

### Expected behavior
The parsed year should remain scoped to the parsed era:

```text
平成09: 1997-01-01
平成10: 1998-01-01
令和01: 2019-05-01
民國01: 1912-01-01
民國58: 1969-01-01
民國59: 1970-01-01
```
### Configuration (please complete the following information):
 - Swift Version: Apple Swift version 6.3 (swiftlang-6.3.0.123.5)
 - OS: macOS
 - OS Version: 26.4.1

### Regression information:
Not confirmed.

### Additional context
Also reproduced against `swift-foundation` `release/6.4.x`.

The affected APIs all use `ICUDateFormatter`. `ICUDateFormatter` currently calls `udat_set2DigitYearStart` unconditionally, using the parse strategy's `twoDigitStartDate` default of `1970-01-01`.

ICU applies the two-digit year window when a `y` pattern with width less than 3 parses exactly two digits. That is appropriate for two-digit Gregorian years, but it also affects explicit year-of-era values such as `平成10` and `民國01`.

JapaneseCalendar normally disables default-century pivoting in ICU, but the unconditional `udat_set2DigitYearStart` call re-enables it for this formatter.

One possible fix is to avoid applying ICU's two-digit year window to explicit era-year parsing, for example by using a parse-only wider year field for Japanese and ROC patterns that contain an era field.

This behavior was also independently observed and discussed on X here: [URL](https://x.com/ukokkei95toyama/status/2069599686634459569)

vistar941:
Fix two-digit era year parsing for Japanese and ROC calendars.
Fixes https://github.com/swiftlang/swift-foundation/issues/2061

### Motivation:

`ICUDateFormatter` sets a two-digit-year start date for parsing. ICU then treats exactly two parsed year digits with a narrow `y` pattern as a two-digit year window, even when the value is already scoped by an explicit era such as `平成` or `民國`.

This causes Japanese and ROC era-year inputs such as `平成10年` and `民國01年` to parse into the wrong century.

### Modifications:

- Use a parse-only date pattern for `Date.FormatStyle` parsing while preserving the formatting pattern.
- Apply the same parse-only pattern adjustment to `Date.ParseStrategy` fixed-format parsing.
- For Japanese and ROC calendars with an explicit unquoted era field, widen only affected unquoted `y`/`yy` year-of-era fields to `yyyy` for parsing.
- Leave `yyy`/`yyyy`, patterns without an era field, quoted pattern text, and escaped apostrophes unchanged.
- Add coverage for `Date.ParseStrategy`, `VerbatimFormatStyle`, `Date.FormatStyle.parse`, ROC edge cases, and focused parse-pattern rewriting behavior.

### Result:

Japanese and ROC era-year parsing no longer applies ICU's two-digit-year window to explicit year-of-era values. Formatting output remains unchanged because the wider year field is used only for parsing.

### Testing:

- `swift test --filter FoundationInternationalizationTests.DateFormatStyleTests/eraYearParsePatternOnlyWidensAffectedYearFields`
- `swift test --filter FoundationInternationalizationTests.DateVerbatimFormatStyleTests/parseableEraYearsDoNotUseTwoDigitYearWindow`
- `swift test --filter FoundationInternationalizationTests.DateFormatStyleTests`
- `swift test --filter FoundationInternationalizationTests.DateVerbatimFormatStyleTests`

### DIFF
--- a/Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
+++ b/Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
@@ -282,6 +282,10 @@ final class ICUDateFormatter : @unchecked Sendable {
         cachedFormatter(for: .init(format))
     }
 
+    static func cachedFormatter(forParsing format: Date.FormatStyle) -> ICUDateFormatter? {
+        cachedFormatter(for: .init(format, forParsing: true))
+    }
+
     static func cachedFormatter(for format: Date.VerbatimFormatStyle) -> ICUDateFormatter? {
         cachedFormatter(for: .init(format))
     }
@@ -293,7 +297,7 @@ final class ICUDateFormatter : @unchecked Sendable {
 }
 
 extension ICUDateFormatter.DateFormatInfo {
-    init(_ format: Date.FormatStyle) {
+    init(_ format: Date.FormatStyle, forParsing: Bool = false) {
         let calendarIdentifier = format.calendar.identifier
         let datePatternOverride: String?
 #if FOUNDATION_FRAMEWORK
@@ -323,6 +327,13 @@ extension ICUDateFormatter.DateFormatInfo {
             }
         }
 
+        let dateFormatPattern: String
+        if forParsing {
+            dateFormatPattern = Self.parsePattern(for: pattern, calendarIdentifier: calendarIdentifier)
+        } else {
+            dateFormatPattern = pattern
+        }
+
         let firstWeekday: Int
         if let forceFirstWeekday = format.locale.forceFirstWeekday(calendarIdentifier) {
             firstWeekday = forceFirstWeekday.icuIndex
@@ -330,7 +341,7 @@ extension ICUDateFormatter.DateFormatInfo {
             firstWeekday = format.calendar.firstWeekday
         }
 
-        self.init(localeIdentifier: format.locale.identifier, timeZoneIdentifier: format.timeZone.identifier, calendarIdentifier: calendarIdentifier, firstWeekday: firstWeekday, minimumDaysInFirstWeek: format.calendar.minimumDaysInFirstWeek, capitalizationContext: format.capitalizationContext, pattern: pattern, parseLenient: format.parseLenient)
+        self.init(localeIdentifier: format.locale.identifier, timeZoneIdentifier: format.timeZone.identifier, calendarIdentifier: calendarIdentifier, firstWeekday: firstWeekday, minimumDaysInFirstWeek: format.calendar.minimumDaysInFirstWeek, capitalizationContext: format.capitalizationContext, pattern: dateFormatPattern, parseLenient: format.parseLenient)
     }
 
     init(_ format: Date.VerbatimFormatStyle) {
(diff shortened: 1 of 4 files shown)


itingliu:
My understanding of the problem, like you said, is that we are setting the two-digit year start date unconditionally. That causes ICU to treat "10" in "平成10/01/01" as an offset from the two-digit year base and ignores the era.

In that case, would a better fix be not setting the two-digit start year when there's no ambiguity in parsing the year, like when era is specified? 

Although I'm not sure what ICU uses as the default value when it is left unset. We'll probably need to check that too. And if ICU logic is also causing problems in the case of parsing, I'd be interested in knowing where they're doing wrong so we can follow up with the ICU owners

vistar941:
@itingliu 
Thank you for the suggestion. I checked what happens when the two-digit-year start date is left unset.

For the Japanese calendar, ICU’s `JapaneseCalendar::haveDefaultCentury()` returns `false`, so ICU normally disables century pivoting for that calendar. Therefore, avoiding Foundation’s explicit `udat_set2DigitYearStart` call would fix the Japanese case.

However, that does not fix the ROC calendar. ICU’s `TaiwanCalendar` uses its system-default-century behavior, which initializes a 100-year window starting 80 years before the current date. As a result, leaving the value unset still causes values such as `民國01` to be interpreted as `民國101` (2012), rather than `民國1` (1912).

I also considered whether ICU itself should treat a normal-width `y` as a literal year whenever an explicit era field (`G`) is present. However, that would change ICU’s currently documented behavior. `SimpleDateFormat` documents that exactly two input digits parsed with `y` or `yy` are interpreted using the default-century window, while year fields wider than two characters are interpreted literally. Likewise, `udat_set2DigitYearStart` is documented as applying to all two-digit years, with no exception for an explicit era. Introducing such an exception would therefore be an observable compatibility change that would require a separate ICU specification and compatibility discussion.

Given these differences between the Japanese and ROC calendar defaults, I think the current parse-only pattern approach is the safer and more contained fix in Foundation. It uses ICU’s existing documented behavior for wider year fields, applies only when parsing an explicit year-of-era in the affected calendars, and leaves the formatting pattern unchanged. 

A broader change in ICU, including an update to the documented semantics, may be worth considering in the future, but I think it is outside the scope of this Foundation fix.

I would appreciate your thoughts on whether this reasoning and the current approach make sense 😊

vistar941:
Hi @itingliu, just a gentle ping on this when you have a chance. Thanks!

Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
```diff
@@ -343,6 +354,80 @@ extension ICUDateFormatter.DateFormatInfo {
...
+        // ICU applies its two-digit year window only to year fields narrower
+        // than 3. Era years are already scoped by the parsed era, so widen
+        // only those affected fields for parsing.
+        return replacingUnquotedFieldRuns(in: pattern, matching: "y") { count in
+            guard count < 3 else {
+                return String(repeating: "y", count: count)
+            }
+            return "yyyy"
+        }
+    }
```
itingliu: Can we avoid iterating the string twice, once in `containsUnquotedField` and `replacingUnquotedFieldRuns`? can we collapse them into one loop? Something like

```swift
        var containsEra = false
        var result = ""
        forEachUnquotedFieldRun(in: pattern) { runField, runLength in
            switch runField {
            case "G":
                containsEra = true
                result += String(repeating: "G", count: runLength)
            case "y":
                result += runLength < 3 ? "yyyy" : String(repeating: "y", count: runLength)
            default:
                result += String(repeating: String(runField), count: runLength)
            }
        } quotedRun: { quotedText in
            result += quotedText
        }

        return containsEra ? result : pattern
```

Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
```diff
@@ -343,6 +354,80 @@ extension ICUDateFormatter.DateFormatInfo {
...
+        }
+        return result
+    }
+
+    private static func forEachUnquotedFieldRun(in pattern: String, _ body: (Character, Int) -> Void, quotedRun: ((String) -> Void)? = nil) {
+        var index = pattern.startIndex
+        var isInQuote = false
+
+        while index < pattern.endIndex {
+            let character = pattern[index]
```
itingliu: If I remember correctly, Chinese and Japanese patterns are all single unicode scalars, as in there are no grapheme cluster characters used in the patterns. In that case would it be safe to iterate through `pattern.unicodeScalars` instead of `Character` by character (so that we can skip the cost of Unicode grapheme-cluster-break algorithm)?

itingliu (commented):
Thanks for following up on this! Had some comments

Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
vistar941: Done. I combined era detection and year-field rewriting into a single pass.

Sources/FoundationInternationalization/Formatting/Date/ICUDateFormatter.swift
vistar941: Yes, that is safe here. I switched the scanner to pattern.unicodeScalars and added coverage for combining scalars.

itingliu (commented):
Thanks! The change looks good to me. But we're past 6.4.x release now. Do you mind changing the target to main?

vistar941:
> Thanks! The change looks good to me. But we're past 6.4.x release now. Do you mind changing the target to main?

@itingliu 
Updated the PR to target main and rebased the branch. Thanks!

itingliu (approved):

PR #1836, opened 2026-03-18
UUID v7 - Proposal, implementation, and performance enhancements

parkera:
Add a new API proposal for UUID, plus corresponding implementation.

### Motivation:

See the proposal for this.

### Modifications:

New proposal, new implementation changes to support it, plus associated performance changes.

### Result:

Benchmark shows 185% improvement in UUIDCreate, 23% improvement in UUIDString. Also removes the need for C shims for UUID.

### Testing:

New unit tests, new benchmarks.

PR #2064, opened 2026-06-26
Implement FOU-0041 - UUID v7 and other improvements

parkera:
Implement FOU-0041, UUID v7 and other improvements.

Includes updated availability macros from [the previous PR](https://github.com/swiftlang/swift-foundation/pull/1836).

Other changes: uses SE-0525 versions of some of Span APIs, where possible.


parkera:
@glessard the compiler is crashing on this change, but in the meantime - can you review the bytes-and-span-related changes?

Sources/FoundationEssentials/UUID.swift
```diff
@@ -9,35 +9,62 @@
...
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
+
+internal import Synchronization
 
 public typealias uuid_t = (UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8, UInt8)
 public typealias uuid_string_t = (Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8, Int8)
 
+private extension OutputSpan where Element: BitwiseCopyable {
```
glessard: This should probably be
```suggestion
private extension OutputSpan where Element: ConvertibleFromBytes {
```

Sources/FoundationEssentials/UUID.swift
```diff
@@ -46,38 +73,124 @@ public struct UUID : Hashable, Equatable, CustomStringConvertible, Sendable {
...
+    /// Creates a UUID by copying exactly 16 bytes from a `RawSpan`.
+    ///
+    /// - Precondition: `span.count` must be exactly 16.
+    @available(FoundationPreview 6.5, *)
+    public init(copying span: RawSpan) {
+        precondition(span.byteCount == 16, "UUID requires exactly 16 bytes, but \(span.byteCount) were provided")
+        // Is this safe? Probably not because we don't know the alignment of span
+        // let pieces = Span<UInt64>(viewing: span)
+        // _storage = [pieces[0], pieces[1]]
+        self.init { outputRawSpan in
```
glessard: It should now be possible to load directly (no alignment required) with `load(fromByteOffset: 0, as: InlineArray<2, UInt64>.self)` directly to `_storage`.

Sources/FoundationEssentials/UUID.swift
```diff
@@ -104,26 +217,8 @@ public struct UUID : Hashable, Equatable, CustomStringConvertible, Sendable {
...
-            UInt8(truncatingIfNeeded: secondBits >> 32),
-            UInt8(truncatingIfNeeded: secondBits >> 24),
-            UInt8(truncatingIfNeeded: secondBits >> 16),
-            UInt8(truncatingIfNeeded: secondBits >> 8),
-            UInt8(truncatingIfNeeded: secondBits)
-        )
-
-        return UUID(uuid: uuidBytes)
+        let inline: [2 of UInt64] = [firstBits.bigEndian, secondBits.bigEndian]
+        return UUID(copying: inline.span.bytes)
```
glessard, line 212: We could also have an internal initializer that takes the `[2 of UInt64]` directly.

glessard (approved):

parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
@swift-ci test

parkera:
Rebased to fix conflict in CMakeLists

parkera:
@swift-ci test

Sources/FoundationEssentials/UUID.swift
```diff
@@ -46,38 +73,117 @@ public struct UUID : Hashable, Equatable, CustomStringConvertible, Sendable {
...
+    public init(copying span: RawSpan) {
+        precondition(span.byteCount == 16, "UUID requires exactly 16 bytes, but \(span.byteCount) were provided")
+        _storage = span.load(fromByteOffset: 0, as: [16 of UInt8].self)
+    }
+
+    /// Creates a UUID by filling its 16 bytes using a closure that writes into an `OutputRawSpan`.
+    ///
+    /// The closure must write exactly 16 bytes into the output span.
+    @available(FoundationPreview 6.5, *)
+    public init<E: Error>(initializingWith initializer: (inout OutputRawSpan) throws(E) -> ()) throws(E) {
```
kperryua: Isn't `-> Void` preferred over `-> ()`?

Sources/FoundationEssentials/UUID.swift
```diff
@@ -46,38 +73,117 @@ public struct UUID : Hashable, Equatable, CustomStringConvertible, Sendable {
...
+                try initializer(&outputRawSpan)
+                let count = outputRawSpan.byteCount
+                precondition(count == 16, "UUID requires exactly 16 bytes, but \(count) were provided")
+            }
+        })
+    }
+
+    // Hex lookup tables for UUID string formatting. Each byte is converted to two hex characters via table lookup.
+    private static let _upperHex: StaticString = "0123456789ABCDEF"
+    private static let _lowerHex: StaticString = "0123456789abcdef"
```
kperryua: Should these be `InlineArray` per recent trends?

Sources/FoundationEssentials/UUID.swift
```diff
@@ -46,38 +73,117 @@ public struct UUID : Hashable, Equatable, CustomStringConvertible, Sendable {
...
+
+    /// Writes the UUID as a 36-character hex string into `buffer` using the given hex digit lookup table. Returns 36.
+    private func _unparse(into buffer: UnsafeMutableBufferPointer<UInt8>, hexTable: StaticString) -> Int {
```
kperryua: Should this debug-assert that `buffer` is at least 36 bytes?

Sources/FoundationEssentials/UUID.swift
```diff
@@ -174,17 +261,253 @@ extension UUID : Codable {
...
+extension UUID {
+    /// The version of this UUID, derived from the version bits (bits 48–51) as defined by RFC 9562.
+    public var version: Int {
+        get {
+            Int(_storage.span.bytes[6] >> 4)
+        }
+        set {
+            var s = _storage.mutableSpan
+            var b = s.mutableBytes
+            b[6] = (b[6] & 0x0F) | (UInt8(newValue & 0x0F) << 4)
```
kperryua: For a single mutation like this, is there any benefit to going through mutableBytes as opposed to just using the InlineArray's own subscript?

Sources/FoundationEssentials/UUID.swift
```diff
@@ -174,17 +261,253 @@ extension UUID : Codable {
...
+        return UUID(copying: inline.span.bytes)
+    }
+
+    /// For version 7 UUIDs, returns the `Date` encoded in the most significant 48 bits. Returns `nil` for all other versions.
+    ///
+    /// The returned date has millisecond precision, as specified by RFC 9562.
+    ///
+    /// - Note: Even though this implementation, or others, may choose to encode more precision into other bytes of the `UUID`, this method may only return the portion of the timestamp stored in the RFC-specified bytes.
+    public var date: Date? {
+        guard version == 7 else { return nil }
```
kperryua, line 449: Unless the optimizer can rewrite this whole getter, the additional implicit load from the `_storage` to access the `version` is a little unfortunate.

Sources/FoundationEssentials/UUID.swift
parkera: Good question - we use both in various places, but void is more common so let's go with that.

Sources/FoundationEssentials/UUID.swift
parkera: I like that, because we can use a Span for the unparse function.

Sources/FoundationEssentials/UUID.swift
parkera: Good question - this is leftover from going back and forth on if the storage was UInt64, UInt128, or UInt8. I simplified these to just use the InlineArray accessor.

kperryua: Ah I see. In that light, it might make sense to rewrite this in terms of `self.mutableBytes` then, which would then abstract this code away from that if it were to change again in the future. Assuming you don't end up with the same compiler issues mentioned in the context of that accessor.

kperryua (approved):

parkera:
@swift-ci test

PR #2065, opened 2026-06-26
Fix a few warnings

parkera:
Fix 2 warnings in Foundation.

### Motivation:

We previously had no way to silence the `mktemp` deprecation warning in Data+Writing.swift. Now we do!

Fixed another platform-specific warning in NotificationCenter too.

### Modifications:

Moved the deprecated function into a helper so we scope the warning silencing to a small surface area.

### Result:

Two less warnings.

### Testing:

Unit tests are sufficient.


jmschonfeld (approved):

parkera:
@swift-ci test

PR #2072, opened 2026-07-01
Support var default values as replacement for @CodableDefault

### LINKED ISSUE #2001: Consider using `var` default values as replacement for `CodableDefault`
Based on forum discussions [here](https://forums.swift.org/t/new-codable-prototype-available-for-feedback/85186/68), consider using `var` default values as a replacement for `@CodableDefault` macros:

```
struct BlogPost {
    var required: Int
    var items: [String] = []
}

decode(BlogPost.self, from: data) // { "required" : 1, "items": ["hello"] }
// -> BlogPost(required: 1, items: ["hello"])

decode(BlogPost.self, from: data) // { "required" : 1 }
// -> BlogPost(required: 1, items: [])

decode(BlogPost.self, from: data) // { "items": ["hello"] }
// ❌ error, missing field
```

YeoSeongil:
Resolves #2001
Allow var initializer expressions to be used as default values during macro expansion

### Motivation:
When analyzing the macro expansion flow, I traced how `extractDetailedStoredProperties` builds each `DetailedStoredProperty`. The key field is `defaultExpr: String?`, which feeds into `isRequired`:
```swift
var isRequired: Bool { !isOptional && defaultExpr == nil }
```

I found that `defaultExpr` is populated solely by `defaultValueExpression(from:)`, which only looks for `@CodableDefault` attributes. The var binding's initializer expression (e.g. `= []`) is never read.

This means `var items: [String] = []` produces `defaultExpr: nil`, making `isRequired: true`, and decoding fails when `items` is absent from the payload — even though the developer explicitly provided a default value.

The fix is one line: fall back to `binding.initializer?.value.trimmedDescription` when no `@CodableDefault` is present. This value is then passed into the generated decode code as:
```swift
items: items ?? []
```
making the field optional in the decoded payload as intended.

### Modifications:
In `extractDetailedStoredProperties` (SharedMacroInfrastructure.swift), after attempting to extract a default value from `@CodableDefault`, fall back to the var binding's initializer expression if no attribute is present:
```swift
let resolvedDefaultExpr = defaultExpr ?? binding.initializer?.value.trimmedDescription
```

This string is then used downstream in code generation, producing expressions like:
```swift
items: items ?? []
```

### Result:
```swift
var items: [String] = []  // now recognized as non-required
var required: Int // still required, no change
```
```swift
decode(BlogPost.self, from: #"{"required": 1}"#)
// Before: error, `items` treated as required field
// After: success, BlogPost(required: 1, items: [])
```

### Testing:
- `NewCodableMacros` and `NewCodableMacrosTests` targets build successfully with the Swift nightly toolchain (swift-DEVELOPMENT-SNAPSHOT-2026-06-24-a).

- However, running tests was not possible locally — `swift test` attempts to build the full package including the `NewCodable` runtime, which fails to compile with the available nightly toolchain (swift-DEVELOPMENT-SNAPSHOT-2026-06-24-a). This appears to be a pre-existing issue unrelated to this change.

- Happy to add or run tests once the runtime build is unblocked.

### DIFF
--- a/Sources/NewCodableMacros/SharedMacroInfrastructure.swift
+++ b/Sources/NewCodableMacros/SharedMacroInfrastructure.swift
@@ -373,14 +373,16 @@ func extractDetailedStoredProperties(
                 isOptional = false
                 typeName = type.trimmedDescription
             }
-
+          
+            let resolvedDefaultExpr = defaultExpr ?? binding.initializer?.value.trimmedDescription
+          
             properties.append(DetailedStoredProperty(
                 name: propertyName,
                 key: key,
                 aliases: aliases,
                 typeName: typeName,
                 isOptional: isOptional,
-                defaultExpr: defaultExpr,
+                defaultExpr: resolvedDefaultExpr,
                 encodingStrategy: encodingStrat,
                 decodingStrategy: decodingStrat
             ))
--- a/Tests/NewCodableMacrosTests/CommonCodableMacroTests.swift
+++ b/Tests/NewCodableMacrosTests/CommonCodableMacroTests.swift
@@ -394,6 +394,91 @@ struct CommonCodableMacroTests {
             macros: codableTestMacros
         )
     }
+    
+    @Test func varDefaultValue() {
+        AssertMacroExpansion(
+            """
+            @CommonCodable
+            struct Config {
+                let name: String
+                var locale: String = "en"
+            }
+            """,
+            expandedSource: """
+            struct Config {
+                let name: String
+                var locale: String = "en"
+            }
+            
+            extension Config {
+                enum CommonCodingFields: StaticStringCodingField {
+                    case name
+                    case locale
+                    case unknown
+
+                    @_transparent
+                    var staticString: StaticString {
+                        switch self {
+                        case .name:
+                            "name"
+                        case .locale:
+                            "locale"
+                        case .unknown:
+                            fatalError()
+                        }
+                    }
+
+                    static func field(for key: UTF8Span) throws(CodingError.Decoding) -> CommonCodingFields {
+                        switch UTF8SpanComparator(key) {
+                        case "name":
+                            .name
+                        case "locale":
+                            .locale
+                        default:
+                            .unknown
+                        }
+                    }
+                }
+            }
+
+            extension Config: CommonEncodable {
+                func encode(to encoder: inout some CommonEncoder & ~Copyable & ~Escapable) throws(CodingError.Encoding) {
+                    try encoder.encodeStructFields(count: 2) { structEncoder throws(CodingError.Encoding) in
+                        try structEncoder.encode(field: CommonCodingFields.name, value: self.name)
+                        try structEncoder.encode(field: CommonCodingFields.locale, value: self.locale)
+                    }
+                }
+            }
+
+            extension Config: CommonDecodable {
+                static func decode(from decoder: inout some CommonDecoder & ~Escapable) throws(CodingError.Decoding) -> Config {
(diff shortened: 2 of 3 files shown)


kperryua:
> However, running tests was not possible locally — swift test attempts to build the full package including the NewCodable runtime, which fails to compile with the available nightly toolchain (swift-DEVELOPMENT-SNAPSHOT-2026-06-24-a). This appears to be a pre-existing issue unrelated to this change.

This is news to me! (Attention has been elsewhere in the last couple of weeks.)
Care to file a `new-codable` labeled Issue for this with any details?

kperryua:
Will the macro already produce a worthy diagnostic for an illegal implicitly typed property declaration like `var items = <concretely-typed expression>`, or does that need to be added to this change?

YeoSeongil:
> > However, running tests was not possible locally — swift test attempts to build the full package including the NewCodable runtime, which fails to compile with the available nightly toolchain (swift-DEVELOPMENT-SNAPSHOT-2026-06-24-a). This appears to be a pre-existing issue unrelated to this change.
> 
> This is news to me! (Attention has been elsewhere in the last couple of weeks.) Care to file a `new-codable` labeled Issue for this with any details?

Sure! I just re-verified with the same toolchain (swift-DEVELOPMENT-SNAPSHOT-2026-06-24-a) and the `NewCodable` runtime still fails to build. I'll file an issue for this.

YeoSeongil:
> Will the macro already produce a worthy diagnostic for an illegal implicitly typed property declaration like `var items = <concretely-typed expression>`, or does that need to be added to this change?

After reviewing the existing code, it looks like the macro already handles this case. In`extractDetailedStoredProperties`, the `guard let typeAnnotation = binding.typeAnnotation` check fires before any default value extraction occurs:
```swift
guard let typeAnnotation = binding.typeAnnotation else {
    context.diagnose(.init(
        node: Syntax(binding),
        message: SharedMacroDiagnostic.missingTypeAnnotation(macroName: node.attributeName.trimmedDescription)
    ))
    return nil
}
```

If a property lacks an explicit type annotation (e.g., `var items = []` or `var items = [String]()`), the guard fails and SharedMacroDiagnostic.missingTypeAnnotation is emitted via context.diagnose, producing:
```
@JSONCodable requires all stored properties to have explicit type annotations
```
The property is then excluded from code generation entirely (return nil), so no additional diagnostic appears to be needed for this change. That said, actual testing would be required to confirm this behavior.

kperryua:
I suspect this change is good, but now that we understand the build failure you were seeing (requiring swift-6.4) can you perhaps add some tests that explicitly exercise this new functionality and guarantee proper output? Both macro expansion tests and integration tests, preferably. Thank you!

YeoSeongil:
@kperryua Added tests covering this new functionality!

**1. `varDefaultValue`**
checks that var locale: String = "en" expands correctly — the initializer value is used as the default, so the decoder outputs locale ?? "en" instead of throwing a missing field error.

**2. `varWithoutTypeAnnotationProducesDiagnostic`**
checks that var locale = "en" (no type annotation) still produces the expected error, so the new code path doesn't accidentally suppress it.

**3. `varDefaultValueUsedWhenFieldMissing`**
field absent in JSON -> default value used.

**4. `varDefaultValueOverriddenByProvidedField`**
field present in JSON -> JSON value wins.

**5. `varDefaultValueRoundTrip`**
encode then decode -> round-trip produces identical struct.

**Note**
I tried running the integration tests locally with the Swift 6.4 toolchain, but the NewCodable runtime failed to build (same issue as #2074), so these will need to be verified by CI.

PR #2077, opened 2026-07-02
Atomic file writes don't update permissions atomically

kperryua:
Switch to `*at` style syscalls where possible to allow us to atomically apply permissions to the file while maintaining secure behaviors around creating and renaming temporary file.

### DIFF
--- a/Tests/FoundationEssentialsTests/DataIOTests.swift
+++ b/Tests/FoundationEssentialsTests/DataIOTests.swift
@@ -212,8 +212,9 @@ private final class DataIOTests {
         #expect(readData == data)
     }
 
-    // Atomic writes must succeed when the destination is read-only, and the resulting file must not retain the read-only state.
+    // Atomic writes must succeed when the destination is read-only.
     // On POSIX this is the behavior of rename(2). On Windows, SetFileInformationByHandle with FileRenameInfoEx returns ERROR_ACCESS_DENIED for a FILE_ATTRIBUTE_READONLY destination even with FILE_RENAME_FLAG_POSIX_SEMANTICS | FILE_RENAME_FLAG_REPLACE_IF_EXISTS. So the read-only attribute is cleared and the rename is retried.
+    // The resulting file retains the read-only attribute. (Not yet on Windows).
     @Test
     func atomicWriteReplacesReadOnlyDestination() throws {
         let initial = Data("initial".utf8)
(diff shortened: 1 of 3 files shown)


kperryua:
@swift-ci Please test

kperryua:
@swift-ci please test

kperryua:
@swift-ci please test

Tests/FoundationEssentialsTests/DataIOTests.swift
```diff
@@ -230,12 +231,11 @@ private final class DataIOTests {
         let read = try Data(contentsOf: url)
         #expect(read == next)
 
-#if !FOUNDATION_FRAMEWORK
-        // The path now resolves to the new temp file (the rename operation detached the old destination). Confirm it's writable by performing a subsequent non-atomic write.
-        // Skipped for FOUNDATION_FRAMEWORK: that path preserves and reapplies the destination's pre-rename mode on the new file, so the new file would still be read-only.
-        // TODO: Preserve the destination's mode across platforms.
-        let later = Data("later".utf8)
-        try later.write(to: url)
+#if !os(Windows)
```
jmschonfeld, line 234: Is there an equivalent check we should be doing on Windows here? Previously, the `try later.write(to: url)` ran even on Windows but with the latest changes we lose any check on Windows. Presumably we're missing copying the mode over on Windows, but maybe we should still do the `try later.write(to: url)` on Windows until we standardize the behavior there too?

kperryua: I'm not really a fan of testing for behavior that we consider incorrect and intend to change entirely anyway.

@clintonpi 's other PR will hopefully follow shortly after this one to fix the behavior on Windows, which will allow us to just remove this conditional.

jmschonfeld (approved):

finagolfin:
This broke both [the Android](https://ci.swift.org/job/oss-swift-6.4.x-package-swift-sdk-for-android/61/console) and [Windows CI](https://ci-external.swift.org/job/swift-6.4.x-windows-toolchain/58/console), strange as the Android and Windows workflows here ran fine. That indicates either something wrong with the workflows here, or some other weird syncing issue perhaps.

kperryua:
> This broke both [the Android](https://ci.swift.org/job/oss-swift-6.4.x-package-swift-sdk-for-android/61/console) and [Windows CI](https://ci-external.swift.org/job/swift-6.4.x-windows-toolchain/58/console), strange as the Android and Windows workflows here ran fine. That indicates either something wrong with the workflows here, or some other weird syncing issue perhaps.

I merged a change that should fix both of these errors. From what I can tell this error only occurred on 32-bit builds on platforms that have a different system of Libc types than Darwin. swift-foundation CI targets appear to all be 64-bit. I filed a bug requesting the addition of one of these 32-bit targets to our CI to prevent this from occurring again.

finagolfin:
You're right: both build errors are hitting only with 32-bit Android armv7. I'll check if your fix in #2084 worked, then submit a pull enabling 32-bit Android armv7 here.

finagolfin:
Submitted in #2087

PR #2085, opened 2026-07-07
Fix URL.description for long data: URLs

jrflat:
Fix `URL.description` for long data: URLs so that it prints the actual bytes instead of `UTF8View` objects.

### Motivation:

For long (>128 character) data: URLs, `description` incorrectly prints the `UTF8View` objects containing the first 120 and last 8 bytes instead of the bytes themselves.

### Modifications:

Convert each `UTF8View` to `Substring` before performing String interpolation.

### Result:

Long data: URLs return the expected `<first 120 bytes> ... <last 8 bytes>` description.

### Testing:

New unit test to validate description format.


jrflat:
@swift-ci please test

Catfish-Man (approved):
lgtm

PR #2084, opened 2026-07-07
Fixed Linux platform build failure

kperryua:

PR #2087, opened 2026-07-08
[6.4] Fix mode_t conversion from #2077 for 32-bit Android armv7

finagolfin:
This was incorrectly fixed in #2084; add 32-bit Android armv7 testing to catch such build issues from now on.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,6 +25,7 @@ jobs:
       enable_cross_pr_testing: true
       enable_android_sdk_build: true
       android_sdk_versions: '["nightly-6.4.x"]'
+      android_sdk_triples: '["aarch64-unknown-linux-android28", "armv7-unknown-linux-android28", "x86_64-unknown-linux-android28"]'
 
   cmake_build:
     name: Build (CMake)
--- a/Sources/FoundationEssentials/Data/Data+Writing.swift
+++ b/Sources/FoundationEssentials/Data/Data+Writing.swift
@@ -599,7 +599,7 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable 
         var preRenameState = stat()
         let result = fstatat(destDirfd, basenameRep, &preRenameState, AT_SYMLINK_NOFOLLOW)
         if result == 0 {
-            mode = mode_t(preRenameState.st_mode & ~S_IFMT)
+            mode = mode_t(preRenameState.st_mode) & ~S_IFMT
         } else if (errno != ENOENT) && (errno != ENAMETOOLONG) {
             throw CocoaError.errorWithFilePath(inPath, errno: errno, reading: false)
         }


.github/workflows/pull_request.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       enable_cross_pr_testing: true
       enable_android_sdk_build: true
       android_sdk_versions: '["nightly-6.4.x"]'
+      android_sdk_triples: '["aarch64-unknown-linux-android28", "armv7-unknown-linux-android28", "x86_64-unknown-linux-android28"]'
```
finagolfin, line 28: We were already checking the 64-bit triples by default, but the official workflow does not build 32-bit armv7 by default, since few will need that. We do need it here though, since this repo is required for a Swift SDK, and we ship a 32-bit armv7 SDK in the official Android SDK bundle.

finagolfin:
@swift-ci test

finagolfin:
@swift-ci test macos

finagolfin:
@swift-ci test

.github/workflows/pull_request.yml
itingliu: thanks for finding this. I did not know we have a arvm7 target

itingliu (dismissed):

Sources/FoundationEssentials/Data/Data+Writing.swift
```diff
@@ -599,7 +599,7 @@ private func writeToFileAux(path inPath: borrowing some FileSystemRepresentable
         var preRenameState = stat()
         let result = fstatat(destDirfd, basenameRep, &preRenameState, AT_SYMLINK_NOFOLLOW)
         if result == 0 {
-            mode = mode_t(preRenameState.st_mode & ~S_IFMT)
+            mode = mode_t(preRenameState.st_mode) & ~S_IFMT
```
itingliu, line 602: What are we fixing here? 

If the build error is the same as the previous one
```
mode = preRenameState.st_mode & ~S_IFMT
`- error: cannot convert value of type 'UInt32' to expected argument type 'UInt16'
```

I would expect the existing code to work

finagolfin: `S_IFMT` is `mode_t` on Android armv7, so the `&` still fails. This gets the types matched.

itingliu: ah so the error would actually be something like `error: cannot '&' on UInt32 and UInt16`?

kperryua: Isn't `.st_mode` already a `mode_t`? I'm confused what the suggested change accomplishes.

Can you spell out what the concrete types of `st_mode` and `S_IFMT` are on this platform? I just want to make sure I get this right for every platform without additional churn.

.github/workflows/pull_request.yml
itingliu: Can we actually check in this yml separately so I can enable it to run on the PR?

Sources/FoundationEssentials/Data/Data+Writing.swift
finagolfin: > Isn't .st_mode already a mode_t?

Not on Android armv7, it isn't, 😉 see [other instances in the code like this](https://github.com/swiftlang/swift-foundation/blob/7989ed002f158f364f18eecd6b92f5bc2abf50e2/Sources/FoundationEssentials/FileManager/FileOperations.swift#L899).

> Can you spell out what the concrete types of st_mode and S_IFMT are on this platform?

Not worth getting into, as 32-bit Android armv7 is pretty much a legacy platform at this point. If you look at all such mode checks in the existing code, we use this `mode_t` conversion idiom to get around what Android armv7 does. On other platforms, the conversion should do nothing.

You shouldn't have to remember it anymore after this pull, as the CI will now flag any of these that you miss for you.

.github/workflows/pull_request.yml
finagolfin: armv7 did build on this PR, see the Android output.

finagolfin:
@swift-ci test

itingliu (approved):

Sources/FoundationEssentials/Data/Data+Writing.swift
finagolfin: Btw, [this is the exact Android armv7 error output without this pull](https://ci.swift.org/job/oss-swift-6.4.x-package-swift-sdk-for-android/63/console), but with the previous conversion change:
```
02:45:26  /source/swift-project/swift-foundation/Sources/FoundationEssentials/Data/Data+Writing.swift:602:50: error: binary operator '&' cannot be applied to operands of type 'UInt32' and 'mode_t' (aka 'UInt16')
02:45:26  600 |         let result = fstatat(destDirfd, basenameRep, &preRenameState, AT_SYMLINK_NOFOLLOW)
02:45:26  601 |         if result == 0 {
02:45:26  602 |             mode = mode_t(preRenameState.st_mode & ~S_IFMT)
02:45:26      |                                                  |- error: binary operator '&' cannot be applied to operands of type 'UInt32' and 'mode_t' (aka 'UInt16')
02:45:26      |                                                  `- note: overloads for '&' exist with these partially matching parameter lists: (UInt16, UInt16), (UInt32, UInt32)
```

PR #2094, opened 2026-07-09
Add SortComparator / KeyPathComparator benchmarks

dnadoba:
# Add SortComparator / KeyPathComparator benchmarks

Adds benchmarks for Foundation's sort-comparator stack: `KeyPathComparator`,
`ComparableComparator`, `OptionalComparator`, and `Sequence.sorted(using:)`,
under `InternationalizationBenchmarks`. These APIs have no benchmark coverage
today, so this adds a baseline to track their performance and catch regressions.
Spoiler: I have a patch that will improve these significantly and resolve
rdar://161022764.

## What it measures

The fixture is `Entry`, a struct with a few `Int`, `String`, `Int?`, a computed
`Double?`, and a `UUID` field, plus `EntryObject`, a `final class` with the same
fields so we cover reference elements too. Data is generated from a seeded
SplitMix64 generator so runs are reproducible. String lengths are picked so
about half land above the small-string boundary and need a heap allocation, and
the key fields have enough repeated values to produce ties, which is what makes
the secondary comparator run in the two-key cases.

Each benchmark does one full `sorted(using:)` of 100,000 elements per iteration.
For both value and reference elements the suite covers:

- `KeyPathComparator`: single `Int`, two `Int` keys, optional `Int?`, single
  `String`, two `String` keys, and a computed property.
- `ComparableComparator`: used directly on `[Int]` and `[String]`.
- `OptionalComparator`: reached through `KeyPathComparator` on an optional key.
- `ClosureComparator`: a small custom `SortComparator` backed by a comparison
  closure. It is built behind an `@inline(never)` factory so the closure stays
  opaque to the sort, the way a caller-provided comparator normally would.
- Floor: plain `sorted(by:)` with an inlined closure and no `SortComparator`.
  This is the lower bound for a comparison-based sort of this data, so the
  difference between a comparator and its floor is the comparator's overhead.

## Baseline numbers

Measured on M4 Max, swift-foundation package build, 100k elements, p50. Times
are ms; malloc is allocations per sort.

### Value element (`Entry`) — time (ms)

| key | KeyPath | Closure | Floor |
|---|--:|--:|--:|
| `single-int` | 435 | 163 | 95 |
| `two-int` | 714 | 394 | 95 |
| `single-optional` | 577 | 160 | 92 |
| `single-string` | 501 | 169 | 96 |
| `two-string` | 1007 | 476 | 107 |
| `single-computed` | 652 | 154 | 90 |

### Value element (`Entry`) — malloc

| key | KeyPath | Closure | Floor |
|---|--:|--:|--:|
| `single-int` | 5.0M | 2.6M | 7 |
| `two-int` | 9.9M | 7.4M | 7 |
| `single-optional` | 4.9M | 2.6M | 7 |
| `single-string` | 4.9M | 2.5M | 7 |
| `two-string` | 12.6M | 9.3M | 7 |
| `single-computed` | 4.8M | 2.5M | 7 |

### Reference element (`EntryObject`) — time (ms)

| key | KeyPath | Closure | Floor |
|---|--:|--:|--:|
| `single-int` | 415 | 93 | 19 |
| `two-int` | 678 | 305 | 20 |
| `single-optional` | 549 | 95 | 20 |
| `single-string` | 483 | 104 | 27 |
| `two-string` | 1017 | 400 | 46 |
| `single-computed` | 567 | 87 | 17 |

### Reference element (`EntryObject`) — malloc

| key | KeyPath | Closure | Floor |
|---|--:|--:|--:|
| `single-int` | 5.0M | 2.6M | 7 |
| `two-int` | 9.9M | 7.4M | 7 |
| `single-optional` | 4.9M | 2.6M | 7 |
| `single-string` | 4.9M | 2.5M | 7 |
| `two-string` | 12.6M | 9.3M | 7 |
| `single-computed` | 4.8M | 2.5M | 7 |

`ComparableComparator` used directly:

| collection | time | malloc |
|---|--:|--:|
| `[Int]` | 87 | 2.6M |
| `[String]` | 111 | 2.5M |

### DIFF
--- a/Benchmarks/Benchmarks/Internationalization/BenchmarkSortComparator.swift
+++ b/Benchmarks/Benchmarks/Internationalization/BenchmarkSortComparator.swift
@@ -1,0 +1,354 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022-2023 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import func Benchmark.blackHole
+
+#if os(macOS) && USE_PACKAGE
+import FoundationEssentials
+import FoundationInternationalization
+#else
+import Foundation
+#endif
+
+#if FOUNDATION_FRAMEWORK
+// FOUNDATION_FRAMEWORK has a scheme per benchmark file, so only include one benchmark here.
+let benchmarks: @Sendable () -> Void = {
+    sortComparatorBenchmarks()
+}
+#endif
+
+/// Number of elements sorted per benchmark iteration.
+private let elementCount = 100_000
+
+// MARK: - Fixtures
+
+/// A representative record with a mix of field types, standing in for a typical
+/// row a caller might sort. Field names are uniform; only the types matter for
+/// exercising the different comparator code paths.
+fileprivate struct Entry {
+    let int0: Int
+    let int1: Int
+    let int2: Int
+    let optionalInt: Int?
+    let str0: String
+    let str1: String
+    let str2: String
+    // Backing store for the computed property; `fileprivate` so the synthesized
+    // memberwise initializer is reachable from the generator below.
+    fileprivate let _computedProperty: Double?
+    /// Computed property: exercises the key-path getter (no stored-field offset).
+    var computedProperty: Double? { _computedProperty }
+    let id: UUID
+}
+
+/// The same fields as `Entry`, but a `final class`, so the element is a
+/// reference: sort reordering copies an 8-byte reference (one retain) instead of
+/// the whole struct, and `MemoryLayout.offset(of:)` is `nil` for its fields.
+fileprivate final class EntryObject: Sendable {
+    let int0: Int
+    let int1: Int
+    let int2: Int
+    let optionalInt: Int?
+    let str0: String
+    let str1: String
+    let str2: String
+    private let _computedProperty: Double?
+    var computedProperty: Double? { _computedProperty }
+    let id: UUID
+
+    init(_ e: Entry) {
+        int0 = e.int0
+        int1 = e.int1
+        int2 = e.int2
+        optionalInt = e.optionalInt
+        str0 = e.str0
+        str1 = e.str1
+        str2 = e.str2
+        _computedProperty = e.computedProperty
+        id = e.id
+    }
+}
+
+/// A custom `SortComparator` backed by a user-supplied comparison closure, used
+/// to benchmark sorting through an arbitrary caller-defined comparator.
+fileprivate struct ClosureComparator<Compared>: SortComparator {
+    let compareClosure: @Sendable (Compared, Compared) -> ComparisonResult
+    var order: SortOrder
+
+    init(order: SortOrder = .forward, _ compareClosure: @escaping @Sendable (Compared, Compared) -> ComparisonResult) {
+        self.compareClosure = compareClosure
+        self.order = order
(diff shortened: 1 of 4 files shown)


Benchmarks/Benchmarks/Internationalization/BenchmarkSortComparator.swift
```diff
@@ -0,0 +1,354 @@
...
+    var s = ""
+    s.reserveCapacity(length)
+    for _ in 0..<length {
+        s.append(alphabet[Int(rng.next() % UInt64(alphabet.count))])
+    }
+    return s
+}
+
+/// A UUID built from the generator so the whole fixture is reproducible.
+private func randomUUID(using rng: inout SeededGenerator) -> UUID {
```
jmschonfeld: Can this use the `UUID.random(using:)` API?

Benchmarks/Benchmarks/Internationalization/InternationalizationBenchmark.swift
```diff
@@ -6,5 +6,6 @@ let benchmarks: @Sendable () -> Void = {
     calendarBenchmarks()
     localeBenchmarks()
     timeZoneBenchmarks()
+    sortComparatorBenchmarks()
```
jmschonfeld: I think it might be best to make this a separate benchmark target instead of combining it with the Calendar/Locale/TimeZone target. There are already other benchmarks that import the i18n module like `BenchmarkFormatting.swift` and `BenchmarkCharacterSet.swift` - I think this target just encompasses the core i18n types which are very intertwined.

@parkera @chloe-yeo you've probably worked in this place most recently, does that make sense to you?

Additionally, only `KeyPathComparator` actually requires anything from FoundationInternationalization - the rest of the comparators live in essentials so that's more of a reason to avoid tying it to internationalization-related things.

jmschonfeld (commented):
I think you'll need to rebase this PR to pickup the latest CI configuration to start the required checks - sorry about that!

Benchmarks/Benchmarks/Internationalization/InternationalizationBenchmark.swift
dnadoba: Let me know where it should go and I can move them there.

Benchmarks/Benchmarks/Internationalization/BenchmarkSortComparator.swift
dnadoba: Done, note though that this is only available in 26.4 and the package currently only requires 26.0.
This isn't an issue if you build Foundation while also running the benchmark which is what I was doing but might impact you if you want to run it without building Foundation.

There are variouse ways to workaround that but I leave it up to you to decide if this something we should fix.

Benchmarks/Benchmarks/Internationalization/InternationalizationBenchmark.swift
parkera: We can just make a new benchmark file.

parkera: Or two - one for the Essentials benchmarks and one for i18n.

dnadoba: Do you mean separate executable targets instead of files? The benchmark itself is already in it's own file. 
Splitting those into multiple targets will require some amount of duplication though unless you want me to create a third target that can hold the shared types/functions. 

Happy do it in whichever way you prefer, just let me know.

dnadoba: @jmschonfeld/@parkera what do you prefer?

parkera: A new target, thank you!

dnadoba: Done.

jmschonfeld (approved):

PR #2098, opened 2026-07-10
Fix mode_t conversion from #2077 for 32-bit Android armv7

finagolfin:
Using this to avoid CI failures on another pull


finagolfin:
Not needed any longer

PR #2105, opened 2026-07-12
Add Buddhist and Japanese calendar implementations

dra8an:
## Add Buddhist and Japanese calendar implementations

Composition over `_CalendarGregorian`: both calendars delegate all arithmetic and forward to the existing Gregorian implementation, overriding only era/year extraction and construction.

### Buddhist

Year = Gregorian + 543, single BE era. ~150 LOC. Composition adds negligible overhead (benchmarks match Gregorian within noise).

### Japanese

237 era table (Taika 645 through Reiwa 2019) from CLDR `supplementalData.txt`. Era lookup via linear scan of the descending sorted table. ~450 LOC (237 of which are the era data).

Era inference: when year is provided without an explicit era (as happens in the generic enumeration slow path), defaults to the latest era (Reiwa) to match ICU behavior. Pre Taika dates clamp to era 0.

### Feature flags

Both calendars are gated behind `foundation_swift_buddhist_calendar_feature_enabled()` and `foundation_swift_japanese_calendar_feature_enabled()`, off by default. `Calendar(identifier: .buddhist/.japanese)` continues routing to `_CalendarICU` until the flags are enabled.

### Testing

- `BuddhistRecurrenceRuleParityProbe.swift`: 4 tests (yearly, monthly, weekly, Thanksgiving shape) comparing ICU backed vs native implementation. Zero divergences.
- `JapaneseRecurrenceRuleParityProbe.swift`: 6 tests including 3 era transition tests (monthly across Heisei/Reiwa boundary, yearly May 1st across eras, daily across boundary). Zero divergences.
- Full suite: 1113 tests pass.

### Benchmarks

15 mirrored benchmarks (Gregorian/Buddhist/Japanese x 5 shapes). Key numbers (p50, release, arm64 M3 Max):

| Calendar | dateComponents | roundTrip | enumerate 1000 |
|---|---|---|---|
| Gregorian | 2 ns | 5 ns | 19 us |
| Buddhist | 2 ns | 5 ns | 21 us |
| Japanese | 2 ns | 5 ns | 12 us |

Zero heap allocations in all cases.

### Known limitation

Meiji era start encoded as 1868-09-08 to match Apple's runtime ICU. CLDR canonical is 1868-10-23. Documented in source.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
@@ -1,0 +1,153 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Buddhist calendar. Year = Gregorian + 543, single era (BE). Delegates to `_CalendarGregorian`.
+internal final class _CalendarBuddhist: _CalendarProtocol, @unchecked Sendable {
+
+    static let yearOffset = 543
+
+    private let gregorian: _CalendarGregorian
+
+    init(identifier: Calendar.Identifier, timeZone: TimeZone?, locale: Locale?, firstWeekday: Int?, minimumDaysInFirstWeek: Int?, gregorianStartDate: Date?) {
+        assert(identifier == .buddhist, "_CalendarBuddhist only handles .buddhist")
+        self.gregorian = _CalendarGregorian(identifier: .gregorian, timeZone: timeZone, locale: locale, firstWeekday: firstWeekday, minimumDaysInFirstWeek: minimumDaysInFirstWeek, gregorianStartDate: gregorianStartDate)
+    }
+
+    let identifier: Calendar.Identifier = .buddhist
+
+    var locale: Locale? {
+        get { gregorian.locale }
+        set { gregorian.locale = newValue }
+    }
+
+    var timeZone: TimeZone {
+        get { gregorian.timeZone }
+        set { gregorian.timeZone = newValue }
+    }
+
+    var firstWeekday: Int {
+        get { gregorian.firstWeekday }
+        set { gregorian.firstWeekday = newValue }
+    }
+
+    var minimumDaysInFirstWeek: Int {
+        get { gregorian.minimumDaysInFirstWeek }
+        set { gregorian.minimumDaysInFirstWeek = newValue }
+    }
+
+    func copy(changingLocale: Locale?, changingTimeZone: TimeZone?, changingFirstWeekday: Int?, changingMinimumDaysInFirstWeek: Int?) -> any _CalendarProtocol {
+        let args = _CalendarUtility.resolvedCopyArgs(
+            currentTimeZone: gregorian.timeZone, changingTimeZone: changingTimeZone,
+            currentLocale: gregorian.locale, changingLocale: changingLocale,
+            currentFirstWeekday: gregorian._firstWeekday, changingFirstWeekday: changingFirstWeekday,
+            currentMinimumDaysInFirstWeek: gregorian._minimumDaysInFirstWeek, changingMinimumDaysInFirstWeek: changingMinimumDaysInFirstWeek
+        )
+        return _CalendarBuddhist(identifier: identifier, timeZone: args.timeZone, locale: args.locale, firstWeekday: args.firstWeekday, minimumDaysInFirstWeek: args.minimumDaysInFirstWeek, gregorianStartDate: nil)
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(identifier)
+        hasher.combine(timeZone)
+        hasher.combine(firstWeekday)
+        hasher.combine(minimumDaysInFirstWeek)
+        hasher.combine(localeIdentifier)
+        hasher.combine(preferredFirstWeekday)
+        hasher.combine(preferredMinimumDaysInFirstweek)
+    }
+
+    func supportsNextDateFastPath(for components: Calendar.ComponentSet) -> Bool { gregorian.supportsNextDateFastPath(for: components) }
+
+    // MARK: - Range
+
+    func minimumRange(of component: Calendar.Component) -> Range<Int>? {
+        if component == .era { return 0..<1 }
+        return gregorian.minimumRange(of: component)
+    }
+
+    func maximumRange(of component: Calendar.Component) -> Range<Int>? {
+        if component == .era { return 0..<1 }
+        return gregorian.maximumRange(of: component)
+    }
+
+    func range(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Range<Int>? {
+        gregorian.range(of: smaller, in: larger, for: date)
+    }
+
+    func ordinality(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Int? {
+        gregorian.ordinality(of: smaller, in: larger, for: date)
+    }
+
+    func dateInterval(of component: Calendar.Component, for date: Date) -> DateInterval? {
+        if component == .era {
+            return DateInterval(start: Date(timeIntervalSinceReferenceDate: -63113904000.0 - Calendar._maxDateIntervalDuration), duration: Calendar._maxDateIntervalDuration)
+        }
+        return gregorian.dateInterval(of: component, for: date)
+    }
+
+    func isDateInWeekend(_ date: Date) -> Bool {
+        gregorian.isDateInWeekend(date)
+    }
+
+    // MARK: - Date / DateComponents conversion
+
+    func date(from components: DateComponents) -> Date? {
+        gregorian.date(from: convertedToGregorian(components))
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from date: Date, in timeZone: TimeZone) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: date, in: timeZone)
+        adjustToBuddhist(&dc, requested: components)
+        return dc
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from date: Date) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: date)
+        adjustToBuddhist(&dc, requested: components)
+        return dc
+    }
+
+    func date(byAdding components: DateComponents, to date: Date, wrappingComponents: Bool) -> Date? {
+        gregorian.date(byAdding: components, to: date, wrappingComponents: wrappingComponents)
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from start: Date, to end: Date) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: start, to: end)
+        if components.contains(.era) { dc.era = 0 }
+        return dc
+    }
+
+    // MARK: - Fast path
+
+    func nextDate(after date: Date, matching components: DateComponents, direction: Calendar.SearchDirection) -> Date? {
+        gregorian.nextDate(after: date, matching: convertedToGregorian(components), direction: direction)
+    }
+
+    // MARK: - Helpers
+
+    private func convertedToGregorian(_ components: DateComponents) -> DateComponents {
+        var dc = components
+        if let y = dc.year { dc.year = y - Self.yearOffset }
+        dc.era = nil
+        return dc
+    }
+
+    private func adjustToBuddhist(_ dc: inout DateComponents, requested: Calendar.ComponentSet) {
+        if requested.contains(.year), let y = dc.year { dc.year = y + Self.yearOffset }
+        if requested.contains(.era) { dc.era = 0 }
+    }
+
+#if FOUNDATION_FRAMEWORK
+    func bridgeToNSCalendar() -> NSCalendar {
+        _NSSwiftCalendar(calendar: Calendar(inner: self))
+    }
+#endif
+}
--- a/Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
@@ -69,12 +69,12 @@ internal final class _CalendarBuddhist: _CalendarProtocol, @unchecked Sendable {
     // MARK: - Range
 
     func minimumRange(of component: Calendar.Component) -> Range<Int>? {
-        if component == .era { return 0..<1 }
+        if component == .era { return Range(0...0) }
         return gregorian.minimumRange(of: component)
     }
 
     func maximumRange(of component: Calendar.Component) -> Range<Int>? {
-        if component == .era { return 0..<1 }
+        if component == .era { return Range(0...0) }
         return gregorian.maximumRange(of: component)
     }
 
--- a/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
@@ -29,7 +29,7 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
 
     /// 237 Japanese eras (Taika 645 → Reiwa 2019), sorted descending. Index values match ICU's era numbering.
     // Meiji (232) uses 1868-09-08 to match Apple's runtime ICU (CLDR canonical is 1868-10-23).
-    private static let eraData: InlineArray<237, (Int32, Int16, Int8, Int8)> = [
+    private static let eraData: InlineArray<237, (index: Int32, year: Int16, month: Int8, day: Int8)> = [
         (236, 2019, 5, 1),
         (235, 1989, 1, 8),
         (234, 1926, 12, 25),
@@ -273,7 +273,7 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
 
     private static func era(at i: Int) -> EraEntry {
         let raw = eraData[i]
-        return EraEntry(raw.0, raw.1, raw.2, raw.3)
+        return EraEntry(raw.index, raw.year, raw.month, raw.day)
     }
 
     private let gregorian: _CalendarGregorian
@@ -414,11 +414,7 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
         guard let era = eraEntry(forGregorianYear: y, month: m, day: d) else {
             return gregorian.dateInterval(of: .era, for: date)
         }
-        var startDC = DateComponents()
-        startDC.year = era.startGregorianYear
-        startDC.month = era.startMonth
-        startDC.day = era.startDay
-        startDC.hour = 0; startDC.minute = 0; startDC.second = 0
+        let startDC = DateComponents(year: era.startGregorianYear, month: era.startMonth, day: era.startDay, hour: 0, minute: 0, second: 0)
         guard let start = gregorian.date(from: startDC) else { return nil }
         let endDate: Date
         var nextEra: EraEntry? = nil
@@ -429,11 +425,7 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
             }
         }
         if let next = nextEra {
-            var endDC = DateComponents()
-            endDC.year = next.startGregorianYear
-            endDC.month = next.startMonth
-            endDC.day = next.startDay
-            endDC.hour = 0; endDC.minute = 0; endDC.second = 0
+            let endDC = DateComponents(year: next.startGregorianYear, month: next.startMonth, day: next.startDay, hour: 0, minute: 0, second: 0)
             guard let e = gregorian.date(from: endDC) else { return nil }
             endDate = e
         } else {
--- a/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
@@ -330,13 +330,13 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
     // MARK: - Range
 
     func minimumRange(of component: Calendar.Component) -> Range<Int>? {
-        if component == .era { return 0..<(Self.era(at: Self.eraCount - 1).index + Self.eraCount) }
-        if component == .year { return 1..<2 }
+        if component == .era { return Range(0...Self.era(at: 0).index) }
+        if component == .year { return Range(1...1) }
         return gregorian.minimumRange(of: component)
     }
 
     func maximumRange(of component: Calendar.Component) -> Range<Int>? {
-        if component == .era { return 0..<(Self.era(at: 0).index + 1) }
+        if component == .era { return Range(0...Self.era(at: 0).index) }
         return gregorian.maximumRange(of: component)
     }
 
--- a/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
@@ -402,10 +402,10 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
     }
 
     private func eraEntry(byIndex index: Int) -> EraEntry? {
-        for i in 0..<Self.eraCount {
-            if Self.era(at: i).index == index { return Self.era(at: i) }
-        }
-        return nil
+        let i = Self.eraCount - 1 - index
+        guard i >= 0 && i < Self.eraCount else { return nil }
+        let era = Self.era(at: i)
+        return era.index == index ? era : nil
     }
 
     private func eraInterval(containing date: Date) -> DateInterval? {
@@ -417,14 +417,9 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
         let startDC = DateComponents(year: era.startGregorianYear, month: era.startMonth, day: era.startDay, hour: 0, minute: 0, second: 0)
         guard let start = gregorian.date(from: startDC) else { return nil }
         let endDate: Date
-        var nextEra: EraEntry? = nil
-        for i in 0..<Self.eraCount {
-            if Self.era(at: i).index == era.index && i > 0 {
-                nextEra = Self.era(at: i - 1)
-                break
-            }
-        }
-        if let next = nextEra {
+        let position = Self.eraCount - 1 - era.index
+        if position > 0 {
+            let next = Self.era(at: position - 1)
             let endDC = DateComponents(year: next.startGregorianYear, month: next.startMonth, day: next.startDay, hour: 0, minute: 0, second: 0)
             guard let e = gregorian.date(from: endDC) else { return nil }
             endDate = e
--- a/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
@@ -1,0 +1,481 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// Japanese imperial calendar. Same arithmetic as Gregorian; year is reckoned within the current era. Delegates to `_CalendarGregorian`.
+internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
+
+    private struct EraEntry {
+        let index: Int
+        let startGregorianYear: Int
+        let startMonth: Int
+        let startDay: Int
+
+        init(_ index: Int32, _ year: Int16, _ month: Int8, _ day: Int8) {
+            self.index = Int(index)
+            self.startGregorianYear = Int(year)
+            self.startMonth = Int(month)
+            self.startDay = Int(day)
+        }
+    }
+
+    /// 237 Japanese eras (Taika 645 → Reiwa 2019), sorted descending. Index values match ICU's era numbering.
+    // Meiji (232) uses 1868-09-08 to match Apple's runtime ICU (CLDR canonical is 1868-10-23).
+    private static let eraData: InlineArray<237, (Int32, Int16, Int8, Int8)> = [
+        (236, 2019, 5, 1),
+        (235, 1989, 1, 8),
+        (234, 1926, 12, 25),
+        (233, 1912, 7, 30),
+        (232, 1868, 9, 8),
+        (231, 1865, 4, 7),
+        (230, 1864, 2, 20),
+        (229, 1861, 2, 19),
+        (228, 1860, 3, 18),
+        (227, 1854, 11, 27),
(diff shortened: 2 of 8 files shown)


richgillam:
You haven't added any reviewers to this PR, and no one's kicked off the CI workflows.  Is this PR ready to review?

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
```diff
@@ -0,0 +1,153 @@
...
+        return gregorian.minimumRange(of: component)
+    }
+
+    func maximumRange(of component: Calendar.Component) -> Range<Int>? {
+        if component == .era { return 0..<1 }
+        return gregorian.maximumRange(of: component)
+    }
+
+    func range(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Range<Int>? {
+        gregorian.range(of: smaller, in: larger, for: date)
```
richgillam, line 82: Don't you have to special-case the era here somehow?

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    func ordinality(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Int? {
+        gregorian.ordinality(of: smaller, in: larger, for: date)
```
richgillam, line 86: Don't you have to special-case the era here somehow?

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from date: Date) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: date)
+        adjustToBuddhist(&dc, requested: components)
+        return dc
+    }
+
+    func date(byAdding components: DateComponents, to date: Date, wrappingComponents: Bool) -> Date? {
+        gregorian.date(byAdding: components, to: date, wrappingComponents: wrappingComponents)
```
richgillam, line 119: Wouldn't you have to adjust the era and year fields here?

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from start: Date, to end: Date) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: start, to: end)
+        if components.contains(.era) { dc.era = 0 }
```
richgillam, line 124: What does this function do?  Do you have to adjust the year and era fields?

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
```diff
@@ -0,0 +1,153 @@
...
+    func nextDate(after date: Date, matching components: DateComponents, direction: Calendar.SearchDirection) -> Date? {
+        gregorian.nextDate(after: date, matching: convertedToGregorian(components), direction: direction)
+    }
+
+    // MARK: - Helpers
+
+    private func convertedToGregorian(_ components: DateComponents) -> DateComponents {
+        var dc = components
+        if let y = dc.year { dc.year = y - Self.yearOffset }
+        dc.era = nil
```
richgillam, line 139: You don't have to worry about the era?  Are we getting AD and BC from the sign of the year here?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        init(_ index: Int32, _ year: Int16, _ month: Int8, _ day: Int8) {
+            self.index = Int(index)
+            self.startGregorianYear = Int(year)
+            self.startMonth = Int(month)
+            self.startDay = Int(day)
+        }
+    }
+
+    /// 237 Japanese eras (Taika 645 → Reiwa 2019), sorted descending. Index values match ICU's era numbering.
+    // Meiji (232) uses 1868-09-08 to match Apple's runtime ICU (CLDR canonical is 1868-10-23).
```
richgillam, line 32: Several questions:
- ICU and CLDR just got rid of (IIRC) all the pre-Meiji eras.  We're keeping them?
- You said your source for this was `supplementalData.txt` in CLDR.  What's the plan for keeping this in sync with CLDR in the future?  Aside from the fact (?) that they got rid of all the pre-Meiji eras, they'd be adding a new era every 75 or so years.  Wouldn't we want some automated way of keeping in sync as CLDR evolves?
- This is an `InlineArray` of tuples?  I was under the (perhaps mistaken) impression that tuples were less efficient than `struct`s.  That's wrong?  Also, I think you can name tuple members, and I think that'd be helpful here for readability.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        (2, 672, 1, 1),
+        (1, 650, 2, 15),
+        (0, 645, 6, 19),
+    ]
+
+    private static let eraCount = 237
+
+    private static func era(at i: Int) -> EraEntry {
+        let raw = eraData[i]
+        return EraEntry(raw.0, raw.1, raw.2, raw.3)
```
richgillam: Again, I think you can name tuple members, and that seems like it'd aid readability.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+/// Japanese imperial calendar. Same arithmetic as Gregorian; year is reckoned within the current era. Delegates to `_CalendarGregorian`.
+internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
+
+    private struct EraEntry {
+        let index: Int
+        let startGregorianYear: Int
+        let startMonth: Int
```
richgillam, line 19: Dumb question, but are the month and day 1-based or 0-based?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        hasher.combine(localeIdentifier)
+        hasher.combine(preferredFirstWeekday)
+        hasher.combine(preferredMinimumDaysInFirstweek)
+    }
+
+    // MARK: - Range
+
+    func minimumRange(of component: Calendar.Component) -> Range<Int>? {
+        if component == .era { return 0..<(Self.era(at: Self.eraCount - 1).index + Self.eraCount) }
+        if component == .year { return 1..<2 }
```
richgillam: I wonder if this'd read better as `1...1`?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        return gregorian.minimumRange(of: component)
+    }
+
+    func maximumRange(of component: Calendar.Component) -> Range<Int>? {
+        if component == .era { return 0..<(Self.era(at: 0).index + 1) }
+        return gregorian.maximumRange(of: component)
+    }
+
+    func range(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Range<Int>? {
+        gregorian.range(of: smaller, in: larger, for: date)
```
richgillam, line 113: Don't you need to account for the special era and year fields here?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    func ordinality(of smaller: Calendar.Component, in larger: Calendar.Component, for date: Date) -> Int? {
+        gregorian.ordinality(of: smaller, in: larger, for: date)
```
richgillam, line 117: Wouldn't you need to account for the Japanese-specific era and year numbering here?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    func dateInterval(of component: Calendar.Component, for date: Date) -> DateInterval? {
```
richgillam, line 120: This function returns the beginning and ending dates of the `component` containing `date`?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from date: Date) -> DateComponents {
+        var dc = gregorian.dateComponents(components, from: date)
+        adjustToJapanese(&dc, date: date, requested: components)
+        return dc
+    }
+
+    func date(byAdding components: DateComponents, to date: Date, wrappingComponents: Bool) -> Date? {
+        gregorian.date(byAdding: components, to: date, wrappingComponents: wrappingComponents)
```
richgillam, line 150: You don't have to do anything special here to account for the Japanese eras?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    func dateComponents(_ components: Calendar.ComponentSet, from start: Date, to end: Date) -> DateComponents {
+        gregorian.dateComponents(components, from: start, to: end)
```
richgillam, line 154: What does this function do?  Do you have to adjust the Gregorian results to account for the Japanese eras?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+    func nextDate(after date: Date, matching components: DateComponents, direction: Calendar.SearchDirection) -> Date? {
+        gregorian.nextDate(after: date, matching: convertedToGregorian(components), direction: direction)
+    }
+
+    // MARK: - Era helpers
+
+    private func eraEntry(forGregorianYear y: Int, month m: Int, day d: Int) -> EraEntry? {
+        for i in 0..<Self.eraCount {
+            let era = Self.era(at: i)
+            if (y, m, d) >= (era.startGregorianYear, era.startMonth, era.startDay) {
```
richgillam, line 166: `if (a, b, c) >= (x, y, z)` is the same as `if a >= x && b >= y && c >= z`?  I hadn't seen that before...

Actually, come to think of it, I'd think what you'd want here is `if a >= x || (a == x && b >= y) || (a == x && b == y && c >= z)`.  Maybe _that's_ what `if (a, b, c) >= (x, y, z)` does?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+                return era
+            }
+        }
+        return nil
+    }
+
+    private func eraEntry(byIndex index: Int) -> EraEntry? {
+        for i in 0..<Self.eraCount {
+            if Self.era(at: i).index == index { return Self.era(at: i) }
+        }
```
richgillam: Can you derive the index mathematically instead of storing it and get out of having to do this loop?

(That'd only work if we _don't_ get rid of all the pre-Meiji eras, though...)

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        startDC.hour = 0; startDC.minute = 0; startDC.second = 0
+        guard let start = gregorian.date(from: startDC) else { return nil }
+        let endDate: Date
+        var nextEra: EraEntry? = nil
+        for i in 0..<Self.eraCount {
+            if Self.era(at: i).index == era.index && i > 0 {
+                nextEra = Self.era(at: i - 1)
+                break
+            }
+        }
```
richgillam: Do you have to search for the next era?  Wouldn't it always be the next one in the array?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        let comps = gregorian.dateComponents([.year, .month, .day], from: date)
+        guard let y = comps.year, let m = comps.month, let d = comps.day else { return nil }
+        guard let era = eraEntry(forGregorianYear: y, month: m, day: d) else {
+            return gregorian.dateInterval(of: .era, for: date)
+        }
+        var startDC = DateComponents()
+        startDC.year = era.startGregorianYear
+        startDC.month = era.startMonth
+        startDC.day = era.startDay
+        startDC.hour = 0; startDC.minute = 0; startDC.second = 0
```
richgillam: I thought there was a method that let you set a bunch of elements of a `DateComponents` in a single operation.  No?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        guard let start = gregorian.date(from: startDC) else { return nil }
```
richgillam, line 196: When would this return nil?  And what will the calling function do if it does?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+                break
+            }
+        }
+        if let next = nextEra {
+            var endDC = DateComponents()
+            endDC.year = next.startGregorianYear
+            endDC.month = next.startMonth
+            endDC.day = next.startDay
+            endDC.hour = 0; endDC.minute = 0; endDC.second = 0
+            guard let e = gregorian.date(from: endDC) else { return nil }
```
richgillam: Same questions as with the start date above: Can't you set all the members of the `DateComponents` in one line?  And when would this actually return nil, and what will the calling function do if it does?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+            endDate = e
+        } else {
+            endDate = start.addingTimeInterval(Calendar._maxDateIntervalDuration)
```
richgillam, line 205: Isn't there some kind of "end of time" constant you could use here for the end date of the interval?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+            }
+        }
+        dc.era = nil
+        return dc
+    }
+
+    private func adjustToJapanese(_ dc: inout DateComponents, date: Date, requested: Calendar.ComponentSet) {
+        guard requested.contains(.era) || requested.contains(.year) else { return }
+        let probe = gregorian.dateComponents([.era, .year, .month, .day], from: date)
+        guard let y = probe.year, let m = probe.month, let d = probe.day else { return }
```
richgillam, line 230: Under what conditions would this happen, and is just going with the unadjusted components the right thing for the calling function to do?

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -0,0 +1,481 @@
...
+        let extendedYear = probe.era == 0 ? 1 - y : y
+        if let era = eraEntry(forGregorianYear: extendedYear, month: m, day: d) {
+            if requested.contains(.era) { dc.era = era.index }
+            if requested.contains(.year) { dc.year = extendedYear - era.startGregorianYear + 1 }
+        } else {
+            // Pre-Taika: ICU clamps to the first era.
+            let first = Self.era(at: Self.eraCount - 1)
+            if requested.contains(.era) { dc.era = first.index }
+            if requested.contains(.year) { dc.year = extendedYear - first.startGregorianYear + 1 }
```
richgillam: Again, didn't ICU and CLDR just get rid of all the pre-Meiji eras?

Tests/FoundationInternationalizationTests/BuddhistRecurrenceRuleParityProbe.swift
```diff
@@ -0,0 +1,123 @@
...
+import Testing
+
+#if FOUNDATION_FRAMEWORK
+@testable import Foundation
+#else
+@testable import FoundationInternationalization
+@testable import FoundationEssentials
+#endif
+
+/// RecurrenceRule parity probe: compares date sequences from `recurrences(of:)` between the ICU backed Buddhist calendar and `_CalendarBuddhist`.
```
richgillam, line 22: The problem I have with this class and with `JapanesrRecurrenceParityProbe` below is that they both expect that the ICU-backed calendar will always be there.  If we actually get rid of it someday, these tests won't work because there'll no longer be an existing working calendar implementation we can compare against.  What's the plan for when ICU4C is no longer available?  Is it worth starting to add tests now that don't rely on the old implementation to check for correctness?

(Come to think of it, I was talking about the CLDR conformance tests earlier today, and it might be worth it to start trying to wire some of the Foundation stuff-- both with and without the new Swift stuff-- into the CLDR conformance tests and use that to augment our own unit testing.  The problem, of course, is that we purposely deviate from what CLDR wants in a lot of spots, but for most things, I'd think the conformance tests would be helpful.)

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Buddhist has a single era (BE). _CalendarBuddhist composes over _CalendarGregorian and overrides only year (= Gregorian + 543) and era (always BE); there's no AD/BC split, so no era special casing, Gregorian does the arithmetic and we remap.

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Same answer as above. Also applies to comments below at lines 119, 124, and 139.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Gregorian does the date math; `adjustToJapanese` and `convertedToGregorian` remap between Japanese era+year and the Gregorian year, so era and year are accounted for. Same for comments at 348, 381, and 385

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Yes, start and end of the component interval containing date (for .era, the current Japanese era's span).

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Swift tuple comparison is lexicographic. `(a,b,c) >= (x,y,z)` is exactly your `a>x || (a==x && b>y) || (a==x && b==y && c>=z)`. It compares (year,month,day) against the era start

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Answered above.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Good point about the pre-Meiji eras. @itingliu , is it safe to remove those?

Regarding tuples vs struct: In a fixed-layout `InlineArray`, a tuple of scalars is stored inline with no heap/boxing which is  equivalent to a struct; named the members for readability now, EraEntry stays as the unpacked accessor

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: I agree. Named.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: 1-based (month 1–12, day 1–31)

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Return type is Range<Int>, so `1...1` would be a ClosedRange mismatch; `1..<2` is the half-open equivalent (matches the protocol/Gregorian).

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: `gregorian.date(from:)` only returns nil for invalid/out-of-range components; here they're always valid, so it's a defensive guard, callers propagate nil (no interval).

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Answered above

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: It already uses `Calendar._maxDateIntervalDuration`, the shared constant for bounding open-ended intervals, same as the other calendars.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Yeah, let's wait for input from Foundation about this.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Done. The table is sorted descending with contiguous indices, so the slot is just `eraCount - 1 - index`. Kept a defensive `.index == index` check so it returns nil rather than wrong data if that ever changes. (As you noted, this direct mapping depends on us not renumbering  the indices, so this is tied to the pre-Meiji discussion)

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Right. Since the array is sorted descending, the next era is always the slot above. Replaced the search with `position - 1`, where `position = eraCount - 1 - era.index`.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Yep. Switched to the memberwise initializer `DateComponents(year:month:day:hour:minute:second:)`. (Did the same for the end date at 437.)

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: `probe` comes from `gregorian.dateComponents([.era, .year, .month, .day], from:)`, which always populates year/month/day for a valid Date, so this guard never trips in practice. If it somehow did, leaving the components unadjusted is a harmless no-op

Tests/FoundationInternationalizationTests/BuddhistRecurrenceRuleParityProbe.swift
dra8an: Agreed. These are parity probes, so they only really work while the ICU calendar is around. We don't have an ICU independent correctness test for Japanese or Buddhist yet. The era boundaries come from CLDR supplementalData, but that's the data source, not a test. My plan for when ICU4C goes away is to add golden value tests covering the era start dates plus a spread of known conversions, and keep the parity probes as a cross check for as long as ICU is still there.

On the CLDR conformance tests, I think that's a good idea for a follow up. The one wrinkle is the deviations you mentioned. We intentionally diverge from CLDR in a few spots, so we'd mark those as known expected divergences, and everything else would be a solid implementation independent signal.

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
richgillam: But I don't see any remapping happening in this function.  You're just calling through to the Gregorian calendar and returning the answer.  That works?

Same thing in the other places where I added a similar comment.  In all of those functions, you were just calling through to the Gregorian calendar and not adjusting either the input or the output.  In each case, is no adjustment necessary (maybe I just don't understand what the function does), or is that happening in some other function that's being called indirectly from the one I'm looking at?

Tests/FoundationInternationalizationTests/BuddhistRecurrenceRuleParityProbe.swift
richgillam: Okay.  It sounds like you have a plan here, and the plan sounds good to me.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
richgillam: Yeah, I keep forgetting they're different types.  I wish the language could allow you to freely intermix and do the appropriate conversions under the covers, but maybe there are good reasons why they don't do this.  Forget I said anything here.  :-)

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
richgillam: Same question I had in the Buddhist calendar: This function isn't calling `adjustToJapanese` or `convertedToGregorian`, so I don't understand your response.  Is something else in the call chain calling them, or does this function not need to do some reason?  Same questions in the other places where I asked this question.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
richgillam: Today I learned...

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
```diff
@@ -402,10 +402,10 @@ internal final class _CalendarJapanese: _CalendarProtocol, @unchecked Sendable {
...
 
     private func eraEntry(byIndex index: Int) -> EraEntry? {
-        for i in 0..<Self.eraCount {
-            if Self.era(at: i).index == index { return Self.era(at: i) }
-        }
-        return nil
+        let i = Self.eraCount - 1 - index
+        guard i >= 0 && i < Self.eraCount else { return nil }
+        let era = Self.era(at: i)
+        return era.index == index ? era : nil
```
richgillam, line 178: Two conflicting comments: First, if you do this, do you still need the `index` field?  And second, if we do get rid of the pre-Meiji eras, maybe you _will_ need the `index` field (or you'll need to update these adjustments), since (in ICU at least) they didn't change the indexes.

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
dra8an: Right, no adjustment is needed in these. Functions like range, ordinality, date(byAdding:), and dateComponents(from:to:) return values that don't depend on the era or year labeling. The number of days in a month, an ordinality, or the difference between two dates is the same whether we call it Buddhist, Japanese, or Gregorian, so we just delegate.

The era and year field values only get produced or consumed in two places, and that is where the remapping lives. `dateComponents(_:from:)` runs the Gregorian result through adjustToJapanese on the way out, and `date(from:)` and `nextDate` run the incoming components through  convertedToGregorian on the way in. The era transition tests in the parity probe cover this and match ICU.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Good point. Right now the indices are contiguous, so the stored index is derivable from the array slot (eraCount - 1 - position) and is technically redundant. But as you say, if we trim the pre-Meiji eras ICU keeps the original numbering, so Meiji stays 232, the slot to index mapping stops being linear, and the stored index becomes necessary again while this direct lookup would go back to a search. Since the pre-Meiji question is still open I would rather keep the index field for now than bake the assumption in any harder. This can be revisited once we know the Meiji decision.

Sources/FoundationEssentials/Calendar/Calendar_Buddhist.swift
richgillam: Okay, I'll accept that.  Thanks for the explanation.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
richgillam: That's fine.  I think they're still continuous after the Meiji change, once you take into account the big leap between 0 and whichever index Meiji is.  But yeah, I'm fine with not making any changes until we have an answer on the Meiji thing.

richgillam (approved):

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Range and ordinality return values that do not depend on the era or year. The number of days in a month, an ordinality, or a component difference is identical to Gregorian, so it just delegates. The remapping only happens where the era and year fields are actually produced or consumed, dateComponents(_:from:) runs the result through adjustToJapanese, and date(from:) and nextDate run the input through convertedToGregorian. The era transition tests in the parity probe cover it.

Sources/FoundationEssentials/Calendar/Calendar_Japanese.swift
dra8an: Based on our discussion on re-Meiji eras, trimmed to the 5 modern eras and inherit the Gregorian eras before Meiji, matching CLDR/ICU (unicode-org/icu#4019, ICU-23341).

Table is now Meiji through Reiwa only, keeping ICU numbering (232 to 236).

Pre-Meiji dates before 1868-09-08 report the inherited Gregorian era, CE (1) or BCE (0) with the Gregorian year, which falls out of the existing `_CalendarGregorian` delegation. Proleptic era-years roll in the same way ICU does (Meiji 1 Jan 1 becomes CE 1868, per JapaneseTest.Test5345 in ICU).

Add/subtract is verified. `date(byAdding` works in Gregorian year space, so crossing the Meiji edge is just Gregorian math. Regression tests cover subtracting day/month/year from Meiji 1 into CE and BCE.

richgillam (approved):
The new code that gets rid of all the historical Japanese eras and just uses Gregorian looks good, as far as I could tell.

dra8an:
Superseded by [#2215,](https://github.com/swiftlang/swift-foundation/pull/2215) which implements these calendars plus Republic of China on a shared era table in `_CalendarGregorian` rather than as separate classes.

PR #2109, opened 2026-07-14
Optimize `SortComparator` by up to 10x

dnadoba:
Uses the benchmarks from: https://github.com/swiftlang/swift-foundation/pull/2094

Every `KeyPathComparator.compare(_:_:)` went through a type-erased path. 
Both fields were boxed into `Any` and dynamically cast back to their concrete type on each comparison, dispatched through `AnySortComparator`. 
The `sorted(using:)` /`sort(using:)` wrappers were not `@inlinable`, so the standard library sort could not specialize on the element type nor the Sequence type, running into issue of Collection.subscript.\_read to allocate. 
## Changes
- `KeyPathComparator` holds a typed `_compare` closure built once at init that reads each field as its concrete type and compares it directly removing per-compare `Any `boxing and dynamic casts.
- `compare(_:_:)` applies `order` on top via `ComparisonResult.withOrder.` the type-erased comparator is retained only for `==` / `hash(into:)`.
- The `sorted(using:)` / `sort(using:)` / `compare` wrappers are now `@inlinable`, letting the standard library sort specialize the operations on the concrete `Sequence`.

## Benchmarks — CPU (p50, ms)
M4 Max, 100k elements per sort.

### `KeyPathComparator` — value element (`Entry`, 168-byte struct)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 414 | 117 | 3.5x |
| two-int | 693 | 129 | 5.4x |
| single-optional | 547 | 155 | 3.5x |
| single-string | 475 | 179 | 2.7x |
| two-string | 972 | 236 | 4.1x |
| single-computed | 624 | 244 | 2.6x |

### `KeyPathComparator` — reference element (`EntryObject`)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 388 | 97 | 4.0x |
| two-int | 664 | 101 | 6.6x |
| single-optional | 525 | 132 | 4.0x |
| single-string | 463 | 156 | 3.0x |
| two-string | 989 | 235 | 4.2x |
| single-computed | 539 | 150 | 3.6x |

### `ComparableComparator` (used directly)

| collection | before | after | speedup |
|---|--:|--:|--:|
| `[Int]` | 85 | 12 | 7.1x |
| `[String]` | 108 | 34 | 3.2x |

### Custom closure-based comparator — value element (`Entry`)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 158 | 100 | 1.6x |
| two-int | 381 | 107 | 3.6x |
| single-optional | 154 | 97 | 1.6x |
| single-string | 164 | 110 | 1.5x |
| two-string | 460 | 129 | 3.6x |
| single-computed | 151 | 95 | 1.6x |

### Custom closure-based comparator — reference element (`EntryObject`)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 90 | 23 | 3.9x |
| two-int | 300 | 28 | 10.7x |
| single-optional | 90 | 23 | 3.9x |
| single-string | 101 | 32 | 3.2x |
| two-string | 390 | 58 | 6.7x |
| single-computed | 85 | 19 | 4.5x |

Those are good improvement but we are still (sometimes far) away from a "hand written" version.
### Lower-bound floor (plain `sorted(by:)`) — value element (`Entry`)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 88 | 92 | 1.0x |
| two-int | 87 | 90 | 1.0x |
| single-optional | 87 | 92 | 0.9x |
| single-string | 92 | 94 | 1.0x |
| two-string | 100 | 103 | 1.0x |
| single-computed | 85 | 87 | 1.0x |

### Lower-bound floor (plain `sorted(by:)`) — reference element (`EntryObject`)

| key | before | after | speedup |
|---|--:|--:|--:|
| single-int | 19 | 19 | 1.0x |
| two-int | 19 | 19 | 1.0x |
| single-optional | 18 | 19 | 0.9x |
| single-string | 25 | 26 | 1.0x |
| two-string | 42 | 43 | 1.0x |
| single-computed | 16 | 16 | 1.0x |

## Benchmarks — allocations (total mallocs per sort)

`mallocCountTotal` is deterministic (identical p0–p100) and essentially as low as we can go.

### `KeyPathComparator` — value element (`Entry`)

| key | before | after |
|---|--:|--:|
| single-int | 4,969,548 | 7 |
| two-int | 9,867,359 | 7 |
| single-optional | 4,890,301 | 7 |
| single-string | 4,915,428 | 7 |
| two-string | 12,603,734 | 7 |
| single-computed | 4,826,292 | 7 |

### `KeyPathComparator` — reference element (`EntryObject`)

| key | before | after |
|---|--:|--:|
| single-int | 4,969,548 | 7 |
| two-int | 9,867,359 | 7 |
| single-optional | 4,890,301 | 7 |
| single-string | 4,915,428 | 7 |
| two-string | 12,603,734 | 7 |
| single-computed | 4,826,292 | 7 |

### `ComparableComparator` (used directly)

| collection | before | after |
|---|--:|--:|
| `[Int]` | 2,582,473 | 7 |
| `[String]` | 2,550,715 | 7 |

### Custom closure-based comparator — value element (`Entry`)

| key | before | after |
|---|--:|--:|
| single-int | 2,582,473 | 7 |
| two-int | 7,437,607 | 7 |
| single-optional | 2,563,911 | 7 |
| single-string | 2,550,715 | 7 |
| two-string | 9,264,661 | 7 |
| single-computed | 2,521,063 | 7 |

### Custom closure-based comparator — reference element (`EntryObject`)

| key | before | after |
|---|--:|--:|
| single-int | 2,582,473 | 7 |
| two-int | 7,437,607 | 7 |
| single-optional | 2,563,911 | 7 |
| single-string | 2,550,715 | 7 |
| two-string | 9,264,661 | 7 |
| single-computed | 2,521,063 | 7 |

### Lower-bound floor (plain `sorted(by:)`) — value element (`Entry`)

| key | before | after |
|---|--:|--:|
| single-int | 7 | 7 |
| two-int | 7 | 7 |
| single-optional | 7 | 7 |
| single-string | 7 | 7 |
| two-string | 7 | 7 |
| single-computed | 7 | 7 |

### Lower-bound floor (plain `sorted(by:)`) — reference element (`EntryObject`)

| key | before | after |
|---|--:|--:|
| single-int | 7 | 7 |
| two-int | 7 | 7 |
| single-optional | 7 | 7 |
| single-string | 7 | 7 |
| two-string | 7 | 7 |
| single-computed | 7 | 7 |

### DIFF
--- a/Sources/FoundationInternationalization/String/KeyPathComparator.swift
+++ b/Sources/FoundationInternationalization/String/KeyPathComparator.swift
@@ -23,33 +23,46 @@ public struct KeyPathComparator<Compared>: SortComparator {
 
     /// The sort order that the comparator uses to compare properties.
     public var order: SortOrder {
-        get {
-            comparator.order
-        }
-        set {
-            comparator.order = newValue
-        }
+        didSet { comparator.order = order }
     }
 
+    /// A type-erased copy of the underlying comparator, retained only for the cold paths: `==`, `hash(into:)`, and keeping `order` observable. It is never consulted by `compare(_:_:)`.
     var comparator: AnySortComparator
 
-    private let extractField: @Sendable (Compared) -> Any
+    /// The hot path: a fully-typed comparison of two `Compared` values that extracts the keyed field and compares it *in forward order*, with no `Any` boxing and no dynamic casts. `compare(_:_:)` applies the current `order` on top via `withOrder`.
+    private let _compare: @Sendable (Compared, Compared) -> ComparisonResult
 
-    /// Get the field at `cachedOffset` if there is one, otherwise
-    /// access the field directly through the keypath.
-    private static func getField<T>(ofType fieldType: T.Type, offset maybeOffset: Int?, from base: Compared, fallback keyPath: KeyPath<Compared, T>) -> T {
-        guard let offset = maybeOffset else {
-            return base[keyPath: keyPath]
-        }
+    /// Reads the field of type `T` at a known stored-property `offset` within `base`. `T` is passed as an argument rather than a metatype captured by the caller's `@Sendable` closure, which would be non-`Sendable`.
+    @inline(always)
+    private static func getField<T>(offset: Int, from base: Compared, as type: T.Type) -> T {
         return withUnsafePointer(to: base) { pointer in
-            let rawPointer = UnsafeRawPointer(pointer)
-            return rawPointer
+            UnsafeRawPointer(pointer)
                 .advanced(by: offset)
-                .assumingMemoryBound(to: fieldType)
+                .assumingMemoryBound(to: T.self)
                 .pointee
         }
     }
-    
+
+    /// Builds the fully-typed forward comparison closure for a field of type `Field` compared by `fieldComparator` (which must be in `.forward` order).
+    ///
+    /// This is the cast-free hot path: the field is read as its concrete type and `fieldComparator.compare` is invoked directly, so no value is ever boxed into `Any` and no dynamic cast occurs per comparison. When the field is a stored property (`offset != nil`) it is read at its byte offset; otherwise it is accessed through the key path.
+    private static func makeCompare<Field, FieldComparator: SortComparator>(
+        keyPath: KeyPath<Compared, Field> & Sendable,
+        offset: Int?,
+        fieldComparator: FieldComparator
+    ) -> @Sendable (Compared, Compared) -> ComparisonResult where FieldComparator.Compared == Field {
+        if let offset {
+            return { lhs, rhs in
+                let lhsField = getField(offset: offset, from: lhs, as: Field.self)
+                let rhsField = getField(offset: offset, from: rhs, as: Field.self)
+                return fieldComparator.compare(lhsField, rhsField)
+            }
+        }
+        return { lhs, rhs in
+            fieldComparator.compare(lhs[keyPath: keyPath], rhs[keyPath: keyPath])
+        }
+    }
+
     // A temporary workaround to a compiler bug that changes the ABI when adding the & Sendable constraint
     // Should be removed and the related functions should be made public when rdar://131764614 is resolved
     @_alwaysEmitIntoClient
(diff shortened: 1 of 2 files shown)


parkera:
@swift-ci test

Sources/FoundationInternationalization/String/KeyPathComparator.swift
```diff
@@ -23,33 +23,58 @@ public struct KeyPathComparator<Compared>: SortComparator {
...
-        get {
-            comparator.order
-        }
-        set {
-            comparator.order = newValue
-        }
+        didSet { comparator.order = order }
     }
 
+    /// A type-erased copy of the underlying comparator, retained only for the
```
parkera: Please don't manually wrap comments at 80 chars - thanks.

parkera (commented):
The performance improvements are really nice. Any ABI stability or availability issues with the new inlined paths?

dnadoba:
I have only added `@inlinable` to the sort family of functions for now. They are very simple functions that mainly just call into the standard library sort function and then call into the public `SortComparator.compare(_:_:)` method which continues to **not** be `@inlineable`.

dnadoba:
@parkera any other comments or is this good to go?

parkera:
LGTM - adding @jmschonfeld as another set of eyes

dnadoba:
@jmschonfeld any comments or does it look good to you?

Sources/FoundationInternationalization/String/KeyPathComparator.swift
```diff
@@ -95,26 +108,26 @@ public struct KeyPathComparator<Compared>: SortComparator {
...
-                ofType: Value.self,
-                offset: cachedOffset,
-                from: $0,
-                fallback: sendableKP)
+            let fieldComparator = ComparableComparator<Value>()
+            self.comparator = AnySortComparator(fieldComparator)
+            self._compare = Self.makeCompare(keyPath: sendableKP, offset: cachedOffset, fieldComparator: fieldComparator)
         }
         self.order = order
+        self.comparator.order = order
```
jmschonfeld, line 130: nit: does the `didSet` on the `self.order = order` do this for us?

jmschonfeld (approved):
Sorry for the delay. One minor nit, but LGTM otherwise

Sources/FoundationInternationalization/String/KeyPathComparator.swift
dnadoba: I don't think so. didSet isn't called during initialization of a variable, just on reassignment. If I don't miss another `self.order =` anywhere we only initialize `self.order` once and don't reassigned it so this is still needed.

jmschonfeld: Oh cool I didn't realize that about `didSet` - makes sense!

PR #2110, opened 2026-07-14
[NewCodable] Fix decodeEachElement dispatching closure on empty arrays

t089:
@kperryua just realized I had this patch lying around for a while but had never opened a PR for it. Seems like it is still relevant.

`JSONParserDecoder.ArrayDecoder.decodeEachElement` uses a `repeat...while` loop, so it dispatches the element closure once even when the array is empty (`hasNext == false`, as computed by the init-time `prepareForArrayElement(first: true)`). The closure then tries to decode an element out of `]`, producing errors like:

```
DecodingError: Data corrupted, at path: scores[0], (Unexpected character ']'), at Line: 1 Column: 29 Offset: 28
```

for any `[]` value decoded through the generic `decodeArray` + `decodeEachElement` API.

The specialized `decode([Element].self, sizeHint:)` fast path already guards this case (which is why plain `[Double].self` decoding is unaffected), but anything using `ArrayDecoder` directly — e.g. the `InlineArray` conformance or hand-written `decode(from:)` implementations — hits the bug.

### Fix

Guard on `hasNext` at the top of `decodeEachElement` so empty arrays return without dispatching the closure.

### Testing

Added a regression test decoding `[]` and `[1.5, 2.5]` through a custom `JSONDecodable` that uses `decodeArray` + `decodeEachElement`. The empty case fails with "Unexpected character ']'" before this change and passes after; the non-empty case passes both before and after.

### DIFF
--- a/Sources/NewCodable/JSON/JSONParserDecoder.swift
+++ b/Sources/NewCodable/JSON/JSONParserDecoder.swift
@@ -424,8 +424,7 @@ public struct JSONParserDecoder: JSONDecoderProtocol, ~Escapable {
         
         @_lifetime(self: copy self)
         public mutating func decodeEachElement(_ closure: (inout ElementDecoder) throws(CodingError.Decoding) -> Void) throws(CodingError.Decoding) {
-            // Honor `hasNext` (set by the init-time `prepareForArrayElement(first: true)`)
-            // so empty arrays don't dispatch the closure once into nothing.
+            // noop for empty array
             guard hasNext else { return }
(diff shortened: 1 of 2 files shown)


kperryua (approved):
Great fix. Thanks!

kperryua:
@swift-ci test

t089:
@kperryua could we get this merged?

kperryua:
Of course. Apologies, I’ve been out on vacation. 😀

PR #2121, opened 2026-07-17
Remove SendableOSLog

jmschonfeld:
Removes `SendableOSLog` in favor of `Logger`

### Motivation:

Instead of creating an unchecked `Sendable` wrapper around `OSLog`, we can just use `Logger` which is already `Sendable`. In fact, every use of a `SendableOSLog` converts it to a `Logger` before calling an API

### Modifications:

Replace `SendableOSLog` with `Logger`.

### Result:

The project no longer uses `SendableOSLog`

### Testing:

No additional tests necessary, CI will confirm the project still builds


itingliu (approved):
Thank you!

PR #2122, opened 2026-07-19
[Android] Build and run the tests

finagolfin:
Push the CI down to Android API ~23~ ~24~ 26, add imports where needed, and disable or modify tests for 32-bit armv7.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -22,8 +22,9 @@ jobs:
       enable_cross_pr_testing: true
       enable_android_sdk_build: true
       android_sdk_versions: '["nightly-main"]'
+      android_sdk_triples: '["aarch64-unknown-linux-android23", "armv7-unknown-linux-android23", "x86_64-unknown-linux-android26"]'
+      enable_android_sdk_checks: true
       enable_macos_swiftly_checks: true
-      android_sdk_triples: '["aarch64-unknown-linux-android28", "armv7-unknown-linux-android28", "x86_64-unknown-linux-android28"]'
 
   cmake_build:
     name: Build (CMake)
--- a/Sources/_FoundationCShims/include/string_shims.h
+++ b/Sources/_FoundationCShims/include/string_shims.h
@@ -39,16 +39,6 @@ inline static int _stringshims_LC_ALL_MASK() {
 #include <strings.h>
 #endif
 
-#if defined(TARGET_OS_ANDROID) && TARGET_OS_ANDROID
-inline int _stringshims_android_strncasecmp_l(const char *s1, const char *s2, size_t n, locale_t locale) {
-#if __ANDROID_API__ < 23
-    return strncasecmp(s1, s2, n);
-#else
-    return strncasecmp_l(s1, s2, n, locale);
-#endif
-}
-#endif
-
 #define _STRINGSHIMS_MACROMAN_MAP_SIZE 129
 INTERNAL const uint8_t _stringshims_macroman_mapping[_STRINGSHIMS_MACROMAN_MAP_SIZE][3];
 
--- a/Tests/FoundationEssentialsTests/DataIOTests.swift
+++ b/Tests/FoundationEssentialsTests/DataIOTests.swift
@@ -245,8 +245,12 @@ extension LargeDataTests {
     @Test func readLargeFile() throws {
         let url = URL.temporaryDirectory.appendingPathComponent("testfile-\(UUID().uuidString)")
         defer { try? FileManager.default.removeItem(at: url) }
-        // More than 2 GB
+        // More than 2 GB, if not 32-bit
+        #if _pointerBitWidth(_32)
+        let size = 0x7FFFFFFF
+        #else
         let size = 0x80010000
+        #endif
         
         let data = generateTestData(count: size)
         
--- a/Tests/FoundationEssentialsTests/DecimalTests.swift
+++ b/Tests/FoundationEssentialsTests/DecimalTests.swift
@@ -18,6 +18,8 @@ import Darwin
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Android)
+import Android
 #elseif os(WASI)
 import WASILibc
 #elseif os(Windows)
@@ -826,7 +828,7 @@ private struct DecimalTests {
         #expect(Decimal(string: "5538.0")! - Decimal(string: "2880.4")! == Decimal(string: "2657.6")!)
         #expect(Decimal(string: "2880.4")! - Decimal(5538) == Decimal(string: "-2657.6")!)
         #expect(Decimal(0x10000) - Decimal(0x1000) == Decimal(0xf000))
-#if !os(watchOS)
+#if _pointerBitWidth(_64)
         #expect(Decimal(0x1_0000_0000) - Decimal(0x1000) == Decimal(0xFFFFF000))
         #expect(Decimal(0x1_0000_0000_0000) - Decimal(0x1000) == Decimal(0xFFFFFFFFF000))
 #endif
--- a/Tests/FoundationEssentialsTests/StringTests.swift
+++ b/Tests/FoundationEssentialsTests/StringTests.swift
@@ -18,6 +18,8 @@ import Darwin
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Android)
+import Android
 #elseif os(WASI)
 import WASILibc
 #elseif os(Windows)
--- a/Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
+++ b/Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
@@ -503,7 +503,7 @@ private struct NumberFormatStyleTests {
     }
 #endif // FOUNDATION_PREVIEW
 
-#if !os(watchOS) // These tests require Int to be Int64, which is not always true on watch OSs yet
+#if _pointerBitWidth(_64) // These tests require Int to be Int64, which is not true on 32-bit watchOS or Android
     @Test func currency_compactName() throws {
         let baseStyle = Decimal.FormatStyle.Currency(code: "USD", locale: Locale(identifier: "en_US")).notation(.compactName)
 
@@ -737,7 +737,7 @@ private struct NumberFormatStyleTests {
         #expect( (-92233720368547 as Decimal).formatted(baseStyle.rounded(rule: .awayFromZero, increment: 100)) == "-$100E14")
         #expect((-922337203685477 as Decimal).formatted(baseStyle.rounded(rule: .awayFromZero, increment: 100)) == "-$100E15")
     }
-#endif // !os(watchOS)
+#endif // _pointerBitWidth(_64)
 }
 
 extension NumberFormatStyleConfiguration.Collection {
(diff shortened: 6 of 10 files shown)


Sources/_FoundationCShims/include/string_shims.h
```diff
@@ -39,16 +39,6 @@ inline static int _stringshims_LC_ALL_MASK() {
 #include <strings.h>
 #endif
 
-#if defined(TARGET_OS_ANDROID) && TARGET_OS_ANDROID
-inline int _stringshims_android_strncasecmp_l(const char *s1, const char *s2, size_t n, locale_t locale) {
```
finagolfin, line 43: Being `inline` means it wasn't emitted to the shims object file compiled from C and then the test runner link couldn't find this function, so the test build would fail. Since none of the official Android SDKs support older than API 23 anyway, just removed it.

Tests/FoundationEssentialsTests/StringTests.swift
```diff
@@ -18,6 +18,8 @@ import Darwin
 import Glibc
 #elseif canImport(Musl)
 import Musl
+#elseif canImport(Android)
+import Android
```
finagolfin, line 22: I have been adding these imports when building the tests natively on Android for years, now finally can use them upstream here since Swift 6.4 onwards can cross-compile the tests.

finagolfin:
The emulator kept stalling on API 23, so try API 24.

.github/workflows/pull_request.yml
```diff
@@ -23,7 +23,8 @@ jobs:
       enable_android_sdk_build: true
       android_sdk_versions: '["nightly-main"]'
       enable_macos_swiftly_checks: true
-      android_sdk_triples: '["aarch64-unknown-linux-android28", "armv7-unknown-linux-android28", "x86_64-unknown-linux-android28"]'
+      android_sdk_triples: '["aarch64-unknown-linux-android26", "armv7-unknown-linux-android26", "x86_64-unknown-linux-android26"]'
```
finagolfin: @marcprux, any idea what's going on here? When I tried API 23 and 24, the emulator kept hanging at startup, until it finally ran with 26.

marcprux: Unlike the API 26 CI run, which [reports](https://github.com/swiftlang/swift-foundation/actions/runs/29696090174/job/88216997965#step:11:369):

```
INFO         | Boot completed in 35145 ms
```

The [API 23](https://github.com/swiftlang/swift-foundation/actions/runs/29687016496/job/88200130931) and [API 24](https://github.com/swiftlang/swift-foundation/actions/runs/29692707477/job/88208131324) jobs never get that far, but don't appear to log any errors.

We'd need probably need to dig in further by adding some more debugging to the workflow script (perhaps by poking at it to see if it crashed in the [script](https://github.com/swiftlang/github-workflows/blob/497d9ab8410e69981b3b2f28da24cf7bf4b8733d/.github/workflows/scripts/android/android-emulator-tests.sh#L110)), but unfortunately I won't be able to get to it for a few days.

finagolfin: Thanks for the detailed explanation on a Sunday, 👍 please take a look whenever you like, as API 26 should suffice here for now.

.github/workflows/pull_request.yml
```diff
@@ -22,8 +22,10 @@ jobs:
       enable_cross_pr_testing: true
       enable_android_sdk_build: true
       android_sdk_versions: '["nightly-main"]'
+      android_ndk_versions: '["r27d", "r28c", "r29", "r30-beta2"]'
```
finagolfin: Once LTS NDK 30 releases in the next couple months, all prior NDKs are unsupported, so I'll remove them here then.

Tests/FoundationEssentialsTests/DataIOTests.swift
```diff
@@ -245,8 +245,12 @@ extension LargeDataTests {
     @Test func readLargeFile() throws {
         let url = URL.temporaryDirectory.appendingPathComponent("testfile-\(UUID().uuidString)")
         defer { try? FileManager.default.removeItem(at: url) }
-        // More than 2 GB
+        // More than 2 GB, if not 32-bit Android
+        #if os(Android) && arch(arm)
```
jmschonfeld: `DataTests.swift` defines `largeCount` for this purpose - could we maybe move that declaration somewhere (maybe like an extension on `Data` or something) and then use it here to avoid duplicating the logic? The other location already uses the appropriate `_pointerBitWidth` conditionals for this (assuming this isn't really something specific to Android and arm)

Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -784,7 +786,7 @@ private struct DecimalTests {
         #expect(Decimal(string: "5538.0")! - Decimal(string: "2880.4")! == Decimal(string: "2657.6")!)
         #expect(Decimal(string: "2880.4")! - Decimal(5538) == Decimal(string: "-2657.6")!)
         #expect(Decimal(0x10000) - Decimal(0x1000) == Decimal(0xf000))
-#if !os(watchOS)
+#if !os(watchOS) && !(os(Android) && arch(arm))
```
jmschonfeld: If this is just something unsupported by 32bit platforms, we should just gate it on `_pointerBitWidth(_64)`

Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
```diff
@@ -503,7 +503,7 @@ private struct NumberFormatStyleTests {
     }
 #endif // FOUNDATION_PREVIEW
 
-#if !os(watchOS) // These tests require Int to be Int64, which is not always true on watch OSs yet
+#if !os(watchOS) && !(os(Android) && arch(arm)) // These tests require Int to be Int64, which is not true on 32-bit watchOS or Android
```
jmschonfeld: Same comment here about `_pointerBitWidth(_64)`

.github/workflows/pull_request.yml
jmschonfeld: Does this mean that for the time being we're doing to have 4 different android runs? cc @itingliu on what our policy should be here - for Linux we only test against one platform, I'm not sure if we want to try to be exhaustive here or just pick the lowest version as a sentinel

finagolfin: Yes, 4 runs on every pull. Prior to this pull, this CI was doing two runs with the default NDKs from the official workflows: the latest LTS 27d and the stable 28c used to build the trunk Swift SDK bundle itself. I thought adding 29 and 30-beta2 would give us more signal before the final LTS 30 release planned later this year, after which I'd submit a pull to only use 30 here.

Completely up to you all what we end up using, just let me know.

Tests/FoundationEssentialsTests/DecimalTests.swift
finagolfin: I'm not sure on watchOS, which also has `arm64_32`, but yeah, only a problem on 32-bit Android.

jmschonfeld: @iCharlesHu - just confirming, the only need for `os(watchOS)` checks here was 32 bit right? Does Decimal have any other 64-bit watchOS specific behavior here?

Tests/FoundationEssentialsTests/DataIOTests.swift
finagolfin: [That testing constant uses completely different numbers](https://github.com/swiftlang/swift-foundation/blob/c617f187912b7f1b3e8e5c474516a6cd9b026a75/Tests/FoundationEssentialsTests/DataTests.swift#L3033) though: the whole point here is to go over `Int32.max` on 64-bit, and I used that max on 32-bit Android instead. Now, I'm not sure exactly what the best numbers to use here are, plus these large data tests [are not running on the Android emulator here anyway](https://github.com/swiftlang/swift-foundation/actions/runs/29703279380/job/88235903287?pr=2122#step:11:1108), because 6.4 GB memory is not allocated (nothing specific to 32-bit Android armv7 here, this will fail to compile on all 32-bit platforms).

jmschonfeld: Ah right - the purpose of this test is not _just_ to get a `Data` that is the large slice representation (which `Int32.max` does on 64 bit platforms) but is to check a case where we previously trapped due to an integer overflow from the file size (which requires `Int32.max + 1` on 64 bit platforms).

So you're right that we shouldn't use `largeSize` here since that won't hit the overflow. That being said, I don't think this test actually tests anything on 32 bit platforms with this change. If we really wanted to test for 32 bit platforms, we need to create a file with at least `Int32.max + 1` bytes and then try to read it and ensure that it throws instead of traps (unlike on 64 bit platforms where it succeeds).

So I think either you can guard this test with `_pointerBitWidth(_64)` to maintain the existing behavior (since it never ran on 32 bit platforms to begin with as you mentioned due to the memory constraints) or in the 32 bit path (`_pointerBitWidth(_32)`) we can create a larger file that tests the overflow check rather than the success path.

.github/workflows/pull_request.yml
jmschonfeld: I spoke with @itingliu and @iCharlesHu about this - right now we only build for one version for each platform (only jammy for linux, only windows-2022 for windows, etc.) and we'd like to be consistent among our platforms for CI. So for now, let's keep this at just `r27d` as the lowest denominator for what we support like the others.

It does, however, seem like it might be beneficial to test against multiple versions since there can be slight differences between them. However, I don't think there's been any guidance given to swiftlang projects around what set of supported platforms we should have CI enabled for, so I don't want to jump ahead on this until we have clear guidance for what we should expand to beyond the consistency we have today.

finagolfin: No guidance, but the current official workflows build for two NDKs by default, so I'll keep it that way.

Tests/FoundationEssentialsTests/DataIOTests.swift
finagolfin: It doesn't run on the 64-bit Android x86_64 emulator too, if you follow my link, but it would run on 32-bit, because `largeCount` is much smaller there. You're right that this test is probably useless on 32-bit, but I kept it checking something, ie the max addressable by the allocation method (don't think it's possible to check overflow on 32-bit because the internal allocator this is passed to only takes `Int`, not `Int64`).

These file size constants should be fine, will just change the compilation condition checked if wanted.

.github/workflows/pull_request.yml
jmschonfeld: Ok yeah since that's what the pipeline does today it seems reasonable to me to keep it that way today to preserve existing behavior

Tests/FoundationEssentialsTests/DataIOTests.swift
jmschonfeld: Yeah I see that it isn't running on the emulator either due to the lack of space in memory. My main point here is that the change added a conditional for "Android" and "arm" but nothing about this test is specific to android or arm - rather it is specific to 32bit platforms in general so we should use `_pointerBitWidth(_32)` if we need to add a new conditional. And that if we're adding a conditional, the test should either do something useful or we should use the conditional to skip the test if it does nothing useful (which I think is the case for 32 bit platforms). I think the reason this didn't matter much before is that it's not probable the LargeDataTests suite will run on any 32 bit platforms due to the constrained memory environment.

finagolfin: I'll change it to `_pointerBitWidth(_32)`, but there's almost no way these tests _won't_ run on 32-bit hardware, as `largeCount` is only 64 KB there, so they will run if 192 KB of memory is allocated, ie pretty much always. Rather, the problem is that nobody is running them on 32-bit hardware, which we may add on Android eventually, and these are compilation errors on 32-bit.

I made this allocate the maximum possible on 32-bit with our `Int-based` allocator now, almost 2GB, think that is still worth checking.

.github/workflows/pull_request.yml
finagolfin: Went back to the two default NDKs

Tests/FoundationEssentialsTests/DataIOTests.swift
finagolfin: Changed the condition, kept the constant

finagolfin:
Rebased and made the changes discussed, let me know if there's anything else.

Tests/FoundationEssentialsTests/DecimalTests.swift
iCharlesHu: No, I believe this was only a 32bit check. Decimal shouldn't haven't any WatchOS specific behaviors.

finagolfin: Alright, will change these two to the 32-bit check then.

finagolfin: Done

Tests/FoundationInternationalizationTests/Formatting/NumberFormatStyleTests.swift
finagolfin: Updated the checks in these tests to 64-bit

jmschonfeld (approved):

PR #2123, opened 2026-07-20
Add a Swift implementation of the Chinese calendar

dra8an:
## Add a Swift implementation of the Chinese calendar 

Follows the Hebrew (#1953/#2028) and Buddhist/Japanese (#2105) pattern: a `_CalendarChinese` implementation gated behind `foundation_swift_chinese_calendar_feature_enabled()`. When the flag is off, `Calendar(identifier: .chinese)` continues to route to the ICU-backed calendar exactly as today.

### Design

For Chinese years beginning in Gregorian 1901 through 2100, month structure comes from a baked table (200 x `UInt32` = 800 bytes) generated from `_CalendarICU(.chinese)` Each `UInt32` packs the 12 or 13 month-length bits (29 vs 30 days), the leap-month index (0 = none), and the new-year offset. In range there is no astronomy at all, a date-to-fields query is a direct index plus a bit extraction.

Outside that range, ICU's own chnsecal rules layer over a Meeus / Reingold-Dershowitz astronomy engine evaluated at UTC+8 (Bretagnon & Simon solar series, periodic-term new-moon series). Field conventions match ICU: era is the 60-year cycle number since the 2637 BCE epoch, year is 1...60 within the cycle, month is 1...12 with `isLeapMonth` distinguishing the repeated month, and the extended year used by `yearForWeekOfYear` is the related Gregorian year plus 2637.

### Table size vs the ICU tables it replaces

Apple's ICU fork already ships static tables for the Chinese calendar over Gregorian 1900 to 2100:

| ICU table (astro.cpp / chnsecal.cpp) | Type | Bytes |
| --- | --- | --- |
| `newMoonDates[]` (every new moon, i.e. every month start) | `int32_t` | 9,956 |
| `winterSolsticeDates[]` | `int32_t` | 808 |
| `sunLongitudeAdjustmts[][4]` | `int8_t` | 808 |
| `newYearAdj[]` | `int8_t` | 201 |
| Total | | ~11,773 (~11.5 KB) |

Our table is 800 bytes for the same in-range span, roughly 12x smaller than ICU's new-moon table alone and ~15x smaller than all four combined. We store month lengths as one bit each (derived from new-moon differences) at about 4 bytes per year, where ICU stores each new moon as a full minute-accurate `int32_t` timestamp at about 50 bytes per year, plus separate solstice and sun-longitude tables. 

### Verification

Exhaustive suites were run on the research branch; a distilled subset ships here as `ChineseCalendarTests` and `ChineseRecurrenceRuleParityProbe`.

- Daily parity vs ICU across 1899 to 2102 (74,510 days): zero divergence in the baked range, except two months where ICU itself emits an impossible `day = 0` field (2057-09, 2097-08), documented as an ICU internal inconsistency.

- Cross-checked against the Hong Kong Observatory's official tables (1901 to 2100): matches except exactly ICU's three known HKO deviations, which we keep for parity.

- Out of range, validated against the promulgated Qing record (four independent sources) and Yuk Tung Liu's DE441 computation for 2101 to 2200: 97 of 100 years exact, the three differences within Liu's own stated uncertainty. The intentional divergences from ICU there stem from ICU's Duffett-Smith astronomy having no delta-T (it errs 25 to 60 minutes and invents leap months in 1794, 1813, and 1889); our engine does not.

### Benchmarks

Release, macOS on Apple Silicon, p50 medians. `_CalendarChinese` (flag on) vs the ICU-backed calendar (flag off), same benchmark code:

| Benchmark | ICU | Native | Speedup |
| --- | --- | --- | --- |
| dateComponents (year/month/day) | 24 ns | 2 ns | 12.0x |
| roundTrip dateComponents | 65 ns | 4 ns | 16.3x |
| allocationsForFixedCalendar | 3,300 ns | 129 ns | 25.6x |
| copyOnWritePerformance | 2,043 ns | 50 ns | 40.9x |
| nextThousandNewYears | 26 us | 13 us | 2.0x |

Allocations (malloc total, p50): copyOnWrite 21 to 1, nextThousandNewYears 5 to 1, allocationsForFixedCalendar 4 to 0. `nextThousandNewYears` shows the smallest speedup because it is bound by the `enumerateDates` machinery rather than the calendar math.

The `nextDate` fast path is deliberately deferred to a follow-up PR (leap month match semantics complicate the matching patterns), as with #2105.

### Week-year semantics: one intentional behavior difference

This is the single place where flipping the flag changes observable behavior on valid input, so it is called out explicitly.

ICU's chnsecal `handleGetExtendedYear` never reads `YEAR_WOY`. As a result the ICU-backed calendar computes a date's `yearForWeekOfYear` correctly but cannot use that field to construct or step dates: `dateInterval(of: .yearForWeekOfYear)` degenerates to nil and `date(byAdding:)` on it is a silent no-op. 

Instead, implement Foundation's documented week-year semantics: week 1 is anchored by `firstWeekday` and `minimumDaysInFirstWeek` on the Chinese year start, intervals tile exactly, and adding is O(1). This is the same implementation shape as the merged Hebrew calendar. Field values still match ICU exactly.

### Testing done

- `swift build` and `swift build -c release`: clean.
- `swift test` full suite: passing (calendar suites all green).
- `ChineseCalendarTests`: 8/8 in debug and release.
- `ChineseRecurrenceRuleParityProbe` and the yearless-birthday recurrence probe: passing.
- CMakeLists to SwiftPM source-list parity: verified.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
@@ -79,33 +79,34 @@ internal enum _CalendarAstronomy {
         let y0 = Double(yearInt) / 100.0
         let y1820 = Double(yearInt - 1820) / 100.0
 
-        if (2051...2150).contains(yearInt) {
+        switch yearInt {
+        case 2051...2150:
             return (-20.0 + 32.0 * Double((yearInt - 1820) * (yearInt - 1820)) / 10000.0
                     + 0.5628 * Double(2150 - yearInt)) / 86400.0
-        } else if (2006...2050).contains(yearInt) {
+        case 2006...2050:
             return (62.92 + 0.32217 * y2000 + 0.005589 * y2000 * y2000) / 86400.0
-        } else if (1987...2005).contains(yearInt) {
+        case 1987...2005:
             return polynomial(y2000, [63.86, 0.3345, -0.060374, 0.0017275,
                                 0.000651814, 0.00002373599]) / 86400.0
-        } else if (1900...1986).contains(yearInt) {
+        case 1900...1986:
             return polynomial(c, [-0.00002, 0.000297, 0.025184, -0.181133,
                             0.553040, -0.861938, 0.677066, -0.212591])
-        } else if (1800...1899).contains(yearInt) {
+        case 1800...1899:
             return polynomial(c, [-0.000009, 0.003844, 0.083563, 0.865736,
                             4.867575, 15.845535, 31.332267, 38.291999,
                             28.316289, 11.636204, 2.043794])
-        } else if (1700...1799).contains(yearInt) {
+        case 1700...1799:
             return polynomial(y1700, [8.118780842, -0.005092142, 0.003336121,
                                 -0.0000266484]) / 86400.0
-        } else if (1600...1699).contains(yearInt) {
+        case 1600...1699:
             return polynomial(y1600, [120.0, -0.9808, -0.01532, 0.000140272128]) / 86400.0
-        } else if (500...1599).contains(yearInt) {
+        case 500...1599:
             return polynomial(y1000, [1574.2, -556.01, 71.23472, 0.319781,
                                 -0.8503463, -0.005050998, 0.0083572073]) / 86400.0
-        } else if (-499...499).contains(yearInt) {
+        case -499...499:
             return polynomial(y0, [10583.6, -1014.41, 33.78311, -5.952053,
                              -0.1798452, 0.022174192, 0.0090316521]) / 86400.0
-        } else {
+        default:
             return (-20.0 + 32.0 * y1820 * y1820) / 86400.0
         }
     }
--- a/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
@@ -1,0 +1,246 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+#if canImport(os)
+internal import os
+#elseif canImport(Bionic)
+@preconcurrency import Bionic
+#elseif canImport(Glibc)
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif canImport(CRT)
+import CRT
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
+
+/// Shared astronomical and Gregorian day-number toolkit for the non-arithmetic calendars (Chinese today; Islamic and Hindu variants can build on it). Solar/lunar theory follows Reingold & Dershowitz, Calendrical Calculations (built on the Meeus and Bretagnon & Simon series); all APIs are calendar-agnostic.
+internal enum _CalendarAstronomy {
+    static let meanSynodicMonth = 29.530588861
+    static let j2000 = 730120.5
+    static let newMoonZero = 11.458922815770109
+
+    static func poly(_ x: Double, _ coefficients: [Double]) -> Double {
+        var result = 0.0
+        var power = 1.0
+        for c in coefficients {
+            result += c * power
+            power *= x
+        }
+        return result
+    }
+
+    static func mod360(_ x: Double) -> Double {
+        var r = x.truncatingRemainder(dividingBy: 360.0)
+        if r < 0 { r += 360.0 }
+        return r
+    }
+
+    static func sinDeg(_ d: Double) -> Double { sin(d * .pi / 180.0) }
+    static func cosDeg(_ d: Double) -> Double { cos(d * .pi / 180.0) }
+
+    // Proleptic Gregorian y/m/d -> Rata Die (day 1 = 0001-01-01).
+    static func gregorianRD(_ y: Int, _ m: Int, _ d: Int) -> Int {
+        func fd(_ a: Int, _ b: Int) -> Int { a >= 0 ? a / b : -((-a + b - 1) / b) }
+        let ym1 = y - 1
+        let leap = (fd(y, 4) * 4 == y && fd(y, 100) * 100 != y) || fd(y, 400) * 400 == y
+        var r = 365 * ym1 + fd(ym1, 4) - fd(ym1, 100) + fd(ym1, 400)
+        r += fd(367 * m - 362, 12) + (m <= 2 ? 0 : (leap ? -1 : -2)) + d
+        return r
+    }
+
+    static func gregorianYear(ofRD day: Int) -> Int {
+        var y = Int((Double(day) / 365.2425).rounded(.down)) + 1
+        while gregorianRD(y, 1, 1) > day { y -= 1 }
+        while gregorianRD(y + 1, 1, 1) <= day { y += 1 }
+        return y
+    }
+
+    // Dynamical-minus-universal time (fraction of a day). Meeus/NASA fits.
+    static func ephemerisCorrection(_ moment: Double) -> Double {
+        let year = moment / 365.2425
+        let yearInt = Int(year > 0 ? year + 1 : year)
+        let fixedMidYear = gregorianRD(yearInt, 7, 1)
+        let c = (Double(fixedMidYear) - 693596.0) / 36525.0
+        let y2000 = Double(yearInt - 2000)
+        let y1700 = Double(yearInt - 1700)
+        let y1600 = Double(yearInt - 1600)
+        let y1000 = Double(yearInt - 1000) / 100.0
+        let y0 = Double(yearInt) / 100.0
+        let y1820 = Double(yearInt - 1820) / 100.0
+
+        if (2051...2150).contains(yearInt) {
+            return (-20.0 + 32.0 * Double((yearInt - 1820) * (yearInt - 1820)) / 10000.0
+                    + 0.5628 * Double(2150 - yearInt)) / 86400.0
+        } else if (2006...2050).contains(yearInt) {
+            return (62.92 + 0.32217 * y2000 + 0.005589 * y2000 * y2000) / 86400.0
+        } else if (1987...2005).contains(yearInt) {
+            return poly(y2000, [63.86, 0.3345, -0.060374, 0.0017275,
+                                0.000651814, 0.00002373599]) / 86400.0
+        } else if (1900...1986).contains(yearInt) {
+            return poly(c, [-0.00002, 0.000297, 0.025184, -0.181133,
+                            0.553040, -0.861938, 0.677066, -0.212591])
+        } else if (1800...1899).contains(yearInt) {
+            return poly(c, [-0.000009, 0.003844, 0.083563, 0.865736,
+                            4.867575, 15.845535, 31.332267, 38.291999,
+                            28.316289, 11.636204, 2.043794])
+        } else if (1700...1799).contains(yearInt) {
+            return poly(y1700, [8.118780842, -0.005092142, 0.003336121,
+                                -0.0000266484]) / 86400.0
+        } else if (1600...1699).contains(yearInt) {
+            return poly(y1600, [120.0, -0.9808, -0.01532, 0.000140272128]) / 86400.0
+        } else if (500...1599).contains(yearInt) {
+            return poly(y1000, [1574.2, -556.01, 71.23472, 0.319781,
+                                -0.8503463, -0.005050998, 0.0083572073]) / 86400.0
+        } else if (-499...499).contains(yearInt) {
+            return poly(y0, [10583.6, -1014.41, 33.78311, -5.952053,
+                             -0.1798452, 0.022174192, 0.0090316521]) / 86400.0
+        } else {
+            return (-20.0 + 32.0 * y1820 * y1820) / 86400.0
+        }
+    }
+
+    static func universalFromDynamical(_ dynamical: Double) -> Double {
+        dynamical - ephemerisCorrection(dynamical)
+    }
+
+    static func julianCenturies(_ moment: Double) -> Double {
+        (moment + ephemerisCorrection(moment) - j2000) / 36525.0
+    }
+
+    static func nutation(_ c: Double) -> Double {
+        let a = 124.90 - 1934.134 * c + 0.002063 * c * c
+        let b = 201.11 + 72001.5377 * c + 0.00057 * c * c
+        return -0.004778 * sinDeg(a) - 0.0003667 * sinDeg(b)
+    }
+
+    static func aberration(_ c: Double) -> Double {
+        0.0000974 * cosDeg(177.63 + 35999.01848 * c) - 0.005575
+    }
+
+    // Static so the hot lunation/solstice paths don't allocate arrays per call.
+    private static let solarCoefficients: [Double] = [
+        403406, 195207, 119433, 112392, 3891, 2819, 1721, 660, 350, 334,
+        314, 268, 242, 234, 158, 132, 129, 114, 99, 93, 86, 78, 72,
+        68, 64, 46, 38, 37, 32, 29, 28, 27, 27, 25, 24, 21, 21,
+        20, 18, 17, 14, 13, 13, 13, 12, 10, 10, 10, 10,
+    ]
+    private static let solarAddends: [Double] = [
+        270.54861, 340.19128, 63.91854, 331.26220, 317.843, 86.631, 240.052, 310.26, 247.23,
+        260.87, 297.82, 343.14, 166.79, 81.53, 3.50, 132.75, 182.95, 162.03, 29.8, 266.4,
+        249.2, 157.6, 257.8, 185.1, 69.9, 8.0, 197.1, 250.4, 65.3, 162.7, 341.5, 291.6, 98.5,
+        146.7, 110.0, 5.2, 342.6, 230.9, 256.1, 45.3, 242.9, 115.2, 151.8, 285.3, 53.3, 126.6,
+        205.7, 85.9, 146.1,
+    ]
+    private static let solarMultipliers: [Double] = [
+        0.9287892, 35999.1376958, 35999.4089666, 35998.7287385, 71998.20261, 71998.4403,
+        36000.35726, 71997.4812, 32964.4678, -19.4410, 445267.1117, 45036.8840, 3.1008,
+        22518.4434, -19.9739, 65928.9345, 9038.0293, 3034.7684, 33718.148, 3034.448,
+        -2280.773, 29929.992, 31556.493, 149.588, 9037.750, 107997.405, -4444.176, 151.771,
+        67555.316, 31556.080, -4561.540, 107996.706, 1221.655, 62894.167, 31437.369,
+        14578.298, -31931.757, 34777.243, 1221.999, 62894.511, -4442.039, 107997.909,
+        119.066, 16859.071, -4.578, 26895.292, -39.127, 12297.536, 90073.778,
+    ]
+
+    // Solar longitude in degrees [0, 360), 49-term Bretagnon & Simon series.
+    static func solarLongitude(at moment: Double) -> Double {
+        let c = julianCenturies(moment)
+        var lambda = 0.0
+        for i in 0..<49 {
+            lambda += solarCoefficients[i] * sinDeg(solarAddends[i] + solarMultipliers[i] * c)
+        }
+        lambda *= 0.000005729577951308232
+        lambda += 282.7771834 + 36000.76953744 * c
+        return mod360(lambda + aberration(c) + nutation(c))
+    }
+
+
+    private static let newMoonSineCoefficients: [Double] = [
+        -0.40720, 0.17241, 0.01608, 0.01039, 0.00739, -0.00514, 0.00208, -0.00111, -0.00057,
+        0.00056, -0.00042, 0.00042, 0.00038, -0.00024, -0.00007, 0.00004, 0.00004, 0.00003,
+        0.00003, -0.00003, 0.00003, -0.00002, -0.00002, 0.00002,
+    ]
+    private static let newMoonSolarFactors: [Double] = [
+        0, 1, 0, 0, -1, 1, 2, 0, 0, 1, 0, 1, 1, -1, 2, 0, 3, 1, 0, 1, -1, -1, 1, 0,
+    ]
+    private static let newMoonLunarFactors: [Double] = [
+        1, 0, 2, 0, 1, 1, 0, 1, 1, 2, 3, 0, 0, 2, 1, 2, 0, 1, 2, 1, 1, 1, 3, 4,
+    ]
+    private static let newMoonArgumentFactors: [Double] = [
+        0, 0, 0, 2, 0, 0, 0, -2, 2, 0, 0, 2, -2, 0, 0, -2, 0, -2, 2, 2, 2, -2, 0, 0,
+    ]
+    private static let newMoonExtraAddends: [Double] = [
+        251.88, 251.83, 349.42, 84.66, 141.74, 207.14, 154.84, 34.52, 207.19, 291.34, 161.72, 239.56, 331.55,
+    ]
+    private static let newMoonExtraMultipliers: [Double] = [
+        0.016321, 26.651886, 36.412478, 18.206239, 53.303771, 2.453732, 7.306860, 27.261239,
+        0.121824, 1.844379, 24.198154, 25.513099, 3.592518,
+    ]
+    private static let newMoonExtraCoefficients: [Double] = [
+        0.000165, 0.000164, 0.000126, 0.000110, 0.000062, 0.000060, 0.000056, 0.000047,
+        0.000042, 0.000040, 0.000037, 0.000035, 0.000023,
+    ]
+
+    // Moment of the nth new moon since the 24724-indexed epoch; Meeus 24+13 terms.
+    static func nthNewMoon(_ n: Int) -> Double {
+        let k = Double(n) - 24724.0
+        let c = k / 1236.85
+        let approx = j2000
+            + (5.09766 + meanSynodicMonth * 1236.85 * c
+               + 0.00015437 * c * c
+               - 0.00000015 * c * c * c
+               + 0.00000000073 * c * c * c * c)
+        let e = 1.0 - 0.002516 * c - 0.0000074 * c * c
+        let solarAnomaly = 2.5534 + 1236.85 * 29.10535670 * c
+            - 0.0000014 * c * c - 0.00000011 * c * c * c
+        let lunarAnomaly = 201.5643 + 385.81693528 * 1236.85 * c
+            + 0.0107582 * c * c + 0.00001238 * c * c * c
+            - 0.000000058 * c * c * c * c
+        let moonArgument = 160.7108 + 390.67050284 * 1236.85 * c
+            - 0.0016118 * c * c - 0.00000227 * c * c * c
+            + 0.000000011 * c * c * c * c
+        let omega = 124.7746 + (-1.56375588) * 1236.85 * c
+            + 0.0020672 * c * c + 0.00000215 * c * c * c
+
+        var correction = -0.00017 * sinDeg(omega)
+        for i in 0..<24 {
+            let ePow = pow(e, abs(newMoonSolarFactors[i]))
+            let arg = newMoonSolarFactors[i] * solarAnomaly + newMoonLunarFactors[i] * lunarAnomaly
+                + newMoonArgumentFactors[i] * moonArgument
+            correction += newMoonSineCoefficients[i] * ePow * sinDeg(arg)
+        }
+        let extra = 0.000325 * sinDeg(299.77 + 132.8475848 * c - 0.009173 * c * c)
+        var additional = 0.0
+        for i in 0..<13 {
+            additional += newMoonExtraCoefficients[i] * sinDeg(newMoonExtraAddends[i] + newMoonExtraMultipliers[i] * k)
+        }
+        return universalFromDynamical(approx + correction + extra + additional)
+    }
+
+    static func numOfNewMoonAtOrAfter(_ moment: Double) -> Int {
+        let rawN = ((moment - newMoonZero) / meanSynodicMonth).rounded()
+        var n = Int(rawN)
+        while nthNewMoon(n) < moment { n += 1 }
+        while nthNewMoon(n - 1) >= moment { n -= 1 }
+        return n
+    }
+
+    static func newMoonAtOrAfter(_ moment: Double) -> Double {
+        nthNewMoon(numOfNewMoonAtOrAfter(moment))
+    }
+
+    static func newMoonBefore(_ moment: Double) -> Double {
+        nthNewMoon(numOfNewMoonAtOrAfter(moment) - 1)
+    }
+}
+
--- a/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
@@ -30,7 +30,7 @@ internal enum _CalendarAstronomy {
     static let j2000 = 730120.5
     static let newMoonZero = 11.458922815770109
 
-    static func poly(_ x: Double, _ coefficients: [Double]) -> Double {
+    static func polynomial(_ x: Double, _ coefficients: [Double]) -> Double {
         var result = 0.0
         var power = 1.0
         for c in coefficients {
@@ -46,23 +46,23 @@ internal enum _CalendarAstronomy {
         return r
     }
 
-    static func sinDeg(_ d: Double) -> Double { sin(d * .pi / 180.0) }
-    static func cosDeg(_ d: Double) -> Double { cos(d * .pi / 180.0) }
+    static func sinDegrees(_ d: Double) -> Double { sin(d * .pi / 180.0) }
+    static func cosDegrees(_ d: Double) -> Double { cos(d * .pi / 180.0) }
 
     // Proleptic Gregorian y/m/d -> Rata Die (day 1 = 0001-01-01).
-    static func gregorianRD(_ y: Int, _ m: Int, _ d: Int) -> Int {
-        func fd(_ a: Int, _ b: Int) -> Int { a >= 0 ? a / b : -((-a + b - 1) / b) }
+    static func gregorianRataDie(_ y: Int, _ m: Int, _ d: Int) -> Int {
+        func floorDivide(_ a: Int, _ b: Int) -> Int { a >= 0 ? a / b : -((-a + b - 1) / b) }
         let ym1 = y - 1
-        let leap = (fd(y, 4) * 4 == y && fd(y, 100) * 100 != y) || fd(y, 400) * 400 == y
-        var r = 365 * ym1 + fd(ym1, 4) - fd(ym1, 100) + fd(ym1, 400)
-        r += fd(367 * m - 362, 12) + (m <= 2 ? 0 : (leap ? -1 : -2)) + d
+        let leap = (floorDivide(y, 4) * 4 == y && floorDivide(y, 100) * 100 != y) || floorDivide(y, 400) * 400 == y
+        var r = 365 * ym1 + floorDivide(ym1, 4) - floorDivide(ym1, 100) + floorDivide(ym1, 400)
+        r += floorDivide(367 * m - 362, 12) + (m <= 2 ? 0 : (leap ? -1 : -2)) + d
         return r
     }
 
-    static func gregorianYear(ofRD day: Int) -> Int {
+    static func gregorianYear(ofRataDie day: Int) -> Int {
         var y = Int((Double(day) / 365.2425).rounded(.down)) + 1
-        while gregorianRD(y, 1, 1) > day { y -= 1 }
-        while gregorianRD(y + 1, 1, 1) <= day { y += 1 }
+        while gregorianRataDie(y, 1, 1) > day { y -= 1 }
+        while gregorianRataDie(y + 1, 1, 1) <= day { y += 1 }
         return y
     }
 
@@ -70,7 +70,7 @@ internal enum _CalendarAstronomy {
     static func ephemerisCorrection(_ moment: Double) -> Double {
         let year = moment / 365.2425
         let yearInt = Int(year > 0 ? year + 1 : year)
-        let fixedMidYear = gregorianRD(yearInt, 7, 1)
+        let fixedMidYear = gregorianRataDie(yearInt, 7, 1)
         let c = (Double(fixedMidYear) - 693596.0) / 36525.0
         let y2000 = Double(yearInt - 2000)
         let y1700 = Double(yearInt - 1700)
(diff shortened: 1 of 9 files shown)


richgillam:
You haven't added any reviewers to this PR, and no one's kicked off the CI workflows. Is this PR ready to review?

richgillam:
>For Chinese years beginning in Gregorian 1901 through 2100, month structure comes from a baked table (200 x UInt32 = 800 bytes) generated from _CalendarICU(.chinese) Each UInt32 packs the 12 or 13 month-length bits (29 vs 30 days), the leap-month index (0 = none), and the new-year offset. In range there is no astronomy at all, a date-to-fields query is a direct index plus a bit extraction.

Do your own astronomical calculations also match the contents of this table?  If so, maybe the comment above should say _that_, rather than saying we match what the old ICU calendar did.  If not, why not?

richgillam:
>ICU's own chnsecal rules layer over a Meeus / Reingold-Dershowitz astronomy engine...

Some of this is over my head, but two things worry me: 1) Markus has told me that there are some clear deficiencies in `chnsecal.cpp`, to the point where it gets _the wrong answer_ for some date or set of dates.  Is that true, and how are we working around it?  I kind of thought (but maybe mistakenly) that Peter's lookup table fixed this, but that he didn't fix the underlying astronomy engine.  Is that right?  2) Reingold-Dershowitz didn't do us any favors with the Indian calendars.  Their code is reliable here?  And how much does it overlap with the stuff you had to put together to handle the Indian calendars?

richgillam:
>...matches except exactly ICU's three known HKO deviations, which we keep for parity.

Are we sure we want to do that?

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
```diff
@@ -0,0 +1,246 @@
...
+@preconcurrency import Bionic
+#elseif canImport(Glibc)
+@preconcurrency import Glibc
+#elseif canImport(Musl)
+@preconcurrency import Musl
+#elseif canImport(CRT)
+import CRT
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
```
richgillam, line 25: What's all this do?  And do all the import statements import the same symbols?

dra8an:
> > For Chinese years beginning in Gregorian 1901 through 2100, month structure comes from a baked table (200 x UInt32 = 800 bytes) generated from _CalendarICU(.chinese) Each UInt32 packs the 12 or 13 month-length bits (29 vs 30 days), the leap-month index (0 = none), and the new-year offset. In range there is no astronomy at all, a date-to-fields query is a direct index plus a bit extraction.
> 
> Do your own astronomical calculations also match the contents of this table? If so, maybe the comment above should say _that_, rather than saying we match what the old ICU calendar did. If not, why not?

Well, the original PR text has exactly that, but I didn't want the PR text to get that verbose. Those values not only matches the calculations, they are generated by our code.

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
```diff
@@ -0,0 +1,246 @@
...
+        let fixedMidYear = gregorianRataDie(yearInt, 7, 1)
+        let c = (Double(fixedMidYear) - 693596.0) / 36525.0
+        let y2000 = Double(yearInt - 2000)
+        let y1700 = Double(yearInt - 1700)
+        let y1600 = Double(yearInt - 1600)
+        let y1000 = Double(yearInt - 1000) / 100.0
+        let y0 = Double(yearInt) / 100.0
+        let y1820 = Double(yearInt - 1820) / 100.0
+
+        if (2051...2150).contains(yearInt) {
```
richgillam: I think you could re-cast this long list of else-ifs as a `switch` statement and it'd look a little nicer (I don't have what effect it'd have on performance)...

dra8an:
> > ICU's own chnsecal rules layer over a Meeus / Reingold-Dershowitz astronomy engine...
> 
> Some of this is over my head, but two things worry me: 1) Markus has told me that there are some clear deficiencies in `chnsecal.cpp`, to the point where it gets _the wrong answer_ for some date or set of dates. Is that true, and how are we working around it? I kind of thought (but maybe mistakenly) that Peter's lookup table fixed this, but that he didn't fix the underlying astronomy engine. Is that right? 2) Reingold-Dershowitz didn't do us any favors with the Indian calendars. Their code is reliable here? And how much does it overlap with the stuff you had to put together to handle the Indian calendars?

Well, I have a huge set of tests in my fork, which I didn't want to port here. Those are in https://github.com/dra8an/swift-foundation/compare/port/buddhist...port/chinese. I provided the reference sources in the PR and how we match them. 

Reingold-Dershowitz did a good job for the Chinese calendar. I didn't want to port a thousand more lines for astronomical calculations we're not using yet. This subset is small enough to be worth using for Chinese. We can revisit as we add more lunisolar calendars.

This part from may reference to the "wrong answer" you mentioned `The intentional divergences from ICU there stem from ICU's Duffett-Smith astronomy having no delta-T (it errs 25 to 60 minutes and invents leap months in 1794, 1813, and 1889); our engine does not.`

dra8an:
> > ...matches except exactly ICU's three known HKO deviations, which we keep for parity.
> 
> Are we sure we want to do that?

Yeah, I was not sure here. I can do it either way

dra8an:
> You haven't added any reviewers to this PR, and no one's kicked off the CI workflows. Is this PR ready to review?

For some reason I can't add reviewers, or merge the PR, too.

@itingliu , help? :)

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1145 @@
...
+            monthCount: leap == 0 ? 12 : 13,
+            leapDisplay: leap)
+    }
+
+    static func year(relatedISOYear: Int) -> _ChineseYear {
+        let idx = relatedISOYear - tableStart
+        if idx >= 0 && idx < table.count {
+            return decodeTableYear(relatedISOYear: relatedISOYear)
+        }
+        return fallbackCache.withLock { state in
```
richgillam: Maybe it's not possible (or worth it), but it'd be so great if we didn't have to lock anything...

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1145 @@
...
+        let leap = UInt8((v >> 13) & 0xF)
+        return _ChineseYear(
+            relatedISOYear: relatedISOYear,
+            newYearRataDie: _CalendarAstronomy.gregorianRataDie(relatedISOYear, 1, 19) + Int((v >> 17) & 0x3F),
+            monthLengthBits: UInt16(v & 0x1FFF),
+            monthCount: leap == 0 ? 12 : 13,
+            leapDisplay: leap)
+    }
+
+    static func year(relatedISOYear: Int) -> _ChineseYear {
```
richgillam: Maybe it's just too late in the evening for my brain to function, but I couldn't follow what this function did.  I wonder if it'd be worth it to add some comments explaining the general algorithm.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1145 @@
...
+        let (v, add) = eraYears.addingReportingOverflow(year)
+        return add ? nil : v
+    }
+
+    // MARK: Range
+
+    func minimumRange(of component: Calendar.Component) -> Range<Int>? {
+        switch component {
+        case .era: return 1..<83334          // ICU chnsecal LIMITS
+        case .year: return 1..<61
```
richgillam: I can't help thinking a lot of these ranges would read better as inclusive ranges (e.g., `1...60`).

richgillam (approved):
I ran out of both time and energy midway through this review.  I can't say anything about the math, so I mostly read for style.  Things generally look good, as far as I can tell, and I feel like I can trust your testing.

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
parkera: This is basically boilerplate for cross-platform libc imports that exists elsewhere in the project. It's normal.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: These have to be Range<Int> to match the protocol return type; 1...60 would be a ClosedRange. I kept the half-open form for consistency with Gregorian/Hebrew.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
parkera: I'm also curious about this cache and lock. When is it populated? Is it when we don't hit the prebuilt data in the `static let`? Is it important to cache because the values will be immediately used again?

parkera: Also, how big does that cache get, and is there a need to clear it at some point?

parkera: One more thing - if we do decide to keep the cache, we should keep the critical section as small as possible.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1145 @@
...
+            var bits: UInt16 = 0
+            var leapDisplay: UInt8 = 0
+            for (i, s) in starts.enumerated() {
+                let next = (i + 1 < starts.count) ? starts[i + 1] : nyNext
+                assert(next - s == 29 || next - s == 30, "non-lunation month length \(next - s) in fallback year \(relatedISOYear)")
+                if next - s == 30 { bits |= UInt16(1) << i }
+                let label = state.rules.monthLabel(startingAt: s, gregorianYear: _CalendarAstronomy.gregorianYear(ofRataDie: s))
+                if label.isLeap { leapDisplay = UInt8(label.month) }
+            }
+            let year = _ChineseYear(
```
parkera: Obligatory comment here to unwrap these arbitrarily 80-column limited function calls. If it makes sense to wrap for some reason manually, go ahead - but there is no reason to wrap it like this.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1145 @@
...
+        case .day: return 1..<31
+        case .hour: return 0..<24
+        case .minute: return 0..<60
+        case .second: return 0..<60
+        case .weekday: return 1..<8
+        case .weekdayOrdinal: return -1..<6
+        case .quarter: return 1..<5
+        case .weekOfMonth: return 1..<7
+        case .weekOfYear: return 1..<56
+        case .yearForWeekOfYear: return Self.extendedYearLowerBound..<(Self.extendedYearUpperBound + 1)
```
parkera: Curious why this isn't the same as the `.year` value? (More of a question than a comment)

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Added a doc comment covering the table vs the computed paths and the new moon math.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: `.year` is the cycle year (1…60), which wraps every 60 years. `.yearForWeekOfYear` has to be an absolute year value so `intervals` and `date(byAdding:)` can step across years. The cycle year can't do that. So those APIs use the extended (continuous) year instead, which is `relatedGregorianYear + 2637`.  This is working the  same as ICU's EXTENDED_YEAR.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Unwrapped.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Reworked cache. It's populated only for range outside of the baked table, pre 1901 and post 2100. In range dates read the baked table and never touch it. It's now bounded to 16 entries (evictIfNeeded), so it never needs clearing. Reworked so the astronomy computes lock free with a local rules and the lock now only guards the cache read/write. Critical section is just two dictionary ops.

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
dra8an: Done, converted it to a switch.

richgillam:
> > > ...matches except exactly ICU's three known HKO deviations, which we keep for parity.
> > 
> > 
> > Are we sure we want to do that?
> 
> Yeah, I was not sure here. I can do it either way

I don't have any idea either, and there are good arguments either way.  I wonder if there's anyone who can give us a more informed answer here?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
richgillam: Yeah, I keep forgetting they're different types.  It's kind of unfortunate, because for a lot of these things, a closed range is more readable (months as `1...12` rather than `1..<13`), but sticking a cast in there to get a closed range would screw up readability even worse, and I think you're right about not wanting to change the actual type you're storing.  Unless the Foundation people have a better way, forget I said anything.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
richgillam: Is that in the most recent commit?  I'm only seeing documentation changes around the caching stuff.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
parkera: I think you can just do `Range(1...60)`, if you think that's more readable or useful for some reason. I suspect it'll optimize to the same code.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Duh! I could swear I had committed those changes. Should be in now

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    static func gregorianYear(ofRataDie day: Int) -> Int {
+        var y = Int((Double(day) / 365.2425).rounded(.down)) + 1
+        while gregorianRataDie(y, 1, 1) > day { y -= 1 }
+        while gregorianRataDie(y + 1, 1, 1) <= day { y += 1 }
+        return y
+    }
+
+    // Dynamical-minus-universal time (fraction of a day). Meeus/NASA fits.
```
itingliu: can you explain what this means?

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
```diff
@@ -0,0 +1,247 @@
...
+    ]
+    private static let solarMultipliers: [Double] = [
+        0.9287892, 35999.1376958, 35999.4089666, 35998.7287385, 71998.20261, 71998.4403,
+        36000.35726, 71997.4812, 32964.4678, -19.4410, 445267.1117, 45036.8840, 3.1008,
+        22518.4434, -19.9739, 65928.9345, 9038.0293, 3034.7684, 33718.148, 3034.448,
+        -2280.773, 29929.992, 31556.493, 149.588, 9037.750, 107997.405, -4444.176, 151.771,
+        67555.316, 31556.080, -4561.540, 107996.706, 1221.655, 62894.167, 31437.369,
+        14578.298, -31931.757, 34777.243, 1221.999, 62894.511, -4442.039, 107997.909,
+        119.066, 16859.071, -4.578, 26895.292, -39.127, 12297.536, 90073.778,
+    ]
```
itingliu: These are actually a tuple right? like they are actually pairs of each other?

Would it make sense to represent them as this, or is it more expensive memory-wise/pref-wise to represent them this way?
```
private static let solarConstants = [
    (403406, 270.54861, 0.9287892),
...
]
```

If it's more expensive, maybe we can add a comment specifying the invariants (that they all need to be aligned with each other and they are expected to have the same count). I'd also like to know how it is more expensive if that is really the case

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+    }
+}
+
+// MARK: - Engine: baked table + computed fallback
+
+internal enum _ChineseCalendarEngine {
+    // One entry per Chinese year, indexed by the Gregorian year in which that Chinese year begins.
+    // Packing: bits 0-12 month lengths (1=30d), 13-16 leap display number (0=none), 17-22 new-year offset from Jan 19 of that Gregorian year.
+    static let tableStart = 1901
+    static let table: [UInt32] = [
```
itingliu: This should be a private inline array, and we should be accessing it with a helper that returns a Span so it doesn't copy

```swift
let arr: InlineArray<3, StaticString> = [
    "apple",
    "banana",
    "cherry",
]

// Returns a Span<UInt8> over the UTF-8 bytes of arr[index].
@_lifetime(immortal)
package func search(_ index: Int) -> Span<UInt8> {
    let v = arr[index]
    let span = unsafe Span(_unsafeStart: v.utf8Start, count: v.utf8CodeUnitCount)
    return unsafe _overrideLifetime(span, copying: ())
}
```

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: That looks better indeed. I will replace all open halves with this contract, looks more readable,

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
itingliu: This applies to all these costs in Calendar_Astronomy

```
private static let solarAddends: [Double] 
```

dra8an: I replaced it with InlineArray. As our elements are scalar, I don't think those will be copied. Made a quick SIL analysis:

The optimized SIL for a subscript on a static let InlineArray<200, UInt32>:

  ```
  %8  = global_addr @…table… : $*InlineArray<200, UInt32>   // address of the static storage
    %16 = struct_element_addr %8, #InlineArray._storage        // into its storage
    %18 = vector_base_addr %16                                 // base of the vector
    %19 = index_addr [projection] %18, %17                     // address of element i
    %20 = load %19                                             // load ONE UInt32
    return %20
```
    
This looks like a pure address projection + single load. There is no copy_addr, no alloc_stack, no memcpy of the array anywhere. The 800-byte array is addressed in place in its global storage.
    
Calendar_Astronomy InlineArray may indeed be the different story, will put that in the next commit

File: Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift (hunk 9 above)
Comment (dra8an) [reply Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift]: They are aligned parallel arrays, term i uses coefficient[i], addend[i], and multiplier[i] together, so grouping them into one array of  tuples is a good idea. It is not more expensive, it is actually a little better. It becomes one contiguous allocation instead of three, the three values for a term sit next to each other which matches exactly how the loop reads them, and the tuple structurally enforces the same count and alignment invariant you mentioned. Combined with InlineArray it lives in constant storage with no heap allocation at all. I am switching the solar and new moon series to this form in the next commit.

File: Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift (hunk 8 above)
Comment (dra8an) [reply Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift]: Sure, but last time I got scolded for being too verbose :)

This converts between two time scales. Astronomy math runs on dynamical time, which is uniform, but calendars run on universal time, which follows the Earth's actual rotation. The two drift apart because the Earth's rotation is irregular and slowly slowing, and the gap is called delta-T. ephemerisCorrection returns that gap as a fraction of a day so we can convert before computing new moons and solstices. Without it, event times would be off by up to an hour for far past or future dates, which is the flaw  in ICU's Duffett-Smith astronomy that has no delta-T.

On source: these are the Espenak and Meeus polynomial expressions for delta-T, published by NASA Goddard (Fred Espenak's eclipse site). They are published polynomial fits to historical delta-T observations. What we use are the coefficients and formulas, which are mathematical facts and published scientific data, and the primary publication is a NASA work in the public domain.

File: Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift (hunk 9 above)
Comment (dra8an) [reply Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift]: > These are actually a tuple right? like they are actually pairs of each other?
> 
> Would it make sense to represent them as this, or is it more expensive memory-wise/pref-wise to represent them this way?
> 
> ```
> private static let solarConstants = [
>     (403406, 270.54861, 0.9287892),
> ...
> ]
> ```
> 
> If it's more expensive, maybe we can add a comment specifying the invariants (that they all need to be aligned with each other and they are expected to have the same count). I'd also like to know how it is more expensive if that is really the case

Done in the latest commit. The solar series is now one InlineArray<49, (coefficient, addend, multiplier)>, the new-moon series is two tuple InlineArrays, and polynomial is generic over the InlineArray count so the ephemerisCorrection coefficient literals are InlineArray rather than [Double].

This is how I checked for heap allocations: I emitted optimized SIL (swiftc -O -emit-sil) for the call sites and looked for _allocateUninitializedArray / alloc_ref, which is what an Array literal lowers to. The InlineArray version has none.

This is interesting, though: I also checked the old [Double] version at -O, and it had no heap allocation either, it seems that the optimizer already promotes those small constant literals. I benchmarked an out-of-range path that exercises this heavily (pre-1901 dates, which force the astronomy fallback and the polynomial branches) and old vs new were identical, 8 mallocs and about 90 microseconds per call. So this isn't a speedup, it makes the no-allocation property explicit in the type instead of relying on the optimizer, and it gives the tuple grouping you suggested with the alignment enforced by construction.

**Comment by itingliu** [comment]
@swift-ci please test

File: Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift [hunk 11]
```diff
@@ -0,0 +1,1146 @@
...
+// Chinese lunisolar calendar engine. Years 1901-2100 come from a baked table generated from ICU (parity by construction); outside that range, month structure is computed with ICU's chnsecal rules over _CalendarAstronomy at UTC+8.
+
+// Bounded memoization caches: over the capacity, evict one arbitrary entry (hash order). LRU is deliberately not attempted, a miss only recomputes.
+private func evictIfNeeded<V>(_ cache: inout [Int: V], capacity: Int) {
+    if cache.count > capacity, let victim = cache.keys.first {
+        cache.removeValue(forKey: victim)
+    }
+}
+
+// MARK: - chnsecal rules over the astronomy (flat UTC+8, matching ICU)
```
itingliu: What does "flat UTC+8" mean?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+        Double(rataDie) - 1.0 + 16.0 / 24.0
+    }
+
+    private func toLocalDay(_ moment: Double) -> Int {
+        Int((moment + 8.0 / 24.0).rounded(.down))
+    }
+
+    // Winter solstice day on or after Dec 1 (chnsecal winterSolstice).
+    mutating func winterSolstice(_ gregorianYear: Int) -> Int {
+        if let cached = winterSolsticeCache[gregorianYear] { return cached }
```
itingliu: this is not locked. I think it's ok because you're only calling _ChineseRules locally in static `func year(relatedISOYear: Int) -> _ChineseYear`. But do we really need to declare `_ChineseRules` as a `internal` struct if it really is just a helper inside a function? 

I'm concerned with that for two reasons
1. Using a struct as an intermediary helper might have discreet performance or memory problems depending on how `func year(relatedISOYear: Int) -> _ChineseYear` is called
2. We might misuse `_ChineseRules` in other places where it's not meant to be used. 

Can we revisit again if we really need this intermediary struct, and if so, perhaps make it a `fileprivate` or even a `private` struct, declared closer to the callsite?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+import CRT
+#elseif os(WASI)
+@preconcurrency import WASILibc
+#endif
+
+internal import Synchronization
+
+// Chinese lunisolar calendar engine. Years 1901-2100 come from a baked table generated from ICU (parity by construction); outside that range, month structure is computed with ICU's chnsecal rules over _CalendarAstronomy at UTC+8.
+
+// Bounded memoization caches: over the capacity, evict one arbitrary entry (hash order). LRU is deliberately not attempted, a miss only recomputes.
```
itingliu: what does "memoization" mean?

itingliu: also, is there really not anything from stdlib or swift-collections that we can reuse here? must we implement our own cache that also arbitrarily evicts stuff?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+    0x0026052D, 0x0010CAAD, 0x00360AB5, 0x002205AA, // 2073-2076
+    0x000A8BA5, 0x00300DA5, 0x001C0D4A, 0x00067C95, // 2077-2080
+    0x002A0C96, 0x0014F94E, 0x003A0556, 0x00240AB5, // 2081-2084
+    0x000EB5B2, 0x003406D2, 0x001E0EA5, 0x000A8E4A, // 2085-2088
+    0x002C068B, 0x00170C97, 0x003C04AB, 0x0026055B, // 2089-2092
+    0x0010CAD6, 0x00360B6A, 0x00220752, 0x000C9725, // 2093-2096
+    0x00300B45, 0x001A0A8B, 0x0004549B, 0x002A04AB, // 2097-2100
+    ]
+
+    // Cross-instance memoization of computed out-of-range years only (pre-1901 / post-2100); in-range dates read the baked table and never touch this. Bounded to 16 entries, so it never needs clearing.
```
itingliu: I do not understand this comment. What is `memoization`? What is the key? when do we never touch this? Where does "16 entries" come from and what does it have to with clearing since it is just a dictionary?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: It means the day is at a fixed +8 hour offset with no daylight saving and no historical time-zone changes. I will rewrite the comment to make it more clear

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: It just means a small cache of results we've already computed. I've reworded the comment to say that plainly and to explain the eviction.

The standard library has no bounded/evicting cache as far as I can tell. This is a tiny dictionary with a hard cap so it can't grow without bound. The eviction is deliberately simple, over the cap we drop one entry, and if it's needed again we recompute it (cheap, and only for the rare out of range years). A real LRU would be more code and more state for no practical gain at this size. I've reworded the comment here, too.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Reworded it to spell out the key (the extended year), the value (that year's computed month structure), why in-range dates never touch it (they read the baked table), and where the 16 comes from (the cap, so it stays small enough to never clear).

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: `_ChineseRules` is only used in this file, inside `year(relatedISOYear:)`, so I've made it `fileprivate`. Nothing outside the file can reach it now, which removes the misuse risk.

On "not locked": each `_ChineseRules` value is built locally in a single `year()` call and never shared, so its two small caches don't need locking. The cache that *is* shared across calls (`fallbackCache`) is guarded by a `Mutex`.

On cost: we only build one for years outside the 1901 to 2100 table, which is the rare path. In-range dates read the baked table and never create one. It's a stack value with no heap allocation; it just groups the two per-year lookups with the code that fills them so `year()` stays readable. I can inline it into the function instead if you'd prefer.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+        let newMoon2 = newMoonNear(newMoon1 + Self.synodicGap, true)
+        let newMoon11 = newMoonNear(solsticeAfter + 1, false)
+        let value: Int
+        if Self.synodicMonthsBetween(newMoon1, newMoon11) == 12 &&
+            (hasNoMajorSolarTerm(newMoon1) || hasNoMajorSolarTerm(newMoon2)) {
+            value = newMoonNear(newMoon2 + Self.synodicGap, true)
+        } else {
+            value = newMoon2
+        }
+        evictIfNeeded(&newYearCache, capacity: 32)
```
itingliu: I still have questions about the cache eviction throughout this file. Why 32?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+        return nil
+    }
+}
+
+// MARK: - Engine: baked table + computed fallback
+
+internal enum _ChineseCalendarEngine {
+    // One entry per Chinese year, indexed by the Gregorian year in which that Chinese year begins.
+    // Packing: bits 0-12 month lengths (1=30d), 13-16 leap display number (0=none), 17-22 new-year offset from Jan 19 of that Gregorian year.
+    static let tableStart = 1901
```
itingliu: This should be called something like "calendarStartYearInGregorian" or something like that. Otherwise the reader would be looking at "tableStart" and be confused about "what table? start of what?"

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+        }
+        return nil
+    }
+}
+
+// MARK: - Engine: baked table + computed fallback
+
+internal enum _ChineseCalendarEngine {
+    // One entry per Chinese year, indexed by the Gregorian year in which that Chinese year begins.
+    // Packing: bits 0-12 month lengths (1=30d), 13-16 leap display number (0=none), 17-22 new-year offset from Jan 19 of that Gregorian year.
```
itingliu: Can you write out how to read the value? Also, what is "leap display number"? 

For example, the first entry `0x003E0752` is 
```
// high bit ---------------   low 
0011 1110 0000 0111 0101 0010
```
so bits 0 - 12: `0 0111 0101 0010`. I assume that means month 2, 5, 7, 9, 10, 11 are big months, and the rest are small months?

bits 13-16 are all 0, so I guess that means there's no leap month in this year? but if say the bits are `0010`, then it would mean that there is a leap month 2?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+    0x0026052D, 0x0010CAAD, 0x00360AB5, 0x002205AA, // 2073-2076
+    0x000A8BA5, 0x00300DA5, 0x001C0D4A, 0x00067C95, // 2077-2080
+    0x002A0C96, 0x0014F94E, 0x003A0556, 0x00240AB5, // 2081-2084
+    0x000EB5B2, 0x003406D2, 0x001E0EA5, 0x000A8E4A, // 2085-2088
+    0x002C068B, 0x00170C97, 0x003C04AB, 0x0026055B, // 2089-2092
+    0x0010CAD6, 0x00360B6A, 0x00220752, 0x000C9725, // 2093-2096
+    0x00300B45, 0x001A0A8B, 0x0004549B, 0x002A04AB, // 2097-2100
+    ]
+
+    // A shared cache of the month structure computed for years outside the baked table (before 1901 or after 2100). Key: the extended year. Value: that year's computed structure. In-range dates (1901-2100) read the baked table directly, so they never reach this cache. It is capped at 16 entries (over that, `evictIfNeeded` drops one), which keeps it small enough that we never need to clear it.
```
itingliu: Again I'm confused about how this 16-entry cap is decided. Why 16, why not 32, or any other number?

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -0,0 +1,169 @@
...
+#if canImport(FoundationEssentials)
+@testable import FoundationEssentials
+#else
+@testable import Foundation
+#endif
+
+@Suite("Chinese Calendar")
+private struct ChineseCalendarTests {
+
+    private static func cal() -> _CalendarChinese {
```
itingliu: Again can we please give this function a meaningful name? Otherwise I'd be wondering "call what"

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+            (est, y, ordinal) = Self.nextMonthStart(after: y, ordinal: ordinal)
+        }
+        let lbl = y.label(ordinal: ordinal)
+        if lbl.month != display || lbl.isLeap != leap {
+            (est, y, ordinal) = Self.nextMonthStart(after: y, ordinal: ordinal)
+        }
+        return est
+    }
+
+    private static func nextMonthStart(after y: _ChineseYear, ordinal: Int) -> (Int, _ChineseYear, Int) {
```
itingliu: Can you explain what the return values of this function mean by either adding a comment or giving them labels?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+        var starts = [ny]
+        var cur = ny
+        while true {
+            let nxt = rules.newMoonNear(cur + _ChineseRules.synodicGap, true)
+            if nxt >= nyNext { break }
+            starts.append(nxt)
+            cur = nxt
+        }
+        // Pack month lengths (30-day months set a bit) and record the leap month, if any.
+        var bits: UInt16 = 0
```
itingliu: Also on descriptive name: should it be called `"var monthLengths"` or something like that?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+}
+
+// MARK: - Year structure
+
+internal struct _ChineseYear: Sendable {
+    let relatedISOYear: Int
+    let newYearRataDie: Int
+    let monthLengthBits: UInt16    // bit i set = ordinal month i+1 has 30 days
+    let monthCount: UInt8          // 12 or 13
+    let leapDisplay: UInt8         // 0 = none; else leap month repeats this number
```
itingliu: On descriptive name: What does "display" mean in this context?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -0,0 +1,1146 @@
...
+
+    var daysInYear: Int {
+        var days = 0
+        for i in 1...Int(monthCount) { days += monthLength(ordinal: i) }
+        return days
+    }
+
+    var endRataDie: Int { newYearRataDie + daysInYear }
+
+    func label(ordinal: Int) -> (month: Int, isLeap: Bool) {
```
itingliu: On descriptive name: label of what...?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Named it `chineseRuleCacheCapacity` with a comment. The number is not tuned. Over the cap we drop one entry and recompute it later, which is cheap, so any small bound works. 32 comfortably covers the handful of neighboring years one fallback computation touches.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed to `firstTableYear`. I did not use `calendarStartYearInGregorian` because the calendar is not limited to 1901; that is only where the baked table starts, and earlier years are computed. So the name refers to the table.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Rewrote the comment with a full breakdown and a worked example. Your reading is correct: for 0x003E0752, months 2, 5, 7, 9, 10, and 11 are 30-day months, and bits 13-16 = 0 means no leap month. And yes, 0010 would mean a leap month sharing number 2. I also renamed "leap display number" to "leap month number".

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Named it `fallbackCacheCapacity` and explained it. Same idea as the other cap: dropped entries are recomputed cheaply, so the exact value is not critical. It is smaller than the per-computation cap because out-of-range lookups are rare and usually clustered, so a small working set is enough.

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Renamed to chineseCalendar()

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Added labels: the function now returns (rataDie:, year:, ordinal:), plus a one-line comment

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed `leapDisplay` to `leapMonthNumber` and clarified the comment. 0 means no leap month; otherwise the leap month shows as this month number.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed to `monthLabel` and added a comment. It returns the displayed month number and leap flag for a given ordinal (1-based) position in the year.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed the local to monthLengthBits to match the field it fills

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
```diff
@@ -129,67 +129,113 @@ internal enum _CalendarAstronomy {
...
-    private static let solarMultipliers: [Double] = [
-        0.9287892, 35999.1376958, 35999.4089666, 35998.7287385, 71998.20261, 71998.4403,
-        36000.35726, 71997.4812, 32964.4678, -19.4410, 445267.1117, 45036.8840, 3.1008,
-        22518.4434, -19.9739, 65928.9345, 9038.0293, 3034.7684, 33718.148, 3034.448,
-        -2280.773, 29929.992, 31556.493, 149.588, 9037.750, 107997.405, -4444.176, 151.771,
-        67555.316, 31556.080, -4561.540, 107996.706, 1221.655, 62894.167, 31437.369,
-        14578.298, -31931.757, 34777.243, 1221.999, 62894.511, -4442.039, 107997.909,
-        119.066, 16859.071, -4.578, 26895.292, -39.127, 12297.536, 90073.778,
+    private static let solarTerms: InlineArray<49, (coefficient: Double, addend: Double, multiplier: Double)> = [
+        (403406, 270.54861, 0.9287892),
```
richgillam, line 133: Would it be useful to have named tuple members here like you do in `newMoonTerms` below?  Seems like it...

For that matter, it might be worth having a small block comment above each of these arrays explaining what it is and where you got the numbers. (I think you have that information in here somewhere already, but it might be nice to have it right on top of each array definition.)

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -28,115 +28,120 @@ internal import Synchronization
 
 // Chinese lunisolar calendar engine. Years 1901-2100 come from a baked table generated from ICU (parity by construction); outside that range, month structure is computed with ICU's chnsecal rules over _CalendarAstronomy at UTC+8.
 
-// Bounded memoization caches: over the capacity, evict one arbitrary entry (hash order). LRU is deliberately not attempted, a miss only recomputes.
+// Small caches of already-computed results, capped at `capacity`. Over the cap, drop one entry (whichever the dictionary lists first). We skip least-recently-used tracking on purpose: a dropped entry is just recomputed next time, and that recompute is cheap, so a smarter cache or a new dependency is not worth it.
 private func evictIfNeeded<V>(_ cache: inout [Int: V], capacity: Int) {
     if cache.count > capacity, let victim = cache.keys.first {
         cache.removeValue(forKey: victim)
```
richgillam: Isn't there a removeFirst() or something like that that you could use instead of doing it in two steps? (Hmmm... maybe not for a dictionary...)

itingliu:
@swift-ci please test

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
itingliu: haha, fair. This probably deserves to live inside a separate document about how this entire thing works

itingliu (approved):

Sources/FoundationEssentials/Calendar/Calendar_Astronomy.swift
dra8an: I'll be glad to provide it! Where is the best place to put that document?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -1156,3 +996,133 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+
+    private static func decodeTableYear(relatedISOYear: Int) -> _ChineseYear {
+        let v = table[relatedISOYear - firstTableYear]
+        let leap = UInt8((v >> 13) & 0xF)
+        return _ChineseYear(relatedISOYear: relatedISOYear, newYearRataDie: _CalendarAstronomy.gregorianRataDie(relatedISOYear, 1, 19) + Int((v >> 17) & 0x3F), monthLengthBits: UInt16(v & 0x1FFF), monthCount: leap == 0 ? 12 : 13, leapMonthNumber: leap)
+    }
+
+    /// Month structure for the Chinese year whose New Year falls in Gregorian `relatedISOYear`.
+    ///
+    /// In the baked range (1901...2100) this is a direct table decode. Outside it the structure is computed from astronomy and memoized: locate this year's New Year and the next year's, walk the new moons between them to get each month's first day, record 29- vs 30-day lengths as bits, and mark the leap month (the month carrying no major solar term). The seam years at the table edges reuse the table's own New Year so the computed and baked spans tile exactly.
```
richgillam, line 1072: You say "Outside it the structure is computed from astronomy and memoized".  But you just took out all the memoization logic, didn't you?

itingliu:
@swift-ci please test

PR #2126, opened 2026-07-23
Correct BOM detection for files with unknown encoding

jmschonfeld:
Fixes a bug in BOM detection logic when decoding file contents as a String for files with unknown encoding

### Motivation:

When decoding a file, if the file has an xattr for the encoding we decode as that encoding, otherwise we use BOM detection to select an encoding. The BOM detection included a few logic bugs around reading from incorrect indices.

### Modifications:

Updates the BOM detection logic to use modern span-based APIs to avoid indexing issues

### Result:

BOM detection behaves as expected

Resolves rdar://182478045

### Testing:

Added a unit test to validate the behavior (unit testing previously only covered cases where the file contained the xattr)

### DIFF
--- a/Sources/FoundationEssentials/String/String+IO.swift
+++ b/Sources/FoundationEssentials/String/String+IO.swift
@@ -279,31 +279,31 @@ extension String {
 
 extension String {
     internal init?(dataOfUnknownEncoding data: Data, usedEncoding: inout Encoding) {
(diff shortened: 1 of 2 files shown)


jmschonfeld:

parkera (approved):

glessard (approved):

PR #2131, opened 2026-07-25
`Decimal`: further performance and correctness improvements

xwu:
As a follow-on to #2022, a second tranche of performance and correctness improvements for the `Foundation.Decimal` type.

### Motivation:

`Decimal` operations that leave substantial room for optimization encourage the erroneous impression that decimal floating-point is intrinsically much less performant than it could be as compared to alternative numeric representations. This PR makes a further push into the implementation of that type.

### Modifications:

Major changes here that iterate on work started in #2022 are:
* Elimination of (most) iterative scaling by factors of 10 or 1e4 for at-most-two-shot scaling by estimation (and, consequently, generalization of some `UInt128` helpers to all powers of 10)
* Optimization of `compact()` with a different approach for divisibility by 10 without computing remainder.

Major new efforts here include:
* A reworking of string conversion (preserving quirks but Unicode correct!)
* A reworking of conversion to integers (adopted by `JSONDecoder`—fixes a decoding bug!)
* Restoration of correct rounding to a given scaling factor for arithmetic operations (only internal for now)
* Proper hashing

For the sake of reviewability, I have deferred `Decimal` to/from `Double` conversion work [to a future PR](https://github.com/xwu/swift-foundation/compare/decimal-performance-2...xwu:swift-foundation:decimal-performance-3?expand=1).

### Result:

On its own, performance improvements are as follows:
| Benchmark | Time (wall clock) (ns, baseline &rarr; revised, p50) | Throughput (improvement %, p50) |
| ---- | ---: | ---: |
| Add | 18 &rarr; 10 | 85 |
| Compare | 7 &rarr; 7 | -6 |
| Divide | 19 &rarr; 11 | 81 |
| Hash | 44 &rarr; 36 | 17 |
| Multiply | 17 &rarr; 13 | 31 |

_Cumulative_ performance improvements, when stacked onto #2022, are as follows (from a prior run):

| Benchmark | Time (wall clock) (ns, baseline &rarr; revised, p50) | Throughput (improvement %, p50) |
| ---- | ---: | ---: |
| Add | 111 &rarr; 10 | 1033 |
| Compare | 32 &rarr; 7 | 350 |
| Divide | 1928 &rarr; 10 | 10100 |
| Hash | 44 &rarr; 37 | 17 |
| Multiply | 169 &rarr; 13 | 1167 |

(Excluding added tests, the cumulative change in lines of code for these two PRs is ~zero.)

I've also run [ordo-one benchmarks](https://github.com/ordo-one/FixedPoint/tree/main/Benchmarks), which favor narrow significands that fit in their fixed-point type, to track cumulative progression (the most recent branch tested is the one linked to above that incorporates conversion work, so I'm excluding the benchmarks that relate to those unincorporated changes):

| Benchmark | &nbsp;baseline `20497148cd` | &nbsp;ordo-one `bd8b276aa0` | xwu (main) `af8c8086e9` | &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;xwu `4ad0e0091a` | xwu | **Total** |
|---|---:|---:|---:|---:|---:|---:|
| addition | 454 | 73 | 19 | 14 | 5.2× | **32.4×** |
| subtraction | 559 | 88 | 23 | 19 | 4.6× | **29.4×** |
| multiplication | 1176 | 367 | 67 | 48 | 7.6× | **24.5×** |
| division | 2410 | 915 | 124 | 93 | 9.8× | **25.9×** |
| comparison | 585 | 84 | 20 | 15 | 5.6× | **39.0×** |
| equality | 584 | 84 | 21 | 16 | 5.2× | **36.5×** |
| hash | 493 | 119 | 35 | 32 | 3.7× | **15.4×** |
| description | 1963 | 635 | 522 | 46 | 13.8× | **42.7×** |
| rounded(scale:) | 1068 | 172 | 51 | 28 | 6.1× | **38.1×** |

### Testing:

All existing and new unit tests pass (with modifications to account for now-corrected behavior). As before, any changes required for the framework build—if any—are TBD.

[Timestamps for each commit are when they were rebased on `main`, some of which required manual editing.]


Tests/FoundationEssentialsTests/DecimalTests.swift
```diff
@@ -258,7 +258,7 @@ private struct DecimalTests {
         #expect(zeroE.isZero)
         zeroE = try #require(Decimal(string: "e"))
         #expect(zeroE.isZero)
-        // Partitally valid strings ending with e shold be parsed
+        // Partitally valid strings ending with e should be parsed
```
xwu:
```suggestion
        // Partially valid strings ending with e should be parsed
```

parkera:
Thanks so much for this contribution @xwu. Bringing in @stephentyrone for a review on this one

xwu:
Thanks. @stephentyrone—let me know if it would make more sense to review the binary floating-point conversion work here in one go (which I separated out and linked to above). If so, I can cherrypick those commits over.

(There's a non-trivial policy decision there—which we don't have to get into here if I don't cherrypick—regarding what the actual semantics should be for `Double` to `Decimal` conversion; the existing implementation was not precise enough for a decision to be made about it. I opted for _shortest_ representation ties-to-even, guaranteeing `Decimal(string: String(double)) == Decimal(double)`, to better accord (I think) with user expectations—i.e.: `Decimal(string: "0.1") == Decimal.init(0.1)`. This is, of course, distinct from the _nearest_ decimal representation, which also happens to be somewhat annoying to implement given that the mantissa is a non-integral number of decades.)

stephentyrone:
Let's keep it separate for now.

I _think_ we probably actually want to provide multiple inits from Double; @tbkka and I were discussing this a few weeks back in another context, and we think it makes sense to have both nearest and shortest forms available. That becomes an API proposal, so let's keep that discussion out of this PR.

xwu:
> That becomes an API proposal, so let's keep that discussion out of this PR.

Sounds good. It's now added to my [laundry list of API changes](https://gist.github.com/xwu/58b05fc102a79c0d0361b10a619ebaad).

Sources/FoundationEssentials/Decimal/Decimal.swift
xwu, line 80: Incidentally—and I might be incorrect—this layout is meant to copy that of the framework type defined in C, but it only does so if `0...7` is read as LSB to MSB (which was likely the intention). However, that is not how the accessors below actually work (for example, `_length` right shifts by 4 rather than masking). The 'blast radius' for any changes made here is actually minimal, if we want to restore parity.

Sources/FoundationEssentials/Decimal/Decimal.swift
```diff
@@ -23,6 +23,54 @@ import ucrt
 /// A structure representing a base-10 number.
 @available(macOS 10.10, iOS 8.0, watchOS 2.0, tvOS 9.0, *)
 public struct Decimal: Sendable {
+    // Some notes on `Decimal` storage
+    // -------------------------------
```
xwu, line 27: These notes reflect investigations into the _historical_ behavior of the type, preserved as fully as possible for compatibility purposes.

Sources/FoundationEssentials/Decimal/Decimal.swift
```diff
@@ -23,6 +23,54 @@ import ucrt
...
+    // bit unset without being malformed. Arithmetic operations accept arguments
+    // that are non-compact (and actually compact but with the bit unset).
+    // Whenever reasonable, arithmetic operations *should* return compact values
+    // with the bit properly set.
+    //
+    // *A value that is not actually 'compact' is malformed if the `_isCompact`
+    // bit is erroneously set*. Operations may exhibit unspecified behavior when
+    // passed such values as arguments. For example: two values, one of which is
+    // malformed in this way, may compare equal but hash differently. Public
+    // APIs exposed on the type *should never* produce such values.
```
xwu, line 41: (Violated by the compatibility path of `init(sign:exponent:significand:)` and by the memberwise initializer—neither of which is publicly available in the non-framework build.)

Sources/FoundationEssentials/Decimal/Decimal.swift
```diff
@@ -23,6 +23,54 @@ import ucrt
...
+    // A value with `_length != 0` is nonzero and finite. In such cases, _all_
+    // of the mantissa's 128 bits are part of the value. To re-emphasize the
+    // point: unlike IEEE decimal types, _all_ such values are regarded as
+    // _nonzero_. Arithmetic operations accept arguments with `_length` nonzero
+    // but 'too short' _without_ masking the mantissa bits. For example,
+    // `_length == 1` and `_mantissa = (0xffff, 0xffff, 0, 0, 0, 0, 0, 0)` is
+    // treated as though `_length == 2`. Arithmetic operations *should* return
+    // values with properly set `_length`; that is, (at least) the number of
+    // 16-bit limbs required to include the most significant nonzero bit in the
+    // mantissa.
```
xwu, line 61: An open question is whether the modern implementation should just set `_length == 8` for all nonzero values, since this field has historically (and now) only been tested for zero or nonzero, saving some work that's only observable when the bit pattern is serialized.

Sources/FoundationEssentials/Decimal/Decimal.swift
```diff
@@ -23,6 +23,54 @@ import ucrt
...
+    //
+    // The behavior of (malformed) values with `_length != 0` and `_mantissa ==
+    // (0, 0, 0, 0, 0, 0, 0, 0)` is undocumented but rationalizable. In
```
xwu, line 64: (Amazingly, the original implementation was extremely consistent about these behaviors, totally uncommented and across various strata of implementation history.)

xwu:
Gentle ping :)

xwu:
Ping again :)

parkera (approved):

parkera:
Thanks @xwu !

xwu:
Thanks for merging, @parkera! Please let me know if you notice anything is needed for the framework build.

PR #2132, opened 2026-07-28
Correct the timezone class in timezone cache

itingliu:
Before, we used `_TimeZoneICU` for timezones such as "GMT+8" for bridged timezone. In a previous commit, we wanted to switched the FOUNDATION_FRAMEWORK to `_TimeZoneGMTICU` to follow the Swift timezone path. However I incorrectly created a `_TimeZoneGMT` instead. This caused the -[localizedName:locale:] to return nil for the affected timezones.

Consolidate the code to use the dynamic func `_timeZoneGMTClass()` everywhere, which resolves to `_TimeZoneGMTICU` on Darwin, for extra benefit to remove some duplicated code. Also update the f`unc localizedName(for style: TimeZone.NameStyle, locale: Locale?)` implementation to match what `_TimeZoneICU` does.

While we're here, change to use the dynamic `_timeZoneICUClass()` for `_timeZoneICU` too.

### DIFF
--- a/Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
+++ b/Sources/FoundationEssentials/TimeZone/TimeZone_Cache.swift
@@ -360,21 +360,17 @@ struct TimeZoneCache : Sendable, ~Copyable {
             }
 
             let bridgedTZ: _NSSwiftTimeZone?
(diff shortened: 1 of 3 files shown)


itingliu:
@swift-ci please test

parkera (approved):

PR #2137, opened 2026-07-29
Make _FoundationInternationalizationData a Swift library

jmschonfeld:
Make _FoundationInternationalizationData a Swift library

### Motivation:

We've determined how we can declare the data in Swift rather than C, so make this a Swift library instead of a C library

### Modifications:

- Removes all C headers / module maps
- Replaces C implementation files with swift files
- Updates CMake/SwiftPM configs to account for the different library type

### Result:

The library is built as Swift source rather than C source

### Testing:

Confirmed the CMake/SwiftPM builds succeed, will perform a full toolchain build to verify


finagolfin:
It would be great if we could also eventually add support for partitioning this data, so app devs that really care about their app's size and want to go to that level of detail, can build this library themselves and fine-tune what's in it.

jmschonfeld:
> It would be great if we could also eventually add support for partitioning this data, so app devs that really care about their app's size and want to go to that level of detail, can build this library themselves and fine-tune what's in it.

That's the direction we're heading with this library - for now we're just establishing the library as a "normal" library/dependency as a place to start building up the Foundation-designed data but in the future the goal is that it can be sliced at client build time

parkera (approved):

itingliu (approved):
Thank you!

PR #2138, opened 2026-07-30
Add missing import for Darwin Extras

jmschonfeld:
Adds an extra import

### Motivation:

In the environment for this build configuration, we need an extra import that was previously not present

### Modifications:

Adds the missing import statement

### Result:

Source builds successfully that required this import

### Testing:

Verified the project still builds


itingliu (commented):
what was the build error?

jmschonfeld:
In this build configuration, declarations like `getpagesize()` come from the `unistd` module rather than a parent/umbrella module like `Darwin`

itingliu (approved):

PR #2139, opened 2026-07-30
Check lowercase "addressbook" scheme only for `URLComponents`

jrflat:
Remove case-insensitivity when checking for `"addressbook"` scheme in a `URLComponents` compatibility path.

### Motivation:

Both `NS/URL` v2 parsing and the pre-Swift `URLComponents` implementation had this compatibility path for lowercase `"addressbook"` scheme specifically. Update this to match the other implementations and simplify the logic.

### Modifications:

- Compare against lowercase `"addressbook"` scheme specifically.

### Result:

- Compatibility path is only taken for lowercase `"addressbook"` scheme, matching other parsers.

### Testing:

- Current unit tests for `NS/URL` and `URLComponents`.

### DIFF
--- a/Sources/FoundationEssentials/URL/URLParser.swift
+++ b/Sources/FoundationEssentials/URL/URLParser.swift
@@ -466,17 +466,7 @@ internal struct RFC3986Parser {
     }
 
     private static func shouldIgnorePort(forSchemeBuffer schemeBuffer: Slice<UnsafeBufferPointer<UInt8>>) -> Bool {
-        let schemeToIgnore = "addressbook".utf8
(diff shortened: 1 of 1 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):
Just confirming since you mentioned:

> Remove case-sensitivity

the goal here is to remove case _insensitivity_ right? i.e. make it case sensitive?

jrflat:
> the goal here is to remove case insensitivity right? i.e. make it case sensitive?

Ah yes, good catch on the double negative! Updated the PR description for posterity.

PR #2140, opened 2026-07-30
Update `URL.appending(component:)` docs for file URL behavior

jrflat:
Updates `URL.appending(component:)` documentation to clarify that the method does not percent-encode path separators (`/`) for file URLs.

### Motivation:

The `URL.appending(component:)` implementation has never percent-encoded path separators for file URLs. We can't update the behavior now because that would break clients unintentionally using it like `.appending(component: "path/with/slashes")`, so update the docs to clarify this behavior instead.

### Modifications:

- Update doc comments to clarify `URL.appending(component:)` does not percent-encode path separators for file URLs.

### Result:

- Doc comment change.

### Testing:

- N/A


jrflat:
@swift-ci please test

jmschonfeld (approved):

PR #2141, opened 2026-07-30
URL/Components.host should not decode %00

jrflat:
Don't decode `%00` to `NUL` in the `URL/URLComponents` host methods.

### Motivation:

Decoding `%00` to `NUL` can lead to unexpected behavior if a client later interops with a C string-taking function.

### Modifications:

Exclude `%00` from the percent-encoded sequences that are decoded.

### Result:

`%00` is not decoded in `URLComponents.host` or `URL.host` methods.

### Testing:

New unit test verifying behavior.


jrflat:
@swift-ci please test

parkera (approved):

PR #2146, opened 2026-07-31
Add additional validation for fields in ISO8601 and HTTP string parsing

parkera:
Add additional validation for fields in ISO8601 and HTTP string parsing

### Motivation:

The existing code was inconsistent in how it validated the various fields of ISO8601 and HTTP strings, which share the same underlying helper functions.

Fixes rdar://182534789

### Modifications:

Add a range and calendar/component argument to the parsing functions, and edit the call sites to use them. This also simplified some of the parsing because the range checking moved to the helper.

The change also revealed a bug in the Gregorian calendar's min and max ranges for `yearForWeekOfYear`.

### Result:

There may be some previously accepted strings that we now throw errors for.

### Testing:

New unit tests added, and some were modified to verify the new behavior.

### DIFF
--- a/Sources/FoundationEssentials/Formatting/DateComponents+ISO8601FormatStyle.swift
+++ b/Sources/FoundationEssentials/Formatting/DateComponents+ISO8601FormatStyle.swift
@@ -475,12 +475,11 @@ extension DateComponents.ISO8601FormatStyle {
             
             if fields.contains(.weekOfYear) {
                 // parse day of week ('ee')
-                // ISO8601 "1" is Monday and "7" is Sunday. For our date components, 2 is Monday and 1 is Sunday. Convert below.
-                // The range is validated here and not with `maximumRange(of: .weekday)`, because the modulo below would map any value into that range.
                 let max = dateSeparator == .omitted ? 2 : nil
-                let value = try it.digits(maxDigits: max, input: inputString, range: 1..<8, onFailure: Date.ISO8601FormatStyle(self).format(Date.now))
+                // Validate that we are in a weekday range (1..<8 for ISO8601/Gregorian)...
+                let value = try it.digits(maxDigits: max, input: inputString, calendar: _calendar, component: .weekday, onFailure: Date.ISO8601FormatStyle(self).format(Date.now))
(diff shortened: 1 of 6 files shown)


Sources/FoundationEssentials/Formatting/DateComponents+ISO8601FormatStyle.swift
```diff
@@ -484,34 +475,20 @@ extension DateComponents.ISO8601FormatStyle {
...
+                // The range is validated here and not with `maximumRange(of: .weekday)`, because the modulo below would map any value into that range.
                 let max = dateSeparator == .omitted ? 2 : nil
-                let value = (try it.digits(maxDigits: max, input: inputString, onFailure: Date.ISO8601FormatStyle(self).format(Date.now)) % 7) + 1
-                
-                guard _calendar.maximumRange(of: .weekday)!.contains(value) else {
-                    throw parseError(inputString, exampleFormattedString: Date.ISO8601FormatStyle(self).format(Date.now))
-                }
-                weekday = value
-                
+                let value = try it.digits(maxDigits: max, input: inputString, range: 1..<8, onFailure: Date.ISO8601FormatStyle(self).format(Date.now))
```
itingliu: Doesn't passing in `range:1..<8` mean that this function only succeeds iff the `value` is in that range? in that case the mod below wouldn't be needed?

itingliu: If that's the case, would you want to use the other function that takes `component: .weekday` like the other fields below?

itingliu (approved):

Sources/FoundationEssentials/Formatting/DateComponents+ISO8601FormatStyle.swift
parkera: We can use the weekday integer validation as you suggest, since the valid range is the same (1..<8). We still need the `(value % 7) + 1` to "wrap" the actual _value_ because the meaning of 1 in ISO8601 spec is Monday, but for `DateComponents` it is Sunday.

I've changed this around a bit to be more readable.

parkera:
@swift-ci test

PR #2148, opened 2026-08-02
Fix copyItem/linkItem mangling item names when the source path ends with a separator

### LINKED ISSUE #1266: FileManager.copyItems(atPath:toPath:) produces incorrect file names when dest doesn't have a trailing slash
When calling `FileManager.copyItems(atPath:toPath:)` with a source folder and a destination folder without a trailing slash, the copied files in the source directory will be renamed to have the destination directory name prepended to the file/directory names.

This is occurring on macOS 15.4.1 (24E263) and it didn’t occur in macOS 15.3 and earlier.

```
swift-driver version: 1.120.5 Apple Swift version 6.1 (swiftlang-6.1.0.110.21 clang-1700.0.13.3)
Target: arm64-apple-macosx15.0
```

Filed feedback FB17323332

---

Script to reproduce the issue

```bash
mkdir -p a/b/c z
touch a/b/foo.txt
echo 'import Foundation; try FileManager.default.copyItem(atPath: "a/b/", toPath: "z/b")' | swift -
```

It will start with the following folder structure before preforming the copy

```
.
├── a
│   └── b
│       ├── c
│       └── foo.txt
└── z
```

After the copy, the resulting folder structure should be

```
.
├── a
│   └── b
│       ├── c
│       └── foo.txt
└── z
    └── b
        ├── c
        └── foo.txt
```

But it is actually, the following

```
.
├── a
│   └── b
│       ├── c
│       └── foo.txt
└── z
    ├── b
    ├── bc
    └── bfoo.txt
```

Adding a trailing slash to the `toPath` to become `"z/b/"` works around the issue.

maxches99:
Fix `copyItem`/`linkItem` mangling item names when the source path ends with a separator.

### Motivation:

`FileManager.copyItem(atPath:toPath:)` writes the copied items under wrong names when the source path has a trailing slash. Copying `a/b/` to `z/b` produces

```
z
├── b
├── bc
└── bfoo.txt
```

instead of nesting `c` and `foo.txt` inside `z/b`. Note that the stray items are created as siblings of the destination, outside of it.

`_linkOrCopyFile` builds each destination path by appending the descendant's path with the source prefix removed, using `strlen(srcPtr)` as the prefix length. However, `fts` appends a descendant's name to the source path and inserts a separator only when the source path does not already end with one:

| source argument | descendant `fts_path` |
| --- | --- |
| `a/b` | `a/b/foo.txt` |
| `a/b/` | `a/b/foo.txt` |

So when the source ends with a separator, the prefix length covers the separator that `fts` reused, and the remainder is appended to the destination without one.

Resolves #1266

### Modifications:

Exclude trailing separators from the prefix length, so the remainder of a descendant's path always begins with a separator.

This affects `linkItem` as well, which shares the same traversal. The paths reported to `FileManagerDelegate` are unchanged: they still come from `fts_path` as given by the caller. The Windows implementation recurses using `appendingPathComponent`, which normalizes separators, and is unaffected.

### Result:

`copyItem(atPath: "a/b/", toPath: "z/b")` now produces the same tree as `copyItem(atPath: "a/b", toPath: "z/b")`.

### Testing:

Added `copyItemAtPathWithTrailingSlash`, which copies a directory containing both a file and a nested subdirectory using a trailing-slash source path, and asserts the full resulting tree — so it also catches the stray entries the bug created next to the destination. Verified that the test fails without the change and passes with it.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileOperations.swift
+++ b/Sources/FoundationEssentials/FileManager/FileOperations.swift
@@ -1118,9 +1118,10 @@ enum _FileOperations {
         #else
         try withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
             let dstLen = Platform.copyCString(dst: buffer.baseAddress!, src: dstPtr, size: FileManager.MAX_PATH_SIZE)
-            // fts builds the path of a descendant by appending its name to the source path, inserting a separator only when the source path does not already end with one. Exclude trailing separators from the prefix length so that the remainder of a descendant's path always begins with a separator.
+            // fts builds the path of a descendant by appending a separator and its name to the source path, but implementations disagree about how the trailing separators of the source path are treated: some drop one of them beforehand and some keep all of them. Ignore them entirely when determining the length of the prefix that the destination path replaces, and re-insert a single separator below.
+            let pathSeparator = CChar(UInt8(ascii: "/"))
             var srcLen = strlen(srcPtr)
-            while srcLen > 0, srcPtr[srcLen - 1] == CChar(UInt8(ascii: "/")) {
+            while srcLen > 0, srcPtr[srcLen - 1] == pathSeparator {
                 srcLen -= 1
             }
             let dstAppendPtr = buffer.baseAddress!.advanced(by: dstLen)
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/FileManager/FileOperations.swift
```diff
@@ -1118,7 +1118,11 @@ enum _FileOperations {
         #else
         try withUnsafeTemporaryAllocation(of: CChar.self, capacity: FileManager.MAX_PATH_SIZE) { buffer in
             let dstLen = Platform.copyCString(dst: buffer.baseAddress!, src: dstPtr, size: FileManager.MAX_PATH_SIZE)
-            let srcLen = strlen(srcPtr)
+            // fts builds the path of a descendant by appending its name to the source path, inserting a separator only when the source path does not already end with one. Exclude trailing separators from the prefix length so that the remainder of a descendant's path always begins with a separator.
```
jmschonfeld: From experimenting with it, it looks like Darwin platforms like macOS do _not_ have this behavior. On macOS it unconditionally appends a separator followed by the name even if the input path ends with a separator. With this change, that means that the `srcLen` is now too short on Darwin so the paths that we pass to `copyfile` would end up looking like `dir2//foo`. It looks like `copyfile` compresses the trailing slashes when processing and it works out fine, but I am a bit worried that this could cause problems.

Did you see that behavior when testing this? Should this be limited to platforms where `fts` has the behavior you mention or is there a better way of replacing the source prefix with the destination prefix here that avoids this problem?

maxches99: You're right that the convention differs — it turns out it isn't even stable within Darwin. Libc's NAPPEND has two branches selected by dyld_get_program_min_os_version() < DYLD_MACOSX_VERSION_15_4: older deployment targets append the separator unconditionally (the dir2//foo you saw), newer ones drop one trailing separator first, which is what glibc's 4.4BSD-derived NAPPEND has always done. Neither drops more than one, so copyItem(atPath: "dir//", ...) produced a doubled separator on every platform with my original patch.

So instead of matching a particular fts, I made this independent of the convention: the prefix length still ignores all trailing separators of the source, and the descendant's relative path is then advanced past whatever separators fts left behind, with exactly one separator written into the destination buffer. That also removes the trailing separator my previous patch left on the destination of the root item itself, which was visible through fileManager(_:shouldCopyItemAtPath:toPath:).

The test is parameterized over dir/ and dir// and now asserts the exact destination paths handed to the delegate, since comparing the resulting tree can't catch dir2//foo (copyfile collapses it). Verified on macOS 26.3: the dir// case fails with the previous patch and passes with this one.

jmschonfeld: Thanks, that seems more robust for varying behavior

jmschonfeld (approved):

Issue #1535, opened 2025-10-05
`base64EncodedData` appends unnecessary `NULL` bytes in certain rare cases. (6.2)


YOCKOW:
**Describe the bug**

`Data`'s `base64EncodedData` appends `NULL` bytes to the last of its output when the following conditions are met:
- Swift 6.2
- OS is Ubuntu
- The length of `Data` is 1993...1995.

**To Reproduce**

Execute the code below with Swift 6.2:

```swift
import Foundation

for length in 1992..<1997 {
  let encoded = Data(
    repeating: 0xFF, count: length
  ).base64EncodedData(
    options: .lineLength76Characters
  )
  print("Length: \(length) | Last byte is NULL?: \(encoded.last == 0x00 ? "YES" : "NO")")
}

// On Darwin:
//
// Length: 1992 | Last byte is NULL?: NO
// Length: 1993 | Last byte is NULL?: NO
// Length: 1994 | Last byte is NULL?: NO
// Length: 1995 | Last byte is NULL?: NO
// Length: 1996 | Last byte is NULL?: NO

// On Ubuntu:
// Length: 1992 | Last byte is NULL?: NO
// Length: 1993 | Last byte is NULL?: YES
// Length: 1994 | Last byte is NULL?: YES
// Length: 1995 | Last byte is NULL?: YES
// Length: 1996 | Last byte is NULL?: NO

```

**Expected behavior**

No unnecessary bytes are added to Base64 encoded data.

**Configuration (please complete the following information):**
 
- Swift Version: 6.2
- OS: Ubuntu 24.04

**Additional context**

This bug doesn't exist in Swift 6.3-dev.
However, let me file this here because Swift 6.2.x is current "RELEASE" version.

parkera:
Was this fixed in https://github.com/swiftlang/swift-foundation/pull/1425? @fabianfett

maxches99:
This appears to be fixed on `main` by #1425 (merged 2025-07-14), which corrected `encodeComputeCapacity` for the case where the last output line exactly fills the line length. With `.lineLength76Characters`, input lengths 1993–1995 hit exactly that boundary: the encoded output is exactly 35 full 76-character lines, so the capacity computation over-allocated one line separator (2 bytes) that the encoder never wrote, and `base64EncodedData` returned them as trailing `NUL` bytes (the buffer of `Data(count:)` is zero-initialized and was not trimmed to the actual encoded length).

[...]

PR #2150, opened 2026-08-02
Honor omitPaddingCharacter when base64 encoding with line breaks

### LINKED ISSUE #2149: Out-of-bounds write when base64 encoding with .omitPaddingCharacter and a line-length option
**Describe the bug**

`Data.base64EncodedString(options:)` and `Data.base64EncodedData(options:)` write past the end of their output buffer when `.omitPaddingCharacter` is combined with `.lineLength64Characters` or `.lineLength76Characters`.

This is a memory-safety bug: the encoder writes 1–2 bytes beyond the buffer it was given. In debug builds this hits the `UnsafeMutableBufferPointer` bounds check and traps, so it is easy to notice. In release builds that check is a `_debugPrecondition` and is compiled out, so the write actually lands past the end of the allocation and corrupts adjacent heap memory.

It is not a rare edge case — it happens for **every input whose length is not a multiple of 3**, with either line-length option and any combination of line-ending options.

**To Reproduce**

```swift
let data = Data(repeating: 0xFF, count: 1)
_ = data.base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter])
```

In a debug build this traps in the `UnsafeMutableBufferPointer` bounds check (signal 5):

```
Swift/UnsafeBufferPointer.swift:791: Fatal error
```

The same happens with `.lineLength64Characters` and with `base64EncodedData(options:)`.

**Expected behavior**

Encoding succeeds and the padding characters are omitted, i.e. the result is the padded output with the trailing `=` characters removed:

```swift
Data(repeating: 0xFF, count: 1)
    .base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter])  // "/w"
```

- Swift Version: swift-foundation `main` (fc629f0), built with a recent trunk toolchain
- OS: macOS
- OS Version: 26.3 — though the affected code path is platform-independent

**Regression information:**

`.omitPaddingCharacter` was introduced in #1195 and is available from FoundationPreview 6.3, so every version that ships the option is affected. There is no earlier working version.

**Additional context**

In `Sources/FoundationEssentials/Data/Data+Base64.swift`, `encodeComputeCapacity(bytes:options:)` reduces the allocated capacity by 1–2 bytes when `.omitPaddingCharacter` is set and `bytes % 3 != 0`:

```swift
let capacityWithoutBreaks = if options.contains(.omitPaddingCharacter) {
    switch bytes % 3 {
    case 0: ((bytes + 2) / 3) * 4
    case 1: ((bytes + 2) / 3) * 4 - 2
    case 2: ((bytes + 2) / 3) * 4 - 1
    default: fatalError()
    }
} else {
    ((bytes + 2) / 3) * 4
}
```

but `_encodeWithLineBreaks(input:buffer:length:options:)` — the path taken whenever a line-length option is set — hardcodes the flag to `false` and therefore always writes the `=` padding:

```swift
static func _encodeWithLineBreaks(
    input: UnsafeBufferPointer<UInt8>,
    buffer: UnsafeMutableBufferPointer<UInt8>,
    length: inout Int,
    options: Data.Base64EncodingOptions
) {
    let omitPaddingCharacter = false
    ...
```

So the buffer is sized for output *without* padding while the encoder writes output *with* padding.

This looks like a missed call site in #1195: that PR replaced the identical `let omitPaddingCharacter = false` line in `_encode` with `options.contains(.omitPaddingCharacter)` and taught `encodeComputeCapacity` about the new option, but left the line in `_encodeWithLineBreaks` as it was.

Worth noting that `encodeToData` discards the length written back by `_encode` and returns a `Data` of the full computed capacity, so the computed capacity has to match the number of bytes written exactly: over-allocating shows up as trailing NUL bytes (#1535), and under-allocating is this out-of-bounds write.

Honoring the option in `_encodeWithLineBreaks` makes `encodeComputeCapacity` exactly right as it is already written. Comparing the computed capacity against the number of bytes `_encodeWithLineBreaks` writes, for every input length in `0...20000` across all eight combinations of line-length and line-ending options, gives an exact match in every case, both with and without `.omitPaddingCharacter`.

I have a fix and tests ready and will open a PR.

maxches99:
_Fix an out-of-bounds write when base64 encoding with `.omitPaddingCharacter` and a line-length option._

### Motivation:

Resolves #2149.

`Data.base64EncodedString(options:)` and `Data.base64EncodedData(options:)` write past the end of their output buffer when `.omitPaddingCharacter` is combined with `.lineLength64Characters` or `.lineLength76Characters`:

```swift
let data = Data(repeating: 0xFF, count: 1)
_ = data.base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter])
// Swift/UnsafeBufferPointer.swift:791: Fatal error
```

`encodeComputeCapacity(bytes:options:)` shrinks the output buffer by 1–2 bytes when `.omitPaddingCharacter` is set and the input length is not a multiple of 3, but `_encodeWithLineBreaks(input:buffer:length:options:)` — the path taken whenever a line-length option is set — hardcodes `let omitPaddingCharacter = false` and always writes the `=` padding. The buffer is sized for output without padding while the encoder writes output with padding.

This affects every input whose length is not a multiple of 3, with either line-length option and any combination of line-ending options. In debug builds it hits the `UnsafeMutableBufferPointer` bounds check and traps; in release builds that check is a `_debugPrecondition` and is compiled out, so the write lands past the end of the allocation.

The hardcoded `false` looks like a missed call site in #1195, which introduced the option: it replaced the identical line in `_encode` with `options.contains(.omitPaddingCharacter)` and taught `encodeComputeCapacity` about the option, but left this one unchanged.

### Modifications:

`_encodeWithLineBreaks` now reads the flag from the options instead of hardcoding `false`, matching `_encode`.

Honoring the option is also what makes the existing capacity computation correct rather than merely safe. `encodeToData` discards the length written back by `_encode` and returns a `Data` of the full computed capacity, so the capacity has to match the number of bytes written exactly — over-allocating leaves trailing NUL bytes and under-allocating is the out-of-bounds write. Comparing `encodeComputeCapacity` against the number of bytes `_encodeWithLineBreaks` writes, for every input length in `0...20000` across all eight combinations of line-length and line-ending options, now matches exactly in every case, with and without `.omitPaddingCharacter`.

The alternative — ignoring `.omitPaddingCharacter` when a line-length option is set and dropping the capacity reduction in that case — would also fix the crash, but the option is documented without any such carve-out, and Darwin's `NSData` offers no precedent since it does not implement this option at all.

### Result:

`.omitPaddingCharacter` works together with the line-length options and produces the padded output minus the trailing `=` characters:

```swift
Data(repeating: 0xFF, count: 1)
    .base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter])  // "/w"

Data(repeating: 0, count: 49)
    .base64EncodedString(options: [.lineLength64Characters, .omitPaddingCharacter])
// AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\r\nAA
```

### Testing:

Two tests added to `DataTests`:

- `base64Encode_omitPaddingWithLineBreaks` checks the expected output for the inputs that previously overflowed, including a case with a full first line followed by a partial last line that needs padding.
- `base64Encode_omitPaddingWithLineBreaksMatchesUnbrokenOutput` sweeps input lengths `0..<260` across both line-length options and all four line-ending combinations, and checks that the output equals the padded output with `=` removed. It asserts this for `base64EncodedString` and for `base64EncodedData` and checks that their lengths agree, so an over-allocated buffer (trailing NUL bytes in the `Data` result, as in #1535) fails the test as well as an under-allocated one.

Both tests fail on `main` and pass with the fix. The existing base64 tests continue to pass.

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Base64.swift
+++ b/Sources/FoundationEssentials/Data/Data+Base64.swift
@@ -380,7 +380,7 @@ extension Base64 {
         length: inout Int,
         options: Data.Base64EncodingOptions
     ) {
-        let omitPaddingCharacter = false
+        let omitPaddingCharacter = options.contains(.omitPaddingCharacter)
 
         assert(options.contains(.lineLength64Characters) || options.contains(.lineLength76Characters))
 
--- a/Tests/FoundationEssentialsTests/DataTests.swift
+++ b/Tests/FoundationEssentialsTests/DataTests.swift
@@ -2556,6 +2556,70 @@ extension DataTests {
 
     }
 
+    @Test func base64Encode_omitPaddingWithLineBreaks() {
+        #expect(
+            Data(repeating: 0xFF, count: 1).base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter]) ==
+            "/w"
+        )
+        #expect(
+            Data(repeating: 0xFF, count: 2).base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter]) ==
+            "//8"
+        )
+        #expect(
+            Data(repeating: 0xFF, count: 3).base64EncodedString(options: [.lineLength76Characters, .omitPaddingCharacter]) ==
+            "////"
+        )
+
+        // a full first line followed by a partial last line that needs padding
+        #expect(
+            Data(repeating: 0, count: 49).base64EncodedString(options: [.lineLength64Characters, .omitPaddingCharacter]) ==
+            """
+            AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\r\n\
+            AA
+            """
+        )
+        #expect(
+            Data(repeating: 0, count: 50).base64EncodedString(options: [.lineLength64Characters, .omitPaddingCharacter]) ==
+            """
+            AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\r\n\
+            AAA
+            """
+        )
+    }
+
+    @Test func base64Encode_omitPaddingWithLineBreaksMatchesUnbrokenOutput() {
+        let lineLengthOptions: [Data.Base64EncodingOptions] = [.lineLength64Characters, .lineLength76Characters]
+        let lineEndOptions: [Data.Base64EncodingOptions] = [
+            [], .endLineWithCarriageReturn, .endLineWithLineFeed, [.endLineWithCarriageReturn, .endLineWithLineFeed]
+        ]
+
+        for count in 0..<260 {
+            let data = Data((0..<count).map { UInt8($0 % 256) })
+
+            for lineLength in lineLengthOptions {
+                for lineEnd in lineEndOptions {
+                    let options: Data.Base64EncodingOptions = [lineLength, lineEnd, .omitPaddingCharacter]
+                    let padded = data.base64EncodedString(options: [lineLength, lineEnd])
+                    let expected = String(padded.filter { $0 != "=" })
+
+                    let string = data.base64EncodedString(options: options)
+                    #expect(string == expected, "count: \(count), options: \(options.rawValue)")
+
+                    // `base64EncodedData` keeps the full allocated capacity, so a mismatch here
+                    // means the computed capacity did not match what was actually written.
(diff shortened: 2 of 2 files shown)


fabianfett (approved):
Thanks for this fix. And what a stupid bug I put in here.

jmschonfeld (approved):

PR #2151, opened 2026-08-02
Set the creation date in setAttributes(_:ofItemAtPath:) where supported

### LINKED ISSUE #1906: FileManager: Support setting .creationDate where possible
**Describe the bug**
FileManager claims to support setting `.creationDate`, but it doesn't.

The implementation of `FileManager.replaceItemAt` in corelibs-foundation uses this API: https://github.com/swiftlang/swift-corelibs-foundation/blob/f9c1d4eee71957c49eb66fd8fb5dac856d0efbf4/Sources/Foundation/FileManager%2BPOSIX.swift#L479-L491

This missing functionality caused us to comment out a test on https://github.com/swiftlang/swift-corelibs-foundation/pull/5443

**To Reproduce**
Sample code to reproduce the behavior.

```swift
let myDate = Date.now
try FileManager.default.setAttributes([.creationDate : myDate], ofItemAtPath: "mypath.txt")
let attrs = try FileManager.default.attributesOfItem(atPath: "mypath.txt")
assert(attrs[.creationDate] == myDate) // fails
```

**Expected behavior**
I expect the attribute to be set or an error to be thrown on platforms where it's not supported. If an error can't be thrown on those platforms due to compatibility, at least some very clear documentation on `.creationDate`.

 - Swift Version: main
 - OS: Ubuntu
 - OS Version: noble

**Regression information:**
If applicable, please list older versions where this issue did not occur.

**Additional context**
It may not be possible to set creation date on Linux (see context: https://github.com/swiftlang/swift-corelibs-foundation/pull/5443#discussion_r3012844994) in which case I'd expect either good documentation or an error to be thrown.

The code path seems to end up at this TODO: https://github.com/swiftlang/swift-foundation/blob/b460a1ff3d9cfefcc9ec52f34cc127932dc800f5/Sources/FoundationEssentials/FileManager/FileManager%2BUtilities.swift#L133-L138

maxches99:
### Motivation

`FileManager` accepts `.creationDate` in `setAttributes(_:ofItemAtPath:)` but silently discards it in swift-foundation:

- the non-framework branch of `_setCatInfoAttributes` returned without doing anything, and
- the Windows implementation passed `nil` for `SetFileTime`'s creation time parameter, so only the modification date was ever written.

Nothing is set and nothing is thrown, so callers have no way to tell the attribute was dropped. This also caused a test to be commented out in swiftlang/swift-corelibs-foundation#5443, since `FileManager.replaceItemAt` relies on this attribute.

Resolves #1906.

### Modifications

Set the attribute on the platforms that can record it:

- **Darwin**: `setattrlist` with `ATTR_CMN_CRTIME`, in both the swift-foundation and `FOUNDATION_FRAMEWORK` builds. `FSOPT_NOFOLLOW` is deliberately not set, because `setAttributes(_:ofItemAtPath:)` follows symlinks — which is also the reason the URL resource value path has to resolve the symlink by hand. `.creationDate` is no longer handed to `setResourceValues`; that path now carries only the Finder info and `hasHiddenExtension` keys.
- **Windows**: the creation time is passed to `SetFileTime` alongside the modification time, so that setting both in a single call applies both.

`stat.creationDate` now reads `st_birthtimespec` instead of `st_ctimespec` on Darwin (the same change as #1648). Without it a creation date written by `setAttributes(_:ofItemAtPath:)` never reads back, because `setattrlist` bumps the inode change time to *now* — so the round trip reported in #1906 would keep failing. Happy to drop that hunk if #1648 lands first.

`_setCatInfoAttributes` gained a `fileSystemRepresentation` parameter; its only call site already runs inside `withFileSystemRepresentation`, so no extra conversion is introduced.

**Platforms that cannot record a creation date keep ignoring the attribute rather than throwing `featureUnsupported`.** The issue asked for an error in that case, but `.creationDate` is unconditionally included in every `attributesOfItem(atPath:)` result, so throwing would break the common pattern of passing such a result straight back into `setAttributes(_:ofItemAtPath:)` — which is exactly what the existing `getSetAttributes` test does. I'm happy to switch to throwing if reviewers prefer it, but it looked like a source-compatibility hazard worth avoiding.

While restructuring the Windows path, a malformed date now only drops that single date instead of returning early from the whole closure and skipping the remaining attributes (previously a non-representable modification date would also skip applying read-only).

### Result

`setAttributes([.creationDate: date], ofItemAtPath:)` sets the creation date on Darwin and Windows, `attributesOfItem(atPath:)` reads it back, and other platforms behave as before.

### Testing

New tests in `FileManagerTests`, both enabled on Darwin and Windows (Linux offers no way to set a creation date):

- `setAttributesCreationDate` — files and directories, plus setting `.creationDate` and `.modificationDate` together to check that Windows' single `SetFileTime` call applies both.
- `malformedCreationDateAttribute` — infinity/NaN dates are dropped rather than throwing or crashing, mirroring `malformedModificationDateAttribute`.

Both read the value back through `attributesOfItem(atPath:)` rather than through platform APIs. `getSetAttributes` additionally asserts that `.creationDate` is present in the result it feeds back in, which is what makes that round trip meaningful for this change.

Verified locally on macOS (the package build takes the non-framework path, so the Darwin change is covered): the full suite passes, and the new assertions fail without the fix. **The Windows path and the `FOUNDATION_FRAMEWORK` build are not covered by my local testing.**

I targeted `main`; happy to rebase onto a release branch if that is the preferred destination.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -127,17 +127,26 @@ extension _FileManagerImpl {
     
     private static let _catInfoKeys: [FileAttributeKey] = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden, .creationDate]
     private static let _swiftFoundationUnsupportedKeys: [FileAttributeKey] = [.hfsCreatorCode, .hfsTypeCode, .busy, .extensionHidden]
-    static func _setCatInfoAttributes(_ attributes: [FileAttributeKey : Any], path: String) throws {
+    static func _setCatInfoAttributes(_ attributes: [FileAttributeKey : Any], path: String, fileSystemRepresentation: UnsafePointer<CChar>) throws {
         let hasRelevantKeys = attributes.keys.contains(where: { _catInfoKeys.contains($0) })
         guard hasRelevantKeys else { return }
-        
+
         #if !FOUNDATION_FRAMEWORK
         // Exclude some attributes (like .creationDate) from this check since they are unconditionally, implicitly included in `attributesForItem(atPath:)` results
         if attributes.keys.contains(where: { _swiftFoundationUnsupportedKeys.contains($0) }) {
             throw CocoaError.errorWithFilePath(.featureUnsupported, path)
-        } else {
-            return // TODO: support relevant cat info keys in swift-foundation
         }
+
+        #if canImport(Darwin)
+        if let creationDate = attributes[.creationDate] as? Date {
+            try Self._setCreationDate(creationDate, path: path, fileSystemRepresentation: fileSystemRepresentation)
+        }
+        #endif
+        // On platforms that provide no way to set a file's creation date, the attribute is
+        // ignored rather than reported as unsupported: `.creationDate` is implicitly present in
+        // every `attributesOfItem(atPath:)` result, so throwing would break passing those
+        // results back to `setAttributes(_:ofItemAtPath:)` unmodified.
+        return // TODO: support the remaining cat info keys in swift-foundation
         #else
         // -setAttributes:ofItemAtPath:error: follows symlinks (<rdar://5815920>), but the NSURL resource value API doesn't, so we have to manually resolve the symlink.
         // We lie to fileURLWithPath:isDirectory: to avoid the extra stat. Since this URL isn't used as a base URL for another URL, it shouldn't make any difference.
@@ -179,6 +188,33 @@ extension _FileManagerImpl {
         try url.setResourceValues(URLResourceValues(values: urlAttributes))
         #endif
     }
+
+#if canImport(Darwin) && !FOUNDATION_FRAMEWORK
+    private static func _setCreationDate(_ date: Date, path: String, fileSystemRepresentation: UnsafePointer<CChar>) throws {
+        let interval = date.timeIntervalSince1970
+        let seconds = interval.rounded(.down)
+        // Malformed dates are dropped rather than reported as an error, matching the behavior of `.modificationDate`.
+        guard let tv_sec = time_t(exactly: seconds),
+              let tv_nsec = Int(exactly: ((interval - seconds) * 1_000_000_000).rounded()) else {
+            return
+        }
+
+        var attrs = attrlist(
+            bitmapcount: u_short(ATTR_BIT_MAP_COUNT),
+            reserved: 0,
+            commonattr: attrgroup_t(ATTR_CMN_CRTIME),
+            volattr: 0,
+            dirattr: 0,
+            fileattr: 0,
+            forkattr: 0
+        )
+        var value = timespec(tv_sec: tv_sec, tv_nsec: tv_nsec)
+        // `setAttributes(_:ofItemAtPath:)` follows symlinks, so FSOPT_NOFOLLOW is intentionally not set here.
+        guard setattrlist(fileSystemRepresentation, &attrs, &value, MemoryLayout<timespec>.size, 0) == 0 else {
+            throw CocoaError.errorWithFilePath(path, errno: errno, reading: false)
+        }
+    }
+#endif
 
 #if !os(Windows) && !os(WASI) && !os(OpenBSD) && !os(Emscripten)
     static func _setAttribute(_ key: UnsafePointer<CChar>, value: Data, at path: UnsafePointer<CChar>, followSymLinks: Bool) throws {
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -138,6 +138,7 @@ extension _FileManagerImpl {
         }
         #endif
 
+        // Platforms with no way to record a creation date ignore `.creationDate` rather than reporting it as unsupported, since it is implicitly included in every `attributesOfItem(atPath:)` result and throwing would break passing such a result straight back into `setAttributes(_:ofItemAtPath:)`.
         #if canImport(Darwin)
         if let creationDate = attributes[.creationDate] as? Date {
             try Self._setCreationDate(creationDate, path: path, fileSystemRepresentation: fileSystemRepresentation)
@@ -182,7 +183,7 @@ extension _FileManagerImpl {
         guard !urlAttributes.isEmpty else { return }
         try url.setResourceValues(URLResourceValues(values: urlAttributes))
         #else
-        // TODO: support the remaining cat info keys in swift-foundation. Platforms with no way to record a creation date ignore `.creationDate` rather than reporting it as unsupported, since it is implicitly included in every `attributesOfItem(atPath:)` result and throwing would break passing such a result straight back into `setAttributes(_:ofItemAtPath:)`.
+        // TODO: support the remaining cat info keys in swift-foundation
         #endif
     }
 
--- a/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
@@ -136,18 +136,15 @@ extension _FileManagerImpl {
         if attributes.keys.contains(where: { _swiftFoundationUnsupportedKeys.contains($0) }) {
             throw CocoaError.errorWithFilePath(.featureUnsupported, path)
         }
+        #endif
 
         #if canImport(Darwin)
         if let creationDate = attributes[.creationDate] as? Date {
             try Self._setCreationDate(creationDate, path: path, fileSystemRepresentation: fileSystemRepresentation)
         }
         #endif
-        // On platforms that provide no way to set a file's creation date, the attribute is
-        // ignored rather than reported as unsupported: `.creationDate` is implicitly present in
-        // every `attributesOfItem(atPath:)` result, so throwing would break passing those
-        // results back to `setAttributes(_:ofItemAtPath:)` unmodified.
-        return // TODO: support the remaining cat info keys in swift-foundation
-        #else
+
+        #if FOUNDATION_FRAMEWORK
         // -setAttributes:ofItemAtPath:error: follows symlinks (<rdar://5815920>), but the NSURL resource value API doesn't, so we have to manually resolve the symlink.
         // We lie to fileURLWithPath:isDirectory: to avoid the extra stat. Since this URL isn't used as a base URL for another URL, it shouldn't make any difference.
         var url = URL(fileURLWithPath: path.resolvingSymlinksInPath, isDirectory: false)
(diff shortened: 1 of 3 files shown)


Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -127,17 +127,26 @@ extension _FileManagerImpl {
...
         // Exclude some attributes (like .creationDate) from this check since they are unconditionally, implicitly included in `attributesForItem(atPath:)` results
         if attributes.keys.contains(where: { _swiftFoundationUnsupportedKeys.contains($0) }) {
             throw CocoaError.errorWithFilePath(.featureUnsupported, path)
-        } else {
-            return // TODO: support relevant cat info keys in swift-foundation
         }
+
+        #if canImport(Darwin)
+        if let creationDate = attributes[.creationDate] as? Date {
+            try Self._setCreationDate(creationDate, path: path, fileSystemRepresentation: fileSystemRepresentation)
```
jmschonfeld, line 144: Is there any difference between doing this and using `URLResourceValues` to set the value? If not, we might as well do this always rather than conditionalizing on `FOUNDATION_FRAMEWORK`

cc @jrflat

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -909,6 +909,87 @@ private struct FileManagerTests {
...
+            try fileManager.setAttributes([.creationDate: creationTestDate, .modificationDate: modificationTestDate], ofItemAtPath: "testFile")
+            if let creationDate = self.observedCreationDate(ofItemAtPath: "testFile") {
+                #expect(abs(creationDate.timeIntervalSince1970 - creationTestDate.timeIntervalSince1970) < 1.0, "Creation date should be set when set alongside a modification date")
+            }
+            let modificationDate = try #require(try fileManager.attributesOfItem(atPath: "testFile")[.modificationDate] as? Date)
+            #expect(abs(modificationDate.timeIntervalSince1970 - modificationTestDate.timeIntervalSince1970) < 1.0, "Modification date should be set when set alongside a creation date")
+        }
+    }
+
+    @Test func setAttributesAcceptsAttributesOfItemResult() async throws {
```
jmschonfeld: How is this different than the existing `getSetAttributes` test?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -909,6 +909,87 @@ private struct FileManagerTests {
...
+                try fileManager.setAttributes(attributes, ofItemAtPath: "testFile")
+            }
+        }
+    }
+
+    @Test func malformedCreationDateAttribute() async throws {
+        try await FilePlayground {
+            "foo"
+        }.test { fileManager in
+            let sentinelDate = self.observedCreationDate(ofItemAtPath: "foo")
```
jmschonfeld: I think we should just use `attributesOfItemAtPath` here to ensure the result remains consistent

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -909,6 +909,87 @@ private struct FileManagerTests {
...
+
+    @Test func testSetAttributesCreationDate() async throws {
+        try await FilePlayground {
+            "testFile"
+            Directory("testDir") {}
+        }.test { fileManager in
+            let testDate = Date(timeIntervalSince1970: 1234567890)
+            for path in ["testFile", "testDir"] {
+                try fileManager.setAttributes([.creationDate: testDate], ofItemAtPath: path)
+                if let creationDate = self.observedCreationDate(ofItemAtPath: path) {
```
jmschonfeld: I'd rather just use the `attributesOf` API here to read it back after setting it rather than embedding platform-specific code into the unit tests. As it stands today, this would not test Windows even though it should

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -909,6 +909,87 @@ private struct FileManagerTests {
...
+                    #expect(abs(creationDate.timeIntervalSince1970 - testDate.timeIntervalSince1970) < 1.0, "Creation date of '\(path)' should be set correctly")
+                }
+            }
+
+            // Setting both dates at once must apply both rather than dropping one of them.
```
jmschonfeld: I'm confused by this comment because it matches neither the implementation nor what the test does below

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -127,17 +127,26 @@ extension _FileManagerImpl {
...
+        }
+        #endif
+        // On platforms that provide no way to set a file's creation date, the attribute is
```
jmschonfeld: We don't manually wrap comments unless it semantically makes sense to do so in this repo

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
maxches99: I couldn't find one. The only thing the URL resource value path does that `setattrlist` doesn't is resolve the symlink by hand — and that's only needed because the NSURL API doesn't follow symlinks, whereas `setattrlist` without `FSOPT_NOFOLLOW` does, which is the behavior we want here anyway.

So I've moved it out of the conditional: `.creationDate` now goes through `_setCreationDate` on all Darwin platforms, and `setResourceValues` is left carrying only the Finder info and `hasHiddenExtension` keys (with an early return added when nothing is left to set). Note I can't build the `FOUNDATION_FRAMEWORK` variant locally, so that half is checked by inspection only.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
maxches99: It isn't — same round trip. I've removed it and moved the one assertion it added over to `getSetAttributes`: that `.creationDate` is actually present in the `attributesOfItem(atPath:)` result being fed back in. That's what makes the existing test cover the "must not throw where we can't set it" behavior rather than passing vacuously.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
maxches99: Done — `malformedCreationDateAttribute` now mirrors `malformedModificationDateAttribute` exactly: it seeds a known creation date through `FilePlayground` and reads it back with `attributesOfItem(atPath:)` on both sides of the malformed writes.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
maxches99: Agreed, and this turned out to be the more interesting part of the review. Reading back through `attributesOfItem(atPath:)` didn't work on Darwin because of the *read* side: `stat.creationDate` reported `st_ctimespec`, and `setattrlist` bumps the inode change time to *now*, so the value written was never observable through the public API. That's also why the round trip in #1906 kept failing.

So I've included the `st_birthtimespec` change (the same one as #1648, which has been open a while — happy to drop this hunk if you'd rather land that separately). Windows already returned `ftCreationTime` from `attributesOfItem(atPath:)`, so nothing was needed there.

Both tests now read back through `attributesOfItem(atPath:)` with no platform-specific code, and are gated with `.enabled(if: isDarwin || isWindows)`. Linux stays out because there's no API to set a birth time there, not because of how the value is read.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
maxches99: Fair — that was meant to describe the Windows path, where both dates go through a single `SetFileTime` call and an early return used to drop one of them, but it read like a claim about the implementation in general. Reworded to say what it's actually guarding.

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
maxches99: Fixed, thanks.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
jmschonfeld: I don't think we need to add the additional expectation to that test. That test is about ensuring the roundtrip is successful - it doesn't really pertain to the creation date in particular. I think it's fine to leave validating the creation date received from `attributesOfitem(atPath:)` to the other tests

Sources/FoundationEssentials/FileManager/FileManager+Utilities.swift
```diff
@@ -173,13 +179,40 @@ extension _FileManagerImpl {
         if let extensionHidden = attributes[.extensionHidden] {
             urlAttributes[.hasHiddenExtensionKey] = extensionHidden
         }
-        if let creationDate = attributes[.creationDate] {
-            urlAttributes[.creationDateKey] = creationDate
-        }
+        guard !urlAttributes.isEmpty else { return }
         try url.setResourceValues(URLResourceValues(values: urlAttributes))
+        #else
+        // TODO: support the remaining cat info keys in swift-foundation. Platforms with no way to record a creation date ignore `.creationDate` rather than reporting it as unsupported, since it is implicitly included in every `attributesOfItem(atPath:)` result and throwing would break passing such a result straight back into `setAttributes(_:ofItemAtPath:)`.
```
jmschonfeld: I think the second half of this comment is better placed on the creation date code rather than separately here at the end

maxches99: Done — moved that sentence up to the creation date block and left just the original TODO here.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
maxches99: Fair enough — removed the extra expectation, `getSetAttributes` is back to its original form.

maxches99:
@jmschonfeld gentle ping — the second round of feedback was addressed on Aug 8th and this has been idle since.

One thread is still open with a question back to you: whether the `setattrlist` path differs in any way from `URLResourceValues`. I couldn't find a difference beyond the manual symlink resolution, which is why the code is now unconditional rather than `#if !FOUNDATION_FRAMEWORK` — but I'd like your confirmation before this lands.

The other thing worth your call: this also changes the read side, `stat.creationDate` from `st_ctimespec` to `st_birthtimespec`. Without it the date written via `ATTR_CMN_CRTIME` isn't observable through `attributesOfItem(atPath:)` at all, since `setattrlist` bumps `st_ctime` to now — so the fix can't be tested without it. That's the same change as #1648, open since January. Happy to keep it here, or to split it out if you'd rather review it on its own.

CI also hasn't run on this PR yet. The branch is 51 commits behind main with no conflicts — say the word and I'll rebase.

PR #2152, opened 2026-08-02
Add an .editorconfig file

### LINKED ISSUE #2119: There should be an editorconfig file
swift project has it https://github.com/swiftlang/swift/pull/78087

It'd be easier for those who work across repos with difference indentation preferences

maxches99:
### Motivation

Resolves #2119.

### Modifications

Adds a root `.editorconfig`. The values are derived from what this repository actually uses rather than copied from [swiftlang/swift's file](https://github.com/swiftlang/swift/pull/78087), which sets two-space indentation because that repository is mostly C++:

| | setting | basis |
|---|---|---|
| default | `indent_size = 4`, `indent_style = space` | Swift sources; also matches `.h`/`.c` under `Sources/_FoundationCShims` and the `CMakeLists.txt` files (456 four-space indents vs 12 two-space). Consistent with the `"indentation": { "spaces": 4 }` proposed in #2052. No tracked source file indents with tabs. |
| `*.{yml,yaml}` | `indent_size = 2` | the workflows and `.spi.yml` |
| default | `end_of_line = lf`, `charset = utf-8`, `insert_final_newline = true` | no tracked file outside test resources uses CRLF, all of them decode as UTF-8, and all but nine end with a newline |

### Test and benchmark resources are excluded

This is the part worth reviewing. Files under `Tests/**/Resources/**` and `Benchmarks/**/Resources/**` are byte-for-byte parser input rather than source, so every property is `unset` for them:

- 90 of the 214 test resource files deliberately end **without** a trailing newline, including the `fail*.json` cases whose whole purpose is to be rejected;
- several are binary property lists;
- `Tests/FoundationEssentialsTests/Resources/JSON5/spec/new-lines/` stores the same content with CR, LF and CRLF endings on purpose. That directory already carries its own `.editorconfig` and a `.gitattributes` marking its contents `binary`.

That nested `.editorconfig` does not set `root = true`, so without this exclusion a root file would leak `insert_final_newline = true` into it and an editor would append a newline to fixtures that exist precisely to test newline handling.

I verified the resolution with the reference `editorconfig` implementation. Spot checks:

```
Sources/FoundationEssentials/Data/Data+Writing.swift      indent_size=4  end_of_line=lf  insert_final_newline=true
.github/workflows/pull_request.yml                        indent_size=2  end_of_line=lf  insert_final_newline=true
Tests/.../Resources/JSON/fail/fail1.json                  all unset
Tests/.../Resources/JSON5/spec/new-lines/comment-cr.json5 end_of_line=cr (nested config still wins), rest unset
Benchmarks/Benchmarks/JSON/Resources/twitter.json         all unset
```

### Deliberate omissions

- **`trim_trailing_whitespace`** — 297 of the 442 Swift files currently contain trailing whitespace. Enabling it would make editors rewrite unrelated lines the moment anyone saves one of those files, producing large incidental diffs. It seemed better to leave this out than to pair the config with a repository-wide whitespace sweep, but I'm happy to add it (with or without that sweep) if you'd prefer.
- **`max_line_length`** — left to `.swift-format` (#2052 proposes `"lineLength": 250`) so that there is one source of truth rather than two that can drift.

Nine tracked files outside test resources do not end with a newline today (`LICENSE.md`, six documents under `Proposals/`, `Data+Writing.swift` and `Sources/FoundationEssentials/Predicate/CMakeLists.txt`). I left them untouched to keep this to a single file — `Proposals/` in particular holds accepted, archived documents. Let me know if you'd rather they were fixed here.

### Result

Editors that support EditorConfig pick up this repository's conventions automatically, without normalizing test fixtures.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,32 @@
+# editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+
+[*.{yml,yaml}]
+indent_size = 2
+
+# Test and benchmark inputs are byte-for-byte fixtures rather than source: many
+# deliberately end without a trailing newline, some use CR or CRLF line endings
+# on purpose (see Tests/FoundationEssentialsTests/Resources/JSON5/spec/new-lines,
+# which has its own .editorconfig and marks its contents binary in .gitattributes),
+# and others are binary property lists. Editors must not normalize any of that.
+[Tests/**/Resources/**]
+charset = unset
+end_of_line = unset
+indent_style = unset
+indent_size = unset
+insert_final_newline = unset
+
+[Benchmarks/**/Resources/**]
+charset = unset
+end_of_line = unset
+indent_style = unset
+indent_size = unset
+insert_final_newline = unset
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,9 +8,15 @@ end_of_line = lf
 indent_style = space
 indent_size = 4
 insert_final_newline = true
+trim_trailing_whitespace = true
 
 [*.{yml,yaml}]
 indent_size = 2
+
+# Markdown gives two trailing spaces the meaning of a hard line break, which the
+# proposals under Proposals/ rely on.
+[*.md]
+trim_trailing_whitespace = false
 
 # Test and benchmark inputs are byte-for-byte fixtures rather than source: many
 # deliberately end without a trailing newline, some use CR or CRLF line endings
@@ -23,6 +29,7 @@ end_of_line = unset
 indent_style = unset
 indent_size = unset
 insert_final_newline = unset
+trim_trailing_whitespace = unset
 
 [Benchmarks/**/Resources/**]
 charset = unset
@@ -30,3 +37,4 @@ end_of_line = unset
 indent_style = unset
 indent_size = unset
 insert_final_newline = unset
+trim_trailing_whitespace = unset


.editorconfig
```diff
@@ -0,0 +1,32 @@
...
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+indent_size = 4
+insert_final_newline = true
+
+[*.{yml,yaml}]
+indent_size = 2
```
iCharlesHu, line 14: We should omit this one for now since SwiftFoundation does not have an official yaml style guide (yet).

iCharlesHu:
>  trim_trailing_whitespace — 297 of the 442 Swift files currently contain trailing whitespace. Enabling it would make editors rewrite unrelated lines the moment anyone saves one of those files, producing large incidental diffs. 

I was worried about this too. However, from my testing, it seems both Xcode and Zed only apply the new rules to new lines that you created, as opposed to reformatting the entire file (unless you explicitly asked it to reformat the entire file) on save. Could you verify if this is also the behavior for you as well? If that is the case, we can/should enable `trim_trailing_whitespace`, as these settings should only affect new code from this point on. We can do a separate "linter" pass if we actually want to update the older files.

maxches99:
@iCharlesHu 
Verified — Xcode matches what you saw, Zed doesn't, and enabling it needs two exclusions I hadn't anticipated.

**Xcode** behaves the way you describe, and it does honour the property. I tested on a copy of `Calendar_Recurrence.swift`, which carries trailing whitespace on 21 lines of code and indentation on 58 blank ones. With `trim_trailing_whitespace = true`, editing one line and saving trimmed that line and nothing else — the other 20 code lines and all 58 indented blank lines came back byte-identical. Flipping the same fixture to `trim_trailing_whitespace = false` and repeating the edit left the file byte-for-byte unchanged, so it really is the property driving this rather than an editor preference that would have applied either way.

**Zed** doesn't, as far as I can tell from the source. `.editorconfig`'s `trim_trailing_whitespace` maps onto `remove_trailing_whitespace_on_save`, whose modified-row ranges are `Some` only when formatting a selection — a plain save passes `None`, which means the whole buffer:

https://github.com/zed-industries/zed/blob/f25b256f2c10e1b638031a3d8d5a524056ebf268/crates/project/src/lsp_store.rs#L1631-L1665

That cuts in favour of enabling it, though: `remove_trailing_whitespace_on_save` already defaults to `true`, so Zed rewrites whole files today with or without this file. Adding the property doesn't make Zed any worse — it just stops being invisible.

**VS Code** is where the behaviour actually changes. `files.trimTrailingWhitespace` defaults to `false`, and the EditorConfig extension maps the property onto `editor.action.trimTrailingWhitespace` (or, when the document isn't the active editor, a loop over every line) — whole-document either way:

https://github.com/editorconfig/editorconfig-vscode/blob/4da6dd253eccf6182f384d3518c3dd72bd197ac4/src/transformations/TrimTrailingWhitespace.ts#L43-L61

So contributors on VS Code would get a full-file diff the first time they save any of the 302 non-Markdown files that currently carry trailing whitespace. Given your point that we can follow up with a linter pass, I've gone ahead and enabled it — that pass would make the question moot.

Two exclusions came out of the check, and I don't think either can be skipped:

- **`[*.md] trim_trailing_whitespace = false`.** Markdown gives two trailing spaces the meaning of a hard line break, and nine lines across `Proposals/0023-progress-manager.md` and `Proposals/0024-CurrentBundle.md` render as line breaks because of it. Trimming them would quietly change how accepted proposals display.
- **`trim_trailing_whitespace = unset` in the two `Resources` sections**, alongside the properties already unset there. Four fixtures carry trailing whitespace — `JSON/pass/pass3.json`, `JSON/pass/pass4.json`, `JSON5/pass/hex.json5` and, fittingly, `JSON5/pass/whitespace.json5`.

I re-ran the reference `editorconfig` implementation over the same spot checks; `Proposals/*.md` now resolves to `trim_trailing_whitespace=false`, everything under `Resources/` to `unset`, and the nested `new-lines/.editorconfig` still wins with `end_of_line=cr`.

Happy to do the sweep as a follow-up PR whenever you'd like it: 6,653 lines across 298 Swift files, plus 13 in one C file and 6 in three `.txt` files. It is tamer than the count makes it sound — 6,576 of those Swift lines are blank lines carrying indentation, and only 77 have code in front of the whitespace. No multiline string literal in the repository has trailing whitespace either, so trimming the Swift sources cannot change the contents of any string.

parkera (changes requested):
I think we should do this as part of a more holistic view of linter rules, which I know @itingliu is working on.

maxches99:
@parkera Understood — I'd rather defer to #2052 than land a second source of style truth.

Two questions so I know what to do with this PR:

1. Hold it open until #2052 lands, or close it and leave #2119 as the tracker? #2052 is still a draft and hasn't been updated since 15 July, so I don't want this sitting in your review queue in the meantime.
2. If it's held: should the `.editorconfig` become part of #2052 itself, or stay a follow-up once the swift-format rules are settled?

Either way, two findings from this PR are worth carrying over to whatever does land: the fixtures under `Tests/**/Resources/**` and `Benchmarks/**/Resources/**` need every property `unset` — many deliberately end without a final newline, some use CR or CRLF on purpose, and some are binary plists — and `Proposals/*.md` needs `trim_trailing_whitespace = false`, since nine lines there rely on two trailing spaces as hard line breaks.

.editorconfig
maxches99: Agreed — I will drop the `[*.{yml,yaml}]` section if this PR carries on (see the question to @parkera above about whether it should).

itingliu:
@maxches99 Thanks for taking a look at this. After talking to @iCharlesHu and @parkera I think we'd want to hold this off until we have a linter/format setup. That way we can fix everything altogether, and avoid creating a bunch of diffs when this editorconfig goes into effect while the rest of the project is still malformatted.

PR #2154, opened 2026-08-03
fr_CA: 24-hour setting is incorrect

itingliu:
Previously we worked around an ICU issue where it didn't handle the "rg" keyword in locale identifiers.
    
Our implementation treats "fr_CA" as "und_CA" under the assumption that the language code doesn't take effect for determining the properties such as the preferred hour cycle of the locale.
    
However this assumption is incorrect; in this case, fr_CA has a different hour cycle preference from en_CA.

Fix: Remove this workaround altogether since the ICU issue has been addressed.


itingliu:
@swift-ci please test

jmschonfeld (approved):

PR #2164, opened 2026-08-06
Switch HAS_FOUNDATION_DARWIN_EXTRAS to canImport(_FoundationDarwinExtras)

jmschonfeld:
Switches the custom define for a canImport check

### Motivation:

Rather than using a custom define to declare when the module exists / needs to be imported, we can just check canImport instead

### Modifications:

Replaces `HAS_FOUNDATION_DARWIN_EXTRAS` with `canImport(_FoundationDarwinExtras)`

### Result:

Removed reliance on a custom define

### Testing:

Validated the project still builds

### DIFF
--- a/Sources/FoundationEssentials/Platform.swift
+++ b/Sources/FoundationEssentials/Platform.swift
@@ -50,7 +50,7 @@ import C.strings
 import C
 #endif
 fileprivate let _pageSize: Int = Int(getpagesize())
-#elseif HAS_FOUNDATION_DARWIN_EXTRAS
+#elseif canImport(_FoundationDarwinExtras)
 internal import _FoundationDarwinExtras
 internal import _FoundationDarwinExtras._string_runtime.xlocale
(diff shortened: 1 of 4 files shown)


Sources/FoundationEssentials/Platform.swift
```diff
@@ -396,16 +396,16 @@ extension Platform {
 }
 
 extension Platform {
-    #if canImport(Darwin) || HAS_FOUNDATION_DARWIN_EXTRAS
+    #if canImport(Darwin) || canImport(_FoundationDarwinExtras)
     private static var cLocale: locale_t? { /* LC_C_LOCALE */ nil }
     #elseif os(Windows)
-    private static var cLocale: _locale_t = {
+    private static let cLocale: _locale_t = {
```
jmschonfeld, line 402: Just a drive-by warning fix while I was here

iCharlesHu (approved):

PR #2165, opened 2026-08-07
Fix month rolling, the week-of-month day range, and week-year date resolution in the Chinese and Hebrew calendars

dra8an:
Fixes two independent Chinese calendar bugs: a month roll that carried into the year, and a nil day range within a week of the month.

### Motivation:

Both showed up when running the Chinese calendar against a larger test suite.

### Modifications:

`date(byAdding:to:wrappingComponents:)` had a wrapping branch for `.day` but none for `.month`, so a month roll fell through to the same month walk a plain add uses, and that walk crosses year boundaries. Added a `.month` wrapping branch that wraps the month's ordinal position over the year's own month count, 12 or 13,.

  `date(from:)` in both calendars required `.year` and had no week of year path, so handing back the week fields `dateComponents` had just reported returned nil.  Added a week year branch built on the `firstDayOfWeekYear` helper both calendars already had, so the first week rule is reused rather than reimplemented. It is only taken when `.year` is absent, so nothing that already worked changes. For the Chinese calendar `yearForWeekOfYear` is an extended year rather than a cycle year, so it is used directly.

### Result:

`(.day, .weekOfMonth)` was not a case in ordinality, so the generic range fallback gave up on its first lookup and returned nil. Added that pair, and (.weekday, .weekOfMonth), to the existing case that computes a day's position inside its week

### Testing:

Six new tests across ChineseCalendarTests and HebrewCalendarTests, all confirmed to fail on the unfixed source and to pass with the fix.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
@@ -548,7 +548,7 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
             let end = localMidnight(extendedYear: extendedYear + 1, ordinal: 1, day: 1, in: timeZone)
             return DateInterval(start: start, duration: end.timeIntervalSince(start))
         case .yearForWeekOfYear:
-            // Deliberate divergence from ICU: ICU's chinese calendar cannot use YEAR_WOY on the fields-to-time side (chnsecal handleGetExtendedYear never reads it), so its interval degenerates to nil and its add is a no-op. We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval). If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:).
+            // Deliberate divergence from ICU: ICU's chinese calendar returns nil for this interval and no-ops an add of yearForWeekOfYear, both measured. We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval). ICU is asymmetric here, it does resolve the week-year form in date(from:), and we match that exactly. If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:), and leave date(from:) as it is.
             let weekYearComps = dateComponents([.yearForWeekOfYear], from: date, in: timeZone)
             guard let weekYear = weekYearComps.yearForWeekOfYear else { return nil }
             let rdStart = firstDayOfWeekYear(weekYear)
@@ -623,12 +623,31 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
     // MARK: Date ↔ DateComponents
 
     func date(from components: DateComponents) -> Date? {
+        var secondsInDay: Double = 0
+        if let hour = components.hour { secondsInDay += Double(hour) * 3600 }
+        if let minute = components.minute { secondsInDay += Double(minute) * 60 }
+        if let second = components.second { secondsInDay += Double(second) }
+        if let nanosecond = components.nanosecond { secondsInDay += Double(nanosecond) / 1e9 }
+
+        let timeZone = components.timeZone ?? timeZone
+
+        // A request carrying yearForWeekOfYear instead of year resolves through the week-year, the way the Gregorian calendar does, so the week fields we report in dateComponents can be handed back to us. yearForWeekOfYear is an extended year for this calendar (see the field conventions note above), so it is used directly rather than combined with era; the Gregorian calendar likewise skips its era adjustment once a week is specified. The week number is not clipped to the week-year: as in the Gregorian calendar, a week past the end simply runs on into the next year.
+        if components.year == nil, let weekYear = components.yearForWeekOfYear {
+            guard weekYear > Self.extendedYearLowerBound && weekYear < Self.extendedYearUpperBound else { return nil }
+            let weekOfYear = components.weekOfYear ?? 1
+            let weekday = components.weekday ?? firstWeekday
+            let dayWithinWeek = ((weekday - firstWeekday) % 7 + 7) % 7
+            let rataDie = firstDayOfWeekYear(weekYear) + (weekOfYear - 1) * 7 + dayWithinWeek
+            return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
+                           repeatedTimePolicy: .former, skippedTimePolicy: .former)
+        }
+
         // Missing era defaults to the CURRENT date's era (ICU fields default from now).
         let era: Int
         if let e = components.era {
             era = e
         } else {
-            let (nowExt, _, _) = fields(for: Date.now, in: components.timeZone ?? timeZone)
+            let (nowExt, _, _) = fields(for: Date.now, in: timeZone)
             era = Self.eraAndYear(extendedYear: nowExt).era
         }
         guard let yearValue = components.year else { return nil }
@@ -655,13 +674,6 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
 
         let rataDie = y.monthStartRataDie(ordinal: ordinal) + day - 1
 
-        var secondsInDay: Double = 0
-        if let hour = components.hour { secondsInDay += Double(hour) * 3600 }
-        if let minute = components.minute { secondsInDay += Double(minute) * 60 }
-        if let second = components.second { secondsInDay += Double(second) }
-        if let nanosecond = components.nanosecond { secondsInDay += Double(nanosecond) / 1e9 }
-
-        let timeZone = components.timeZone ?? timeZone
         return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
                        repeatedTimePolicy: .former, skippedTimePolicy: .former)
     }
--- a/Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
@@ -372,7 +372,7 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
             guard let span = Self.quarterSpan(extendedYear: extendedYear, ordinal: ordinal) else { return nil }
             return Range(span.firstDisplay...span.lastDisplay)
         case (.day, .weekOfMonth):
-            // The day-of-month values this week covers, the same shape the Gregorian calendar reports. A week can start before or end after the month, so clip to the month's own day range.
+            // Which days of the month this week covers, as day-of-month values. A week can start before or end after the month, so clip both ends.
             guard let week = dateInterval(of: .weekOfMonth, for: date) else { return nil }
             let (extendedYear, ordinal, _) = fields(for: date, in: timeZone)
             let monthLength = Self.yearData(extendedYear: extendedYear).monthLength(ordinal: ordinal)
@@ -381,8 +381,7 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
             let (weekStart, _): (Int, Double) = _CalendarUtility.rataDieAndSecondsInDay(localSeconds: localSeconds)
             let firstDay = max(1, weekStart - monthStart + 1)
             let lastDay = min(monthLength, weekStart - monthStart + 7)
-            guard firstDay <= lastDay else { return nil }
-            return Range(firstDay...lastDay)
+            return firstDay..<lastDay + 1
         case (.day, .quarter):
             // ICU counts calendar days here, not 86400 s chunks, the generic interval+ordinality fallback overcounts by one in DST fall-back quarters.
             let (extendedYear, ordinal, _) = fields(for: date, in: timeZone)
@@ -548,7 +547,7 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
             let end = localMidnight(extendedYear: extendedYear + 1, ordinal: 1, day: 1, in: timeZone)
             return DateInterval(start: start, duration: end.timeIntervalSince(start))
         case .yearForWeekOfYear:
-            // Deliberate divergence from ICU: ICU's chinese calendar returns nil for this interval and no-ops an add of yearForWeekOfYear, both measured. We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval). ICU is asymmetric here, it does resolve the week-year form in date(from:), and we match that exactly. If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:), and leave date(from:) as it is.
+            // The year that owns a week, which can differ from the calendar year at either end of a year. ICU returns nil here for the Chinese calendar; we deliberately return a real interval, as the Gregorian and Hebrew calendars do.
             let weekYearComps = dateComponents([.yearForWeekOfYear], from: date, in: timeZone)
             guard let weekYear = weekYearComps.yearForWeekOfYear else { return nil }
             let rdStart = firstDayOfWeekYear(weekYear)
@@ -631,15 +630,14 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
 
         let timeZone = components.timeZone ?? timeZone
 
-        // A request carrying yearForWeekOfYear instead of year resolves through the week-year, the way the Gregorian calendar does, so the week fields we report in dateComponents can be handed back to us. yearForWeekOfYear is an extended year for this calendar (see the field conventions note above), so it is used directly rather than combined with era; the Gregorian calendar likewise skips its era adjustment once a week is specified. The week number is not clipped to the week-year: as in the Gregorian calendar, a week past the end simply runs on into the next year.
+        // `.yearForWeekOfYear` plus `.weekOfYear` plus `.weekday` names a date too. `.era` is unused here because `.yearForWeekOfYear` is already an extended year. A `.weekOfYear` past the end of the year runs on into the next, as in the Gregorian calendar.
         if components.year == nil, let weekYear = components.yearForWeekOfYear {
             guard weekYear > Self.extendedYearLowerBound && weekYear < Self.extendedYearUpperBound else { return nil }
             let weekOfYear = components.weekOfYear ?? 1
             let weekday = components.weekday ?? firstWeekday
             let dayWithinWeek = ((weekday - firstWeekday) % 7 + 7) % 7
             let rataDie = firstDayOfWeekYear(weekYear) + (weekOfYear - 1) * 7 + dayWithinWeek
-            return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
-                           repeatedTimePolicy: .former, skippedTimePolicy: .former)
+            return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone, repeatedTimePolicy: .former, skippedTimePolicy: .former)
         }
 
         // Missing era defaults to the CURRENT date's era (ICU fields default from now).
@@ -647,8 +645,8 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
         if let e = components.era {
             era = e
         } else {
-            let (nowExt, _, _) = fields(for: Date.now, in: timeZone)
-            era = Self.eraAndYear(extendedYear: nowExt).era
+            let (nowExtendedYear, _, _) = fields(for: Date.now, in: timeZone)
+            era = Self.eraAndYear(extendedYear: nowExtendedYear).era
         }
         guard let yearValue = components.year else { return nil }
         guard let extendedYear = Self.extendedYear(era: era, year: yearValue),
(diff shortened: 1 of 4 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -371,6 +371,18 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
             let (extendedYear, ordinal, _) = fields(for: date, in: timeZone)
             guard let span = Self.quarterSpan(extendedYear: extendedYear, ordinal: ordinal) else { return nil }
             return Range(span.firstDisplay...span.lastDisplay)
+        case (.day, .weekOfMonth):
+            // The day-of-month values this week covers, the same shape the Gregorian calendar reports. A week can start before or end after the month, so clip to the month's own day range.
```
itingliu: > the same shape the Gregorian calendar reports

What does this refer to? Also I think the algorithm does look similar to Gregorian calendar. maybe see if we can consolidate these as a follow up?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -371,6 +371,18 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+            guard let week = dateInterval(of: .weekOfMonth, for: date) else { return nil }
+            let (extendedYear, ordinal, _) = fields(for: date, in: timeZone)
+            let monthLength = Self.yearData(extendedYear: extendedYear).monthLength(ordinal: ordinal)
+            let monthStart = Self.rataDie(extendedYear: extendedYear, ordinal: ordinal, day: 1)
+            let localSeconds = week.start.timeIntervalSinceReferenceDate + Double(timeZone.secondsFromGMT(for: week.start))
```
itingliu, line 380: Also as a follow-up idea: We have lots of those use of "local time" concept throughout calendar calculation. It might be worthwhile adding a helper in `Calendar` for something like this
```swift
extension Calendar {
     func localDate(from date: Date) -> Date {
          return timeIntervalSinceReferenceDate + Double(timeZone.secondsFromGMT(for: self))
    }
}

File: Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift [hunk 3]
```diff
@@ -371,6 +371,18 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+            let (weekStart, _): (Int, Double) = _CalendarUtility.rataDieAndSecondsInDay(localSeconds: localSeconds)
+            let firstDay = max(1, weekStart - monthStart + 1)
+            let lastDay = min(monthLength, weekStart - monthStart + 7)
+            guard firstDay <= lastDay else { return nil }
```
itingliu: Are there any scenarios where `firstDay > lastDay`? If there are, would it be more correct to return `lastDay...firstDay` instead of nil?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -371,6 +371,18 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+            return Range(firstDay...lastDay)
```
itingliu: Is there any reason why you don't want to just write this?
```suggestion
            return firstDay..<lastDay+1
```

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -535,7 +548,7 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
             return DateInterval(start: start, duration: end.timeIntervalSince(start))
         case .yearForWeekOfYear:
-            // Deliberate divergence from ICU: ICU's chinese calendar cannot use YEAR_WOY on the fields-to-time side (chnsecal handleGetExtendedYear never reads it), so its interval degenerates to nil and its add is a no-op. We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval). If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:).
+            // Deliberate divergence from ICU: ICU's chinese calendar returns nil for this interval and no-ops an add of yearForWeekOfYear, both measured. We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval). ICU is asymmetric here, it does resolve the week-year form in date(from:), and we match that exactly. If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:), and leave date(from:) as it is.
```
itingliu: > no-ops an add of yearForWeekOfYear, both measured

> If behavior identical to ICU is ever required

Can we write this paragraph in [plain English](https://www.plainenglish.co.uk/free-guides)? 

> We implement the Gregorian-family week-year semantics instead, like Hebrew (precedent: Japanese .era interval).

This is confusing. Unless I miss something, how is Hebrew a Gregorian-family calendar?

> If behavior identical to ICU is ever required: return nil here and delete the yearForWeekOfYear block in date(byAdding:), and leave date(from:) as it is.

I'm not sure if this is helpful. If we think this current behavior is correct, we should just do it, regardless of what ICU does. Also even if I ever want to change this to match that of ICU, I still wouldn't know for sure which "yearForWeekOfYear block in date(byAdding:)" to delete if I want to change the behavior to match that of ICU.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -610,12 +623,31 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
     func date(from components: DateComponents) -> Date? {
+        var secondsInDay: Double = 0
+        if let hour = components.hour { secondsInDay += Double(hour) * 3600 }
+        if let minute = components.minute { secondsInDay += Double(minute) * 60 }
+        if let second = components.second { secondsInDay += Double(second) }
+        if let nanosecond = components.nanosecond { secondsInDay += Double(nanosecond) / 1e9 }
+
+        let timeZone = components.timeZone ?? timeZone
+
+        // A request carrying yearForWeekOfYear instead of year resolves through the week-year, the way the Gregorian calendar does, so the week fields we report in dateComponents can be handed back to us. yearForWeekOfYear is an extended year for this calendar (see the field conventions note above), so it is used directly rather than combined with era; the Gregorian calendar likewise skips its era adjustment once a week is specified. The week number is not clipped to the week-year: as in the Gregorian calendar, a week past the end simply runs on into the next year.
```
itingliu: Similar comment about this being hard to parse. Please explain this in plain English.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -610,12 +623,31 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+        if components.year == nil, let weekYear = components.yearForWeekOfYear {
+            guard weekYear > Self.extendedYearLowerBound && weekYear < Self.extendedYearUpperBound else { return nil }
+            let weekOfYear = components.weekOfYear ?? 1
+            let weekday = components.weekday ?? firstWeekday
+            let dayWithinWeek = ((weekday - firstWeekday) % 7 + 7) % 7
+            let rataDie = firstDayOfWeekYear(weekYear) + (weekOfYear - 1) * 7 + dayWithinWeek
+            return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
+                           repeatedTimePolicy: .former, skippedTimePolicy: .former)
```
itingliu: what's up with this being in a new line?

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -610,12 +623,31 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+        }
+
         // Missing era defaults to the CURRENT date's era (ICU fields default from now).
         let era: Int
         if let e = components.era {
             era = e
         } else {
-            let (nowExt, _, _) = fields(for: Date.now, in: components.timeZone ?? timeZone)
+            let (nowExt, _, _) = fields(for: Date.now, in: timeZone)
```
itingliu: Can we avoid using abbreviations for `nowExt`? Can we just call it `nowExtendedYear`

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -788,26 +813,53 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
 
     // MARK: Adding
 
+    // How many components carry a non-zero value. The roll shortcuts below apply only when a single field is set, so they can wrap that one field and leave everything else alone.
```
itingliu: This comment seems to suggest that we only add this function for the rolling function. In that case I don't think it needs to be a separate `static func` on its own.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -788,26 +813,53 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+        if wrappingComponents, Self.nonZeroComponentCount(components) == 1 {
             let timeZone = self.timeZone
-            let (extendedYear, ordinal, curDay, secondsInDay) = fieldsAndTime(for: result, in: timeZone)
-            let y = Self.yearData(extendedYear: extendedYear)
-            let monthLen = y.monthLength(ordinal: ordinal)
-            let newDay = ((curDay - 1 + d) % monthLen + monthLen) % monthLen + 1
-            let rataDie = y.monthStartRataDie(ordinal: ordinal) + newDay - 1
-            return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
-                           repeatedTimePolicy: .former, skippedTimePolicy: .former)
+            if let d = components.day, d != 0 {
```
itingliu: Similar comment about avoiding abbreviation for `d`

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -788,26 +813,53 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+            if let m = components.month, m != 0 {
+                let (extendedYear, ordinal, curDay, secondsInDay) = fieldsAndTime(for: result, in: timeZone)
+                let y = Self.yearData(extendedYear: extendedYear)
+                let monthCount = Int(y.monthCount)
+                let newOrdinal = ((ordinal - 1 + m) % monthCount + monthCount) % monthCount + 1
+                // A shorter target month clamps the day, matching the non-wrapping month add.
+                let clampedDay = min(curDay, y.monthLength(ordinal: newOrdinal))
+                let rataDie = y.monthStartRataDie(ordinal: newOrdinal) + clampedDay - 1
+                return _CalendarUtility.utcDate(fromRataDie: rataDie, secondsInDay: secondsInDay, in: timeZone,
+                               repeatedTimePolicy: .former, skippedTimePolicy: .former)
```
itingliu: Similar comment about consistent line breaking

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
```diff
@@ -788,26 +813,53 @@ internal final class _CalendarChinese: _CalendarProtocol, @unchecked Sendable {
...
+            }
+            // The month wraps around this year's own month count, 12 or 13 when there is a leap month, so the year never carries. Walking months the way a plain add does would cross into the next year.
+            if let m = components.month, m != 0 {
+                let (extendedYear, ordinal, curDay, secondsInDay) = fieldsAndTime(for: result, in: timeZone)
+                let y = Self.yearData(extendedYear: extendedYear)
+                let monthCount = Int(y.monthCount)
+                let newOrdinal = ((ordinal - 1 + m) % monthCount + monthCount) % monthCount + 1
+                // A shorter target month clamps the day, matching the non-wrapping month add.
```
itingliu: > the non-wrapping month add.

Is this grammar correct?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -587,6 +587,31 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
+        // Time-of-day in local seconds (defaults to midnight).
+        var secondsInDay: Double = 0
+        if let hour = components.hour { secondsInDay += Double(hour) * 3600 }
+        if let minute = components.minute { secondsInDay += Double(minute) * 60 }
+        if let second = components.second { secondsInDay += Double(second) }
+        if let nanosecond = components.nanosecond { secondsInDay += Double(nanosecond) / 1e9 }
+
+        let tz = components.timeZone ?? timeZone
+
+        // A request carrying yearForWeekOfYear instead of year resolves through the
```
itingliu: Similar comment about consistent styling. As you can tell we don't have a line length limit in this repo, so no need wrapping the comments like this

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -587,6 +587,31 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
+        // week-year, the way the Gregorian calendar does, so the week fields we report
```
itingliu: What does "week-year" mean? Can you map this to a `DateComponent` or `Calendar.Component` that we use this in our project?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -587,6 +587,31 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
+        // in dateComponents can be handed back to us. `firstDayOfWeekYear` already
+        // applies firstWeekday and minimumDaysInFirstWeek. The week number is not
+        // clipped to the week-year: as in the Gregorian calendar, a week past the end
+        // simply runs on into the next year.
+        if components.year == nil, let weekYearValue = components.yearForWeekOfYear {
+            guard weekYearValue >= Int(Int32.min) && weekYearValue <= Int(Int32.max) else { return nil }
```
itingliu, line 602: Why cap this at Int32's range?

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
```diff
@@ -587,6 +587,31 @@ internal final class _CalendarHebrew: _CalendarProtocol, @unchecked Sendable {
...
+            let weekOfYear = components.weekOfYear ?? 1
+            let weekday = components.weekday ?? firstWeekday
+            let dayWithinWeek = ((weekday - firstWeekday) % 7 + 7) % 7
+            let rd = firstDayOfWeekYear(Int32(weekYearValue)) &+ Int64((weekOfYear - 1) * 7 + dayWithinWeek)
+            return utcDate(fromRataDie: rd, secondsInDay: secondsInDay, in: tz,
+                          repeatedTimePolicy: .former, skippedTimePolicy: .former)
```
itingliu: ditto about arbitrary line break

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -115,7 +115,7 @@ private struct ChineseCalendarTests {
         #expect(c.maximumRange(of: .weekOfYear) == 1..<56)
     }
 
-    // Deliberate divergence from ICU, guarded: ICU's chinese calendar cannot use YEAR_WOY on the fields-to-time side (chnsecal handleGetExtendedYear ignores it), yielding a nil interval and a no-op add; we implement Gregorian-family week-year semantics instead (precedent: the Japanese calendar's .era interval). If this test is changed to expect nil/no-op, that reversion must be an explicit decision.
+    // Deliberate divergence from ICU, guarded: ICU's chinese calendar returns nil for the yearForWeekOfYear interval and no-ops an add of it, so we implement Gregorian-family week-year semantics instead (precedent: the Japanese calendar's .era interval). ICU is asymmetric here, it does resolve the week-year form in date(from:), which dateFromWeekYearComponents pins against ICU's own answers. If this test is changed to expect nil/no-op, that reversion must be an explicit decision.
```
itingliu: same comment as the one above. Also how does this comment help? What am I going to do with this information in the future?

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
             #expect((dc.day ?? 0) >= 1, "\(gy)-\(gm)-\(gd)")
         }
     }
+
+    // A roll wraps the month inside its own year and never carries into the year, and it wraps around this year's own month count, 12 or 13 when there is a leap month. Regression: the wrapping path had a branch for .day but none for .month, so a roll fell through to the month walk a plain add uses, and that walk crosses year boundaries.
```
itingliu: > Regression: the wrapping path had a branch for .day but none for .month, so a roll fell through to the month walk a plain add uses, and that walk crosses year boundaries.

I don't understand how this comment is helpful either. It sounds like it's explaining the bug itself, but that will have been irrelevant after this fix lands

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+        for (iso, monthCount, amount, wantMonth) in cases {
+            let y = _CalendarChinese.year(relatedISOYear: iso)
+            #expect(y.monthCount == monthCount, "fixture \(iso): month count is \(y.monthCount)")
+            let from = Self.date(rataDie: y.monthStartRataDie(ordinal: 3) + 4)   // ordinal month 3, day 5
+            let before = c.dateComponents([.year, .month, .day], from: from, in: .gmt)
+            var dc = DateComponents()
+            dc.month = amount
+            guard let rolled = c.date(byAdding: dc, to: from, wrappingComponents: true) else {
+                #expect(Bool(false), "\(iso) roll by \(amount) returned nil")
+                continue
```
itingliu: Can we take advantage of swift-testing's `#require` here?

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+            }
+            let after = c.dateComponents([.year, .month, .day], from: rolled, in: .gmt)
+            #expect(after.year == before.year, "\(iso) roll by \(amount) changed the year, \(before.year ?? -1) to \(after.year ?? -1)")
+            #expect(after.month == wantMonth, "\(iso) roll by \(amount): month \(after.month ?? -1), wanted \(wantMonth)")
+            #expect(after.day == 5, "\(iso) roll by \(amount): day \(after.day ?? -1)")
+        }
+        // The same amount as a plain add does carry into the next year. That difference is the whole point of the wrapping branch, so pin it too.
```
itingliu: What does "plain" mean in "plain add"? Can we just use the common terminology that we use in this repo instead of arbitrary adjective like "plain", e.g. "add without wrapping"?

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+        let y = _CalendarChinese.year(relatedISOYear: 2024)
+        let from = Self.date(rataDie: y.monthStartRataDie(ordinal: 3) + 4)
+        var dc = DateComponents()
+        dc.month = 11
+        let added = c.date(byAdding: dc, to: from, wrappingComponents: false)
+        let before = c.dateComponents([.year], from: from, in: .gmt)
+        let after = added.map { c.dateComponents([.year, .month], from: $0, in: .gmt) }
+        #expect(after?.year == (before.year ?? 0) + 1 && after?.month == 2,
+                "add of 11 months: y\(after?.year ?? -1)/m\(after?.month ?? -1)")
```
itingliu: unnecessary and arbitrary line break

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+    }
+
+    // range(of: .day, in: .weekOfMonth) reports the day-of-month values the week covers, clipped to the month at both ends because a week can start before the month or run past its end. Regression: the (.day, .weekOfMonth) pair was missing from ordinality, so the generic range fallback gave up on the first lookup and returned nil.
```
itingliu: same note about the description after "Regression:" being unhelpful

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+        let y = _CalendarChinese.year(relatedISOYear: 2024)   // 12 months, new year Feb 10; firstWeekday is Sunday here
+        // (ordinal month, day of month, expected range)
+        let cases: [(Int, Int, Range<Int>)] = [
+            (1, 1, 1..<2),      // the month starts on a Saturday, so its first week holds a single day
+            (1, 15, 9..<16),    // a whole week, clipped at neither end
+            (3, 1, 1..<6),      // first week, clipped by the month start
+            (2, 30, 29..<31),   // last week, clipped by the month end
+            (3, 29, 27..<30),   // last week, clipped by the month end
+        ]
+        for (ordinal, day, want) in cases {
```
itingliu: Can we take advantage of parameterized testing for these cases?

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
```diff
@@ -166,4 +166,146 @@ private struct ChineseCalendarTests {
...
+                for day in 1...length {
+                    let d = Self.date(rataDie: start + day - 1)
+                    guard let r = c.range(of: .day, in: .weekOfMonth, for: d) else {
+                        failures.append("\(iso)/o\(ordinal)/d\(day): nil range")
+                        continue
+                    }
+                    if r.isEmpty || r.lowerBound < 1 || r.upperBound > length + 1 || !r.contains(day) {
+                        failures.append("\(iso)/o\(ordinal)/d\(day): \(r) in a \(length) day month")
+                    }
+                    guard let o = c.ordinality(of: .day, in: .weekOfMonth, for: d) else {
```
itingliu: same comment about using "#require". Also applies to rest of the tests

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: It meant that Gregorian returns day-of-month values here rather than 1...7, and I copied that. Reworded to say it directly, without the comparison. Agreed on consolidating, the two are close enough to share; I have not done it here to keep this PR to the fixes.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Agreed, this pattern is repeated across the calendars. Left out of this PR and will address it in a separate one.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: No, it cannot. I measured the week start offset from the month start over 1950 to 2050 across six `firstWeekday` and `minimumDaysInFirstWeek `combinations: it stays within `-6...29`, and `firstDay > lastDay` needs an offset below -6 or above monthLength - 1. So the guard was dead code and I removed it rather than adding a reversed range case.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: No reason. Updated.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Rewritten and shortened.

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Rewritten

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Moved to a nested function inside date(byAdding:to:wrappingComponents:)

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Renamed

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Chinese.swift
dra8an: Reworded

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: The comment now names .yearForWeekOfYear, .weekOfYear and .weekday directly so they can be grepped.

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: the Hebrew arithmetic underneath works in Int32 years: `firstDayOfWeekYear(_ year: Int32)` and `HebrewArithmetic.fixedFromHebrew(year: Int32, ...)`

Sources/FoundationEssentials/Calendar/Calendar_Hebrew.swift
dra8an: Fixed

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Rewritten and shortened

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Agreed. Removed, here and in the other two places I had written one.

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Done. Used in the pinned tests. Not used in the two sweep tests, which deliberately collect every failure into a list and report the count and the first few; #require would stop at the first and hide the rest. Happy to change that if you would rather they fail fast.

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Reworded

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Fixed

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Removed

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Done. rollMonthWrapsWithinYear and dayRangeInWeekOfMonth are now @Test(arguments:).

Tests/FoundationEssentialsTests/ChineseCalendarTests.swift
dra8an: Done.

itingliu (approved):
thanks!

PR #2166, opened 2026-08-09
Honor `repeatedTimePolicy` when matching nanoseconds

### LINKED ISSUE #483: `Calendar.dates(byMatching: ...)` does not respect `repeatedTimePolicy` when `.nanosecond` is set
When searching by dates matching date components, the `repeatedTimePolicy` field is not respected if the `nanoseconds` field is set. For example, we want to find the date `expected` which is just after a daylight switch:

```swift
let start    = Date(timeIntervalSince1970: 1730535600.0) // 2024-11-02T01:20:00-0700
let expected = Date(timeIntervalSince1970: 1730625600.0) // 2024-11-03T01:20:00-0800
        
var dc = DateComponents()
dc.hour = 1
dc.minute = 20
dc.second = 0
dc.nanosecond = 0
        
var sequenceIterator = gregorian.dates(byMatching: dc, startingAt: start,  matchingPolicy: .strict, repeatedTimePolicy: .last).makeIterator()
let result = sequenceIterator.next()
assert(result == expected)
```

when `nanosecond` is set to a value, enumeration always returns the 01:20 before the switch and the assertion fails.

The bug occurs in `Calendar.dateAfterMatchingNanosecond(...)` in `Calendar_Enumerate.swift`. This method simply returns a new date using `Calendar.date(from: DateComponents)`, but in the test above there are two dates matching the components, and we need a way to specify if we want the first or second one.

cayde-6:
Fix `Calendar.dates(byMatching:)` returning the wrong occurrence of a repeated hour when `DateComponents.nanosecond` is set.

### Motivation:

Resolves #483 (rdar://124640028).

`Calendar.enumerateDates` and `Calendar.dates(byMatching:)` accept a `repeatedTimePolicy` that chooses which occurrence to return when a wall clock time happens twice, at the end of daylight saving time. The policy is honored when `DateComponents.nanosecond` is `nil` and silently ignored when it is set to any value, including `0`, so callers that spell out every time component always get the earlier occurrence.

```swift
var cal = Calendar(identifier: .gregorian)
cal.timeZone = TimeZone(identifier: "America/Los_Angeles")!

var dc = DateComponents()
dc.hour = 1; dc.minute = 20; dc.second = 0
dc.nanosecond = 0   // remove this line and the result is correct

let start = Date(timeIntervalSince1970: 1730536000.0)  // 2024-11-02T01:26:40-0700
var iterator = cal.dates(byMatching: dc, startingAt: start,
                         matchingPolicy: .strict, repeatedTimePolicy: .last).makeIterator()

// expected 2024-11-03T01:20:00-0800, got 2024-11-03T01:20:00-0700
```

### Modifications:

`dateAfterMatchingHour` already applies the policy, so by the time the nanosecond step runs the candidate is the correct instant. `dateAfterMatchingNanosecond` then decomposed it into era/year/month/day/hour/minute/second, set the nanoseconds, and resolved the whole thing again through `date(from:)`. Inside a repeated hour one set of wall clock components maps to two instants and `date(from:)` always resolves to the earlier one, so the selected occurrence could not survive the round trip — the choice simply is not encoded in the components.

Substituting a nanosecond value is arithmetic on the absolute time scale rather than a calendrical operation, so the function now truncates to the start of the containing second and adds the nanoseconds. This matches the neighboring helpers: `dateAfterMatchingSecond` already truncates with `dateInterval(of: .second, for:)`, and none of them round-trip through `date(from:)`.

### Result:

`repeatedTimePolicy` is honored regardless of whether `nanosecond` is set. The fix is calendar- and timezone-agnostic, needs no public API change, and removes a `date(from:)` call from the enumeration path. Behavior outside repeated hours is unchanged.

### Testing:

Three tests added to `CalendarTests`, all around the DST transition in `America/Los_Angeles` on 2024-11-03, where local 01:20 occurs twice.

The forward test is parameterized over `nanosecond` of `nil`, `0` and `500_000_000` paired with the sub-second offset each should produce, and pins both policies to their expected instants. The `nil` case passes before the change and guards against regressing the already-working path; the `.last` cases with `nanosecond` set are the bug.

Searching backwards, the same round trip misroutes `.first` rather than `.last`. Which occurrence each policy selects in that direction is pre-existing behavior that looks surprising on its own, so rather than pin it down the backward test asserts the property this change is actually about: setting `nanosecond` must not change which instant is selected. It is parameterized over both policies and compares against the `nanosecond: nil` result. A third test asserts the two policies disagree, so the second cannot pass vacuously.

Full `FoundationEssentials`, `FoundationInternationalization` and `FoundationMacros` suites pass on Linux (1584 tests, `swiftlang/swift:nightly-main-jammy`).

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -2370,13 +2370,7 @@ extension Calendar {
             return nil
         }
 
-        // Substituting nanoseconds is arithmetic on the absolute time scale, so do it
-        // directly instead of decomposing into date components and resolving them
-        // again. Within a repeated hour, one set of wall clock components maps to two
-        // instants and date(from:) always picks the first, which would discard the
-        // occurrence that dateAfterMatchingHour selected via repeatedTimePolicy.
-        // This also avoids treating nanoseconds like seconds and iterating over them,
-        // which causes a hang. <rdar://problem/30229247>
+        // Substitute nanoseconds on the absolute time scale, instead of decomposing into date components and resolving them again. Within a repeated hour, one set of wall clock components maps to two instants. date(from:) always picks the first, which would discard the occurrence that dateAfterMatchingHour selected via repeatedTimePolicy. This also avoids treating nanoseconds like seconds and iterating over them, which causes a hang. <rdar://problem/30229247>
         guard let secondRange = dateInterval(of: .second, for: startingAt) else {
             return nil
         }
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -2370,11 +2370,18 @@ extension Calendar {
             return nil
         }
 
-        // This taken directly from the old algorithm.  We don't have great support for nanoseconds in general and trying to treat them like seconds causes a hang. :-/
-        // <rdar://problem/30229247>
-        var dateComp = _dateComponents(.init(.era, .year, .month, .day, .hour, .minute, .second), from: startingAt)
-        dateComp.nanosecond = nanosecond
-        return date(from: dateComp)
+        // Substituting nanoseconds is arithmetic on the absolute time scale, so do it
+        // directly instead of decomposing into date components and resolving them
+        // again. Within a repeated hour, one set of wall clock components maps to two
+        // instants and date(from:) always picks the first, which would discard the
+        // occurrence that dateAfterMatchingHour selected via repeatedTimePolicy.
+        // This also avoids treating nanoseconds like seconds and iterating over them,
+        // which causes a hang. <rdar://problem/30229247>
+        guard let secondRange = dateInterval(of: .second, for: startingAt) else {
+            return nil
+        }
+
+        return secondRange.start + Double(nanosecond) / 1_000_000_000
     }
 
     // MARK: -
--- a/Tests/FoundationInternationalizationTests/CalendarTests.swift
+++ b/Tests/FoundationInternationalizationTests/CalendarTests.swift
@@ -526,15 +526,12 @@ private struct CalendarTests {
         #expect(result?.timeIntervalSinceReferenceDate == 657967500)
     }
 
-    /// Matches 01:20:00 on the day daylight saving time ends in Los Angeles, when
-    /// that local time occurs twice: once at -0700 and again an hour later at -0800.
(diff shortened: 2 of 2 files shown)


Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -2370,11 +2370,18 @@ extension Calendar {
...
-        var dateComp = _dateComponents(.init(.era, .year, .month, .day, .hour, .minute, .second), from: startingAt)
-        dateComp.nanosecond = nanosecond
-        return date(from: dateComp)
+        // Substituting nanoseconds is arithmetic on the absolute time scale, so do it
+        // directly instead of decomposing into date components and resolving them
+        // again. Within a repeated hour, one set of wall clock components maps to two
+        // instants and date(from:) always picks the first, which would discard the
+        // occurrence that dateAfterMatchingHour selected via repeatedTimePolicy.
+        // This also avoids treating nanoseconds like seconds and iterating over them,
+        // which causes a hang. <rdar://problem/30229247>
```
itingliu: Can we avoid hard wrapping these comments? See https://github.com/swiftlang/swift-foundation/blob/main/CONTRIBUTION_GUIDELINE.md#code-style

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
```diff
@@ -2370,11 +2370,18 @@ extension Calendar {
...
+        guard let secondRange = dateInterval(of: .second, for: startingAt) else {
+            return nil
+        }
+
+        return secondRange.start + Double(nanosecond) / 1_000_000_000
```
itingliu, line 2378: So is this essentially truncating `startingAt` to whole second, then add nanosecond part back?

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
itingliu: Also I'd appreciate if we can write this paragraph in an easier-to-read language: https://github.com/swiftlang/swift-foundation/blob/main/CONTRIBUTION_GUIDELINE.md#writing-style

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
cayde-6: Yes. `dateInterval(of: .second, for:)` returns `floor(timeIntervalSinceReferenceDate)` in every backend and ignores the time zone, so it rounds down to the whole second and adds the nanoseconds back.

Rebuilding from components does not work here: `date(from:)` returns the earlier of the two repeated times, which drops the one `repeatedTimePolicy` picked.

Comments unwrapped and simplified too.

hristost (approved):

PR #2168, opened 2026-08-09
Document that `RepeatedTimePolicy` is applied in search order

### LINKED ISSUE #2167: Calendar.RepeatedTimePolicy docs do not say the policy is applied in search order
`Calendar.RepeatedTimePolicy` documents `.first` as returning "the first occurrence" and `.last` as returning "the last occurrence", without saying relative to what. The implementation applies the policy in the order the search encounters the repeated times, so the two cases swap meaning when `SearchDirection.backward` is used.

Searching around the end of daylight saving time in `America/Los_Angeles` on 2024-11-03, where local 01:20 occurs twice:

```swift
import Foundation

var cal = Calendar(identifier: .gregorian)
cal.timeZone = TimeZone(identifier: "America/Los_Angeles")!

var dc = DateComponents()
dc.hour = 1; dc.minute = 20; dc.second = 0

let forwardStart = Date(timeIntervalSince1970: 1730536000.0)  // 2024-11-02T01:26:40-0700
let backwardStart = Date(timeIntervalSince1970: 1730712000.0) // 2024-11-04T01:20:00-0800

cal.nextDate(after: forwardStart,  matching: dc, matchingPolicy: .strict, repeatedTimePolicy: .first, direction: .forward)
// 2024-11-03T01:20:00-07:00 — the earlier of the two

cal.nextDate(after: backwardStart, matching: dc, matchingPolicy: .strict, repeatedTimePolicy: .first, direction: .backward)
// 2024-11-03T01:20:00-08:00 — the later of the two
```

| | `.forward` | `.backward` |
|---|---|---|
| `.first` | 01:20 −0700 (earlier) | 01:20 −0800 (later) |
| `.last` | 01:20 −0800 (later) | 01:20 −0700 (earlier) |

This is deliberate rather than accidental: `dateAfterMatchingHour` walks hour by hour with `direction == .backward ? -1 : 1` when resolving the policy, and the behavior matches the Foundation currently shipping in macOS, so it is not something to change. Only the documentation is ambiguous about it.

Suggested wording, as a note on the enum rather than on the individual cases:

> The policy is applied in the order the search encounters the repeated times. When searching with `SearchDirection.backward`, the "first" occurrence is therefore the later of the two in absolute time, and the "last" is the earlier.

I noticed this while working on #483 and initially mistook it for a second bug, so it seems worth spelling out. Happy to open a PR with the above if the wording looks right.

cayde-6:
Document that `Calendar.RepeatedTimePolicy` is applied in the order the search encounters the repeated times.

### Motivation:

Resolves #2167.

`RepeatedTimePolicy` says `.first` returns "the first occurrence" and `.last` "the last occurrence", but not relative to what. The policy is applied in the order the search runs into the repeated times, so the two swap meaning when searching backward.

Around the end of DST in `America/Los_Angeles` on 2024-11-03, where local 01:20 happens twice:

| | `.forward` | `.backward` |
|---|---|---|
| `.first` | 01:20 -0700 (earlier) | 01:20 -0800 (later) |
| `.last` | 01:20 -0800 (later) | 01:20 -0700 (earlier) |

I hit this while working on #483 and first took it for another bug, which is what prompted writing it down.

`enumerateDates` and `dates(byMatching:)` said the opposite, that "first" is defined from the point of view of searching forwards. Confirmed in review that the documentation should be corrected rather than the implementation changed, so this PR corrects those two sentences as well. The same sentence is in `NSCalendar.h`, which is not in this repository and would need the same correction there.

### Modifications:

A note on the `RepeatedTimePolicy` doc comment, the corrected sentence on `enumerateDates` and `dates(byMatching:)`, plus a test for the four direction/policy combinations. The per-case docs are left alone, since the ordering applies to both.

### Result:

No functional change.

### Testing:

`repeatedTimePolicyIsAppliedInSearchOrder` matches 01:20:00 across the Los Angeles transition, starting a day out on either side so the first match lands inside the repeated hour either way. It fails if the two expected instants are swapped, or if the direction term in the search loop is replaced with a constant.


parkera (commented):
This seems correct as far as I can remember. @itingliu can you also verify?

parkera:
It also might be nice to make sure we have a test that validates this behavior explicitly (sounds like you generated one based on the description).

cayde-6:
Hi. @parkera 
> It also might be nice to make sure we have a test that validates this behavior explicitly (sounds like you generated one based on the description).

Added `repeatedTimePolicyIsAppliedInSearchOrder` in `CalendarTests`, pinning all four direction/policy combinations.


Also found this while writing the test: `enumerateDates`, `dates(byMatching:)` and `NSCalendar.h` all say that "first" means first from the point of view of a forward search, which is the opposite of what the code does. Worth fixing too?

parkera:
Yes, if by fixing you mean fixing the documentation and not changing the implementation.

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -526,6 +526,35 @@ private struct CalendarTests {
         #expect(result?.timeIntervalSinceReferenceDate == 657967500)
     }
 
+    /// Matches 01:20:00 on the day daylight saving time ends in Los Angeles, when
+    /// that local time occurs twice: once at -0700 and again an hour later at -0800.
+    /// Both start dates sit outside the repeated hour, one day away from it, so the
+    /// first match lands inside it in either direction.
```
parkera: Please unwrap the comments per https://github.com/swiftlang/swift-foundation/blob/main/CONTRIBUTION_GUIDELINE.md#code-style - thanks

cayde-6:
Done — comments unwrapped, docs corrected on `enumerateDates` and `dates(byMatching:)`, behavior unchanged.

`NSCalendar.h` carries the same sentence on `enumerateDatesStartingAfterDate:`, but that header is not in this repository, so I could not fix it here.

cayde-6:
Rebased on main now that #2166 has landed. It added a `firstDateMatchingRepeatedHour` helper with a `nanosecond:` parameter, so this branch drops its own copy and the test uses that one instead. Only the test's call sites changed; the documentation edits are the same as before.

Tests/FoundationInternationalizationTests/CalendarTests.swift
```diff
@@ -570,6 +570,17 @@ private struct CalendarTests {
         #expect(first != last)
     }
 
+    /// Pins which occurrence each `RepeatedTimePolicy` selects in each search direction. The policy is applied in the order the search encounters the repeated times, so the two cases swap meaning in absolute time under `.backward`.
```
itingliu: This comment reads like it's just describing what the test does. It does not add extra information that cannot be inferred from the test expectation. In that case can we remove the comment

cayde-6: Removed in `388fd6a`. The test name and the four expectations carry it.

Tests/FoundationInternationalizationTests/CalendarTests.swift
cayde-6: Sorry, I answered this in the main thread back then and never replied here. The comments were unwrapped in 34a58a5, d4c54d8 and 9cb71a6 on 18 Aug, and the docs on `enumerateDates` and `dates(byMatching:)` were corrected in the same round.

cayde-6:
Everything raised in review is addressed, latest commits are from 25 Aug.

CI has never run on this PR. If it looks reasonable, could someone kick it off with `@swift-ci test`? Happy to rebase first if you'd rather.

parkera (approved):

parkera:
@swift-ci test

cayde-6:
Everything requested is addressed, the PR is approved, and Linux, Windows, and macOS CI are green. @parkera, is anything else needed from me before merge?

PR #2170, opened 2026-08-11
Refactor handling of invalid errno values in POSIXError

jmschonfeld:
Updates invalid errno handling to avoid crashes in production

### Motivation:

In some cases, an invalid errno indicates a bug in Foundation or the kernel. However, it's possible that it was created by a misbehaving module called by the kernel which passes through the errno value. Foundation should be resilient to these modules rather than terminating

### Modifications:

Change the existing assertion to a debug-only assertion so it fires in Foundation's unit tests but not in the toolchain (dropping the POSIXError underlying value when one cannot be formed)

### Result:

Foundation does not crash when given an invalid errno value

### Testing:

Verified existing unit tests pass (it is not very feasible to craft a situation in which this errno is passed through in a unit test)

### DIFF
--- a/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
+++ b/Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
@@ -73,10 +73,11 @@ extension CocoaError.Code {
 
 extension POSIXError {
     internal init?(errno: Int32) {
-        // (130280235) POSIXError.Code does not have a case for EOPNOTSUPP
-        guard errno != EOPNOTSUPP else { return nil }
         guard let code = POSIXError.Code(rawValue: errno) else {
-            fatalError("Invalid posix errno \(errno)")
+            // Assert in debug mode to catch cases in Foundation unit tests where we create a POSIXError with invalid error information (for example when the errno has already been set back to 0 after a subsequent successful API call)
+            // Avoid asserting in production because invalid errnos can be produced by arbitrary code run by the kernel and Foundation should be resilient if that code misbehaves. Unfortunately we don't have a way to form a POSIXError in Swift with the invalid code, so we instead just drop the underlying error
+            assertionFailure("Unable to form POSIXError from invalid errno \(errno)")
+            return nil
         }
         self.init(code)
     }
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Error/CocoaError+FilePath.swift
```diff
@@ -73,10 +73,11 @@ extension CocoaError.Code {
 
 extension POSIXError {
     internal init?(errno: Int32) {
-        // (130280235) POSIXError.Code does not have a case for EOPNOTSUPP
-        guard errno != EOPNOTSUPP else { return nil }
         guard let code = POSIXError.Code(rawValue: errno) else {
-            fatalError("Invalid posix errno \(errno)")
+            // Assert in debug mode to catch cases in Foundation unit tests where we create a POSIXError with invalid error information (for example when the errno has already been set back to 0 after a subsequent successful API call)
+            // Avoid asserting in production because invalid errnos can be produced by arbitrary code run by the kernel and Foundation should be resilient if that code misbehaves. Unfortunately we don't have a way to form a POSIXError in Swift with the invalid code, so we instead just drop the underlying error
```
parkera, line 78: I feel like we should pick some placeholder error instead of returning nil here.

parkera: `ENOSYS` or `EINVAL` seem like plausible choices, although I acknowledge it could be confused with that actual error when it happens.

jmschonfeld: Do you have an error code in mind? I couldn't find a good one. Alternatively, we could set the underlying error to some new `UnknownPosixError` structure when this initializer returns `nil` if we don't have a good errno value to replace it with as long as we're ok returning that struct out as the underlying error, but we don't have a domain for that kind of error today

jmschonfeld: Yeah that was my main worry was that using `ENOSYS` or `EINVAL` could be confused with the function actually returning `ENOSYS` or `EINVAL`. It seemed to me like it was better to report no information than potentially confusable information, but happy to switch to one of those if you think it would be better

jmschonfeld: Updated to report the unknown errno value in a separate user info dictionary key when we can't create the underlying error so that the information is still present in places that log the contents of the error

parkera (approved):

PR #2171, opened 2026-08-11
[Base64] Use Span instead UnsafeBufferPointer for lookup tables

fabianfett:
#1196 follow up.

### Motivation:

We like to use safe Swift as much as possible. 

### Modifications:

- This PR removes lots of uses of UnsafeBufferPointer and replaces them with local `Base64[En/De]codeLookupTable`s which are backed by Span.

### Result:

- No performance regressions.
- Safer, easier to reason about code.

### Testing:

- All tests pass
- No performance regressions


Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -154,6 +154,46 @@ extension Data {
 
 enum Base64 {}
 
+/// A base64 encoding lookup table.
+///
+/// The table holds exactly 256 bytes and is only indexable by `UInt8`. Every possible index is
+/// therefore in bounds, which is why the subscript can forward to the span's unchecked subscript:
```
itingliu: Can we not wrap these comments? I find them difficult to read when I review the PR side by side

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -542,18 +576,17 @@ extension Base64 {
...
-        let encoding1 = options.contains(.base64URLAlphabet) ? Self.encoding1url : Self.encoding1
-
-        assert(encoding0.count == 256)
-        assert(encoding1.count == 256)
-
-        return encoding0.withUnsafeBufferPointer { e0 in
-            encoding1.withUnsafeBufferPointer { e1 in
-                body(e0, e1)
-            }
+    private static func withEncodingTables<R, E: Swift.Error>(
```
parkera, line 591: Do we need this at all, or can the tables be just passed directly to the function instead of using a closure?

parkera (commented):
Can you post the results of running the benchmarks? Coincidentally, I had also written almost exactly this same patch and found the throughput regressed to half of the previous values.

Sources/FoundationEssentials/Data/Data+Base64.swift
parkera: In my benchmarking I also found that this `stride` costs quite a lot for the abstraction. A simple addition was much faster.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -490,8 +524,8 @@ extension Base64 {
     }
 
     private static func loopEncode(
-        _ e0: UnsafeBufferPointer<UInt8>,
```
parkera, line 493: For full replacement of unsafe, this should use `OutputSpan` instead of `outIndex` and `UnsafeMutableBufferPointer`.

Sources/FoundationEssentials/Data/Data+Base64.swift
parkera: Note that to do so, you may need to use a bit of pre-validation of the indexes and values.

Mine looked something like this, but I didn't use the abstraction for the table that you have here.

```
    private static func loopEncode(
        _ e0: Span<UInt8>,
        _ e1: Span<UInt8>,
        input: Span<UInt8>,
        from: Int,
        to: Int,
        output: inout OutputSpan<UInt8>
    ) {
        guard to > from else {
            // nothing to do
            return
        }
        
        precondition(to - from <= output.freeCapacity, "OutputSpan capacity overflow")
        // We will index input up to the following value; striding by 3
        let maxIndex = to - (to % 3)
        precondition(input.count >= maxIndex, "Input did not have sufficient count")
        
        // These are spans from inline arrays, so the following two checks should never fail.
        precondition(e0.count >= 256)
        precondition(e1.count >= 256)
        
        output.withUnsafeMutableBufferPointer { buffer, count in
            for index in stride(from: from, to: to, by: 3) {
                let i1 = input[unchecked: index]
                let i2 = input[unchecked: index + 1]
                let i3 = input[unchecked: index + 2]
                
                buffer[count + 0] = e0[unchecked: Int(i1)]
                buffer[count + 1] = e1[unchecked: Int(((i1 & 0x03) << 4) | ((i2 >> 4) & 0x0F))]
                buffer[count + 2] = e1[unchecked: Int(((i2 & 0x0F) << 2) | ((i3 >> 6) & 0x03))]
                buffer[count + 3] = e1[unchecked: Int(i3)]
                count += 4
            }
        }
    }
```

Sources/FoundationEssentials/Data/Data+Base64.swift
jmschonfeld: I believe the input can also be made a `Span`/`RawSpan` instead of an `UnsafeBufferPointer`

Sources/FoundationEssentials/Data/Data+Base64.swift
fabianfett: when we don't use the tables we have `unsafe unchecked` everywhere. And it feels much harder to review. The tables 100% codify our promises. I started with just passing the Spans around. In the nio team we have the rule that we need to annotate every unsafe call and explain why it is safe. This is what I did. Only then did I realize that the LookupTable structure, which should 100% be compiled out is much more elegant as it codifies the UInt8 to 256 elements promise.

Sources/FoundationEssentials/Data/Data+Base64.swift
fabianfett: Let's do input and output span in a separate follow-up. I have a small proposal ready to go.

Sources/FoundationEssentials/Data/Data+Base64.swift
fabianfett: Let's do input and output span in a separate follow-up. I have a small proposal ready to go.

fabianfett:
@parkera I've run benchmarks on main and this branch. 

  Apple M-series (16 cores, arm64), p50 values. Instructions is the primary metric; malloc counts were identical in every benchmark.

  | Benchmark | Instructions (main) | Instructions (branch) | Δ | CPU time (main) | CPU time (branch) | Δ |
  |---|---:|---:|---:|---:|---:|---:|
  | decode-1MB-fromData-noOptions | 7,864 K | 7,860 K | −0.05% | 269 µs | 269 µs | 0% |
  | decode-1MB-fromData-noOptions-invalidAfter257bytes | 6,205 | 4,858 | **−21.7%** | 251 ns | 194 ns | **−22.7%** |
  | decode-1MB-fromString-lineLength64 | 11 M | 11 M | ~0% | 358 µs | 358 µs | 0% |
  | decode-1MB-fromString-noOptions | 7,864 K | 7,864 K | 0% | 268 µs | 267 µs | −0.4% |
  | decode-jwtHeader-fromString-noOptions | 1,593 | 1,593 | 0% | 83 ns | 81 ns | −2.4% |
  | encode-1MB-toData-noOptions | 10 M | 10 M | ~0% | 356 µs | 358 µs | +0.6% |
  | encode-1MB-toString-lineLength64 | 12 M | 12 M | ~0% | 321 µs | 328 µs | +2.2% |
  | encode-1MB-toString-noOptions | 9,265 K | 9,265 K | 0% | 282 µs | 281 µs | −0.4% |
  | encode-jwtHeader-toString-noOptions | 1,336 | 1,332 | −0.3% | 68 ns | 67 ns | −1.5% |

  **Summary:** no instruction-count regressions. Everything is within noise except early-exit decoding of invalid input, which drops ~22% (6,205 → 4,858 instructions). Malloc counts unchanged across the board.

Sources/FoundationEssentials/Data/Data+Base64.swift
fabianfett: > In my benchmarking I also found that this `stride` costs quite a lot for the abstraction. A simple addition was much faster.

Based on this branch I've tested your `stride` statement. Very true. Once this PR is in, I'm happy to propose a follow up, further reducing instructions significantly.

fabianfett:
I've added another perf optimization for encoding:

| Benchmark | Instructions (main) | Instructions (branch) | Δ | CPU time (main) | CPU time (branch) | Δ |
  |---|---:|---:|---:|---:|---:|---:|
  | decode-1MB-fromData-noOptions | 7,864 K | 7,864 K | 0% | 269 µs | 269 µs | 0% |
  | decode-1MB-fromData-noOptions-invalidAfter257bytes | 6,205 | 6,062 | −2.3% | 251 ns | 247 ns | −1.6% |
  | decode-1MB-fromString-lineLength64 | 11 M | 11 M | ~0% | 358 µs | 359 µs | +0.3% |
  | decode-1MB-fromString-noOptions | 7,864 K | 7,864 K | 0% | 268 µs | 268 µs | 0% |
  | decode-jwtHeader-fromString-noOptions | 1,593 | 1,594 | ~0% | 83 ns | 80 ns | −3.6% |
  | encode-1MB-toData-noOptions | ~10 M | 8,331 K | **≈ −17%** | 356 µs | 356 µs | 0% |
  | encode-1MB-toString-lineLength64 | ~12 M | 7,827 K | **≈ −35%** | 321 µs | 287 µs | **−10.6%** |
  | encode-1MB-toString-noOptions | 9,265 K | 7,557 K | **−18.4%** | 282 µs | 278 µs | −1.4% |
  | encode-jwtHeader-toString-noOptions | 1,336 | 1,286 | −3.7% | 68 ns | 69 ns | +1.5% |

**Summary:** no instruction-count regressions. The 1 MB encode paths get substantially cheaper (17–35% fewer instructions; the `lineLength64` case also shows a 10.6% CPU-time win). Note that main's `M`-scaled rows are rounded by the harness, so those two deltas are approximate. Decoding is unchanged within noise. Malloc counts unchanged across the board.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -420,24 +454,30 @@ extension Base64 {
...
                 self.loopEncode(e0, e1, input: input, from: 0, to: lineLength, output: buffer, outIndex: &outIndex)
             }
 
             // following full lines
-            for lineInputIndex in stride(from: lineLength, to: lines * lineLength, by: lineLength) {
+            var lineInputIndex = lineLength
+            while lineInputIndex < lines * lineLength {
                 buffer[outIndex] = separatorByte1
-                outIndex += 1
+                outIndex &+= 1
```
parkera, line 473: I've noticed that some tools love to make this conversion, but it's not 100% obvious that it's actually safe to do so.

fabianfett: I actually made this conversion. We are working with offsets here on a pointer, that we validated before.

parkera:
I'm ok with doing the `OutputSpan` and additional `Span` work in a follow up, because I suspect we'll need some more work to keep performance acceptable there.

Sources/FoundationEssentials/Data/Data+Base64.swift
glessard: It would be nice for readers to include a comment pointing at the operation that validates the `outIndex` arithmetic.

glessard (commented):
Nice improvement!

glessard (approved):

fabianfett:
@glessard can you merge this pr, please?

glessard:
@fabianfett merged.

PR #2174, opened 2026-08-13
Fix build failure in SortComparatorBenchmarks

jmschonfeld:
Removes an API usage with availability from the benchmark to allow it to build

### Motivation:

#2094 introduced a new SortComparatorBenchmarks target which fails to build in the default configuration on macOS because it uses an API unavailable on all deployment targets.

### Modifications:

Removed the usage of the API since it wasn't needed (it doesn't look like it was used) to allow the target to build

### Result:

The benchmark package builds on macOS successfully with no env vars set

### Testing:

Validated the build locally

### DIFF
--- a/Benchmarks/Benchmarks/SortComparator/BenchmarkSortComparator.swift
+++ b/Benchmarks/Benchmarks/SortComparator/BenchmarkSortComparator.swift
@@ -41,7 +41,6 @@ fileprivate struct Entry {
     fileprivate let _computedProperty: Double?
     /// Computed property: exercises the key-path getter (no stored-field offset).
     var computedProperty: Double? { _computedProperty }
-    let id: UUID
 }
 
 /// The same fields as `Entry`, but a `final class`, so the element is a
@@ -57,7 +56,6 @@ fileprivate final class EntryObject: Sendable {
     let str2: String
     private let _computedProperty: Double?
     var computedProperty: Double? { _computedProperty }
-    let id: UUID
 
     init(_ e: Entry) {
         int0 = e.int0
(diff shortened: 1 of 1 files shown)


parkera (approved):

dnadoba (commented):
Ah yeah, that is what I have mentioned here when we moved to the new API: https://github.com/swiftlang/swift-foundation/pull/2094#discussion_r3625414187

Benchmarks/Benchmarks/SortComparator/BenchmarkSortComparator.swift
```diff
@@ -41,7 +41,6 @@ fileprivate struct Entry {
     fileprivate let _computedProperty: Double?
     /// Computed property: exercises the key-path getter (no stored-field offset).
     var computedProperty: Double? { _computedProperty }
-    let id: UUID
```
dnadoba, line 44: I think it is still somewhat used because UUID is fragile (aka not frozen) which in turn makes the layout of Entry not fully known at compile time and more expensive to copy/destroy etc. 
Now that I think about it again, cross module optimization might have already broken this if we compile with Foundation together?

Either way, we it increased the size of `Entry` which is defiantly has an impact. 

As we don't actually sort with it we can still get the same benefit (or I guess downside :D) of UUID's fragility by just initializing it to a constant like all zeros. 
WDYT?

jmschonfeld: Ah I didn't think about the frozen / optimization aspect. Yeah initializing it to a constant like all zeroes sounds good to me if that works for you. I can add that back and add a comment explaining why it's there.

PR #2175, opened 2026-08-13
Use empty UUID in SortComparatorBenchmark

jmschonfeld:
Followup to the conversation in #2174 to re-add the UUID property but initialize it to all zeroes to preserve the non-resilient behavior while avoiding the API with availability


dnadoba (approved):
Sweet, thank you!

iCharlesHu (approved):

PR #2183, opened 2026-08-18
Add a UTF8Span helper

itingliu:
Add two UTF8Span helpers so we can use these when utf8Span is not available

Credit @jmschonfeld 

Resolves 185219160


jmschonfeld (approved):

itingliu:
cmake build failure is unrelated.

PR #2187, opened 2026-08-19
Resolve `weekOfMonth` when neither day nor weekday is given

### LINKED ISSUE #532: `Calendar.date(from:)` incorrectly sets `UCAL_DAY_OF_MONTH`
I was trying to iterate through all the weeks in a month, like so:

```swift
for w in 1 ... 5 {
    let components = DateComponents(year: 2024, month: 1, weekOfMonth: w)
    let date = Calendar.current.date(from: components)!
    print(date.debugDescription)
}
```

However, this printed some unexpected results:

```
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
```

After some digging into [the implementation](https://github.com/apple/swift-foundation/blob/main/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift#L1146), I discovered that the `date(from:)` method is incorrectly setting the `UCAL_DAY_OF_MONTH` field on the cleared `UCalendar`, which is messing up the calculation.

I verified this by testing the logic myself:

```c
UErrorCode error = U_ZERO_ERROR;
UCalendar *ucalendar = ucal_open(NULL, 0, "en_US_POSIX", UCAL_GREGORIAN, &error);

for (int32_t week = 1; week <= 5; week++) {
    ucal_clear(ucalendar);

    ucal_set(ucalendar, UCAL_ERA, 1);
    ucal_set(ucalendar, UCAL_IS_LEAP_MONTH, 0);
    ucal_set(ucalendar, UCAL_YEAR, 2024);
    ucal_set(ucalendar, UCAL_MONTH, 0); // 0 = January
    ucal_set(ucalendar, UCAL_WEEK_OF_MONTH, week);
//    ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1);
    
    ucal_set(ucalendar, UCAL_HOUR_OF_DAY, 0);
    ucal_set(ucalendar, UCAL_MINUTE, 0);
    ucal_set(ucalendar, UCAL_SECOND, 0);
    ucal_set(ucalendar, UCAL_MILLISECOND, 0);
    
    UDate udate = ucal_getMillis(ucalendar, &error);
    
    NSDate *d = [NSDate dateWithTimeIntervalSince1970:udate / 1000];
    printf("%s\n", d.debugDescription);
}
ucal_close(ucalendar);
```

When the `UCAL_DAY_OF_MONTH` field is set, this prints:
```
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
2024-01-01 07:00:00 +0000
```

When the line is commented out, it prints:
```
2023-12-31 07:00:00 +0000
2024-01-07 07:00:00 +0000
2024-01-14 07:00:00 +0000
2024-01-21 07:00:00 +0000
2024-01-28 07:00:00 +0000
```

Setting the `UCAL_DAY_OF_MONTH` field on the `UCalendar` causes operations around finding the weekOfMonth and weekOfYear to fail and return the *first* week of the month/year, instead of the desired week.

cayde-6:
`Calendar.date(from:)` now honors `weekOfMonth` when the components carry no day and no weekday.

### Motivation:

`Calendar.date(from:)` returns the first of the month for every `weekOfMonth`
value when the components carry a week of month but no day and no weekday,
so the weeks of a month cannot be iterated:

```swift
for w in 1...5 {
    let d = Calendar.current.date(from: DateComponents(year: 2024, month: 1, weekOfMonth: w))!
    print(d)  // 2024-01-01 five times
}
```

Adding a weekday hides the problem, and there is no `yearForWeekOfMonth`
spelling to fall back on, so the combination has no working alternative.

Resolves #532.

### Modifications:

Both calendar backends had their own cause for the same symptom.

`_CalendarGregorian`: `ResolvedDateComponents.init` matched the `year != nil`
branch before the `weekOfMonth` branch and resolved to `.day` with a nil day,
which falls back to day 1. The `weekOfMonth` branch is now reached first.

`_CalendarICU`: `date(from:)` seeded `UCAL_DAY_OF_MONTH` with 1 before setting
the requested fields. ICU then resolves the ambiguous field combination in
favor of day of month and never looks at week of month. The default is now
skipped when the components ask for a week of month without a day or weekday.

### Result:

```
weekOfMonth 1...5 of January 2024, first weekday Sunday, minimum days in first week 1

before: 01-01  01-01  01-01  01-01  01-01
after:  12-31  01-07  01-14  01-21  01-28
```

The expected dates match the ICU output quoted in #532, and ICU4J 76.1 returns
the same five dates for the same input.

### Testing:

New regression tests for both backends, plus a check that the two backends
agree on the input from the issue.

`testDateFromComponents` in `GregorianCalendarTests` captured the previous
behavior, with several pairs of different `weekOfMonth` values expecting the
same date (for example `year: 1828, month: 7` with `weekOfMonth: 2` and
`weekOfMonth: 5` both expecting 1828-07-01). Those 17 expectations are updated.
The two rows that already held a correct date are untouched.

Full test suite passes.

### Two things worth a maintainer decision

**The Gregorian cutover year.** With the fix in place, the two backends
disagree when the components name 1582 as the year: `_CalendarICU` returns
dates 14 days later than `_CalendarGregorian`. The disabled
`GregorianCalendarCompatibilityTests` suite reports 416 such cases when run
locally. This is not new arithmetic on our side: `gregocal.cpp` adds a
hardcoded `jd += 14` for `UCAL_WEEK_OF_MONTH` in the cutover year, a line that
dates back to 2003 and has no counterpart in ICU4J. ICU4J and
`_CalendarGregorian` agree on these dates; ICU4C stands alone. Previously both
backends returned the first of the month here, so the divergence was masked
rather than absent. I have not touched cutover handling in either backend and
am happy to follow whichever direction you prefer.

**`DateComponents(year:weekOfYear:)`** has the same shape of problem and still
returns the first day of the year. `CalendarTests` documents that as
intentional ("year and weekOfYear, not valid setting so the result is just the
first day of the year"), and `yearForWeekOfYear` is the working spelling, so I
left it alone. I have the fix ready if you want it.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -148,6 +148,8 @@ enum ResolvedDateComponents {
             self = .dayOfYear(year: year, dayOfYear: dayOfYear)
         } else if components.yearForWeekOfYear != nil  {
             self = .weekOfYear(year: year, weekOfYear: components.weekOfYear, weekday: components.weekday)
+        } else if let weekOfMonth = components.weekOfMonth {
+            self = .weekOfMonth(year: year, month: month, weekOfMonth: weekOfMonth, weekday: components.weekday)
         } else if components.year != nil {
             self = .day(year: year, month: month, day: components.day, weekOfYear: components.weekOfYear)
         } else if let weekOfYear = components.weekOfYear {
--- a/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
@@ -154,8 +154,6 @@ enum ResolvedDateComponents {
             self = .day(year: year, month: month, day: components.day, weekOfYear: components.weekOfYear)
         } else if let weekOfYear = components.weekOfYear {
             self = .weekOfYear(year: year, weekOfYear: weekOfYear, weekday: components.weekday)
-        } else if let weekOfMonth = components.weekOfMonth {
-            self = .weekOfMonth(year: year, month: month, weekOfMonth: weekOfMonth, weekday: components.weekday)
         } else if let weekdayOrdinal = components.weekdayOrdinal {
             self = .weekdayOrdinal(year: year, month: month, weekdayOrdinal: weekdayOrdinal, weekday: components.weekday)
         } else if let weekday = components.weekday {
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1174,9 +1174,7 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 #if FOUNDATION_FRAMEWORK // FIXME: https://github.com/swiftlang/swift-foundation-icu/issues/62
             ucal_set(ucalendar, UCAL_IS_REPEATED_DAY, 0)
 #endif
-            // ICU resolves the day-of-month / week-of-month ambiguity in favor of
-            // day-of-month, so defaulting it to 1 would make week-of-month be ignored
-            // entirely. See https://github.com/swiftlang/swift-foundation/issues/532
+            // ICU resolves the day-of-month and week-of-month ambiguity in favor of day-of-month, so defaulting the day to 1 would ignore week-of-month entirely.
             let resolvesByWeekOfMonth = components.weekOfMonth != nil && components.day == nil && components.weekday == nil
             if !resolvesByWeekOfMonth {
                 ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1174,7 +1174,13 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 #if FOUNDATION_FRAMEWORK // FIXME: https://github.com/swiftlang/swift-foundation-icu/issues/62
             ucal_set(ucalendar, UCAL_IS_REPEATED_DAY, 0)
 #endif
-            ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
+            // Don't default day-of-month to 1 when the components specify a week of month
+            // (and don't also specify a day or weekday), or ICU will always resolve the
+            // ambiguity in favor of day-of-month and ignore week-of-month entirely.
+            // See https://github.com/swiftlang/swift-foundation/issues/532
+            if !(components.weekOfMonth != nil && components.day == nil && components.weekday == nil) {
+                ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
+            }
             ucal_set(ucalendar, UCAL_HOUR_OF_DAY, 0)
             ucal_set(ucalendar, UCAL_MINUTE, 0)
             ucal_set(ucalendar, UCAL_SECOND, 0)
--- a/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
+++ b/Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
@@ -1174,11 +1174,11 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 #if FOUNDATION_FRAMEWORK // FIXME: https://github.com/swiftlang/swift-foundation-icu/issues/62
             ucal_set(ucalendar, UCAL_IS_REPEATED_DAY, 0)
 #endif
-            // Don't default day-of-month to 1 when the components specify a week of month
-            // (and don't also specify a day or weekday), or ICU will always resolve the
-            // ambiguity in favor of day-of-month and ignore week-of-month entirely.
-            // See https://github.com/swiftlang/swift-foundation/issues/532
-            if !(components.weekOfMonth != nil && components.day == nil && components.weekday == nil) {
+            // ICU resolves the day-of-month / week-of-month ambiguity in favor of
+            // day-of-month, so defaulting it to 1 would make week-of-month be ignored
+            // entirely. See https://github.com/swiftlang/swift-foundation/issues/532
+            let resolvesByWeekOfMonth = components.weekOfMonth != nil && components.day == nil && components.weekday == nil
+            if !resolvesByWeekOfMonth {
                 ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
             }
             ucal_set(ucalendar, UCAL_HOUR_OF_DAY, 0)
--- a/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
+++ b/Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
@@ -220,7 +220,7 @@ private struct GregorianCalendarTests {
     // Regression test for https://github.com/swiftlang/swift-foundation/issues/532:
     // year + month + weekOfMonth (without day or weekday) resolved to the 1st of the month
     // for every weekOfMonth value instead of the correct week.
-    @Test func testDateFromComponentsWeekOfMonthWithoutWeekday() {
+    @Test func testDateFromComponentsWeekOfMonth() {
         let cal = _CalendarGregorian(identifier: .gregorian, timeZone: .gmt, locale: nil, firstWeekday: 1, minimumDaysInFirstWeek: 1, gregorianStartDate: nil)
         func test(_ dateComponents: DateComponents, expected: Date, sourceLocation: SourceLocation = #_sourceLocation) {
             let date = cal.date(from: dateComponents)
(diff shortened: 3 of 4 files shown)


parkera:
@dra8an can you also take a look please?

DraganBesevic:
> @dra8an can you also take a look please?

Related work: https://unicode-org.atlassian.net/browse/ICU-23489

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -148,6 +148,8 @@ enum ResolvedDateComponents {
             self = .dayOfYear(year: year, dayOfYear: dayOfYear)
         } else if components.yearForWeekOfYear != nil  {
             self = .weekOfYear(year: year, weekOfYear: components.weekOfYear, weekday: components.weekday)
+        } else if let weekOfMonth = components.weekOfMonth {
+            self = .weekOfMonth(year: year, month: month, weekOfMonth: weekOfMonth, weekday: components.weekday)
```
dra8an, line 152: There is a duplicated code for this at line 157 that is dead now. Can you delete it?

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1174,7 +1174,13 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 #if FOUNDATION_FRAMEWORK // FIXME: https://github.com/swiftlang/swift-foundation-icu/issues/62
             ucal_set(ucalendar, UCAL_IS_REPEATED_DAY, 0)
 #endif
-            ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
+            // Don't default day-of-month to 1 when the components specify a week of month
+            // (and don't also specify a day or weekday), or ICU will always resolve the
+            // ambiguity in favor of day-of-month and ignore week-of-month entirely.
+            // See https://github.com/swiftlang/swift-foundation/issues/532
+            if !(components.weekOfMonth != nil && components.day == nil && components.weekday == nil) {
```
dra8an: Suggestion: Can we name this condition instead of negating a three part &&

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
```diff
@@ -217,6 +217,23 @@ private struct GregorianCalendarTests {
         test(.init(weekOfYear: 43, yearForWeekOfYear: 2935), expected: Date(timeIntervalSince1970: 30477945600.0))
     }
 
+    // Regression test for https://github.com/swiftlang/swift-foundation/issues/532:
+    // year + month + weekOfMonth (without day or weekday) resolved to the 1st of the month
```
dra8an: Can you add a couple of additional tests: one with `day` and `weekOfMonth` both set, and one with `weekOfMonth` and `weekday` both set?

Tests/FoundationInternationalizationTests/CalendarICUTests.swift
```diff
@@ -72,4 +72,39 @@ private struct CalendarICUTests {
...
+
+        let expected: [Date] = [
+            Date(timeIntervalSince1970: 1703980800), // 2023-12-31 00:00:00 +0000
+            Date(timeIntervalSince1970: 1704585600), // 2024-01-07 00:00:00 +0000
+            Date(timeIntervalSince1970: 1705190400), // 2024-01-14 00:00:00 +0000
+            Date(timeIntervalSince1970: 1705795200), // 2024-01-21 00:00:00 +0000
+            Date(timeIntervalSince1970: 1706400000), // 2024-01-28 00:00:00 +0000
+        ]
+
+        for (i, weekOfMonth) in (1...5).enumerated() {
```
dra8an: Nit: can you use `@Test(arguments:)` over pairs instead of `(1...5).enumerated()`?

Something like

```
     @Test(arguments: [
         (1, Date(timeIntervalSince1970: 1703980800)), // 2023-12-31
         (2, Date(timeIntervalSince1970: 1704585600)), // 2024-01-07
         (3, Date(timeIntervalSince1970: 1705190400)), // 2024-01-14
         (4, Date(timeIntervalSince1970: 1705795200)), // 2024-01-21
         (5, Date(timeIntervalSince1970: 1706400000)), // 2024-01-28
     ])
     func dateFromComponentsWeekOfMonth(weekOfMonth: Int, expected: Date) {
         let icuCalendar = _CalendarICU(identifier: .gregorian, timeZone: .gmt, locale: nil, firstWeekday: 1, minimumDaysInFirstWeek: 1, gregorianStartDate: nil)
         let components = DateComponents(year: 2024, month: 1, weekOfMonth: weekOfMonth)
         #expect(icuCalendar.date(from: components) == expected)
     }
```

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
cayde-6: Done in `ab7e24e`, the branch is deleted.

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
cayde-6: Done in `ab7e24e`. The condition is now named `resolvesByWeekOfMonth`, so the `if` carries a single negation.

Tests/FoundationEssentialsTests/GregorianCalendarTests.swift
cayde-6: Added both in `ab7e24e`. `day` + `weekOfMonth` resolves to the day (2024-01-20, week of month ignored), `weekOfMonth` + `weekday` resolves to the weekday within that week (2024-01-17). Both go through pre-existing branches, so these pin that the new branch does not hijack them.

Renamed the test to `testDateFromComponentsWeekOfMonth`, since it no longer covers only the weekday free case.

Tests/FoundationInternationalizationTests/CalendarICUTests.swift
cayde-6: Done in `ab7e24e`, switched to `@Test(arguments:)` over pairs as suggested.

cayde-6:
All four points from the review are addressed, in `ab7e24e` and `5572107`.

CI hasn't run on this one yet. Could someone start it with `@swift-ci test` when you get a moment? Happy to rebase first if that helps.

parkera:
@swift-ci test

cayde-6:
All four review points are addressed, and Linux, Windows, and macOS CI are green. @dra8an @itingliu, when you have a chance, could you re-review? The remaining threads refer to the earlier revision or have replies. Happy to update anything still missing.

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
```diff
@@ -1174,7 +1174,11 @@ internal final class _CalendarICU: _CalendarProtocol, @unchecked Sendable {
 #if FOUNDATION_FRAMEWORK // FIXME: https://github.com/swiftlang/swift-foundation-icu/issues/62
             ucal_set(ucalendar, UCAL_IS_REPEATED_DAY, 0)
 #endif
-            ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
+            // ICU resolves the day-of-month and week-of-month ambiguity in favor of day-of-month, so defaulting the day to 1 would ignore week-of-month entirely.
+            let resolvesByWeekOfMonth = components.weekOfMonth != nil && components.day == nil && components.weekday == nil
+            if !resolvesByWeekOfMonth {
+                ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
+            }
```
itingliu, line 1182: Does it make sense to only set `UCAL_WEEK_OF_MONTH` if we do not set `UCAL_DAY_OF_MONTH`? Like, consolidate the `UCAL_WEEK_OF_MONTH` branch below into this one here.

Conceptually something like 
```
if resolvesByWeekOfMonth {
    ucal_set(ucalendar, UCAL_WEEK_OF_MONTH, components.weekOfMonth)
} else if let dayOfMonth = components.dayOfMonth {
    ucal_set(ucalendar, UCAL_DAY_OF_MONTH, dayOfMonth)
} else {
    ucal_set(ucalendar, UCAL_DAY_OF_MONTH, 1)
}
```

cayde-6:
Addressed in 9233e88d. `UCAL_WEEK_OF_MONTH` is now set in the default-day branch only when `components.day` is nil; the separate assignment was removed, so an explicit day keeps priority. `swift test --filter CalendarICUTests` passes (8 tests).

cayde-6:
The previous Linux, Windows, and macOS results are for the old revision; no checks are currently reported for new head `9233e88d`. Could someone run `@swift-ci test` for this revision when convenient?

parkera:
@swift-ci test

Sources/FoundationInternationalization/Calendar/Calendar_ICU.swift
cayde-6: Makes sense, consolidated in `9233e88d`.

cayde-6: Follow-up in `241e69ae`: setting `UCAL_WEEK_OF_MONTH` in the default-day branch changed which field ICU prefers when week of month is combined with `weekOfYear` or `weekdayOrdinal`, since ICU favors the most recently set field. The two backends then disagreed, e.g. `year: 2024, month: 3, weekOfMonth: 2, weekOfYear: 20` gave 2024-05-12 from ICU and 2024-03-03 from Gregorian. The default day is still skipped in that branch, but week of month is now set where its priority matches `_CalendarGregorian`. The backend agreement test now covers 16 combinations across three week settings.

cayde-6:
Squashed into one commit and rebased on `main`, as `CONTRIBUTING.md` asks. The code is the same as the previous head `241e69ae`. Changes since the last review (`5572107`): https://github.com/swiftlang/swift-foundation/compare/5572107d35f31b0dd77e9937fe02876d8d7f35f5...241e69aed1851b4952c33ca21e3ff84cda0bad2d

The full test suite passes locally. Could someone run `@swift-ci test` for the new head when convenient?

PR #2191, opened 2026-08-20
[Base64] Use Span for Input and Output when Encoding

fabianfett:
### Motivation:

We love fast and safe code.

### Modifications:

- Use the new Span type as the internal input and output type.

### Result:

- Code that is as fast as before but safer

### Testing:

- The existing test suite validates these changes

### DIFF
--- a/Sources/FoundationEssentials/Data/Data+Base64.swift
+++ b/Sources/FoundationEssentials/Data/Data+Base64.swift
@@ -133,7 +133,7 @@ extension Data {
     /// - parameter options: The options to use for the encoding. Default value is `[]`.
     /// - returns: The Base-64 encoded string.
     public func base64EncodedString(options: Base64EncodingOptions = []) -> String {
-        Base64.encodeToString(bytes: self.span, options: options)
+        Base64.encodeToString(bytes: self.bytes, options: options)
     }
 
     /// Returns Base-64 encoded data.
@@ -141,7 +141,7 @@ extension Data {
     /// - parameter options: The options to use for the encoding. Default value is `[]`.
     /// - returns: The Base-64 encoded data.
     public func base64EncodedData(options: Base64EncodingOptions = []) -> Data {
-        Base64.encodeToData(bytes: self.span, options: options)
+        Base64.encodeToData(bytes: self.bytes, options: options)
     }
 }
 
@@ -309,33 +309,26 @@ extension Base64 {
         UInt8(ascii: "6"), UInt8(ascii: "7"), UInt8(ascii: "8"), UInt8(ascii: "9"), UInt8(ascii: "-"), UInt8(ascii: "_"),
     ]
 
-    static func encodeToString(bytes: Span<UInt8>, options: Data.Base64EncodingOptions = []) -> String {
-        let newCapacity = self.encodeComputeCapacity(bytes: bytes.count, options: options)
+    static func encodeToString(bytes: RawSpan, options: Data.Base64EncodingOptions = []) -> String {
+        let newCapacity = self.encodeComputeCapacity(bytes: bytes.byteCount, options: options)
 
         return String(unsafeUninitializedCapacity: newCapacity) { buffer -> Int in
-            var outputSpan = OutputRawSpan(buffer: UnsafeMutableRawBufferPointer(buffer), initializedCount: 0)
+            let ptr = UnsafeMutableRawBufferPointer(buffer)
+            var outputSpan = OutputRawSpan(buffer: ptr, initializedCount: 0)
             Self._encode(input: bytes, buffer: &outputSpan, options: options)
-            return outputSpan.byteCount
+            return outputSpan.finalize(for: ptr)
         }
     }
 
-    static func encodeToData(bytes: Span<UInt8>, options: Data.Base64EncodingOptions = []) -> Data {
-        let newCapacity = self.encodeComputeCapacity(bytes: bytes.count, options: options)
+    static func encodeToData(bytes: RawSpan, options: Data.Base64EncodingOptions = []) -> Data {
+        let newCapacity = self.encodeComputeCapacity(bytes: bytes.byteCount, options: options)
 
-        return Data(capacity: newCapacity) { (buffer: inout OutputBuffer<UInt8>) in
-            var outputSpan = OutputRawSpan(
-                buffer: UnsafeMutableRawBufferPointer(
-                    start: UnsafeMutableRawPointer(buffer.start),
-                    count: buffer.capacity
-                ),
-                initializedCount: 0
-            )
-            Self._encode(input: bytes, buffer: &outputSpan, options: options)
-            buffer.initialized = outputSpan.byteCount
+        return Data(capacity: newCapacity) { (span: inout OutputRawSpan) in
+            Self._encode(input: bytes, buffer: &span, options: options)
         }
     }
 
-    static func _encode(input: Span<UInt8>, buffer: inout OutputRawSpan, options: Data.Base64EncodingOptions) {
+    static func _encode(input: RawSpan, buffer: inout OutputRawSpan, options: Data.Base64EncodingOptions) {
         if options.contains(.lineLength64Characters) || options.contains(.lineLength76Characters) {
             return self._encodeWithLineBreaks(input: input, buffer: &buffer, options: options)
         }
@@ -343,16 +336,16 @@ extension Base64 {
         let omitPaddingCharacter = options.contains(.omitPaddingCharacter)
 
         Self.withEncodingTables(options: options) { (e0, e1) throws(Never) -> Void in
-            let to = input.count / 3 * 3
+            let to = input.byteCount / 3 * 3
 
             self.loopEncode(e0, e1, input: input.extracting(0..<to), output: &buffer)
 
-            if to < input.count {
+            if to < input.byteCount {
                 let index = to
 
                 let i1 = input[unchecked: index] // fine, since index = to and to < input.count
-                let i2 = index &+ 1 < input.count ? input[unchecked: index &+ 1] : nil // range check in the same line
-                let i3 = index &+ 2 < input.count ? input[unchecked: index &+ 2] : nil // range check in the same line
+                let i2 = index &+ 1 < input.byteCount ? input[unchecked: index &+ 1] : nil // range check in the same line
+                let i3 = index &+ 2 < input.byteCount ? input[unchecked: index &+ 2] : nil // range check in the same line
 
                 buffer.append(e0[i1])
 
@@ -379,7 +372,7 @@ extension Base64 {
     }
 
     static func _encodeWithLineBreaks(
-        input: borrowing Span<UInt8>,
+        input: borrowing RawSpan,
         buffer: inout OutputRawSpan,
         options: Data.Base64EncodingOptions
     ) {
@@ -393,7 +386,7 @@ extension Base64 {
             57
         }
 
-        let lines = input.count / lineLength
+        let lines = input.byteCount / lineLength
 
         let separatorByte1: UInt8
         let separatorByte2: UInt8?
@@ -416,7 +409,7 @@ extension Base64 {
             //       can never wrap.
 
             // first full line
-            if input.count >= lineLength {
+            if input.byteCount >= lineLength {
                 self.loopEncode(e0, e1, input: input.extracting(0..<lineLength), output: &buffer)
             }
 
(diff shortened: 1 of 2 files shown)


fabianfett:
Performance numbers from my machine:

```
  Instructions (p50) — primary metric

  ┌───────────────────────────────┬──────┬─────────┬──────┬────────┐
  │           Benchmark           │ main │ current │  Δ   │   Δ%   │
  ├───────────────────────────────┼──────┼─────────┼──────┼────────┤
  │ 1MB-toData-noOptions (K)      │ 8327 │    8331 │   +4 │ +0.05% │
  ├───────────────────────────────┼──────┼─────────┼──────┼────────┤
  │ 1MB-toString-lineLength64 (K) │ 7827 │    8188 │ +361 │ +4.61% │
  ├───────────────────────────────┼──────┼─────────┼──────┼────────┤
  │ 1MB-toString-noOptions (K)    │ 7557 │    7561 │   +4 │ +0.05% │
  ├───────────────────────────────┼──────┼─────────┼──────┼────────┤
  │ jwtHeader-toString-noOptions  │ 1286 │    1287 │   +1 │ +0.08% │
  └───────────────────────────────┴──────┴─────────┴──────┴────────┘

  Instructions across percentiles

  ┌───────────────────────────────┬─────────┬──────┬──────┬──────┬──────┬──────┬──────┬──────┐
  │           Benchmark           │ Branch  │  p0  │ p25  │ p50  │ p75  │ p90  │ p99  │ p100 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │ 1MB-toData-noOptions (K)      │ main    │ 8130 │ 8319 │ 8327 │ 8352 │ 8360 │ 8372 │ 8441 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │                               │ current │ 8183 │ 8323 │ 8331 │ 8352 │ 8364 │ 8376 │ 8788 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │ 1MB-toString-lineLength64 (K) │ main    │ 7824 │ 7827 │ 7827 │ 7827 │ 7832 │ 7848 │ 8867 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │                               │ current │ 8187 │ 8188 │ 8188 │ 8188 │ 8196 │ 8212 │ 9263 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │ 1MB-toString-noOptions (K)    │ main    │ 7556 │ 7557 │ 7557 │ 7557 │ 7557 │ 7578 │ 8531 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │                               │ current │ 7557 │ 7561 │ 7561 │ 7561 │ 7565 │ 7578 │ 8532 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │ jwtHeader-toString-noOptions  │ main    │ 1285 │ 1286 │ 1286 │ 1286 │ 1286 │ 1296 │ 1322 │
  ├───────────────────────────────┼─────────┼──────┼──────┼──────┼──────┼──────┼──────┼──────┤
  │                               │ current │ 1287 │ 1287 │ 1287 │ 1287 │ 1287 │ 1297 │ 1371 │
  └───────────────────────────────┴─────────┴──────┴──────┴──────┴──────┴──────┴──────┴──────┘

  Time (total CPU, p50)

  ┌───────────────────────────────────┬──────┬─────────┬───────┐
  │             Benchmark             │ main │ current │  Δ%   │
  ├───────────────────────────────────┼──────┼─────────┼───────┤
  │ 1MB-toData-noOptions (μs)         │  337 │     338 │ +0.3% │
  ├───────────────────────────────────┼──────┼─────────┼───────┤
  │ 1MB-toString-lineLength64 (μs)    │  278 │     283 │ +1.8% │
  ├───────────────────────────────────┼──────┼─────────┼───────┤
  │ 1MB-toString-noOptions (μs)       │  268 │     266 │ −0.7% │
  ├───────────────────────────────────┼──────┼─────────┼───────┤
  │ jwtHeader-toString-noOptions (ns) │   67 │      65 │ −3.0% │
  └───────────────────────────────────┴──────┴─────────┴───────┘

  Malloc (total)

  Unchanged everywhere: 2 for toData, 1 for all toString benchmarks.
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,78 @@ extension Base64 {
...
-                rawBuffer.withMemoryRebound(to: UInt8.self) { buffer in
-                    var length = newCapacity
-                    Self._encode(input: input, buffer: buffer, length: &length, options: options)
-                    return length
-                }
-            }
-            return data
-        }) {
-            return result
+        return Data(capacity: newCapacity) { (buffer: inout OutputBuffer<UInt8>) in
```
jmschonfeld: Instead of using the `OutputBuffer` initializer, we should use the `OutputRawSpan` initializer directly

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,78 @@ extension Base64 {
...
-        if let result = bytes.withContiguousStorageIfAvailable({ input -> [UInt8] in
-            [UInt8](unsafeUninitializedCapacity: newCapacity) { buffer, length in
-                Self._encode(input: input, buffer: buffer, length: &length, options: options)
-            }
-        }) {
-            return result
+        return String(unsafeUninitializedCapacity: newCapacity) { buffer -> Int in
+            var outputSpan = OutputRawSpan(buffer: UnsafeMutableRawBufferPointer(buffer), initializedCount: 0)
+            Self._encode(input: bytes, buffer: &outputSpan, options: options)
+            return outputSpan.byteCount
```
jmschonfeld: Although I think this works, we should use `outputSpan.finalize()` for this

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,78 @@ extension Base64 {
...
-
-            length = outIndex
         }
     }
 
     static func _encodeWithLineBreaks(
-        input: UnsafeBufferPointer<UInt8>,
-        buffer: UnsafeMutableBufferPointer<UInt8>,
-        length: inout Int,
+        input: borrowing Span<UInt8>,
```
jmschonfeld: I think we likely want to be using `RawSpan` here rather than assuming the memory should be bound to `UInt8`s

jmschonfeld (commented):
Haven't done a detailed review yet - will come back and do another more detailed pass but wanted to leave a few initial comments

fabianfett:
@jmschonfeld thanks for the first round. updated

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,71 @@ extension Base64 {
...
+            self.loopEncode(e0, e1, input: input.extracting(0..<to), output: &buffer)
 
-            if to < input.count {
+            if to < input.byteCount {
                 let index = to
 
-                let i1 = input[index]
-                let i2 = index &+ 1 < input.count ? input[index &+ 1] : nil
-                let i3 = index &+ 2 < input.count ? input[index &+ 2] : nil
+                let i1 = input[unchecked: index] // fine, since index = to and to < input.count
```
jmschonfeld, line 346: Out of curiosity: is the compiler smart enough to optimize away the bounds checking here without this (and this is just for the guarantee) or can the compiler actually not reason about that?

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -455,104 +404,94 @@ extension Base64 {
...
-            let i1 = input[index]
-            let i2 = input[index &+ 1]
-            let i3 = input[index &+ 2]
-            output[outIndex] = e0[i1]
-            output[outIndex &+ 1] = e1[((i1 & 0x03) &<< 4) | ((i2 &>> 4) & 0x0F)]
-            output[outIndex &+ 2] = e1[((i2 & 0x0F) &<< 2) | ((i3 &>> 6) & 0x03)]
-            output[outIndex &+ 3] = e1[i3]
-            outIndex &+= 4
-            index &+= 3
+        output.withUnsafeMutableBytes { outPtr, initializedCount in
```
jmschonfeld, line 480: I think this deserves a comment explaining why we need to drop down to unsafe APIs. When I was originally playing around with this I did see that the `append` bounds checking caused some severe performance issues so I'm assuming this is needed to eliminate that bounds checking (which we know isn't required since we calculate the buffer size correctly prior to this) but I think we should explain that. Additionally, is there something the compiler can/should optimize here so that we don't need to do this?

jmschonfeld:
Overall looks good, thanks for tackling this! A few comments about the use of unchecked subscripts but the pattern makes sense to me in general

Sources/FoundationEssentials/Data/Data+Base64.swift
MahdiBM: I would expect LLVM can reason about that but not sure in this specific case (Fabian might have tried).
In my experience these kinds of simple stuff are easily done by LLVM. Here it can reason that `input` is not mutable, `to < input.byteCount` -> `index < input.byteCount` -> index is always in bounds.
Also `to` is clearly `> 0` since `byteCount` is wrapped in `_assumeNonNegative` which makes LLVM assume it as such (see the impl below).

The `RawSpan` underlying impl doesn't look to be obstructing the way for that either, apparently they had to add dedicated SIL semantics to ensure _something_ about the index checks although I don't know why they'd be needed.

```swift
  @export(implementation)
  @_semantics("fixed_storage.get_count")
  public var byteCount: Int { _assumeNonNegative(_count) }

  public var byteOffsets: Range<Int> {
    unsafe Range(_uncheckedBounds: (0, byteCount))
  }

  @_semantics("fixed_storage.check_index")
  @export(implementation) @inline(__always)
  internal func _checkIndex(_ position: Int) {
    _precondition(byteOffsets.contains(position), "Index out of bounds")
  }

  @export(implementation) @inline(__always)
  public subscript(_ byteOffset: Int) -> UInt8 {
    _checkIndex(byteOffset)
    return unsafe self[unchecked: byteOffset]
  }
```

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,71 @@ extension Base64 {
...
-            if to < input.count {
+            if to < input.byteCount {
                 let index = to
 
-                let i1 = input[index]
-                let i2 = index &+ 1 < input.count ? input[index &+ 1] : nil
-                let i3 = index &+ 2 < input.count ? input[index &+ 2] : nil
+                let i1 = input[unchecked: index] // fine, since index = to and to < input.count
+                let i2 = index &+ 1 < input.byteCount ? input[unchecked: index &+ 1] : nil // range check in the same line
+                let i3 = index &+ 2 < input.byteCount ? input[unchecked: index &+ 2] : nil // range check in the same line
```
MahdiBM, line 348: This for example, I personally would file a bug against the compiler if the `unchecked` here is needed. It would mean Swift/`RawSpan` are not exposing enough info to LLVM.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -309,122 +309,71 @@ extension Base64 {
...
 
-                buffer[outIndex] = e0[i1]
+                buffer.append(e0[i1])
 
                 if let i2 = i2 {
-                    buffer[outIndex &+ 1] = e1[((i1 & 0x03) &<< 4) | ((i2 &>> 4) & 0x0F)]
+                    buffer.append(e1[((i1 & 0x03) &<< 4) | ((i2 &>> 4) & 0x0F)])
                     if let i3 = i3 {
-                        buffer[outIndex &+ 2] = e1[((i2 & 0x0F) &<< 2) | ((i3 &>> 6) & 0x03)]
```
MahdiBM, line 398: I expect these `append`s to have a negative impact. However this is just a one-off not in the loop so it's ok.

Sources/FoundationEssentials/Data/Data+Base64.swift
MahdiBM: Correct, not only the bounds checks but also: https://github.com/swiftlang/swift-foundation/pull/2191/changes#r4007444905.
Bounds checks though have a bigger impact here.

MahdiBM (approved):
Waiting on this so I can rebase https://github.com/swiftlang/swift-foundation/pull/2236 on top of it 🙂.

PR #2194, opened 2026-08-20
Don't allow brackets in a non-IP-literal URL host

jrflat:
Don't allow `[` or `]` anywhere in a URL host that's not an IP-literal.

### Motivation:

The non-special-scheme case of host parsing currently disallows a partial IP literal (starting with `[` but not ending with `]`), but otherwise allows `[` and `]` to be percent-encoded within. This means a URL like `scheme://[v1.a@example]` would be parsed as having encoded host `example%5D`.

The previous (v1) URL parsing rejected this because of its wider application of IDNA-encoding, which would fail for `[` or `]`. We should restore that behavior by rejecting `[` and `]` anywhere in a non-IP-literal host. This is also consistent with strict RFC 3986, which doesn't allow `[` and `]` in `reg-name`, and with the WHATWG URL spec.

### Modifications:

Fail URL parsing if a non-IP-literal host contains `[` or `]`.

### Result:

`URL(string:)` returns `nil` for non-IP-literal hosts that contain `[` or `]`.

### Testing:

Added unit test.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Parsing.swift
+++ b/Sources/FoundationEssentials/URL/URL_Parsing.swift
@@ -595,9 +595,9 @@ extension URL {
                         // host, it would remain in the final string, so fail now.
                         if containsInvalidASCII(host: host) { return false }
                     } else {
-                        // Don't allow an unterminated IP-literal for any scheme
+                        // Don't allow partial IP-literal brackets for any scheme.
                         assert(!host.isEmpty) // Validation failed, so host is non-empty
-                        if host[0] == UInt8(ascii: "[") { return false }
+                        if containsIPLiteralBracket(host: host) { return false }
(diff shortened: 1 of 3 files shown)


jrflat:
@swift-ci please test

jmschonfeld (approved):

jrflat:
Failures are unrelated test runner issues:
CMake:
```
CMake Error at build/_deps/swiftsyntax-src/CMakeLists.txt:9 (cmake_minimum_required):
  CMake 3.29 or higher is required.  You are running version 3.28.3
```
macOS via Swiftly:
```
Swiftly is currently unlinked and will not manage the active toolchain. You can
run
the following command to link swiftly to the active toolchain:

    $ swiftly link
```

PR #2195, opened 2026-08-20
Add compact long number notation

garywade:
### Summary
- Adds `.compactLong` to `NumberFormatStyleConfiguration.Notation`, the full-name sibling of the existing `.compactName` short-form notation, backed by ICU's
`compact-long` skeleton stem.
- Evolution proposal: Proposals/0043-compact-long-number-notation.md (abbreviated review)
- Known limitation: currency formatting currently falls back to `.compactName`'s short-form output, since ICU/CLDR has no long-form compact currency pattern data
yet — documented in the proposal and in the API doc comment. 

### Motivation:
- The compact long option is missing.  Users wish to use it.

### Modifications:
- Added the option so that FormatStyle calls can make use of it

### Result:
- Users will be able to format values with the long-form name of values such as "1 billion"

### Test plan
- [x] `swift test --filter NumberFormatStyleTests` passes (42/42), including `currency_compactLong` and `plainStyle_compactLong`
- [x] `swift test --filter FoundationInternationalizationTests` passes (401/401, no regressions)

### DIFF
--- a/Proposals/NNNN-compact-long-number-notation.md
+++ b/Proposals/NNNN-compact-long-number-notation.md
@@ -1,0 +1,99 @@
+# Compact Long Number Notation
+
+* Proposal: [SF-NNNN](NNNN-compact-long-number-notation.md)
+* Authors: [Gary Wade](https://github.com/garywade)
+* Review Manager: TBD
+* Status: **Awaiting review**
+* Implementation: TBD
+
+## Introduction/Motivation
+
+When formatting numbers, percents, or currencies, Foundation already supports a compact notation that abbreviates large values using locale-appropriate symbols or short suffixes:
+
+```swift
+let count = 1_500_000
+let result = count.formatted(.number.notation(.compactName))
+assert(result == "1.5M")
+```
+
+ICU's number-skeleton syntax supports two flavors of compact notation: `compact-short` (which `.compactName` already uses) and `compact-long`, which spells out the same magnitude using full unit names instead of abbreviated symbols, e.g. "1.5 million" instead of "1.5M". Foundation does not currently expose `compact-long`.
+
+The long form is useful whenever an abbreviated symbol is ambiguous, unfamiliar, or simply less appropriate for the context than a spelled-out word — for example, some locales' abbreviated compact symbols are already combined or overloaded at very large magnitudes, where the spelled-out form reads more clearly.
+
+## Proposed solution
+
+Add a `compactLong` notation option alongside the existing `compactName` option, available anywhere `Notation` is used today (integer, floating-point, and currency format styles):
+
+```swift
+let compactLongFormatted = 1234.formatted(.number
+    .locale(Locale(identifier: "en_US"))
+    .notation(.compactLong)) // "1.2 thousand"
+```
+
+## Detailed design
+
+```swift
+extension NumberFormatStyleConfiguration.Notation {
+    /// A locale-appropriate compact long name notation.
+    ///
+    /// A compact long name notation, when available in the format style's locale, that uses the full-length
+    /// unit names corresponding to powers of ten, such as "thousand" or "million", instead of the abbreviated
+    /// symbols that ``compactName`` uses. The following example shows a compact long name notation in the
+    /// `en_US` locale:
+    ///
+    /// ```swift
+    /// let compactLongFormatted = 1234.formatted(.number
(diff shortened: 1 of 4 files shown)


Proposals/NNNN-compact-long-number-notation.md
```diff
@@ -0,0 +1,99 @@
...
+
+Since `CurrencyFormatStyleConfiguration.Notation` is a `typealias` for `NumberFormatStyleConfiguration.Notation` (added in [SF-0008](0008-notation-formatting-for-currencies.md)), this single addition is automatically available to the integer, floating-point, and decimal currency format styles as well, with no further API surface needed:
+
+```swift
+let price = Decimal(1_500_000.59)
+let result = price.formatted(.currency(code: "USD").notation(.compactLong))
+assert(result == "$1.5M") // short-form fallback, not "$1.5 million" — see "Known limitation" below
+```
+
+Internally, this maps to ICU's `compact-long` number-skeleton stem, the direct sibling of the `compact-short` stem that `.compactName` already uses — no new ICU data or plumbing is required for the plain-number case.
```
itingliu, line 67: This entire paragraph sounds like implementation detail. I'd probably drop those

Proposals/NNNN-compact-long-number-notation.md
```diff
@@ -0,0 +1,99 @@
...
+    public static var compactLong: Self { .init(option: .compactLong) }
+}
+```
+
+Since `CurrencyFormatStyleConfiguration.Notation` is a `typealias` for `NumberFormatStyleConfiguration.Notation` (added in [SF-0008](0008-notation-formatting-for-currencies.md)), this single addition is automatically available to the integer, floating-point, and decimal currency format styles as well, with no further API surface needed:
+
+```swift
+let price = Decimal(1_500_000.59)
+let result = price.formatted(.currency(code: "USD").notation(.compactLong))
+assert(result == "$1.5M") // short-form fallback, not "$1.5 million" — see "Known limitation" below
```
itingliu, line 64: This is contradicting the statement that "currency format gets this for free" statement above if it actually doesn't work for currency format

Proposals/NNNN-compact-long-number-notation.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+Internally, this maps to ICU's `compact-long` number-skeleton stem, the direct sibling of the `compact-short` stem that `.compactName` already uses — no new ICU data or plumbing is required for the plain-number case.
+
+### Known limitation: currency formatting falls back to short form
+
+ICU/CLDR currently ships long-form (`patternsLong`) compact pattern data only for plain decimal formatting, not for currency formatting (only `patternsShort` exists for currencies, across all locales). As a result, `notation(.compactLong)` applied to a currency format style silently produces the same output as `notation(.compactName)` today (e.g. `"$1.5M"` rather than `"$1.5 million"`), rather than an error. This is an ICU/CLDR-level gap, not something this proposal's implementation can work around at the Foundation layer — there's no ICU skeleton combination that yields symbol-prefixed currency amounts with spelled-out magnitude words. Plain number and percent formatting are unaffected and get genuine long-form output.
```
itingliu, line 71: I appreciate the limitation callout. But a big chunk of this paragraph is talking about implementation again. Do these add values to this document? If not can we simplify it and reduce what doesn't add value?

itingliu: Another limitation worth calling out is that this isn't parseable

Proposals/NNNN-compact-long-number-notation.md
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// ```swift
+    /// let compactLongFormatted = 1234.formatted(.number
+    ///     .locale(Locale(identifier: "en_US"))
+    ///     .notation(.compactLong)) // "1.2 thousand"
+    /// ```
+    ///
+    /// - note: We do not support parsing a number string containing localized prefixes or suffixes.
+    /// - note: When combined with a currency format style, this currently falls back to the same abbreviated
+    ///   output as ``compactName``, since ICU does not yet provide long-form compact patterns for currency;
```
itingliu, line 52: How and why ICU supports and doesn't support isn't strictly relevant to Foundation users. Unless that makes a difference here I would not mention that

PR #2197, opened 2026-08-20
Remove OutputBuffer

jmschonfeld:
Remove `OutputBuffer` in favor of `OutputSpan`

### Motivation:

`OutputBuffer` was an experimental, early version of `OutputSpan`. Only a few pieces still use it, and now most of the APIs it provides are provided by `OutputSpan` itself. At this point, we can transition the final use sites over to `OutputSpan` and remove `OutputBuffer`

### Modifications:

- Transition final uses of `OutputBuffer` to `OutputSpan`
- Remove `OutputBuffer`

### Result:

`OutputBuffer` is no longer present an its extensions on `Data` and `String` are no longer confusingly ambiguous with the "real" `OutputSpan` versions

### Testing:

Verified all existing tests still pass. This removal does not change behavior so no additional tests are required.


glessard (approved):

itingliu (approved):
Thank you! I've been wanting to do this too

Sources/FoundationEssentials/Span+Utils.swift
```diff
@@ -119,6 +163,14 @@ extension String {
             #endif
         }
     }
+
+    package init<E>(_capacity capacity: Int, initializingWith body: (inout OutputSpan<UTF8.CodeUnit>) throws(E) -> Void) throws(E) {
```
parkera, line 167: Hm, if we have a `UTF8.CodeUnit` type inside the `OutputSpan`, does that mean we can also skip UTF8 validation?

jmschonfeld: Unfortunately it does not. `UTF8.CodeUnit` is just a type alias to `UInt8`, so it's valid to have a series of `UTF8.CodeUnit` scalars that combine to malformed UTF-8. Theoretically we could design some `UTF8OutputSpan` that guarantees valid contents as you append (by requiring that you append full unicode scalars rather than individual UTF-8 scalars) but `String` does not offer any entry points (even unsafely) to skip the validation IIRC.

PR #2202, opened 2026-08-22
Move string encoding tables to Swift

jmschonfeld:
Moves the macOS Roman + NextStep encoding tables to Swift rather than C Shims

### Motivation:

In #1991 we determined a way to put static binary data into Swift code rather than C code. We can adopt that here as well to remove a large C shim we have

### Modifications:

- Moves the string tables from C to Swift
- Updates the calling code to operate on spans rather than unsafe pointers

### Result:

Behavior should be the same, but performance should improve (based on godbolt inspections of generated code) from reduced copies

### Testing:

Existing unit tests cover this behavior


parkera (approved):
Great!

PR #2208, opened 2026-08-26
Keep the directory slash in standardizedFileURL for paths ending in a dot segment

### LINKED ISSUE #2207: _URL.hasDirectoryPath is false for paths ending in "." or "..", so standardizedFileURL drops the directory trailing slash (differs from _SwiftURL and Swift 6.3)
### Summary

Follow-up from #2198 (as suggested by @jmschonfeld there). For a relative file URL whose path ends in a `.` or `..` component, the v2 `_URL` backing returns `hasDirectoryPath == false`, where `_SwiftURL` and Swift 6.3 return `true`. `standardizedFileURL` inherits the wrong hint (`hasDirectoryPath ? .isDirectory : .notDirectory`), so the standardized result loses its directory trailing slash while `.standardized` keeps it.

### Reproduction

```swift
import Foundation
let root = URL(filePath: "/srv/root/", directoryHint: .isDirectory)
for p in ["a/x/..", "a/x/."] {
    let u = URL(filePath: p, relativeTo: root)
    print(p, u.hasDirectoryPath, u.standardized.absoluteString, u.standardizedFileURL.absoluteString)
}
```

On a `_URL`-backed Foundation (Linux `swiftlang/swift:nightly-6.4.x-jammy` and `nightly-main-jammy` 2026-08-21, and `FoundationEssentials` built from main 767ff1c on macOS):

```
a/x/.. false file:///srv/root/a/  file:///srv/root/a    <- stdFile lost the slash
a/x/.  false file:///srv/root/a/x/ file:///srv/root/a/x <- same
```

On `_SwiftURL`-backed Foundation (macOS 27.0 beta 26A5416b) and on Swift 6.3.3 (macOS 26.5.2 and Linux `swift:6.3-jammy`):

```
a/x/.. true  file:///srv/root/a/  file:///srv/root/a/
```

So `.standardized` and `.standardizedFileURL` agree everywhere except on `_URL`, and `_URL` disagrees with both `_SwiftURL` and Swift 6.3 about `hasDirectoryPath` itself. (This is why it initially looked platform-dependent in the #2198 discussion — the macOS 27 beta routes through `_SwiftURL`, the Linux nightlies through `_URL`; building `FoundationEssentials` from main on macOS reproduces the Linux behavior, so it is a backing-class difference, not a platform one.)

### Where

- `_SwiftURL.hasDirectoryPath` treats a last path component of `"."` or `".."` as a directory (URL_Swift.swift, `path.lastPathComponent == "." || path.lastPathComponent == ".."`).
- `_URL` sets its `.hasDirectoryPath` flag at parse time from `hasTrailingSeparator()`, which only checks for a trailing `/` (and `\` for Windows paths) — the `.`/`..` rule is absent (URL_Impl.swift).
- `_URL.standardizedFileURL` then builds `URL(filePath: fileSystemPath.standardizingPath, directoryHint: hasDirectoryPath ? .isDirectory : .notDirectory)`, so the flag decides the trailing slash.

### Expected

`hasDirectoryPath == true` for paths ending in `.`/`..` (a path ending in a dot segment necessarily refers to a directory), and `standardizedFileURL` preserving the trailing slash, matching `_SwiftURL`, `.standardized`, and Swift 6.3 behavior.

Found while property-testing path-resolution invariants in [hegel-swift's SecurityProperties example](https://github.com/nassersala/hegel-swift/tree/main/Examples/SecurityProperties) (the same metamorphic relation that surfaced #2198).

nassersala:
Fixes #2207.

### Motivation:

For a file URL whose path ends in a `.` or `..` component, `standardizedFileURL` returns a URL without the trailing directory slash, while `.standardized` keeps it:

```swift
let base = URL(filePath: "/base/dir/", directoryHint: .isDirectory)
let url = URL(filePath: "a/x/..", relativeTo: base)
url.standardized.absoluteString        // "file:///base/dir/a/"
url.standardizedFileURL.absoluteString // "file:///base/dir/a"
```

A path ending in a dot segment always refers to a directory, so the standardized URL should keep the slash. `_SwiftURL` and Swift 6.3 return `file:///base/dir/a/` for both properties.

The cause: `_URL.standardizedFileURL` derives its directory hint from `hasDirectoryPath`, which reports `false` for file paths ending in a dot segment. That flag value is intentional (`testDeletingLastPathComponent` pins it for compatibility with pre-`_SwiftURL` Foundation), so this change leaves `hasDirectoryPath` alone and computes the directory hint locally.

### Modifications:

- `_URL.standardizedFileURL` now also treats the URL as a directory when its own path ends in a `.` or `..` component, using the existing `URL.hasDirectoryPath` helper that the URL string parser uses. The check runs on the URL's own path rather than `fileSystemPath`, because resolving a relative URL against its base already removes the trailing dot segments.

### Result:

`standardizedFileURL` keeps the trailing directory slash for paths ending in `.` or `..`, matching `.standardized`, `_SwiftURL`, and Swift 6.3. `hasDirectoryPath` is unchanged.

### Testing:

- New test `standardizedFileURLDotSegmentEndings` covering relative `a/x/..`, `a/x/.`, `a/..` against a directory base and an absolute path ending in `..`.
- Full test suite passes with a 2026-08-11 main snapshot toolchain on macOS.

### DIFF
--- a/Sources/FoundationEssentials/URL/URL_Impl.swift
+++ b/Sources/FoundationEssentials/URL/URL_Impl.swift
@@ -1470,10 +1470,8 @@ extension _URL {
         guard !path.isEmpty else {
             return nil
         }
-        // A path ending in a "." or ".." component refers to a directory,
-        // even though hasDirectoryPath reports false for such file paths for
-        // compatibility. Check the URL's own path, since resolving a relative
-        // URL against its base already removes the trailing dot segments.
+        // The file-path initializers don't set .hasDirectoryPath for a trailing
+        // "." or ".." component, so re-check the unresolved path here.
         let isDirectory = hasDirectoryPath || withPathSpan { pathSpan in
             pathSpan.withUnsafeBufferPointer { buffer in
                 URL.hasDirectoryPath(buffer, pathEnd: buffer.count, pathLength: buffer.count)
--- a/Sources/FoundationEssentials/URL/URL_Impl.swift
+++ b/Sources/FoundationEssentials/URL/URL_Impl.swift
@@ -1470,7 +1470,16 @@ extension _URL {
         guard !path.isEmpty else {
             return nil
         }
-        return _URL(filePath: path.standardizingPath, directoryHint: hasDirectoryPath ? .isDirectory : .notDirectory).url
+        // A path ending in a "." or ".." component refers to a directory,
+        // even though hasDirectoryPath reports false for such file paths for
+        // compatibility. Check the URL's own path, since resolving a relative
+        // URL against its base already removes the trailing dot segments.
+        let isDirectory = hasDirectoryPath || withPathSpan { pathSpan in
+            pathSpan.withUnsafeBufferPointer { buffer in
+                URL.hasDirectoryPath(buffer, pathEnd: buffer.count, pathLength: buffer.count)
+            }
+        }
+        return _URL(filePath: path.standardizingPath, directoryHint: isDirectory ? .isDirectory : .notDirectory).url
     }
 
     func resolvingSymlinksInPath() -> URL? {
--- a/Sources/FoundationEssentials/URL/URL_Impl.swift
+++ b/Sources/FoundationEssentials/URL/URL_Impl.swift
@@ -1470,8 +1470,7 @@ extension _URL {
         guard !path.isEmpty else {
             return nil
         }
-        // The file-path initializers don't set .hasDirectoryPath for a trailing
-        // "." or ".." component, so re-check the unresolved path here.
+        // The file-path initializers don't set .hasDirectoryPath for a trailing "." or ".." component, so re-check the unresolved path here.
         let isDirectory = hasDirectoryPath || withPathSpan { pathSpan in
             pathSpan.withUnsafeBufferPointer { buffer in
                 URL.hasDirectoryPath(buffer, pathEnd: buffer.count, pathLength: buffer.count)
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -4493,10 +4493,7 @@ private struct URLTests {
     }
 
     @Test func standardizedFileURLDotSegmentEndings() throws {
-        // A path ending in a "." or ".." component refers to a directory,
-        // but the file path initializers don't set .hasDirectoryPath for
-        // these trailing components. The standardized file URL must still
-        // end in a directory slash, consistent with .standardized.
+        // A path ending in a "." or ".." component refers to a directory, but the file path initializers don't set .hasDirectoryPath for these trailing components. The standardized file URL must still end in a directory slash, consistent with .standardized.
         let base = URL(filePath: "/base/dir/", directoryHint: .isDirectory)
         let cases = [
             ("a/x/..", "file:///base/dir/a/"),
--- a/Tests/FoundationEssentialsTests/URLTests.swift
+++ b/Tests/FoundationEssentialsTests/URLTests.swift
@@ -561,6 +561,27 @@ private struct URLTests {
         try FileManager.default.removeItem(at: URL(filePath: "\(tempDirectory.path)/tmp-dir"))
     }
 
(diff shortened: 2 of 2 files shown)


Sources/FoundationEssentials/URL/URL_Impl.swift
```diff
@@ -1470,7 +1470,16 @@ extension _URL {
         guard !path.isEmpty else {
             return nil
         }
-        return _URL(filePath: path.standardizingPath, directoryHint: hasDirectoryPath ? .isDirectory : .notDirectory).url
+        // A path ending in a "." or ".." component refers to a directory,
+        // even though hasDirectoryPath reports false for such file paths for
+        // compatibility. Check the URL's own path, since resolving a relative
+        // URL against its base already removes the trailing dot segments.
```
jrflat: Could we tighten this comment a bit and clarify that it's only the file path initializers (e.g. `URL(filePath:)` and friends) that don't store the `.hasDirectoryPath` flag? Maybe something like:

```
// The file-path initializers don't set .hasDirectoryPath for a trailing
// "." or ".." component, so re-check the unresolved path here.
```

Tests/FoundationEssentialsTests/URLTests.swift
```diff
@@ -561,6 +561,27 @@ private struct URLTests {
         try FileManager.default.removeItem(at: URL(filePath: "\(tempDirectory.path)/tmp-dir"))
     }
 
+    @Test func standardizedFileURLDotSegmentEndings() throws {
+        // A path ending in a "." or ".." component refers to a directory.
+        // hasDirectoryPath reports false for such file paths for compatibility
+        // (see deletingLastPathComponent), but the standardized file URL must
+        // still end in a directory slash, consistent with .standardized (#2207).
```
jrflat: Few minor nits: 
- Let's also clarify here that it's the file path _initializers_ that don't set `.hasDirectoryPath` for these trailing components, and let's remove the reference to `deletingLastPathComponent` since it's a bit confusing that it's referring to a comment in the `deletingLastPathComponent` test case and not the URL method itself.
- Generally, we won't include an issue or PR number in a comment since that comes with the git history.
- Could we move this test case near the other `standardizedFileURLAndResolvingSymlinks` test towards the bottom of this file for some (admittedly loose) organization?

jrflat (commented):
Nice targeted fix, thank you! I left a few minor nits on comment clarifications but overall the change looks good. I'll run some cross-repo testing to make sure this doesn't accidentally break `swift-corelibs-foundation` tests (which also test `URL`)

jrflat:
@swift-ci please test

Sources/FoundationEssentials/URL/URL_Impl.swift
nassersala: Done in 3ac9530.

Tests/FoundationEssentialsTests/URLTests.swift
nassersala: Done in 3ac9530.

nassersala: One correction in cf54ea3: the move in 3ac9530 accidentally placed the test inside the `#if !os(Windows)` guard that belongs to `standardizedFileURLAndResolvingSymlinks`. It now sits just above the guard, so it still runs on Windows as it did in its original location.

Tests/FoundationEssentialsTests/URLTests.swift
```diff
@@ -4492,6 +4492,27 @@ private struct URLTests {
         #expect(url.query() == nil)
     }
 
+    @Test func standardizedFileURLDotSegmentEndings() throws {
+        // A path ending in a "." or ".." component refers to a directory,
+        // but the file path initializers don't set .hasDirectoryPath for
+        // these trailing components. The standardized file URL must still
+        // end in a directory slash, consistent with .standardized.
```
parkera: Please don't manually wrap comments in the Foundation repo - thanks

nassersala: Unwrapped both comments in 65f8496 (the one in URL_Impl.swift as well).

jrflat (approved):
Thanks!

PR #2210, opened 2026-08-27
[FoundationEssentials] Fix JSONDecoder silently parsing JSON5 Infinity into Decimal.quietNaN

ArihantMarwaha:
### Motivation:

Decimal itself doesn't have support for Infinity values yet. Even the part of the old `NSJSONSerialization` implementation that would try to reinterpret an `NaN` or `Infinity` value as an `NSDecimalNumber` did not have very predictable behavior.

Currently, when `JSONDecoder` is configured to parse JSON5 (`allowsJSON5 = true`), encountering `+Infinity`, 
`-Infinity`, or `Infinity` for a `Decimal` value, silently returns `Decimal.quietNaN`. Because `Decimal` does not have a representation for infinity, the decoder should throw an error instead of silently falling back to `NaN`, which is mathematically incorrect.

### Modifications:

- Updated `unwrapDecimal` logic in `JSONDecoder.swift`
- The string is already pre-validated by `JSON5Scanner.prevalidateJSONNumber` as a valid special double value.
- the code now simply checks if the buffer contains the `I` character.
- If it contains `I`, it correctly throws `JSONError.numberIsNotRepresentableInSwift`.
- If it does not contain `I` (meaning it is `NaN`), it continues to safely return `Decimal.quietNaN`.

### Result:

Decoding `Infinity` into a `Decimal` using JSON5 correctly throws a `DecodingError.dataCorrupted` error (wrapping the `numberIsNotRepresentableInSwift` error) instead of silently succeeding with `quietNaN`.

### Testing:

- Added a new test `decodeDecimalInfinityAndNaN` in `JSONEncoderTests.swift` that verifies `NaN` successfully decodes as `quietNaN`, and that `Infinity` and `-Infinity` successfully throw an error.

### DIFF
--- a/Sources/FoundationEssentials/JSON/JSONDecoder.swift
+++ b/Sources/FoundationEssentials/JSON/JSONDecoder.swift
@@ -800,8 +800,11 @@ extension JSONDecoderImpl: Decoder {
                         return decimal
                     }
                 } else if isSpecialJSON5DoubleValue {
-                    // Decimal itself doesn't have support for Infinity values yet. Even the part of the old NSJSONSerialization implementation that would try to reinterpret an NaN or Infinity value as an NSDecimalNumber did not have very predictable behavior.
-                    // TODO: Proper handling of Infinity and NaN Decimal values.
+                    
+                    if numberBuffer.contains(UInt8(ascii: "I")) {
+                        throw JSONError.numberIsNotRepresentableInSwift(parsed: String(decoding: numberBuffer, as: UTF8.self))
+                    }
+                    
                     return Decimal.quietNaN
                 } else {
                     switch Decimal._decimal(from: numberBuffer, matchEntireString: true) {
--- a/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
+++ b/Tests/FoundationEssentialsTests/JSONEncoderTests.swift
@@ -2870,7 +2870,7 @@ extension JSONEncoderTests {
     
     @Test func decodeDecimalInfinityAndNaN() throws {
         let decoder = JSONDecoder()
-        decoder.allowsJSON5 = true  // <-- Changed this line
+        decoder.allowsJSON5 = true
         
         let nanData = "NaN".data(using: .utf8)!
         let nanDecimal = try decoder.decode(Decimal.self, from: nanData)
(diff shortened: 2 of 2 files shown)


ArihantMarwaha:
@parkera could you please run tests for these.

Sources/FoundationEssentials/JSON/JSONDecoder.swift
```diff
@@ -800,8 +800,11 @@ extension JSONDecoderImpl: Decoder {
                         return decimal
                     }
                 } else if isSpecialJSON5DoubleValue {
-                    // Decimal itself doesn't have support for Infinity values yet. Even the part of the old NSJSONSerialization implementation that would try to reinterpret an NaN or Infinity value as an NSDecimalNumber did not have very predictable behavior.
-                    // TODO: Proper handling of Infinity and NaN Decimal values.
+                    
+                    if numberBuffer.contains(UInt8(ascii: "I")) {
+                        throw JSONError.numberIsNotRepresentableInSwift(parsed: String(decoding: numberBuffer, as: UTF8.self))
+                    }
+
```
parkera, line 807: Why is it better to throw than to return `NaN`?

ArihantMarwaha: My main thought process was just trying to keep the behavior consistent with how `Decimal` already works in Swift, and avoiding silent data loss.

If you try to do `Decimal(string: "Infinity")` anywhere else, it just returns `nil`.It doesn't try to force it into `NaN`.

 I figured `JSONDecoder` should match that same strictness and fail instead of pretending it's `NaN`

also`.overlargeValue` follows a similar structure and throws the same `numberIsNotRepresentableInSwift` error.

ArihantMarwaha:
@kperryua can you please review this TODO

xwu:
@ArihantMarwaha Can you check again the behavior on main with the reworked `Decimal`?

Also, I agree with @parkera that I'm not sure that throwing is more correct here: when doing math with `Decimal`, overflow is to `nan` in place of (unrepresentable) infinity so the currently shipping behavior here seems aligned with that.

If I were starting from scratch, I'd want to modify `Decimal(string: "inf")` to return `nan` but there'd be binary compatibility issues with doing so and it'd require a proposal review. But I don't think that means we should change over other behaviors to match that API.

PR #2213, opened 2026-08-28
[FoundationEssentials] Safely unwrap Calendar month adjustment during enumeration

ArihantMarwaha:
Safely unwrap the Calendar month adjustment during enumeration

### Motivation:

In `Calendar_Enumerate.swift`, computing the adjusted components during a backward day enumeration forces an unwrap on the result of `date(byAdding: .month, value: -1, to: date)`.

 If this adjustment fails (for example, if the search reaches the extreme minimum boundary of the calendar's supported range), it will cause a fatal crash.

### Modifications:

Replaced the force-unwrap (`!`) with a safe `guard let`.

If the month subtraction returns `nil`, the function now safely falls back to returning the unadjusted components, allowing the enumeration loop to gracefully handle the boundary rather than crashing the application.

### Result:

Boundary date searches in `Calendar` enumeration are now structurally safe from force-unwrap crashes when traversing months.

### Testing:

Successfully ran `FoundationEssentialsTests` via Swift Package Manager locally. 

Since this mitigates an extreme out-of-bounds crash, no new explicit boundary test cases were required, but all existing`Calendar` regression tests pass successfully.

### DIFF
--- a/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
+++ b/Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
@@ -695,7 +695,9 @@ extension Calendar {
                 let dateDay = component(.day, from: date)
                 // We need to make sure we don't surpass the day we want
                 if comps.day ?? Int.max >= dateDay {
-                    let tempDate = self.date(byAdding: .month, value: -1, to: date)! // TODO: Check force unwrap here
+                    guard let tempDate = self.date(byAdding: .month, value: -1, to: date) else {
+                         return adjusted
+                    }
                     adjusted.month = component(.month, from: tempDate)
                 } else {
                     // adjusted is the date components we're trying to match against; dateDay is the current day of the current search date.


parkera:
@swift-ci test

parkera (commented):
Seems reasonable to me. Note we have another TODO about `nil` a few lines up as well.

@itingliu I think my main question is if this will result in an infinite loop (I know we protect against this higher up in the algorithm, though)

ArihantMarwaha:
@itingliu and @parkera If `date(byAdding:)` returns nil at the absolute limits of the calendar and we safely return adjusted without modifying month, `_matchingDate` will attempt to search using the unadjusted month. If that causes it to return nil, the while true loop in enumerateDates will catch it and attempt to try again.

However, because that loop is strictly guarded by `STOP_EXHAUSTIVE_SEARCH_AFTER_MAX_ITERATIONS` (which is capped at 100), even in the absolute worst-case mathematical scenario where it fails to bump the date, it will safely bail out after 100 iterations. Triggering an exhaustive exit is vastly preferable to a fatal force unwrap crash in production!

Regarding the other TODO: can year ever be nil here? on line 683: It looks like the code directly underneath it 

`if let adjustedDate = self.date(from: adjusted)` already handles the nil scenario gracefully, so there isn't an active risk of a crash there.

If you'd like, I can remove both TODO comments in this PR since the logic safely handles both cases now?

itingliu (commented):
> If this adjustment fails (for example, if the search reaches the extreme minimum boundary of the calendar's supported range), it will cause a fatal crash.

This fix seems reasonable, but do you know if we can reproduce this hypothetical crash? I think we had this force unwrap because we were fairly confident that this force unwrap would succeed. So if it actually can fail I think it's useful to know what circumstances this would fail

ArihantMarwaha:
@itingliu That is a very fair question! I

Actually ran a battery of boundary tests (testing `Date.distantPast`, `-.greatestFiniteMagnitude`, and exotic ICU calendars), and you are exactly right: standard built-in ICU calendars actively clamp extreme boundaries rather than returning `nil`. For a standard `Calendar(identifier: .gregorian)`, subtracting 1 month will essentially never fail in isolation.

However, the force-unwrap is still architecturally unsafe for two reasons:

1. Calendar is a protocol that can be backed by custom NSCalendar implementations bridged from Objective-C. A custom calendar is perfectly allowed to return nil for unsupported out-of-bounds dates, which would trigger a fatal crash here.

2. If a future ICU update changes behavior to return `U_ILLEGAL_ARGUMENT_ERROR` at absolute mathematical limits instead of silently clamping, this line becomes a ticking time bomb.

Since swapping the ! for a guard lets seamlessly defer any hypothetical failure to the algorithm's existing 100-iteration exhaust loop, it guarantees safety against custom bridged calendars and future ICU changes without altering standard behavior. I don't have a reproducible crash for built-in calendars, but I think the resilience upgrade is worth the guard let swap!

Sources/FoundationEssentials/Calendar/Calendar_Enumerate.swift
itingliu, line 675: We could also change this function to `throw` and handle the failure on the callsite. But that can be done as a separate PR

itingliu (approved):

itingliu:
This seems fine. Thanks!

PR #2214, opened 2026-08-28
Add missing Data.resetBytes lower bounds check

jmschonfeld:
Adds a missing bounds check

### Motivation:

`Data.resetBytes` validated that `range.lowerBound >= 0`, but not `range.lowerBound >= startIndex`

### Modifications:

Defer bounds checking to the representation types and ensure that `startIndex` is used instead of 0 where applicable

### Result:

`Data.resetBytes` correctly checks bounds

### Testing:

Added additional exit tests to validate preconditions


jmschonfeld:

parkera (approved):

PR #2215, opened 2026-08-29
Add Buddhist, Japanese and Republic of China calendar implementations

dra8an:
### Description:

  Adds the Buddhist, Japanese and Republic of China calendars as an era table on `_CalendarGregorian`, rather than as three new classes.

  ### Motivation:

  These three calendars use Gregorian arithmetic. They differ from it only in how they label eras and number years inside them. Written as separate classes, each one would repeat the whole of Gregorian's date and component conversion, and only about five members actually differ.

  ### Modifications:

  `_CalendarGregorian` gains a stored `eraTable`. `_calendarClass(identifier:)` routes `.buddhist`, `.japanese` and `.republicOfChina` to that same class, so one class serves five identifiers alongside `.gregorian` and `.iso8601`. The era table is the only thing that varies, so it lives on the class as data, the same way `.iso8601` already gets different defaults inside that one class.

  `CalendarEra.swift` is new. It holds the era types and the four era tables as static values. The tables are static so that a calendar copy does not allocate one.

  Era codes match the established numbering. Buddhist BE is 0, ROC is Before Minguo 0 and Minguo 1, and Japanese runs Meiji 232 through Reiwa 236. Codes 0 and 1 stay reserved for the Gregorian BCE and CE eras, which lets the Japanese calendar inherit them before Meiji.

  `hash(into:)` and `isDateInWeekend(_:)` were duplicated across calendars, so they moved to `_CalendarUtility` and became defaults on `_CalendarProtocol`.

  ### Result:

  Three calendars become available behind feature flags, off by default, matching Hebrew and Chinese. Nothing changes for existing callers while the flags are off.

  Two Japanese behaviors are worth calling out:

  * An era ends where the next one begins, so successive eras meet exactly, with no overlap and no gap.
  * Dates before Meiji report the inherited Gregorian era, following CLDR (unicode-org/icu#4019, ICU-23341). `maximumRange(of: .era)` stays `0..<237` to match the reserved code space.

  We can change either behavior if reviewers would rather have strict parity with the currently bundled data.

  ### Testing:

  Seven new suites, all passing with no known issues.

  Five compare all three calendars against `_CalendarICU` for era, year, month, day, `date(from:)`, era intervals, component ranges, `ordinality(of:in: .era)` and the  Hashable contract. The parity probes run 18 dates from 1000 BCE to 2025, covering every era boundary. A recurrence rule probe compares full `recurrences(of:)` sequences  and crosses each calendar's era boundary.

  Two suites hold golden values rather than comparing. `ROC Backward Era` covers the only era in the family that counts years backward. `Japanese Gregorian Era Inheritance` covers the two behaviors above, including that consecutive era intervals meet exactly.

  The dates affected by those two behaviors are skipped from the ICU comparison, since the golden-value suite pins them and it does not depend on which data version is bundled.


Sources/FoundationEssentials/Calendar/Calendar_Cache.swift
```diff
@@ -39,9 +39,27 @@ internal func foundation_swift_chinese_calendar_feature_enabled() -> Bool {
     // _foundation_swift_chinese_calendar_feature_enabled() — once Apple adds the underscored binding
     return false
 }
+
+internal func foundation_swift_buddhist_calendar_feature_enabled() -> Bool {
+    // Call _foundation_swift_buddhist_calendar_feature_enabled() once Apple adds the underscored binding.
```
richgillam, line 44: Can you explain what this means?

richgillam (approved):
Didn't review everyone in great detail, but it looks sound, and I love the improvement in design!

Sources/FoundationEssentials/Calendar/Calendar_Cache.swift
itingliu: I think it's more helpful to write this as a TODO:

```suggestion
    // TODO: return the runtime feature flag value
```

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -86,8 +86,14 @@ enum ResolvedDateComponents {
             rawYear = 1
         }
 
-        if adjustEra && components.era == 0 /* BC */{
-           rawYear = 1 - rawYear
+        // The era table turns an era-relative year into an extended one. A code this calendar's own table does not list, like Japanese reading a pre-Meiji era, is an inherited Gregorian era instead.
+        // Only an era the caller actually set counts as inherited BCE. An absent era must not: the empty table used to find an era's own boundary defaults to code 0, which would flip the sign.
```
itingliu: I don't understand this comment. What does "Only an era the caller actually set counts as inherited BCE." mean? Also, this sentence seems incomplete, "An absent era must not:"?

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,154 @@
...
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// The direction an era counts its years.
+internal enum _CalendarEraDirection: Sendable {
```
itingliu: Can we nest this into `_CalendarEraEntry` since it's only used there? This comes with the benefit that we can just call it `enum Direction`

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,154 @@
...
+        self.erasCanEnd = entries.filter { $0.direction == .forward && !$0.labelsEveryDate }.count > 1
+    }
+
+    /// The era labeling the given date, or nil when the date falls before every era in the table.
+    func entry(extendedYear year: Int, month: Int, day: Int) -> _CalendarEraEntry? {
+        entries.first { $0.labels(extendedYear: year, month: month, day: day) }
+    }
+
+    func entry(code: Int) -> _CalendarEraEntry? {
+        entries.first { $0.code == code }
```
itingliu: Should we consider using a dictionary like `[Int /*code*/: _CalendarEraEntry]` so we can have a O(1) lookup? Or `[(Int, _CalnedarEraEntry)]` for binary search?

(Looking at the callsites I think we really want to make these O(1) since previously these were constant checks for Gregorian Calendar, but now we have to search the table)

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /// The era that follows `entry`, or nil when `entry` is the newest one.
+    func eraAfter(_ entry: _CalendarEraEntry) -> _CalendarEraEntry? {
+        guard let index = entries.firstIndex(where: { $0.code == entry.code }), index > 0 else { return nil }
```
itingliu: Similarly it seems unnecessary to look up the matching entry using `firstIndex` when we control all the passed-in`[_CalendarEraEntry]` which we know is already sorted by `code`

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,164 @@
...
+/// The direction an era counts its years.
+internal enum _CalendarEraDirection: Sendable {
+    /// Years count up from the anchor year, like every Japanese era and the Buddhist and Minguo eras.
+    case forward
+    /// Years count down from the anchor year, like ROC's Before-Minguo era.
+    case backward
+}
+
+/// One era of a Gregorian-family calendar.
+internal struct _CalendarEraEntry: Sendable {
```
itingliu: Should we call this something like `_GregorianFamilyCalendarEra` per your comment?

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,164 @@
...
+    func extendedYear(fromEraYear year: Int) -> Int {
+        switch direction {
+        case .forward: return year + anchorYear - 1
+        case .backward: return anchorYear - year
+        }
+    }
+}
+
+/// The eras of one Gregorian-family calendar, newest first.
+internal struct _CalendarEraTable: Sendable {
```
itingliu: Similarly, maybe we can call this something like `_GregorianFamilyCalendarEras`?

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,160 @@
...
+    /// Years count up from the anchor year, like every Japanese era and the Buddhist and Minguo eras.
+    case forward
+    /// Years count down from the anchor year, like ROC's Before-Minguo era.
+    case backward
+}
+
+/// One era of a Gregorian-family calendar.
+internal struct _CalendarEraEntry: Sendable {
+    /// The value `DateComponents.era` carries for this era. These match ICU's numbering.
+    let code: Int
```
itingliu: Nit: Is "code" a term of art? I find it a little confusing because it's in fact just an ordinal number, whereas "code" sounds like it's some artificially assigned label. If not, can we call it something that explains that, like ... idk, `ordering`, `ordinal`?

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,160 @@
...
+
+/// The eras of one Gregorian-family calendar, newest first.
+internal struct _CalendarEraTable: Sendable {
+    let entries: [_CalendarEraEntry]
+
+    /// The highest era code. Era ranges run from 0 up to this value, because codes 0 and 1 stay reserved for the Gregorian BCE and CE eras a calendar inherits.
+    let highestCode: Int
+
+    /// The code to assume when `DateComponents.era` is absent, which is the newest era in the table.
+    let defaultCode: Int
```
itingliu: Similarly on the naming of "code": in this context it feels it'd be more natural to just call it something like "defaultEraNumber` or something like that

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,160 @@
...
+        }
+    }
+}
+
+/// The eras of one Gregorian-family calendar, newest first.
+internal struct _CalendarEraTable: Sendable {
+    let entries: [_CalendarEraEntry]
+
+    /// The highest era code. Era ranges run from 0 up to this value, because codes 0 and 1 stay reserved for the Gregorian BCE and CE eras a calendar inherits.
+    let highestCode: Int
```
itingliu: Similarly, on the use of "code" in this name: based on your comment it sounds like we can call it something like `rangeUpperBound` or `eraRangeUpperBound` or `maxEraNumber`... something along those lines

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -1631,11 +1637,66 @@ package final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
...
+            guard let next = eraTable.eraAfter(entry), let end = eraBoundary(of: next) else {
+                return DateInterval(start: boundary, duration: Calendar._maxDateIntervalDuration)
+            }
+            return DateInterval(start: boundary, end: end)
+        }
+
+        // Older than every forward era. A backward era covers that, otherwise the era is inherited.
+        if let backward = eraTable.entries.first(where: { $0.direction == .backward }),
+           let boundary = eraBoundary(of: backward) {
+            return DateInterval(start: boundary - Calendar._maxDateIntervalDuration, end: boundary)
```
itingliu, line 1667: Nit: is it me or is this indentation weird?

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -1631,11 +1637,66 @@ package final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
...
+        // Find the era by instant rather than by extended year. At the CE boundary the two disagree, because the boundary is proleptic while the arithmetic is Julian aware.
+        for entry in eraTable.entries where entry.direction == .forward {
+            guard let boundary = eraBoundary(of: entry), date >= boundary else { continue }
+
+            // An era ends where the next one begins, so successive eras meet exactly and never overlap.
+            guard let next = eraTable.eraAfter(entry), let end = eraBoundary(of: next) else {
+                return DateInterval(start: boundary, duration: Calendar._maxDateIntervalDuration)
+            }
+            return DateInterval(start: boundary, end: end)
+        }
```
itingliu: Would it make sense to build the era boundaries in advance, and store those along/inside `_CalendarEraEntry`? That way we can do a simple binary search to find the entry given a `Date`

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -2379,7 +2442,7 @@ package final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
         case .yearForWeekOfYear:
             var dc = dateComponents(weekBasedComponents, from: dateInWholeSecond, in: timeZone)
             var amount = amount
-            if let era = dc.era, era == 0 {
+            if eraCountsBackward(dc.era) {
```
itingliu: I think we would really want to make `eraCountsBackward` a O(1)

Sources/FoundationEssentials/Calendar/Calendar_Cache.swift
dra8an: Replaced with TODO

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
dra8an: Rewritten to say why the fallback reads `components.era` rather than the defaulted number

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done, now `_GregorianFamilyCalendarEra.Direction`.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Renamed to `eraNumber`. Not a fan of `ordinal`,CLDR assigns these as keys in `supplementalData`, so Meiji is 232 rather than a count, and `japancal.h` notes the numbering is arbitrary and may change.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done, `maxEraNumber`.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done, `defaultEraNumber`.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done, as a dense array indexed by era number rather than a dictionary, so there's no hashing. Verified it matches the old scan for every number across all five tables.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Deleted the function. The one caller already loops, so iterating by index removes the search entirely.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
dra8an: Era boundaries can't live on the era itself because they depend on the calendar's time zone. The Meiji boundary lands on a different instant in GMT, Tokyo and Los Angeles.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
dra8an: Fixed, folded onto one line.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
dra8an: Done, now two integer compares against values precomputed once per table.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+// Copyright (c) 2026 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+/// One era of a Gregorian-family calendar.
+internal struct _GregorianFamilyCalendarEra: Sendable {
```
parkera: We haven't been prefixing internal types with `_` in Swift - the type system gives us enough info to know if it is public or not.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+
+    /// The direction an era counts its years.
+    enum Direction: Sendable {
```
parkera: We already have `Calendar.SearchDirection`

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+    func entry(eraNumber: Int) -> _GregorianFamilyCalendarEra? {
+        let index = eraNumber - lowestEraNumber
+        guard erasByNumber.indices.contains(index) else { return nil }
+        return erasByNumber[index]
+    }
+}
+
+/// The era tables themselves, one per calendar that `_CalendarGregorian` serves.
+///
+/// These are stored properties, so each table is built once for the process. A function returning a fresh table would allocate on every calendar copy.
```
parkera: We don't really need the play by play comments here.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+    ])
+
+    /// One era, which numbers every date, including those before its own start. 1 CE is 544 BE.
+    static let buddhist = _GregorianFamilyCalendarEras([
+        _GregorianFamilyCalendarEra(eraNumber: 0, anchorYear: -542, startMonth: 1, startDay: 1, direction: .forward)
+    ], coversEveryDate: true)
+
+    /// The five modern eras, newest first, with ICU's numbering (Meiji 232 through Reiwa 236).
+    ///
+    /// CLDR and ICU dropped the pre-Meiji eras (unicode-org/icu#4019, ICU-23341), so earlier dates keep the Gregorian era and numbers 2 through 231 are unused.
```
parkera: Similarly, let's not put references to PRs or other repos into our comments. The comments should describe what *we do here* not *what other projects did elsewhere*.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+    ///
+    /// Meiji starts 1868-09-08 to match Apple's runtime ICU, where the CLDR canonical date is 1868-10-23.
```
parkera: Comments again - avoid describing what we are trying to match, and describe why we would use this value.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
```diff
@@ -0,0 +1,190 @@
...
+    static let japanese = _GregorianFamilyCalendarEras([
+        _GregorianFamilyCalendarEra(eraNumber: 236, anchorYear: 2019, startMonth: 5, startDay: 1, direction: .forward),
+        _GregorianFamilyCalendarEra(eraNumber: 235, anchorYear: 1989, startMonth: 1, startDay: 8, direction: .forward),
+        _GregorianFamilyCalendarEra(eraNumber: 234, anchorYear: 1926, startMonth: 12, startDay: 25, direction: .forward),
+        _GregorianFamilyCalendarEra(eraNumber: 233, anchorYear: 1912, startMonth: 7, startDay: 30, direction: .forward),
+        _GregorianFamilyCalendarEra(eraNumber: 232, anchorYear: 1868, startMonth: 9, startDay: 8, direction: .forward),
+    ], erasCanEnd: true)
+
+    /// Two eras sharing the 1912 boundary, with ICU's numbering from `taiwncal.h` (Before-Minguo 0, Minguo 1).
```
parkera: Why do we need to match ICU's numbering? Let's not name random header files from other projects.

Tests/FoundationInternationalizationTests/GregorianFamilyRecurrenceRuleParityProbe.swift
```diff
@@ -0,0 +1,256 @@
...
+
+    private static func divergences(icu icuDates: [Date], ours ourDates: [Date], label: (Int) -> String) -> [String] {
+        var failures: [String] = []
+        for idx in 0..<max(icuDates.count, ourDates.count) where idx >= icuDates.count || idx >= ourDates.count || icuDates[idx] != ourDates[idx] {
+            failures.append(label(idx))
+        }
+        return failures
+    }
+
+    @available(macOS 15, iOS 18, tvOS 18, watchOS 11, *)
```
parkera: Why do we need availability on the tests?

Tests/FoundationInternationalizationTests/GregorianFamilyRecurrenceRuleParityProbe.swift
```diff
@@ -0,0 +1,256 @@
...
+    @Test(arguments: Family.allCases)
+    private func yearly_christmas(_ family: Family) {
```
parkera: This and below functions have a strange naming convention for Swift.

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
```diff
@@ -0,0 +1,188 @@
...
+import Testing
+
+#if FOUNDATION_FRAMEWORK
+@testable import Foundation
+#else
+@testable import FoundationInternationalization
+@testable import FoundationEssentials
+#endif
+
+/// Golden-value tests for the era the Japanese calendar reports before Meiji (unicode-org/icu#4019, ICU-23341). A pre-Meiji date reports the Gregorian era instead of a pre-Meiji Japanese era.
```
parkera: Same kind of commentary here about referencing ICU or other projects in general. It just doesn't age well when we come back in the future and want to understand this code.

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
```diff
@@ -0,0 +1,188 @@
...
+///
+/// These values are not compared against ICU, because only unreleased ICU behaves this way. The bundled ICU still carries the old 237-era data.
```
parkera: Again, let's not talk about "bundled ICU", because that's not really a thing that we need to think about when evaluating this code in isolation.

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
```diff
@@ -0,0 +1,188 @@
...
+    @Test(arguments: eraCases)
+    func eraAndYear(_ eraCase: EraCase) throws {
+        let date = try Self.gregorianDate(era: eraCase.sourceEra, eraCase.year, eraCase.month, eraCase.day)
+        let components = Self.japanese().dateComponents([.era, .year], from: date)
+        #expect(components.era == eraCase.expectedEra)
+        #expect(components.year == eraCase.expectedYear)
+    }
+
+    @Test func prolepticEraYearRollsIntoGregorian() throws {
+        // ICU JapaneseTest.Test5345: setting Meiji year 1, Jan 1 resolves to CE, because Meiji 1 Jan 1 = Gregorian 1868-01-01, before Meiji's Sept 8 start.
```
parkera: Let's not reference other projects tests.

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
```diff
@@ -1631,11 +1637,67 @@ package final class _CalendarGregorian: _CalendarProtocol, @unchecked Sendable {
         }
     }
 
+    // MARK: - Era lookup
+
+    /// Whether this era counts its years down instead of up, like BCE or ROC's Before-Minguo. Adding or wrapping a year has to move opposite the requested amount in that case.
+    ///
+    /// A code this calendar's own table does not list, like Japanese reading a pre-Meiji date, is an inherited Gregorian era: code 0 is BCE and counts backward, code 1 or absent is CE.
+    func eraCountsBackward(_ eraNumber: Int?) -> Bool {
```
parkera: Should the "era number" be a bespoke type for the era table, instead of passing it around as an Int? Then you could have "counts backward" as a _property_ on the type instead of an API that takes an arbitrary Int argument.

Sources/FoundationEssentials/Calendar/CalendarUtility.swift
```diff
@@ -98,6 +98,30 @@ internal enum _CalendarUtility {
     /// Default weekend range (region 001): Sat–Sun, full day.
     static let defaultWeekendRange = WeekendRange(onsetTime: 0, ceaseTime: 86400, start: 7, end: 1)
 
+    // MARK: - hash(into:)
+
+    /// Combines exactly the state that `Calendar`'s `==` compares, so calendars that compare equal always hash equally.
```
parkera, line 103: It might be a bit strange to put hash here but not equal (do we have that too? I can't recall)

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Removed the prefix.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Oh, ok. Updated the code to reuse it.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Done. All the PR links and ticket numbers are gone.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Fixed

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Removed the header file reference. The numbers themselves have to stay, because `DateComponents.era` is public API and existing callers already round-trip 232 for Meiji.

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
dra8an: Updated to remove the ICU reference

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
dra8an: Same here

Tests/FoundationInternationalizationTests/JapaneseGregorianEraInheritanceTests.swift
dra8an: Removed

Tests/FoundationInternationalizationTests/GregorianFamilyRecurrenceRuleParityProbe.swift
dra8an: Removed all eight.

Tests/FoundationInternationalizationTests/GregorianFamilyRecurrenceRuleParityProbe.swift
dra8an: Renamed to camelCase

Sources/FoundationEssentials/Calendar/Calendar_Gregorian.swift
dra8an: I like that. I will do that in the next commit

Sources/FoundationEssentials/Calendar/CalendarUtility.swift
dra8an: No, there isn't one. `==` is on `Calendar` itself and always has been, so there was never a duplicate to move here. `hash(into:)` is different: each calendar had written its own copy. Both compare the same seven fields, so equal calendars still hash equally.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
dra8an: Trimmed here and in three other places.

Sources/FoundationEssentials/Calendar/CalendarEra.swift
richgillam: Dumb question: `Calendar.SearchDirection` has the same members as Dragan's new `Direction` class, but the two classes have different meanings and are used in different places.  Is this okay?

dra8an: Not a dumb question at all, I had the same hesitation. `SearchDirection` documents its cases as searching for a date earlier or later in time, whereas here it means which way an era numbers its years. However, both cover the same thing, which is a direction. So in order to avoid a second forward/backward type, I think the intent is reasonably obvious to reuse the existing type.

parkera: I agree it's not a dumb question. If we had some other type that was Direction but semantically "left" and "right" - I would create a new type.

It's just that this one felt really close to the other one in both naming and purpose.

Issue #991, opened 2024-10-18
FileManager creating temporary directory has differing behavior between FoundationEssentials/Foundation (Linux)


WFT:
This call has differing behavior when importing `Foundation` vs when importing `FoundationEssentials` on Linux:
```swift
try FileManager.default.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: documents, create: true)
```

# Reproduction

The following snippet exits successfully when importing Foundation (macOS or Linux), but if you instead import FoundationEssentials it throws an error when creating `temp`.

```swift
import Foundation // Comment this line and uncomment the next to see the issue
// import FoundationEssentials

print("Creating documents...")
let documents = try FileManager.default.url(
    for: .documentDirectory,
    in: .userDomainMask,
    appropriateFor: nil,
    create: true
)
print("Creating temporary...")
// Raises FoundationEssentials.CocoaError(code: FoundationEssentials.CocoaError.Code(rawValue: 256), userInfo: [:])
let temp = try FileManager.default.url(
    for: .itemReplacementDirectory,
    in: .userDomainMask,
    appropriateFor: documents,
    create: true
)

print("Results")

print(temp)
print(FileManager.default.fileExists(atPath: temp.path(percentEncoded: false)))
```

You can run this with the `swift:6.0.1` Docker image to see the issue.

# Expected behavior

I think `temp` should be a temporary directory which has been created, probably somewhere like `/tmp/abc123/NSIRD_test.swift_xyz456/`. I feel that’s the best approximation of what happens on macOS (where we get something like `/var/folders/q0/abc123/T/TemporaryItems/NSIRD_test.swift_xyz456/`). The particular format isn’t particularly important — I’m basically just looking for a portable form of `mkdtemp`.

Or at the very least the behavior shouldn’t change when the imported module changes.

# Actual behavior

An error is raised from the indicated line (FoundationEssentials only).

tothambrus11:
This issue is still present on Windows. Our GitHub CI passes but I get a failure when running locally. Maybe GitHub's processes run with default permissions?

Anyways, here is a minimal reproducer: https://github.com/tothambrus11/repro-swift-foundation-essentials-inconsistency

Hokila:
[...]

PR #2219, opened 2026-08-31
Handle item replacement directories in FoundationEssentials

Hokila:
### Motivation

Resolves #991.

FoundationEssentials currently falls through to normal search path lookup for `.itemReplacementDirectory`. That lookup returns no URL, so `FileManager.url(for:in:appropriateFor:create:)` throws on Linux when callers ask for a replacement directory.

### Modifications

Add a FoundationEssentials path for `.itemReplacementDirectory` when the domain is `.userDomainMask` and `appropriateFor` is supplied. It creates a `TemporaryItems` container under the process temporary directory when that directory is on the same volume as the reference URL. Otherwise it falls back to a writable directory based on the reference URL, matching the older corelibs behavior.

The replacement directory itself is always created, including for `create: false`, so callers get a unique directory and avoid races.

Add a regression test that covers `create: false` and `create: true`, verifies the returned directories exist, and checks that repeated calls return distinct directories.

### Validation

- `git diff --check`
- `swiftc -parse Sources/FoundationEssentials/FileManager/FileManager+Directories.swift Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift`

I could not run `swift test --filter FileManager/itemReplacementDirectory` locally with Xcode Swift 6.2.3. The package starts building on the `release/6.4.x` base, but fails in existing source that requires a newer compiler (`@specialized` and `@_lifetime` diagnostics in `String+Internals.swift`) before it reaches this test.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -60,6 +60,45 @@ extension _FileManagerImpl {
     var temporaryDirectory: URL {
         URL(filePath: String.temporaryDirectoryPath, directoryHint: .isDirectory)
     }
+
+    private func itemReplacementDirectory(appropriateFor reference: URL) throws -> URL {
+        guard reference.isFileURL else {
+            throw CocoaError.errorWithFilePath(.fileWriteUnsupportedScheme, reference)
+        }
+
+        let temporaryItemsDirectory = temporaryDirectory.appending(component: "TemporaryItems", directoryHint: .isDirectory)
+        let referenceDirectory = reference.deletingPathExtension()
+        let useTemporaryDirectory: Bool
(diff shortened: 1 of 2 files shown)


Hokila:
Thanks for taking a look.

I checked for existing work before opening this. #991 is still open with no assignee, and I did not find an open PR that handles `.itemReplacementDirectory` for the FoundationEssentials path.

This PR keeps the change scoped to that path and adds a regression test for both `create: false` and `create: true`. I could run `git diff --check` and parse the touched files locally. A full `swift test --filter FileManager/itemReplacementDirectory` build starts on `release/6.4.x`, but my local Xcode Swift 6.2.3 toolchain fails on existing newer language features before it reaches the test, so I am relying on CI for the full run.

Hokila:
Reopened as #2220 after renaming the fork branch to remove the codex prefix.

PR #2220, opened 2026-08-31
Handle item replacement directories in FoundationEssentials

### LINKED ISSUE #991: FileManager creating temporary directory has differing behavior between FoundationEssentials/Foundation (Linux)

Hokila:
### Motivation

Resolves #991.

FoundationEssentials did not handle `.itemReplacementDirectory` in `FileManager.url(for:in:appropriateFor:create:)`. It fell through to normal search-path lookup, which could fail to return a replacement directory.

### Modifications

- Add FoundationEssentials handling for `.itemReplacementDirectory` with a user-domain reference URL.
- Select the replacement container based on the reference volume, using `TemporaryItems` when appropriate and a writable reference location otherwise.
- Always create a unique replacement directory, including when `create` is `false`, matching corelibs Foundation behavior and avoiding races.
- Use the existing FoundationEssentials `.systemNumber` file attribute for volume comparison.
- Add regression coverage for both `create: true` and `create: false`, including existence and uniqueness checks.

### Validation

Using `6.4.x-snapshot-2026-08-28` (`Apple Swift 6.4-dev`):

```text
swift test --filter itemReplacementDirectory

Test itemReplacementDirectory(create:) with 2 test cases passed.
```

Both `create: true` and `create: false` cases pass.

### DIFF
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -115,7 +115,7 @@ extension _FileManagerImpl {
         create shouldCreate: Bool
     ) throws -> URL {
         #if FOUNDATION_FRAMEWORK
-        // TODO: Support correct trash/replacement locations in swift-foundation
+        // TODO: Support correct trash locations in FoundationEssentials
         #if os(macOS) || os(iOS)
         if let url, directory == .trashDirectory {
             return try fileManager._URLForTrashingItem(at: url, create: shouldCreate)
--- a/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
+++ b/Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
@@ -62,6 +62,51 @@ extension _FileManagerImpl {
     var temporaryDirectory: URL {
         URL(filePath: String.temporaryDirectoryPath, directoryHint: .isDirectory)
     }
+
+    #if !FOUNDATION_FRAMEWORK
+    private func itemReplacementDirectory(appropriateFor reference: URL) throws -> URL {
+        guard reference.isFileURL else {
+            throw CocoaError.errorWithFilePath(.fileWriteUnsupportedScheme, reference)
+        }
+
+        let temporaryDirectory = fileManager.temporaryDirectory
+        let temporaryItemsDirectory = temporaryDirectory.appending(component: "TemporaryItems", directoryHint: .isDirectory)
+        let referenceDirectory = reference.deletingPathExtension()
+        let useTemporaryDirectory: Bool
+        if let temporaryVolumeIdentifier = systemNumber(for: temporaryDirectory),
+           let referenceVolumeIdentifier = systemNumber(for: reference) {
+            useTemporaryDirectory = temporaryVolumeIdentifier == referenceVolumeIdentifier
+        } else {
+            useTemporaryDirectory = !fileManager.isWritableFile(atPath: referenceDirectory.path)
+        }
+
+        let containerDirectory = useTemporaryDirectory ? temporaryItemsDirectory : referenceDirectory
+        try fileManager.createDirectory(at: containerDirectory, withIntermediateDirectories: true, attributes: [.posixPermissions: 0o700])
+
+        var attempt = 0
+        while true {
+            let replacementDirectory = containerDirectory.appending(component: itemReplacementDirectoryName(forAttempt: attempt), directoryHint: .isDirectory)
+            do {
+                try fileManager.createDirectory(at: replacementDirectory, withIntermediateDirectories: false)
+                return replacementDirectory
+            } catch let error as CocoaError where error.code == .fileWriteFileExists {
+                attempt += 1
+            }
+        }
+
+        func systemNumber(for url: URL) -> UInt? {
+            try? fileManager.attributesOfItem(atPath: url.path)[.systemNumber] as? UInt
+        }
+    }
+
+    private func itemReplacementDirectoryName(forAttempt attempt: Int) -> String {
+        let processName = ProcessInfo.processInfo.processName.filter { $0.isLetter || $0.isNumber }
+        if attempt == 0 {
+            return "(A Document Being Saved By \(processName))"
+        }
+        return "(A Document Being Saved By \(processName) \(attempt + 1))"
+    }
+    #endif
     
     func url(
         for directory: FileManager.SearchPathDirectory,
@@ -76,10 +121,16 @@ extension _FileManagerImpl {
             return try fileManager._URLForTrashingItem(at: url, create: shouldCreate)
         }
         #endif
+        #endif
         if let url, domain == .userDomainMask, directory == .itemReplacementDirectory {
+            #if FOUNDATION_FRAMEWORK
             // The only place we need to do this is for certain operations, namely the replacing item API.
             return try fileManager._URLForReplacingItem(at: url)
+            #else
+            return try itemReplacementDirectory(appropriateFor: url)
+            #endif
         }
+        #if FOUNDATION_FRAMEWORK
         var domain = domain
         if domain == .systemDomainMask {
             domain = ._partitionedSystemDomainMask
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -1110,12 +1110,22 @@ private struct FileManagerTests {
                 try? fileManager.removeItem(at: secondReplacementDirectory)
             }
 
-            var isDirectory = false
-            #expect(fileManager.fileExists(atPath: replacementDirectory.path, isDirectory: &isDirectory))
-            #expect(isDirectory)
-            isDirectory = false
-            #expect(fileManager.fileExists(atPath: secondReplacementDirectory.path, isDirectory: &isDirectory))
-            #expect(isDirectory)
+            #if FOUNDATION_FRAMEWORK
+            var isDir: ObjCBool = false
+            func isDirBool() -> Bool {
+                isDir.boolValue
+            }
+            #else
+            var isDir: Bool = false
+            func isDirBool() -> Bool {
+                isDir
+            }
+            #endif
+            #expect(fileManager.fileExists(atPath: replacementDirectory.path, isDirectory: &isDir))
+            #expect(isDirBool())
+            isDir = false
+            #expect(fileManager.fileExists(atPath: secondReplacementDirectory.path, isDirectory: &isDir))
+            #expect(isDirBool())
             #expect(replacementDirectory != reference)
             #expect(secondReplacementDirectory != replacementDirectory)
         }
--- a/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
+++ b/Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
@@ -1096,6 +1096,31 @@ private struct FileManagerTests {
         assertSearchPaths([.itemReplacementDirectory], exists: false)
     }
 
+    @Test(arguments: [false, true])
+    func itemReplacementDirectory(create: Bool) async throws {
+        try await FilePlayground {
+            Directory("Documents") {}
+        }.test { fileManager in
+            let reference = URL.currentDirectory()
+                .appending(component: "Documents", directoryHint: .isDirectory)
+            let replacementDirectory = try fileManager.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: reference, create: create)
+            let secondReplacementDirectory = try fileManager.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: reference, create: create)
+            defer {
+                try? fileManager.removeItem(at: replacementDirectory)
+                try? fileManager.removeItem(at: secondReplacementDirectory)
+            }
+
+            var isDirectory = false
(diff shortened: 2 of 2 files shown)


Hokila:
Thanks for taking a look.

I checked for existing work before opening this. #991 is still open with no assignee, and I did not find an open PR that handles `.itemReplacementDirectory` for the FoundationEssentials path.

This PR keeps the change scoped to that path and adds a regression test for both `create: false` and `create: true`. I could run `git diff --check` and parse the touched files locally. A full `swift test --filter FileManager/itemReplacementDirectory` build starts on `release/6.4.x`, but my local Xcode Swift 6.2.3 toolchain fails on existing newer language features before it reaches the test, so I am relying on CI for the full run.

Hokila:
Validation update:

- The touched source and regression test pass Swift parser/type checks with local Xcode Swift 6.2.3, and `git diff --check` is clean.
- The implementation was cross-checked against the upstream swift-corelibs-foundation implementation for `.itemReplacementDirectory` (volume selection, `TemporaryItems`, unconditional creation, and collision handling).
- The package-level `swift test --filter FileManager/itemReplacementDirectory` still cannot be executed locally because Xcode Swift 6.2.3 fails in pre-existing `String+Internals.swift` code before reaching this test.
- The PR workflow is currently `action_required` with no jobs, so upstream CI has not yet run for this fork PR and needs maintainer approval.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
...
+    private func itemReplacementDirectory(appropriateFor reference: URL) throws -> URL {
+        guard reference.isFileURL else {
+            throw CocoaError.errorWithFilePath(.fileWriteUnsupportedScheme, reference)
+        }
+
+        let temporaryItemsDirectory = temporaryDirectory.appending(component: "TemporaryItems", directoryHint: .isDirectory)
+        let referenceDirectory = reference.deletingPathExtension()
+        let useTemporaryDirectory: Bool
+        if let temporaryVolumeIdentifier = systemNumber(for: temporaryDirectory),
+           let referenceVolumeIdentifier = systemNumber(for: reference) {
```
jmschonfeld, line 77: This diverges from the existing implementation in SCL-F which uses the `volumeIdentifierKey` URL Resource Value. How does the behavior differ between these two?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
     var temporaryDirectory: URL {
         URL(filePath: String.temporaryDirectoryPath, directoryHint: .isDirectory)
     }
+
+    private func itemReplacementDirectory(appropriateFor reference: URL) throws -> URL {
+        guard reference.isFileURL else {
+            throw CocoaError.errorWithFilePath(.fileWriteUnsupportedScheme, reference)
+        }
+
+        let temporaryItemsDirectory = temporaryDirectory.appending(component: "TemporaryItems", directoryHint: .isDirectory)
```
jmschonfeld, line 73: This should delegate to `fileManager.temporaryDirectory`

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
...
+                attempt += 1
+            }
+        }
+
+        func systemNumber(for url: URL) -> UInt? {
+            guard let attributes = try? fileManager.attributesOfItem(atPath: url.path),
+                  let value = attributes[.systemNumber] else {
+                return nil
+            }
+            return _readFileAttributePrimitive(value, as: UInt.self)
```
jmschonfeld: There's no need for `_readFileAttributePrimitive` here because this isn't coming from the user, we should just use the type provided by `attributesOfItem(atPath:)`

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
...
+        }
+    }
+
+    private func itemReplacementDirectoryName(forAttempt attempt: Int) -> String {
+        let processName = ProcessInfo.processInfo.processName.filter { $0.isLetter || $0.isNumber }
+        let sanitizedProcessName = processName.isEmpty ? "Process" : processName
```
jmschonfeld: When would the process name be empty?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
...
+            useTemporaryDirectory = temporaryVolumeIdentifier == referenceVolumeIdentifier
+        } else {
+            useTemporaryDirectory = !fileManager.isWritableFile(atPath: referenceDirectory.path)
+        }
+
+        let containerDirectory = useTemporaryDirectory ? temporaryItemsDirectory : referenceDirectory
+        try fileManager.createDirectory(at: containerDirectory, withIntermediateDirectories: true, attributes: [.posixPermissions: 0o700])
+
+        var attempt = 0
+        while true {
```
jmschonfeld, line 87: Is my understanding correct that this is just taking the existing implementation of this from SCL-F and lowering it to `FoundationEssentials`?

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -60,6 +60,53 @@ extension _FileManagerImpl {
     var temporaryDirectory: URL {
         URL(filePath: String.temporaryDirectoryPath, directoryHint: .isDirectory)
     }
+
+    private func itemReplacementDirectory(appropriateFor reference: URL) throws -> URL {
```
jmschonfeld, line 67: Can you please put these helper functions in `!FOUNDATION_FRAMEWORK` for now? I'll need to combine the `FOUNDATION_FRAMEWORK` functionality into these and until that happens I want to make sure that the framework build continues using the internal implementation of this function (which does more work like localizing the file name, handling sandboxing, etc.)

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
```diff
@@ -83,6 +130,10 @@ extension _FileManagerImpl {
             domain = ._partitionedSystemDomainMask
         }
         #endif
+
+        if let url, domain == .userDomainMask, directory == .itemReplacementDirectory {
```
jmschonfeld: Can you combine this with the call above to make it clear that `fileManager._URLForReplacingItem(at: url)` is used in `FOUNDATION_FRAMEWORK` and otherwise `itemReplacementDirectory(appropriateFor: url)` is used?

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1074,6 +1074,31 @@ private struct FileManagerTests {
         assertSearchPaths([.itemReplacementDirectory], exists: false)
     }
 
+    @Test(arguments: [false, true])
+    func itemReplacementDirectory(create: Bool) async throws {
+        try await FilePlayground {
+            Directory("Documents") {}
+        }.test { fileManager in
+            let reference = URL(filePath: fileManager.currentDirectoryPath, directoryHint: .isDirectory)
```
jmschonfeld:
```suggestion
            let reference = URL.currentDirectory()
```

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: Moved the helper functions under #if !FOUNDATION_FRAMEWORK.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: Updated to use fileManager.temporaryDirectory

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: `URLResourceValues.volumeIdentifier` is not currently available in the non-framework FoundationEssentials build. `.systemNumber` provides the filesystem identifier exposed by `attributesOfItem(atPath:)` (st_dev on POSIX and the volume serial number on Windows), so it provides the same-volume comparison needed here.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: Updated to use the concrete UInt value returned by `attributesOfItem(atPath:)` directly.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: The fallback was unnecessary, so I removed it to match the SCL-F implementation. The sanitized process name is only empty if the original process name contains no alphanumeric characters.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: Yes. This is the SCL-F item-replacement-directory behavior lowered into `FoundationEssentials`, with `.systemNumber` used for volume comparison because URL resource values are not available in the non-framework build.

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
Hokila: Done
Combined the branch so framework builds call `_URLForReplacingItem(at:)`, while FoundationEssentials builds call `itemReplacementDirectory(appropriateFor:).`

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
Hokila: done

Sources/FoundationEssentials/FileManager/FileManager+Directories.swift
jmschonfeld: This comment is now outdated and needs to be updated

Hokila: Thanks.
Updated, replacement directories are now handled above.
implementing trash locations is outside the scope of this PR.

Tests/FoundationEssentialsTests/FileManager/FileManagerTests.swift
```diff
@@ -1096,6 +1096,31 @@ private struct FileManagerTests {
...
+            let reference = URL.currentDirectory()
+                .appending(component: "Documents", directoryHint: .isDirectory)
+            let replacementDirectory = try fileManager.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: reference, create: create)
+            let secondReplacementDirectory = try fileManager.url(for: .itemReplacementDirectory, in: .userDomainMask, appropriateFor: reference, create: create)
+            defer {
+                try? fileManager.removeItem(at: replacementDirectory)
+                try? fileManager.removeItem(at: secondReplacementDirectory)
+            }
+
+            var isDirectory = false
```
jmschonfeld: There's a discrepancy here between the type of this value in `FoundationEssentials` and what's available in Foundation, so this test fails to compile in `FOUNDATION_FRAMEWORK` mode. Could you please update this test to follow the convention set by `fileExistsAtPath` for handling this difference?

Hokila: Updated the test to follow the existing `fileExistsAtPath` convention, using `ObjCBool` for `FOUNDATION_FRAMEWORK` and `Bool` for FoundationEssentials.

jmschonfeld (approved):

PR #2223, opened 2026-09-03
Add >= and <= to Date

stephentyrone:
Without these, Date falls back on the default implementations from Comparable, which do not handle NaN correctly.

Before this change, both `Date(timeIntervalSince1970: .nan) <= Date(timeIntervalSince1970: 0)` and `Date(timeIntervalSince1970: .nan) >= Date(timeIntervalSince1970: 0)` would be `true`. Now they are false.

### DIFF
--- a/Sources/FoundationEssentials/Date.swift
+++ b/Sources/FoundationEssentials/Date.swift
@@ -199,11 +199,13 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
     }
   
     /// Returns true if the left hand `Date` is earlier in time than or equal to the right hand `Date`.
+    @export(implementation)
     public static func <=(lhs: Date, rhs: Date) -> Bool {
         return lhs.timeIntervalSinceReferenceDate <= rhs.timeIntervalSinceReferenceDate
     }
 
     /// Returns true if the left hand `Date` is later in time than or equal to the right hand `Date`.
+    @export(implementation)
     public static func >=(lhs: Date, rhs: Date) -> Bool {
         return lhs.timeIntervalSinceReferenceDate >= rhs.timeIntervalSinceReferenceDate
     }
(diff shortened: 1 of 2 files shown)


Sources/FoundationEssentials/Date.swift
```diff
@@ -197,6 +197,16 @@ public struct Date : Comparable, Hashable, Equatable, Sendable {
     public static func >(lhs: Date, rhs: Date) -> Bool {
         return lhs.timeIntervalSinceReferenceDate > rhs.timeIntervalSinceReferenceDate
     }
+  
+    /// Returns true if the left hand `Date` is earlier in time than or equal to the right hand `Date`.
+    public static func <=(lhs: Date, rhs: Date) -> Bool {
```
jmschonfeld, line 203: I think these will need to be `@export(implementation)` or `@backDeployed` since they can't have stricter availability than the protocol conformance, is that correct?

stephentyrone: IIRC the symbols are already there because the protocol requires them and provides defaults, so we shouldn't need availability, but we can export them and get the fix to more targets.

jmschonfeld: Oh right sorry, I confused declarations that the compiler synthesizes (and creates the ABI symbol for) with default implementations inherited from extensions on the protocol (which don't create a symbol since it uses the symbol from the default implementation instead) - my bad.

I think `@export(implementation)` would actually now remove that symbol right? So maybe just leaving it as it was without availability is the way to go 😅

stephentyrone: I think export(implementation) is right here. We'll still have the protocol witnesses, they'll just hit the new implementations.

jmschonfeld (approved):

PR #2225, opened 2026-09-05
Rtg locale wip

richgillam:
**DRAFT**

**WORK IN PROGRESS**

This is the beginning of my work on eliminating ICU from the Foundation `Locale` class.  At the moment, I have Swift code for the manipulation of base locale IDs (the part before the @ sign), and I've got some very early code for handling the key-value pairs (the part after the @ sign).  This is nowhere near done yet, and is hidden behind a compile-time flag, with the old ICU-based implementation undisturbed.  This is both so I can do side-by-side benchmarking comparisons and because I don't want the new code replacing the old code until everything is actually implemented (or I've got things wired up so the new stuff calls through to the old stuff for things I haven't implemented yet).  I'm posting a PR now just so that people can see what I'm doing and comment on the overall direction.

### DIFF
--- a/Sources/FoundationInternationalization/Locale/Locale_Impl.swift
+++ b/Sources/FoundationInternationalization/Locale/Locale_Impl.swift
@@ -1,0 +1,309 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2026- Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+//
+//===----------------------------------------------------------------------===//
+
+// TODO: Is there boilerplate legalese we have to add to point back to ICU/CLDR as the source of the algorithms?
+
+#if canImport(FoundationEssentials)
+import FoundationEssentials
+#endif
+
+#if FOUNDATION_FRAMEWORK
+// for CFXPreferences call
+internal import _ForSwiftFoundation
+// For Logger
+internal import os
+#endif
+
+internal import _FoundationICU
+internal import Synchronization
+
+#if canImport(Glibc)
+@preconcurrency import Glibc
+#endif
+
+#if FOUNDATION_LOCALE_EXPERIMENTAL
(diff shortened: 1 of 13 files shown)


Sources/FoundationInternationalization/Locale/Locale_Impl.swift
```diff
@@ -0,0 +1,803 @@
...
+        _originalIdentifier = identifier
+        _prefs = prefs
+        
+        let (language, script, region, variant) = Self.parseBaseLocaleID(identifier)
+        var normalizedIdentifier = ""
+        if let language {
+            normalizedIdentifier.append(Self.normalizedLanguageCode(language))
+        }
+        if let script {
+            normalizedIdentifier.append("_\(Self.normalizedScriptCode(script))")
```
itingliu, line 57: I have a feeling that these `.append` will need to be rewritten using `OutputSpan` instead of `String`

Sources/FoundationInternationalization/Locale/Locale_Impl.swift
```diff
@@ -0,0 +1,803 @@
...
+        
+        if let keyValuePairs = Self.normalizeKeyValuePairs(identifier) {
+            // TODO: We might want to use firstIndexOf() rather than two calls to split() here to avoid the transient array creation
+            // TODO: We might also be able to take advantage of the fact that the keys are sorted to speed up the switch (or maybe not-- I don't know how switch is implemented here)
+            for pair in keyValuePairs.dropFirst().split(separator: ";") {
+                let kv = pair.split(separator: "=")
+                let (key, value) = (String(kv[0]), String(kv[1]))
+                
+                // TODO: Is ICULegacyKey really buying us anything here?  Can we get rid of it?  As things stand, we have the mappings between BCP47 and kegacy keys in two spots in the code that have to be kept in sync-- is there a good way to improve on that?
+                switch ICULegacyKey(key) {
```
itingliu, line 102: `ICULegacyKey` was added to make sure we don't accidentally pass in non-legacy-key to ICU API that expects legacy keys ("calendar" vs "ca"). 

Looking at your usage here, it looks like now we are comparing the `key` with a list of known `cases` below to construct a `Locale.Component`, all of which are controlled and owned by us. In that case we probably don't need any of the "ICU" stuff nor "Legacy" stuff.

PR #2236, opened 2026-09-11
Make Base64 encoding faster

MahdiBM:
### Motivation:

The Base64 encoding logic can be made ~3 times more performant.

### Modifications:

Make the Base64 encoding more performant.

### Result:

Better performance.

Result of the current benchmarks, from my local runs are attached below.

benchmark | p90 before | p90 after | speedup | instructions before | after | ratio
-- | -- | -- | -- | -- | -- | --
1MB-toData-noOptions | 411 µs | 144 µs | **2.85x** | 7258 K | 1481 K | **4.90x**
1MB-toString-noOptions | 409 µs | 150 µs | **2.73x** | 7569 K | 1796 K | **4.21x**
1MB-toString-lineLength64 | 423 µs | 162 µs | **2.61x** | 7840 K | 2343 K | **3.35x**
jwtHeader-toString-noOptions | 100 ns | 105 ns | **0.95x** | 1580 | 1662 | **0.95x**

Across different sizes:

input | old µs | new µs | speedup | old GB/s | new GB/s
-- | -- | -- | -- | -- | --
64 KB | 24.2 | 7.9 | 3.06x | 2.71 | 8.29
256 KB | 97.3 | 31.4 | 3.10x | 2.69 | 8.35
1 MB | 390.7 | 126.9 | 3.08x | 2.68 | 8.26
4 MB | 1581.9 | 516.0 | 3.07x | 2.65 | 8.13
16 MB | 6252.3 | 2042.6 | 3.06x | 2.68 | 8.21
64 MB | 37299.1 | 8107.7 | 4.60x | 1.80 | 8.28

On my M1 Macbook Pro this vectorized version looks to simply be maxing out the SIMD compute my Macbook has, for non-trivial amount of encodes.

### Testing:

Existing tests cover most of the code but I added a new test for full coverage and ensuring no edge-case is missed.

### Even Faster:

The implementation can be made even faster similar to the base64 impl over at https://github.com/simdutf/simdutf/.

Currently I've moved the impl to be auto-vectorized by LLVM, but I had to remove the table lookups and manually calculate the values instead.

The faster way is to simply have both SIMD AND table lookups together.

The blocker for that is that LLVM currently quits when it sees a table lookup (tbl) is needed, instead of auto-vectorizing the loop with appropriate instructions such as Arm's `vqtbl4q_u8`.
Furthermore Swift's `SIMD` stuff are not good enough to help here.

One proper way we can do this is to have a dependency on [google's Highway](https://github.com/google/highway) so we don't have to manually deal with SIMD instructions in this package. Highway contains b64 implementations as well as it's not a barebones SIMD helper, but we can simply keep those in Swift and just leverage the SIMD helpers.
In my tests this'll 2x+ the performance compared to this PR.
Then perhaps at some point Swift's SIMD gets better and we can just move to that.

There are also some more issues but those can be worked around. For example some table lookups in Swift can (not always do) cost decently more than just a simple C array lookup, as well as emit other behavior such as not being fully known at compile time, but that can be worked around to an ok extent by using `StaticString` as an storage for static data in a way that LLVM can also "identify array bounds" so it can therefore try to vectorize the loop.
Another workaround would be to simply have a C module and declare the table there.


MahdiBM:
@fabianfett as the one who's worked on the existing Base64 implementations.

MahdiBM:
When reviewing the files on GitHub, this "Hide whitespace" toggle will be clutch since a bunch of lines only changed in indentation.

<kbd> Screenshot 2026-09-12 at 2 47 48 AM </kbd>

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -154,23 +154,6 @@ extension Data {
 
 enum Base64 {}
 
-/// A base64 encoding lookup table.
```
MahdiBM, line 157: Lookup tables no longer useful. See the description of the PR.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -377,48 +244,41 @@ extension Base64 {
 
         let omitPaddingCharacter = options.contains(.omitPaddingCharacter)
 
-        Self.withEncodingTables(options: options) { (e0, e1) throws(Never) -> Void in
-            let to = input.count / 3 * 3
-            var outIndex = 0
+        let (char62, char63) = Self.encodingCharacters(options: options)
```
MahdiBM, line 247: Hoist calc for char62/63.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -542,18 +395,35 @@ extension Base64 {
...
+    @inline(always)
+    private static func encodeCharacter(_ value: UInt8, _ char62: UInt8, _ char63: UInt8) -> UInt8 {
+        var offset = UInt8(ascii: "A")
+        if value >= 26 { offset = UInt8(ascii: "a") &- 26 }
+        if value >= 52 { offset = UInt8(ascii: "0") &- 52 }
+        var character = value &+ offset
+        if value == 62 { character = char62 }
+        if value == 63 { character = char63 }
+        return character
     }
```
MahdiBM, line 426: Manually encode character with no lookup.

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -377,48 +244,41 @@ extension Base64 {
...
-
-                let i1 = input[index]
-                let i2 = index &+ 1 < input.count ? input[index &+ 1] : nil
-                let i3 = index &+ 2 < input.count ? input[index &+ 2] : nil
-
-                buffer[outIndex] = e0[i1]
-
-                if let i2 = i2 {
-                    buffer[outIndex &+ 1] = e1[((i1 & 0x03) &<< 4) | ((i2 &>> 4) & 0x0F)]
-                    if let i3 = i3 {
```
MahdiBM, line 397: `i3` was logically always `nil`.

```swift
let to = input.count / 3 * 3

...

if to < input.count { /// Requires `(input.count - to) ~= [1, 2]` == `true`
    let index = to /// `to == index` so `(input.count - index) ~= [1, 2]`

    let i1 = input[index]
    let i2 = index &+ 1 < input.count ? input[index &+ 1] : nil
    let i3 = index &+ 2 < input.count ? input[index &+ 2] : nil
```

`i3` == `nil` iff `index &+ 2 < input.count` == `false` => `!(index &+ 2 < input.count)` == `true`.

`!(index &+ 2 < input.count)` == `index &+ 2 >= input.count` == `index &+ 3 > input.count`
== `3 > input.count - index` and we already have `(input.count - index) ~= [1, 2]` so
`3 > input.count - index` == `3 > (1 or 2)` which is always `true`.

Therefore `i3` == `nil` is always `true`.

Tests/FoundationEssentialsTests/DataTests.swift
```diff
@@ -2945,6 +2945,76 @@ extension DataTests {
...
+            [.base64URLAlphabet, .omitPaddingCharacter],
+            [],
+        ]
+    )
+    func base64Encode_matchesReferenceEncoderAcrossInputLengths(options: Data.Base64EncodingOptions) {
+        for count in 0..<260 {
+            // Create random-looking bytes
+            let bytes = (0..<count).map { UInt8(truncatingIfNeeded: $0 &* 7 &+ 13) }
+            let data = Data(bytes)
+            let expected = self.slowBase64Encoded(bytes, options: options)
```
parkera, line 3005: I'm not sure what this test is useful for - because it assumes the slow encoder is always correct. In order to prove that we would need tests for the test code. :)

Sources/FoundationEssentials/Data/Data+Base64.swift
```diff
@@ -542,18 +395,35 @@ extension Base64 {
...
+            output[out &+ 1] = Self.encodeCharacter(((i1 & 0x03) &<< 4) | ((i2 &>> 4) & 0x0F), char62, char63)
+            output[out &+ 2] = Self.encodeCharacter(((i2 & 0x0F) &<< 2) | ((i3 &>> 6) & 0x03), char62, char63)
+            output[out &+ 3] = Self.encodeCharacter(i3 & 0x3F, char62, char63)
         }
+        outIndex &+= triples &* 4
+    }
+
+    /// The base64 alphabet computed arithmetically rather than through a lookup table.
+    /// The code below is compiled as branchless, which is needed for LLVM
+    /// to vectorize the loop this func is called in.
```
parkera: nit pick - please unwrap this comment

parkera (commented):
Thank you for this work @MahdiBM - overall this looks great. Just one formatting nitpick and a comment on the tests. I'd like @fabianfett to also review

parkera:
@swift-ci test

parkera:
> Furthermore Swift's SIMD stuff are not good enough to help here.

cc @stephentyrone on this part

Tests/FoundationEssentialsTests/DataTests.swift
MahdiBM: I'd be fine with removing it but the idea is _surely_ we don't have _2_ incorrect encoders.
Because the real encoder has actual tests, and real encoder is tested against this simple encoder.
Also this is a safe encoder that won't have OOB writes or such for sure.
Essentially a 3-way test, real-encoder vs simple-encoder vs tests.

I like having such a thing for a double check. Even if there is a small chance of it possibly catching an issue then it'll be worth it.

Sources/FoundationEssentials/Data/Data+Base64.swift
MahdiBM: done.

Tests/FoundationEssentialsTests/DataTests.swift
parkera: > surely we don't have 2 incorrect encoders

Of course that's possible.

PR #2237, opened 2026-09-14
Data.write(options: .atomic) creates its temp file with 0200 -- break…

kperryua:
Don't attempt to open temporary files with 0200 permissions on anything but Darwin.

### Motivation:

Some file systems on Linux, and potentially other platforms, reject opening a writable file descriptor with just 0200 permissions with an EACCES error. While 0200 is the theoretical minimum permissions we need to write this file, it's not required to achieve the security guarantees we're seeking here.

### Modifications:

On all platforms besides Darwin, open the temporary file as 0600 for maximum compatibility.

### Result:

Linux atomic writing succeeds.

### Testing:

Tested with various file systems that have or were suspected to encounter the error. I do not think we have the infrastructure in CI to automate testing of these particular scenarios at the moment.


jmschonfeld (approved):
LGTM, let's make sure we cherry-pick this to `release/6.4.2` after merging since that's where this original (causing) change landed

jmschonfeld:
@swift-ci please test macOS

